PR #10, opened 2014-08-16
Added basic tests for ipfs-cli

malandrew:
Basic `ipfs-cli` tests. Currently covers all the commands in the readme. There is a fair amount of repetition in the test file right now. 

I've seen this repetition before with CLI tests with `isaacs/node-tap` that I contribute to. IMHO the right solution is a generalized TAP test tool for CLIs like `sstephenson/bats`, but without having to use `bash` (ugghh!). I may end up making such a tool at some point, at which time I'll revisit these tests and DRY everything up.

### DIFF
--- a/submodules/ipfs-cli/add.js
+++ b/submodules/ipfs-cli/add.js
@@ -34,7 +34,7 @@ Adder.prototype.addPath = function(p, cb) {
       if (self.opts.recursive)
         self.addTree(p, cb)
       else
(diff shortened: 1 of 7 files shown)


diasdavid:
Thank you for the contribution @malandrew and my apologies for leaving this one hanging here. We've been refactoring everything in the Node.js IPFS land and this code no longer apply, check the new README for the updates! :)

PR #28, opened 2015-09-15
Cleaned readme, added discussion links

RichardLitt:
@diasdavid How does this look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,38 +27,38 @@ IPFS in Node is a work in progress. As such, there's a few things you can do rig
 # Roadmap
 
 - Network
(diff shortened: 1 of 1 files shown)


diasdavid:
It does look good! :) Thank you @RichardLitt 

We want to have a webpage for libp2p similar to http://iswebrtcreadyyet.com, but instead of browsers have different languages and instead of API features, have the different modules. I'm not sure if it would look good in a README, but feel free to take a stab at it if you have time + interest :)

RichardLitt:
@diasdavid Is that idea documented anywhere else? And by different languages, isn't there just go and node at the moment?

PR #34, opened 2015-11-01
Fixed broken links

masylum:
`s/ipfs/libp2p`


RichardLitt:
LGTM! @diasdavid You made these changes, do they look right? You should be able to eyeball, where I'd have to check each one for veracity.

diasdavid:
Thank you :)

PR #40, opened 2015-12-11
update roadmap

diasdavid:


jbenet:
thank you for this!! helps understand what's going on very well :+1: :smile: 

@whyrusleeping and I need to do it for go-ipfs too.

PR #43, opened 2016-01-07
version

diasdavid:
- [x] cli
- [x] http-api
- [x] core
- [x] tests cli
- [x] tests http-api 
- [x] tests core


src/http-api/index.js
```diff
@@ -0,0 +1,37 @@
...
+var IPFS = require('../ipfs-core')
+var debug = require('debug')
+var log = debug('api')
+
+exports = module.exports
+
+exports.start = function (callback) {
+  // start IPFS and exports.ipfs = new IPFS()
+
+  exports.ipfs = new IPFS()
```
Dignifiedquire: Hmm, this looks like asking for trouble when the export is only available after `start` was called, also what will happen to module caching plus calling `start` twice?

src/http-api/resources/version.js
```diff
@@ -0,0 +1,14 @@
+var ipfs = require('./../index.js').ipfs
+var boom = require('boom')
+
+exports = module.exports
```
Dignifiedquire: Why is this needed?

diasdavid:
going to make it all es6'ish before merging it :)

src/http-api/index.js
```diff
@@ -0,0 +1,37 @@
...
+      }
+    }
+  })
+
+  server.connection({
+    port: 9000
+  })
+
+  // load routes
+  require('./routes/version.js')
```
Dignifiedquire: This could be much cleaner:

```js
server.route(require('./routes/version.js')

// version.js
var resources = require('./../resources')

module.exports = {
  method: 'GET',
  path: '/api/v0/version',
  handler: resources.version.get
}
```

src/ipfs-core/index.js
```diff
@@ -41,13 +22,15 @@ function IPFS () {
...
-    repo.config.read(function (err, config) {
-      if (err) {
-        return callback(err)
-      }
-      callback(null, config.Version.Current)
+    repo.exists(function (err, exists) {
+      if (err) { return callback(err) }
+
+      repo.config.get(function (err, config) {
+        if (err) { return callback(err) }
```
Dignifiedquire: Inconsistent style, braces + same line

src/http-api/index.js
diasdavid: the .ipfs is to be used by the http-api resources, which are only loaded after start is called (from routes), so there is no conflict. Did I get your question right?

Dignifiedquire: Kind of, but why not just pass these resources through explicitly, this could easily break, or be miss used by someone not understanding these exports

src/http-api/resources/version.js
diasdavid: because although they point to the same thing initially, if not binded, once exports changes it will stop pointing to module.exports and therefore none of the functions that gets exported by `exports.` will be exposed

more info and examples:
https://blog.tableflip.io/the-difference-between-module-exports-and-exports/

Dignifiedquire: Hmm but you don't use `module.exports`

diasdavid: not in this case, but it is kind of the same why I always add `{}` to if statements, keeping the code uniform, avoid having to debug someday cause 'I forgot one' or something.

Dignifiedquire: okay

src/http-api/index.js
diasdavid: why is that _"much"_ cleaner? If there was more than one route on that file I would have to go and

```JavaScript
server.route(require('./routes/version.js').routeA)
server.route(require('./routes/version.js').routeB)
// etc
```

Meaning that I have to alter the route file and index.js, the way I have, I only have to open and modify one file. Also, having the server object available means I always have server methods at my hand.

src/cli/commands/id.js
```diff
@@ -1,5 +1,8 @@
-var Command = require('ronin').Command
-var IPFS = require('../../ipfs-core')
+const Command = require('ronin').Command
+const IPFS = require('../../ipfs-core')
+const debug = require('debug')
+let log = debug('cli:id')
```
Dignifiedquire: Can be `const`

src/cli/commands/id.js
```diff
@@ -11,12 +14,10 @@ module.exports = Command.extend({
...
   },
 
-  run: function (name) {
-    var node = new IPFS()
-    node.id(function (err, id) {
-      if (err) {
-        return console.error(err)
-      }
+  run: name => {
+    let node = new IPFS()
```
Dignifiedquire: Why `Let`?

src/ipfs-core/index.js
diasdavid: fixed now

diasdavid:
Thank you for the CR @Dignifiedquire :)

src/ipfs-core/index.js
```diff
@@ -1,99 +1,79 @@
...
-    opts.url = process.env.APIURL
-  }
-
-  // var api = ipfsAPIclt(config.url)
-  // extend(self, api)
-
   var repo = new IPFSRepo(config.repoPath)
 
-  self.daemon = function (callback) {
+  this.daemon = (callback) => {
```
Dignifiedquire: You can drop the parens around callback

src/ipfs-core/index.js
```diff
@@ -1,99 +1,79 @@
...
+const IPFSRepo = require('ipfs-repo')
 
 exports = module.exports = IPFS
 exports.config = config
 
 function IPFS () {
-  var self = this
-
-  if (!(self instanceof IPFS)) {
+  if (!(this instanceof IPFS)) {
```
Dignifiedquire: If you do this check anyway, why not go `class IPFS` instead?

tests/test-cli/test-version.js
```diff
@@ -0,0 +1,16 @@
+/* globals describe, it */
+
+var expect = require('chai').expect
+var nexpect = require('nexpect')
+
+describe('cli: version', function () {
+  it('get the version', function (done) {
```
Dignifiedquire: I want an arrow function as well

tests/test-cli-offline/test-version.js
```diff
@@ -0,0 +1,16 @@
+/* globals describe, it */
+
+var expect = require('chai').expect
+var nexpect = require('nexpect')
+
+describe('cli-offline: version', function () {
+  it('get the version', function (done) {
```
Dignifiedquire: me toooo

tests/test-api/test-version.js
```diff
@@ -0,0 +1,21 @@
+/* globals describe, it */
+
+var expect = require('chai').expect
+
+process.env.IPFS_PATH = process.cwd() + '/tests/repo-example'
+var api = require('../../src/http-api')
+
+describe('api: version', function () {
+  it('get the version', function (done) {
```
Dignifiedquire: and what about me?

src/http-api/index.js
Dignifiedquire: It's not about that, could also change it to

```js
```

Dignifiedquire:
> Thank you for the CR @Dignifiedquire :)

I see code, I review

src/ipfs-core/index.js
diasdavid: I was not sure how classes worked underneath, was afraid that it would cause the same problem of method allocation during object initialisation and take a lot of unnecessary memory, but after checking, it seems there is no reason why to be concerned since it moves all the methods to .prototype of class.

src/http-api/index.js
diasdavid: we could DI the server obj (and the IPFS core object), but I kind of like making available by knowing where to pick it from other then move it around. I've been doing this for a while, if it becomes complicated over time, I will reconsider your suggestion :)

Issue #50, opened 2016-01-14
block


diasdavid:
Interact with raw blocks in the datastore. It uses IPFS-repo to store and get blocks

#### To learn more about the Block API, you can

- check out the go impl https://github.com/ipfs/go-ipfs/tree/master/blocks
- read datastore spec https://github.com/ipfs/specs/tree/master/repo#datastore

#### Implementation Roadmap

- [x] ipfs-blocks JavaScript implementation https://github.com/ipfs/js-ipfs-blocks
- jsipfs block get <key>  - Get a raw IPFS block.
  - [x] core
  - [x] http-api
  - [x] cli
- jsipfs block put <data> - Stores input as an IPFS block.
  - [x] core
  - [x] http-api
  - [x] cli
- jsipfs block del <data> - Deletes a stored block.
  - [x] core
  - [x] http-api
  - [x] cli
- jsipfs block stat <key> - Print information of a raw IPFS block.
  - [x] core
  - [x] http-api
  - [x] cli

diasdavid:
need `remove` in [js-ipfs-repo](https://github.com/ipfs/js-ipfs-repo) for `del`

diasdavid:
blocks http-api spec - https://github.com/ipfs/http-api-spec/pull/25

RichardLitt:
Let me know if I can help.

RichardLitt:
Why isn't `del` in go-ipfs?

diasdavid:
@RichardLitt go-ipfs doesn't expose it (yet). But it is what gc does internally. Exposing it is nice for testing and convenience.

diasdavid:
Booooom! :D done, thank @RichardLitt and @xicombd for getting this done ! :D

PR #52, opened 2016-01-14
WIP: feat/blocks

diasdavid:
issue: https://github.com/ipfs/js-ipfs/issues/50


diasdavid:
- [ ] needs https://github.com/webpack/webpack/pull/1930 to be merged
- [ ] needs karma-webpack to be updated (`Peer karma-webpack@1.7.0 wants webpack@^1.4.0`)

tests/test-core/test-block.js
```diff
@@ -0,0 +1,73 @@
...
+const base58 = require('bs58')
+const fs = require('fs')
+const IPFS = require('../../src/ipfs-core')
+const Block = require('ipfs-merkle-dag').Block
+
+const isNode = !global.window
+
+const fileA = isNode
+  ? fs.readFileSync(process.cwd() + '/tests/repo-example/blocks/12207028/122070286b9afa6620a66f715c7020d68af3d10e1a497971629c07606bfdb812303d.data')
+  : new Buffer(require('raw!./../repo-example/blocks/12207028/122070286b9afa6620a66f715c7020d68af3d10e1a497971629c07606bfdb812303d.data'))
```
diasdavid: `bug` found (this took me a bit to figure out :))

When using raw-loader, a string is returned with the contents of the file, if the file content was text, it would be ok, but since it is an encoded protobuf, it will add extra encoding data (utf8) adding extra bytes to the content of the Buffer, increasing the size of the Block and by so, failing the test. Need to get another way to load the blocks into localstorage

Dignifiedquire: Maybe this could work for you? https://github.com/martijnvermaat/binary-loader

PR #55, opened 2016-01-24
Run core tests in the browser

xicombd:
This is mostly the same as https://github.com/ipfs/js-ipfs-repo/pull/26.

Changed IPFS core to allow it to be instantiated with a custom `ipfs-repo`, but if none is provided it will use the default with `fs-blob-store` for `node`, or with `local-storage-blob-store` for the browser.

Also had to replace `let`'s by `const`'s on the tests [because of Firefox](https://travis-ci.org/ipfs/js-ipfs/jobs/104435637#L867).

### DIFF
--- a/src/ipfs-core/default-repo/index.js
+++ b/src/ipfs-core/default-repo/index.js
@@ -1,0 +1,5 @@
+const isNode = !global.window
+
+module.exports = isNode
+  ? require('./node')
+  : require('./browser')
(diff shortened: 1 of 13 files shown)


src/ipfs-core/default-repo/index.js
```diff
@@ -0,0 +1,5 @@
+const isNode = !global.window
+
+module.exports = isNode
```
diasdavid: nice one!

diasdavid:
@xicombd this is pretty awesome! thank you once again for contributing :+1: 🎉

PR #61, opened 2016-01-27
WIP: move to latest ipfs-repo

diasdavid:
- [ ]  Still need to fix the problem that blocks get mangled when stored in idb/local storage because they are converted to string, changing their encoding


dignifiedquire:
It is storing the binary, not sure where the changes are coming from :/

PR #62, opened 2016-01-28
Use buffer-loader for tests

xicombd:
I implemented a simple [`buffer-loader`](https://github.com/xicombd/buffer-loader) based on the `binary-loader` which fixes the two missing tests.

However I got a weird bug with webpack as it doesn't seem to define `Buffer` if it is not used outside the tests, so  [I added an hack on top of the file](https://github.com/ipfs/js-ipfs/compare/update/ipfs-repo...xicombd:use-buffer-loader?expand=1#diff-9c147124754a0b48ca1d40235e76a0b6R14). 
@dignifiedquire do you have any idea how to avoid this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
   "homepage": "https://github.com/ipfs/js-ipfs#readme",
   "devDependencies": {
     "async": "^1.5.2",
-    "binary-loader": "0.0.1",
     "brfs": "^1.4.3",
+    "buffer-loader": "0.0.1",
     "chai": "^3.4.1",
     "fs-blob-store": "^5.2.1",
     "idb-plus-blob-store": "^1.0.0",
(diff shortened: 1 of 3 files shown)


diasdavid:
@xicombd the buffer thing is weird indeed. Did you tried with the WebPack version that uses node-libs 1.0.0?

xicombd:
Tried with latest webpack and with your fork, and neither seem to work :worried:

dignifiedquire:
@xicombd take a look at this page: http://webpack.github.io/docs/how-to-write-a-loader.html#mark-dependencies you might have to flag `Buffer` explicitly as a dependency in your loader

xicombd:
@dignifiedquire hum, that doesn't seem to be the problem as it loads the file properly, and I can log the buffer size correctly. The problem is that Buffer isn't defined on the browser context, what makes the two tests that use it fail:
screen shot 2016-01-28 at 13 09 38

dignifiedquire:
@xicombd looks like I found a bug in webpack: https://github.com/webpack/webpack/issues/1944

xicombd:
@dignifiedquire Oh, nice catch! That explains it, I'll change my PR to use regular functions in the meanwhile then. Thanks!

diasdavid:
:+1: @dignifiedquire :D @xicombd wanna update the PR?

xicombd:
@diasdavid done!

diasdavid:
rad! thank you :)

Issue #58, opened 2016-01-26
object


diasdavid:
The object API enables IPFS users to create graph structures with IPFS were links are MerkleLinks (multihashes)

#### To learn more about the object API, you can

- check go implementation of the MerkleDAG https://github.com/ipfs/go-ipfs/tree/master/merkledag
- check the MerkleDAG spec https://github.com/ipfs/specs/tree/master/merkledag
- see the http-api spec for the object API https://github.com/ipfs/http-api-spec/pull/32

#### Implementation Roadmap

- [x] MerkleDAG JS implementation https://github.com/vijayee/js-ipfs-merkle-dag
- [x] jsipfs object new [template] - Creates a new object from an ipfs template.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object get <key>        - Get and serialize the DAG node named by <key>.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object put <data>       - Stores input as a DAG object, outputs its key.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object stat <key>       - Get stats for the DAG node named by <key>.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object data <key>       - Outputs the raw bytes in an IPFS object.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object links <key>      - Outputs the links pointed to by the specified object.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object patch append-data <root> <data>    - Append data to the data segment of a dag node.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object patch add-link <root> <name> <ref> - Add a link to a given object.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object patch rm-link <root> <link>        - Remove a link from an object.
  - [x] core
  - [x] http-api
  - [x] cli
- [x] jsipfs object patch set-data <root> <data>       - Set data field of an ipfs object.
  - [x] core
  - [x] http-api
  - [x] cli

diasdavid:
**updated this issue**

xicombd:
Started working on this on [feature/object](https://github.com/ipfs/js-ipfs/tree/feature/object) branch :)

diasdavid:
@xicombd :+1: :D

xicombd:
[...]

PR #63, opened 2016-01-28
feature object

diasdavid:
ref: https://github.com/ipfs/js-ipfs/issues/58


src/cli/commands/config.js
```diff
@@ -24,7 +24,7 @@ module.exports = Command.extend({
 
   run: (bool, json, key, value) => {
     if (!key) {
-      throw new Error('argument \'key\' is required')
+      throw new Error("argument 'key' is required")
```
diasdavid: this is `standard` codestyle things

dignifiedquire: that's stupid :(

tests/test-cli/test-config.js
```diff
@@ -77,7 +77,7 @@ describe('config', () => {
   it('call config with no arguments', done => {
     nexpect.spawn('node', [process.cwd() + '/src/cli/bin.js', 'config'])
       .run((err, stdout, exitcode) => {
-        const expected = 'error	argument \'key\' is required'
+        const expected = "error	argument 'key' is required"
```
RichardLitt: error argument is two words

src/cli/commands/config.js
RichardLitt: maybe lint in a separate commit?

tests/test-cli/test-config.js
diasdavid: interesting, apparently both vim and node understood that there was a tab char there, but github just ignored it, changing to `\t`:)

diasdavid:
@xicombd @dignifiedquire I'm getting the same "Buffer is not defined" error we were having in test-block.js, but this time, not using arrow functions is not solving the problem, ideas?

src/ipfs-core/index.js
```diff
@@ -18,6 +20,7 @@ function IPFS (repo) {
     repo = defaultRepo()
   }
   const bs = new BlockService(repo)
+  const ds = new DAGService(bs)
```
dignifiedquire: `ds` and `bs` are really short and hard to distinguish (and understand) :/

src/ipfs-core/index.js
```diff
@@ -153,4 +156,118 @@ function IPFS (repo) {
...
+        })
+      })
+    },
+    patch: {
+      appendData: (multihash, data, callback) => {
+        this.object.get(multihash, (err, obj) => {
+          if (err) { return callback(err) }
+          obj.data = Buffer.concat([obj.data, data])
+          ds.add(obj, (err) => {
+            if (err) { return callback(err) }
```
dignifiedquire: let's be consistent and make this onto multiple lines as above

src/ipfs-core/index.js
```diff
@@ -153,4 +156,118 @@ function IPFS (repo) {
...
+          })
+        })
+      },
+      rmLink: (multihash, multihashLink, callback) => {
+        this.object.get(multihash, (err, obj) => {
+          if (err) { return callback(err) }
+          obj.links = obj.links.filter((link) => {
+            if (link.hash.equals(multihashLink)) {
+              return false
+            } else {
```
dignifiedquire: no need for `else` if you `return` above

tests/test-core/test-object.js
```diff
@@ -0,0 +1,130 @@
...
+'use strict'
+
+const expect = require('chai').expect
+const IPFS = require('../../src/ipfs-core')
+const bs58 = require('bs58')
+const mDAG = require('ipfs-merkle-dag')
+const DAGNode = mDAG.DAGNode
+const DAGLink = mDAG.DAGLink
+
+// TODO use arrow funtions again when https://github.com/webpack/webpack/issues/1944 is fixed
```
dignifiedquire: You can use arrow functions now with webpack 2

tests/test-core/test-object.js
```diff
@@ -0,0 +1,130 @@
...
+describe('object', function () {
+  var ipfs
+
+  before(function (done) {
+    ipfs = new IPFS()
+    done()
+  })
```
dignifiedquire: No need for `done` if no async:

```js
before(() => {
 ipfs = new IPFS()
})
```

tests/test-core/test-object.js
```diff
@@ -0,0 +1,130 @@
...
+    ipfs.object.new(function (err, obj) {
+      expect(err).to.not.exist
+      expect(obj).to.have.property('Size')
+      expect(obj.Size).to.equal(0)
+      expect(obj).to.have.property('Name')
+      expect(obj.Name).to.equal('')
+      expect(obj).to.have.property('Hash')
+      expect(bs58.encode(obj.Hash).toString())
+         .to.equal('QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n')
+      expect(obj.Size).to.equal(0)
```
dignifiedquire: double assertion with l. 25

tests/test-core/test-object.js
```diff
@@ -0,0 +1,130 @@
...
+  })
+
+  it('new', function (done) {
+    ipfs.object.new(function (err, obj) {
+      expect(err).to.not.exist
+      expect(obj).to.have.property('Size')
+      expect(obj.Size).to.equal(0)
+      expect(obj).to.have.property('Name')
+      expect(obj.Name).to.equal('')
+      expect(obj).to.have.property('Hash')
```
dignifiedquire: This could be nicer like this:

```js
expect(obj).to.have.property('Size')
  .that.is.equal(0)

expect(obj)
  .to.have.property('Name')
  .that.is.equal('')
```

dignifiedquire:
@diasdavid some small nitpicks but other than that LGTM

src/ipfs-core/index.js
```diff
@@ -2,10 +2,12 @@
 
 const defaultRepo = require('./default-repo')
 // const bl = require('bl')
```
noffle: any reason to keep this around?

noffle: any reason to keep this around?

src/ipfs-core/index.js
noffle: yeah, +1 for explicitly naming them `blockService` and `dagService` respectively.

tests/test-cli/test-config.js
```diff
@@ -77,7 +77,7 @@ describe('config', () => {
   it('call config with no arguments', done => {
     nexpect.spawn('node', [process.cwd() + '/src/cli/bin.js', 'config'])
       .run((err, stdout, exitcode) => {
-        const expected = 'error	argument \'key\' is required'
+        const expected = "error\targument 'key' is required"
```
noffle: single quotes on the outside, `\'` on the inside (standard)

tests/test-core/test-config.js
```diff
@@ -90,12 +90,12 @@ describe('config', () => {
...
-  // })
+// cli only feature built with show and replace
+// it.skip('edit', done => {
+//   const ipfs = new IPFS()
+//  ipfs.config((err, config) => {
+//    expect(err).to.not.exist
+//    done()
+//  })
+// })
 })
```
noffle: let's just delete this? it's all in git so we can bring it back if we need it.

tests/test-cli/test-config.js
diasdavid: `standard --format` caused that change

diasdavid:
greeeen! :D

PR #67, opened 2016-02-13
feature/files add

diasdavid:
This PR adds `files add` and `files add -r`

Missing:
- [ ] tests
- [x] figure out why adding a dir with multiple files is calling the success callback for each of those files (might be a bug with data-importing itself)


RichardLitt:
@diasdavid Would this need this issue to be solved? https://github.com/ipfs/http-api-spec/issues/57

diasdavid:
Yes, that would be 🌟

diasdavid:
I'm going to merge this PR to unlock `jsipfs init` which needed `files add` for its base docs.

PR #69, opened 2016-02-21
WIP: Make http-api endpoints as specified by the http-api-spec, for the features present in core 

diasdavid:
This is a big PR, still a WIP, it includes:

- generic
  - [x] being able to start and stop the daemon (graciously using SIGINT)
  - [x] upgrade the http-api to have 2 conns (one for API and the other for gateway)
  - [x] lock the repo if a daemon is being run (a la go-ipfs)
  - [x] update the tests to follow this new pattern
  - [x] merge cli offline and cli with api running tests
  - [x] make the cli use the http-api if it is running
  - [x] add tests with js-ipfs-api for extra level of interop guarantees
- http-api endpoints that follow with the http-api-spec
  - [x] version
  - [x] id
  - [x] bootstrap
  - [x] config


tests/test-http-api/test-version.js
```diff
@@ -0,0 +1,52 @@
...
+  describe('using js-ipfs-api', () => {
+    var ctl
+
+    it('start IPFS API ctl', (done) => {
+      ctl = APIctl('/ip4/127.0.0.1/tcp/6001')
+      done()
+    })
+
+    it('get the version', (done) => {
+      ctl.version((err, result) => {
```
diasdavid: @dignifiedquire I'm hitting an issue here that I would love to have your review. I implemented the http-api endpoint following the spec and you can see the test above (line 17) that it is indeed callable and returns the data as expected, however, the problem is that this js-ipfs-api call is giving me the result as a 'buffer', instead of the object correctly parsed, my first intuition was it was lacking headers, but I confirmed that I was adding the same headers as go-ipfs and still I got nothing. Any ideas? Thank you.

tests/test-http-api/test-config.js
```diff
@@ -0,0 +1,25 @@
+/* globals describe, it */
```
dignifiedquire: There is a nicer way for this by the way: https://github.com/feross/standard#what-about-mocha-jasmine-qunit-etc

tests/test-http-api/test-version.js
dignifiedquire: @diasdavid there is your fix :) https://github.com/ipfs/js-ipfs-api/pull/215

diasdavid: sweeeet! :D

PR #78, opened 2016-03-04
Add ipfs commands to cli

### LINKED ISSUE #54: commands
The `commands` command simply prints out all the available commands (and their subcommands). It is part of tooling.

- [x] jsipfs commands

xicombd:
Closes #54

### DIFF
--- a/src/cli/commands/commands.js
+++ b/src/cli/commands/commands.js
@@ -1,10 +1,22 @@
 'use strict'
 
 const Command = require('ronin').Command
+const path = require('path')
+const ronin = require('ronin')
 
 module.exports = Command.extend({
   desc: '',
 
   run: name => {
+    const cli = ronin(path.resolve(__dirname, '..'))
+
+    cli.setupCommands()
+
+    const commands = ['']
+      .concat(Object.keys(cli.commands))
+      .map((command) => 'ipfs ' + command)
+      .join('\n')
+
+    console.log(commands)
   }
 })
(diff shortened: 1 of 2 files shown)


src/cli/commands/commands.js
```diff
@@ -1,10 +1,22 @@
...
 
 const Command = require('ronin').Command
+const path = require('path')
+const ronin = require('ronin')
 
 module.exports = Command.extend({
   desc: '',
 
   run: name => {
+    const cli = ronin(path.resolve(__dirname, '..'))
```
diasdavid: a ronin inside a ronin

src/cli/commands/commands.js
```diff
@@ -1,10 +1,22 @@
...
+
+    cli.setupCommands()
+
+    const commands = ['']
+      .concat(Object.keys(cli.commands))
+      .map((command) => 'ipfs ' + command)
+      .join('\n')
+
+    console.log(commands)
```
diasdavid: wouldn't it be sweet if arrays had by default a streams interface :)

src/cli/commands/commands.js
xicombd: True, but couldn't find out any other way of doing it :worried: 
Requiring `src/cli/index` didn't work as it only got the commands afterwards. 

However this is not a full ronin as `.run()` isn't called :)

src/cli/commands/commands.js
dignifiedquire: take a look at http://highlandjs.org/ :)

Issue #30, opened 2015-09-22
captain.log - IPFS Node.js implementation


diasdavid:
Aye mates!

This is your captain speaking, the IPFS Node.js ship as been to set sail and good winds are blowing! Thank you for all the messages and enthusiasm shared so far, we truly appreciate your support and since some of you are eager to get their hands on code and help with the development, I'll keep this issue updated every week with what has been going on and where you can help the most. To make sure you don't miss any updates, click the 'watch' button on the repo.

### LOG - Week of 21st of September 

**Plan:**

- Finish Swarm revisit 
  - https://github.com/diasdavid/node-ipfs-swarm/issues/8
  - https://github.com/diasdavid/node-ipfs-swarm/tree/revisit
- Implement node-libp2p bindings for every other module that is part of libp2p and have a alpha release people can experiment with
  - https://github.com/diasdavid/node-libp2p

**Where you can help during this week:**

- If crypto is your thing, take a look at 
  - https://github.com/diasdavid/webcrypto/issues/1
  - https://github.com/nodejs/node/issues/2833

- If dream with tests and code coverage, there is a bunch of repos that could use some love
  - https://github.com/ipfs/node-ipfs#roadmap

diasdavid:
Aye! Check my new entry, arr :)

### LOG - Week of 21st of September

**Plan:**

- Have a interoperable version of libp2p in Go, to that the primitives and coverage become stronger so that then the remaining parts of IPFS can implemented with confidence (cli, bitswap, chunker, etc)

**Where you can help during this week:**

- Making SPDY become a duplex stream - https://github.com/indutny/spdy-transport/issues/14

- MOAR transports. Create other transports that swarm can use and tests to see them working at the same time
  - [libp2p-udp](https://github.com/diasdavid/node-libp2p-udp)
  - [libp2p-udt](https://github.com/diasdavid/node-libp2p-udt)
  - [libp2p-utp](https://github.com/diasdavid/node-libp2p-utp)

- If dream with tests and code coverage, there is a bunch of repos that could use some love
  - https://github.com/ipfs/node-ipfs#roadmap

diasdavid:
### LOG - Week of December 21

**Update:**

- We have now IPFS Repo implemented in JS https://github.com/ipfs/js-ipfs-repo
- Bitswap work has started 
  - https://github.com/diasdavid/js-ipfs-bitswap
  - https://github.com/ipfs/specs/pull/53

**Next:**

[...]

PR #81, opened 2016-03-10
jsipfs block http-api and cli

RichardLitt:
So, I stubbed out a couple of files, but I'm not sure this is what I am actually supposed to be doing here. It is hard without more examples, and I don't know how to run, or what sort of development environment I should be using to test the code as I go along.

### DIFF
--- a/src/cli/commands/block/get.js
+++ b/src/cli/commands/block/get.js
@@ -1,0 +1,42 @@
+const Command = require('ronin').Command
+const utils = require('../../utils')
+const bs58 = require('bs58')
+const debug = require('debug')
+const log = debug('cli:block')
+log.error = debug('cli:block:error')
+
+module.exports = Command.extend({
+  desc: 'Get a raw IPFS block',
+
+  options: {},
+
+  run: (key) => {
+    if (!key) {
+      throw new Error("Argument 'key' is required")
+    }
+
+    utils.getIPFS((err, ipfs) => {
+      if (err) {
+        throw err
+      }
+
+      const mh = utils.isDaemonOn()
+        ? key
+        : new Buffer(bs58.decode(key))
+
+      ipfs.block.get(mh, (err, block) => {
+        if (err) {
+          log.error(err)
+          throw err
+        }
+
+        if (block.data) {
+          console.log(block.data.toString())
+          return
+        }
+
+        console.log(block.toString())
+      })
+    })
+  }
+})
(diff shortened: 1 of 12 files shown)


RichardLitt:
Ping @diasdavid, of course. What should I be doing next?

RichardLitt:
Related to #50

diasdavid:
@RichardLitt didn't the "Good foundations" section on https://github.com/ipfs/js-ipfs/issues/30#issuecomment-187950929 helped? Let's chat today to clear out any questions :)

RichardLitt:
Ah, that's interesting. I'll take a look through this again and see what I can do.

diasdavid:
@RichardLitt do you still need some help here?

diasdavid:
ping @RichardLitt

xicombd:
Just finished implementing this, ping @diasdavid

src/cli/commands/block/get.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    utils.getIPFS((err, ipfs) => {
+      if (err) {
+        throw err
+      }
+
+      const mh = utils.isDaemonOn()
+        ? key
+        : new Buffer(bs58.decode(key))
```
diasdavid: soon all these hacks will be gone (once we get core spec \o/)

src/cli/commands/block/rm.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    utils.getIPFS((err, ipfs) => {
+      if (err) {
+        throw err
+      }
+
+      if (utils.isDaemonOn()) {
+        // TODO implement this once `js-ipfs-api` supports it
+        throw new Error('rm block with daemon running is not yet implemented')
```
diasdavid: why?

diasdavid: got it, http-api-spec doesn't have it spec'ed out. Could you change the message to say that `rm` is not available through `http-api-spec`

diasdavid:
@xicombd sweet! Great work :D Only one comment :)

src/cli/commands/block/rm.js
xicombd: There's no ipfs block del or rm on `go-ipfs`, so `js-ipfs-api` doesn't have it either. I could make the raw request here, but thought it would be better to wait and avoid Yet Another Hack™. What are your thoughts on this?

diasdavid: sounds good. We definitely need to push it to be part of the spec (go-ipfs can delete things on disk directly, js-ipfs-api has to interact with disk through the blob-store, always)

PR #83, opened 2016-03-12
Add getIPFS util and refactor id command to use it

xicombd:
Add `getIPFS` method to cli utils that returns a `js-ipfs-api` instance if a daemon is running or a new core instance if not. 

Simplifies the cli commands as the two APIs are mostly equal :)


diasdavid:
I like it :)

Issue #70, opened 2016-02-23
http-api - get version test with js-ipfs-api fails **only** on Travis


diasdavid:
**TEST**

https://github.com/ipfs/js-ipfs/blob/master/tests/test-http-api/test-version.js#L38-L47

**LOG** 

```bash
Debug: internal, implementation, error 
    AssertionError: expected [Error: Client request error: Parse Error] to not exist
    at /home/travis/build/ipfs/js-ipfs/tests/test-http-api/test-version.js:40:27
    at /home/travis/build/ipfs/js-ipfs/node_modules/ipfs-api/src/request-api.js:23:14
    at finish (/home/travis/build/ipfs/js-ipfs/node_modules/wreck/lib/index.js:146:20)
    at wrapped (/home/travis/build/ipfs/js-ipfs/node_modules/hoek/lib/index.js:867:20)
    at ClientRequest.onError (/home/travis/build/ipfs/js-ipfs/node_modules/wreck/lib/index.js:155:16)
    at ClientRequest.g (events.js:260:16)
    at emitOne (events.js:77:13)
    at ClientRequest.emit (events.js:169:7)
    at Socket.socketOnData (_http_client.js:307:9)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
    at Socket.Readable.push (_stream_readable.js:110:10)
    at TCP.onread (net.js:523:20)
```

PR #84, opened 2016-03-13
Refactor HTTP API routes

xicombd:
Use dependency injection on routes so that we can start and stop the http-api multiple times without issues on the tests. 
~~Updated `jsipfs version` to work with the api running and added a test to make sure it works :)~~ Got similar bug to #70

### DIFF
--- a/src/http-api/index.js
+++ b/src/http-api/index.js
@@ -46,7 +46,7 @@ exports.start = (callback) => {
     server.connection({ host: gateway[2], port: gateway[4], labels: 'Gateway' })
 
     // load routes
-    require('./routes')
+    require('./routes')(server)
 
     server.start((err) => {
       if (err) {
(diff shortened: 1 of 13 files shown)


src/http-api/index.js
```diff
@@ -46,7 +46,7 @@ exports.start = (callback) => {
     server.connection({ host: gateway[2], port: gateway[4], labels: 'Gateway' })
 
     // load routes
-    require('./routes')
+    require('./routes')(server.select('API'))
```
diasdavid: pass the server object completely and let the routes decide where to attach

xicombd: @diasdavid done!

xicombd:
Added the ipfs version test with the daemon running again, now fixed

diasdavid:
thank you yet again @xicombd :D :D

PR #85, opened 2016-03-14
Add object endpoints and cli commands

xicombd:
Implemented most of #58, except for the `object patch` commands as the http API changed for 0.4.0 and the spec isn't not finished yet (https://github.com/ipfs/http-api-spec/pull/32#issuecomment-189491059)

### DIFF
--- a/src/cli/commands/object/data.js
+++ b/src/cli/commands/object/data.js
@@ -1,0 +1,41 @@
+'use strict'
+
+const Command = require('ronin').Command
+const utils = require('../../utils')
+const bs58 = require('bs58')
+const debug = require('debug')
+const log = debug('cli:object')
+log.error = debug('cli:object:error')
+
+module.exports = Command.extend({
+  desc: 'Outputs the raw bytes in an IPFS object',
+
+  options: {},
+
+  run: (key) => {
+    if (!key) {
+      throw new Error("Argument 'key' is required")
+    }
+
+    var ipfs = utils.getIPFS()
+
+    const mh = utils.isDaemonOn()
+      ? key
+      : new Buffer(bs58.decode(key))
+
+    ipfs.object.data(mh, (err, data) => {
+      if (err) {
+        log.error(err)
+        throw err
+      }
+
+      if (data instanceof Buffer) {
+        console.log(data.toString())
+        return
+      }
+
+      // js-ipfs-api output (http stream)
+      data.pipe(process.stdout)
+    })
+  }
+})
--- a/src/cli/commands/object/get.js
+++ b/src/cli/commands/object/get.js
@@ -1,0 +1,50 @@
+'use strict'
+
+const Command = require('ronin').Command
+const utils = require('../../utils')
+const bs58 = require('bs58')
+const debug = require('debug')
+const log = debug('cli:object')
+log.error = debug('cli:object:error')
+
+module.exports = Command.extend({
+  desc: 'Get and serialize the DAG node named by <key>',
+
+  options: {},
+
+  run: (key) => {
+    if (!key) {
+      throw new Error("Argument 'key' is required")
+    }
+
+    var ipfs = utils.getIPFS()
+
+    if (utils.isDaemonOn()) {
+      return ipfs.object.get(key, (err, obj) => {
+        if (err) {
+          log.error(err)
+          throw err
+        }
+
+        console.log(JSON.stringify(obj))
+      })
+    }
+
+    const mh = new Buffer(bs58.decode(key))
+    ipfs.object.get(mh, (err, obj) => {
+      if (err) {
+        log.error(err)
+        throw err
+      }
+
+      console.log(JSON.stringify({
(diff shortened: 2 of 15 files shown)


src/cli/commands/object/get.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    if (utils.isDaemonOn()) {
+      return ipfs.object.get(key, (err, obj) => {
+        if (err) {
+          log.error(err)
+          throw err
+        }
+
+        console.log(JSON.stringify(obj))
```
diasdavid: does this print in the same way that go-ipfs does? I believe go-ipfs doesn't some formatting rather then printing the obj directly

src/cli/commands/object/get.js
```diff
@@ -0,0 +1,50 @@
...
+        throw err
+      }
+
+      console.log(JSON.stringify({
+        Links: obj.links.map((link) => ({
+          Name: link.name,
+          Hash: bs58.encode(link.hash).toString(),
+          Size: link.size
+        })),
+        Data: obj.data.toString()
```
diasdavid: which seems you do it here :)

src/cli/commands/object/links.js
```diff
@@ -0,0 +1,44 @@
...
+module.exports = Command.extend({
+  desc: 'Outputs the links pointed to by the specified object',
+
+  options: {},
+
+  run: (key) => {
+    var ipfs = utils.getIPFS()
+
+    if (!key) {
+      throw new Error("Argument 'key' is required")
```
diasdavid: can you do this before utils.getIPFS() ?

src/cli/commands/object/get.js
xicombd: Yes: 
screen shot 2016-03-14 at 16 05 13
(last command was with a daemon running)

src/cli/commands/object/put.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    process.stdin.pipe(bl((err, input) => {
+      if (err) {
+        log.error(err)
+        throw err
+      }
+
+      parseAndAddNode(input)
+    }))
```
diasdavid: this whole command could be more simple. Also, seems like I can `cat file | jsipfs object put` and it would get confused

src/cli/commands/object/get.js
xicombd: That's because core returns an object with lowercase keys and the multihashes as a buffer, on the other hand the `js-ipfs-api` returns the http API response which returns the object like `go-ipfs`

src/cli/commands/object/data.js
```diff
@@ -0,0 +1,41 @@
...
+
+    ipfs.object.data(mh, (err, data) => {
+      if (err) {
+        log.error(err)
+        throw err
+      }
+
+      if (data instanceof Buffer) {
+        console.log(data.toString())
+        return
```
diasdavid: what is this doing?

xicombd: `js-ipfs-api`'s `object.data` accepts multihashes as a string, core needs the `Buffer` instead. 

`js-ipfs-api` returns a `stream`, core returns a `Buffer`

src/cli/commands/object/put.js
xicombd: According to `go-ipfs`, you can:
```
➜  js-ipfs git:(feature/object) ipfs object put --help

    ipfs object put <data> - Stores input as a DAG object, outputs its key.

ARGUMENTS:

    <data> - Data to be stored as a DAG object.

OPTIONS:

    --inputenc string - Encoding type of input data, either "protobuf" or "json".

DESCRIPTION:

    'ipfs object put' is a plumbing command for storing DAG nodes.
    It reads from stdin, and the output is a base58 encoded multihash.

    Data should be in the format specified by the --inputenc flag.
    --inputenc may be one of the following:
    	* "protobuf"
    	* "json" (default)

    Examples:

    	$ echo '{ "Data": "abc" }' | ipfs object put

    This creates a node with the data 'abc' and no links. For an object with links,
    create a file named 'node.json' with the contents:

        {
            "Data": "another",
            "Links": [ {
                "Name": "some link",
                "Hash": "QmXg9Pp2ytZ14xgmQjYEiHjVjMFXzCVVEcRTWJBmLgR39V",
                "Size": 8
            } ]
        }

    And then run:

    	$ ipfs object put node.json
```

diasdavid: got it. I meant that it doesn't seem the code to prevent something like

```
cat file | jsipfs object put <another object> 
```

xicombd: `go-ipfs` doesn't prevent it either, if you pass a file path it simply takes priority over the `stdin` :)

src/cli/commands/object/get.js
diasdavid: cool :)

diasdavid:
LGTM, going to merge this. Thank you @xicombd :D

PR #94, opened 2016-03-22
tests use random ports for gateways now

nginnever:
Added port zero to config file so the tests pull it and use random ports.

Also updated data-importing dep


diasdavid:
sweet :) thank you @nginnever !

PR #98, opened 2016-03-25
moves temp-repo creation into its own module

noffle:


diasdavid:
sweet, thank you @noffle ! I was planning to do this on the plan to refactor tests a bit, this way it just makes things faster :)

PR #102, opened 2016-03-31
Update README.md

kumavis:
:dragon:


RichardLitt:
Thar be dragoons everywar!

PR #107, opened 2016-04-03
feat: pin

AdamStone:
- [x] core
  - [x] add
  - [x] rm
  - [x] ls
  - [x] isPinned
  - [x] isPinnedWithType
  - [x] directKeys / recursiveKeys / internalKeys
  - [x] flush (persist keys to datastore)
  - [x] load (restore keys from datastore)
- [x] cli
  - [x] add
  - [x] rm
  - [x] ls
- [x] http
  - [x] add
  - [x] rm
  - [x] ls

Updated core interface to align with [interface-ipfs-core](https://github.com/ipfs/interface-ipfs-core), still need to update cli and http.


diasdavid:
@AdamStone, js-ipfs has gone through a good amount of updates, including bug fixes and new features. Would you like to revisit pinning again? rebase master on your branch and continue the PR

AdamStone:
Sure, I guess I should have been rebasing rather than merging to begin with. I think the core implementation is basically done, so I'll update the description. The failing test should pass once [this issue](https://github.com/vijayee/js-ipfs-merkle-dag/commit/0e4361f88002583322565b3ce27b26c6876d952c) is fixed.

package.json
```diff
@@ -63,6 +63,7 @@
     "bs58": "^3.0.0",
     "debug": "^2.2.0",
     "detect-node": "^2.0.3",
+    "fnv": "^0.1.3",
```
dignifiedquire: Are we sure about the licence for this module, can't find info on npm nor the repo.

AdamStone: Ah, good point. I'll swap in something with MIT license.

dignifiedquire: thanks :)

diasdavid:
This is looking good @AdamStone :)

Would you also like to lead the https://github.com/ipfs/interface-ipfs-core/pull/34, so that we have the same battery of tests for js-ipfs and js-ipfs-api for the `pinning API`.

Also, `go-ipfs` stores the root hash of the pinset inside level-db (datastore). Currently, js-ipfs doesn't use leveldb (datastore) because it is not implemented in ipfs-repo, would you like to tackle that as well -> https://github.com/ipfs/js-ipfs-repo/issues/13

AdamStone:
@diasdavid Sorry for the delayed response, I see you've already started on [ipfs/interface-ipfs-core#34](https://github.com/ipfs/interface-ipfs-core/pull/34) but I can continue with that if you like. Am I understanding correctly that it basically just involves migrating the core pinning tests from js-ipfs to interface-ipfs-core?

For the datastore, I did notice the discrepancy between js-ipfs and go-ipfs when working out how to store the pin set and ended up just putting it in the block store because I didn't see a good alternative. I just read through [ipfs/js-ipfs-repo#13](https://github.com/ipfs/js-ipfs-repo/issues/13) but found it pretty confusing at first pass. I think I'd better do a little more research before committing to take that on.

diasdavid:
@AdamStone sounds good. Let me know if you would like to go through it on IRC or a hangout, you can also join today's js-ipfs call https://github.com/ipfs/pm

diasdavid:
@AdamStone good news, we've finished the Pin interface in interface-ipfs-core, so now in order to validate this Pin implementation, you just need to make sure it passes those tests 

- https://github.com/ipfs/interface-ipfs-core/tree/master/API/pin
- https://github.com/ipfs/interface-ipfs-core/blob/master/src/pin.js

AdamStone:
Sounds good, I'll take a look tonight.

AdamStone:
Can you clarify, when pin.add succeeds and returns "an array of objects that represent the files that were pinned", should this include indirectly-pinned objects when the add is recursive?

diasdavid:
Since there is the type "recursive pin", we only need to know that the head was recursively pin. So no, we don't need to return them all.

AdamStone:
@diasdavid In that case, are there any circumstances where this should return more than just one object, since `add` expects a single hash? Or does that mean it should return the full pin set, not just the new addition?

diasdavid:
The best is to see how go-ipfs does it through the CLI and match the expectations as closely as possible.

AdamStone:
@diasdavid The go-ifps cli seems pretty flexible. You can pass one or more b58 hashes or ipfs paths like `/ipfs/b58hash/link`. For both `add` and `rm` it prints only the hashes that were changed by that operation, along with the type. I had tried to make the js pin CLI work with an ipfs path and behave like the go CLI, but I didn't notice you could pass more than one item. So I've adjusted the pin core methods to be more consistent with the CLI and to accept one hash/path or an array of them (not committed yet). Not sure if we want to test for those scenarios as well.

I found I needed to make a few modifications to the tests, so I [opened a PR](https://github.com/ipfs/interface-ipfs-core/pull/63) in interface-ipfs-core.

mindworker:
Any updates on pinning?

AdamStone:
@mindworker @diasdavid The pinning code itself has been done for awhile, but it seems the project core and test infrastructure keep changing underneath it. If I remember correctly, the most recent commits were in response to the core tests being refactored into the interface-ipfs-core package. In running this code against the pin tests that were included there, I found some problems with those tests and opened a [PR to interface-ipfs-core](https://github.com/ipfs/interface-ipfs-core/pull/63) to address them. At the time, all the pin tests were passing if those revisions were accepted. That PR is still open, it was labeled 'backlog' but I'm not sure why it was never accepted.

I see that more refactoring of the project and test infrastructure has been done since then, and it looks like it will involve some substantial revisions to this PR to get everything working again. Before I do that, can someone clarify what exactly needs to happen to get these PRs accepted this time around? As long as this stays open, it seems to require significant maintenance just to keep up with changes to the the main branch.

diasdavid:
Apologies for not giving clear feedback in the beginning, @AdamStone. I hoped that we had the 'datastore' ready on ipfs-repo, so that the pinning service could be 100% compatible with what go-ipfs does (storing the roothash of the pinset in the datastore), so that we could verify every time that pinning was done correctly with interop tests.

The implementation of the datastore kept getting deferred as other things got prioritized and so the review and merging of this PR.

Here is the list of things that needs to happen to get datastore - https://github.com/ipfs/js-ipfs-repo/issues/112

diasdavid:
@dignifiedquire once datastore is merged, wanna revisit this PR?

AdamStone:
Sure, sorry I didn't reply to your last post. I figured I'd wait to do anything else with this until that was ready.

zabirauf:
Thanks @AdamStone for working on pinning. I see that the datastore got merged to master. Any idea when the pinning changes would be ready?

hhff:
Hi all!  Is there an example PR of required changes for moving a different part of the js-ipfs wrapper to the new datastore pattern?

Happy to take a look and helping out here!

diasdavid:
@hhff, @pgte has been working on `datastore` and `unixfs` land for a while, he will be able to give you that direction :) 

That being said. I believe the best here is to squash the commits in this PR, rebase master onto it and then start by augmenting the batch of tests available in https://github.com/ipfs/interface-ipfs-core/blob/master/src/pin.js with @AdamStone's work https://github.com/ipfs/interface-ipfs-core/pull/63 in a way that they pass 100% in js-ipfs-api. Does this make sense to you? Happy to go more around it :)

pgte:
@hhff in the next release of js-ipfs you should be able to use `ipfs._repo.datastore`, which provides a [datastore interface](https://github.com/ipfs/interface-datastore#api). What were you looking for?

hhff:
@pgte - mostly just researching what changes this PR is going to need now that datastore is merged.  Happy to help but there's a lot going on here, hoping for some guidance!

pgte:
@hhff this is a very long PR and I'm not familiarised with pinning, so I'm not sure how I can help you.

I suggest that you take a look at [the instance API](https://github.com/ipfs/js-ipfs-repo/tree/master#repos), which I believe is externally accessible via `ipfs._repo`.

Please tell me if you have any question in what regards the internals of that API.

AdamStone:
@diasdavid I'm looking at this again, to see if I can get it working with the new datastore. I'm finding that the CLI "direct to core" tests `init` the repo, but [don't `start` it](https://github.com/ipfs/js-ipfs/blob/master/src/cli/commands/init.js#L35). This leaves the datastore "closed" during the pin tests which throws an error if I try to access it. So it seems like one of the following should happen:

- pin operations should attempt to open the datastore if it's closed
- pin operations should skip reading/writing to datastore if it's closed
- CLI init should also `start` the repo or at least `open` it

Do you know if any of these is correct, or is there something else I'm missing?

diasdavid:
`start` doesn't stand to start the repo, it is for starting the node (bitswap, dht, libp2p and all network stuff)

`init` is what creates the repo, but it doesn't lock onto it as it is a single run.

When you call some other command, it will check if there is a daemon running, if there is then uses the daemon http api, if there is not, then it spawns an in proc node and locks into the repo and therefore opening the datastore. See: https://github.com/ipfs/js-ipfs/blob/master/src/cli/utils.js#L37-L66

Before you advance more, it might be good to rebase master onto this branch, seems that there are multiple merge conflicts already.

AdamStone:
@diasdavid Thanks, I was thinking `start` was where the repo was opened based on the line I linked above, but I see now it can also be opened in [boot](https://github.com/ipfs/js-ipfs/blob/master/src/core/boot.js#L30) and [init](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/init.js#L65). In that case, since both boot and init seem to run before the CLI tests start, I would think the repo should be open at that point. But within the pin code I'm finding `self._repo.closed === true` when running CLI tests. So I still don't understand why.

I did the rebase locally, I can clean up my changes and push what I have so far and then focus on getting the tests passing afterward.

diasdavid:
> But within the pin code I'm finding self._repo.closed === true when running CLI tests. So I still don't understand why.

That sounds like a bug. 

Could you push your rebase so that I test locally the same code that you are testing?

AdamStone:
@diasdavid The repo being closed seems to be a timing problem. At the beginning of the `pin add` call, it is open as expected. By the time I try to access the datastore, within a callback, it's been closed by [this line in bin.js](https://github.com/ipfs/js-ipfs/blob/master/src/cli/bin.js#L65).

My `src/cli/commands/pin/add.js` handler currently looks like this:
```
  handler (argv) {
    const paths = argv['ipfs-path'].split(' ')
    const recursive = argv.recursive
    argv.ipfs.pin.add(paths[0], { recursive }, (err, results) => {
      if (err) { throw err }
      results.forEach((res) => {
        console.log(`pinned ${res.hash} ${res.type}ly`)
      })
    })
  }
```

Since there's no `callback` or `done` here, it looks like the handler just calls `pin.add` and returns, at which point the repo is cleaned up, even though the `add` is still doing async things. Is there some way I can make this yargs framework delay the `cleanup` until the `pin.add` callback actually finishes? (I tried returning the pin.add promise here instead of passing a callback, but it doesn't seem to make a difference)

diasdavid:
> Since there's no callback or done here, it looks like the handler just calls pin.add and returns, 

Got it. Why should the pin.add return before its operation has been completed? pin.add should only return once it finishes all the work it has to do

AdamStone:
I rebased again and pushed what I have so far, with other tests commented out so `npm run test:unit:node:cli` just runs the pin core and cli tests.

I used the same pattern for pin.add as the other core modules, meaning a function that takes a callback, wrapped in a `promisify`. The callback is only called when all the work is done. But when invoked by the CLI in the handler above, I don't see any straightforward way to make it wait for the callback. I did find [this](https://github.com/yargs/yargs/issues/918) which brings up the same issue and offers a hacky workaround.

package.json
```diff
@@ -94,11 +94,13 @@
     "boom": "^5.2.0",
     "cids": "^0.5.1",
     "debug": "^3.0.0",
+    "fnv1a": "^1.0.1",
     "fsm-event": "^2.1.0",
     "glob": "^7.1.2",
     "hapi": "^16.5.2",
     "hapi-set-header": "^1.0.2",
     "hoek": "^4.2.0",
+    "interface-datastore": "^0.3.0",
```
diasdavid: Was this necessary?

package.json
```diff
@@ -94,11 +94,13 @@
     "boom": "^5.2.0",
     "cids": "^0.5.1",
     "debug": "^3.0.0",
+    "fnv1a": "^1.0.1",
```
diasdavid: What is this used for?

diasdavid:
> Or I suppose I could wrap the whole call in a while loop and only break when the callback finishes. This is really strange as Node should continue running until the process is called to exist (or the event loop is empty)

src/http-api/resources/index.js
```diff
@@ -5,6 +5,7 @@ exports.id = require('./id')
 exports.bootstrap = require('./bootstrap')
 exports.repo = require('./repo')
 exports.object = require('./object')
+//exports.pin = require('./pin')
```
diasdavid: This shouldn't be commented out

src/http-api/resources/pin.js
```diff
@@ -0,0 +1,88 @@
...
+    }).code(500)
+  }
+  ipfs.pin.ls(path, { type })
+    .then((result) => {
+      return reply(result)
+    })
+    .catch((err) => {
+      return onError(err)
+    })
+}
```
diasdavid: We default to use callbacks rather then promises everywhere in the code to make things consistent. The Promises API is just for consumers.

test/core/bootstrap.spec.js
```diff
@@ -1,102 +1,102 @@
...
+//      expect(err).to.not.exist()
+//      expect(res).to.be.eql({Peers: ['/ip4/111.111.111.111/tcp/1001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLUVIT']})
+//      node.bootstrap.list((err, list) => {
+//        expect(err).to.not.exist()
+//        expect(list.Peers).to.deep.equal(defaultList)
+//        done()
+//      })
+//    })
+//  })
+//})
```
diasdavid: To run specific tests you can add a `.only` to the set of tests you want to run

diasdavid: example:

`describe.only('tests')` or `it.only('some test')`

test/http-api/test-pin.js
```diff
@@ -0,0 +1,155 @@
...
+            method: 'POST',
+            url: `/api/v0/pin/rm?arg=${keys.root}`
+          }, (res) => {
+            expect(res.statusCode).to.equal(200)
+            expect(res.result[0]).to.equal(keys.root)
+            done()
+          })
+        })
+      })
+    })
```
diasdavid: I see the tests here for the pin routes, yet the endpoint loading is commented out.

test/utils/on-and-off.js
```diff
@@ -64,7 +64,7 @@ function on (tests) {
  */
 exports = module.exports = (tests) => {
   off(tests)
-  on(tests)
+//  on(tests)
```
diasdavid: On the pin tests you can select just off by:

```
const runOff = require('../utils/on-and-off').off
describe('pin', () => runOff((thing) => {
```

See bitswap for a only On example: https://github.com/ipfs/js-ipfs/blob/master/test/cli/bitswap.js#L5-L7

test/cli/pin.js
```diff
@@ -0,0 +1,148 @@
...
+const filenames = [ 'root', 'leaf', 'branch', 'subLeaf' ]
+
+const keys = {
+  root: 'QmWQwS2Xh1SFGMPzUVYQ52b7RC7fTfiaPHm3ZyTRZuHmer',
+  leaf: 'QmaZoTQ6wFe7EtvaePBUeXavfeRqCAq3RUMomFxBpZLrLA',
+  branch: 'QmNxjjP7dtx6pzxWGBRCrgmjX3JqKL7uF2Kjx7ExiZDbSB',
+  subLeaf: 'QmUzzznkyQL7FjjBztG3D1tTjBuxeArLceDZnuSowUggXL'
+}
+
+describe('pin', () => runOnAndOff((thing) => {
```
diasdavid: If you want to test only "offline" do `runOnAndOff.off`

If you want to run just these tests, do: `describe.only`

test/cli/pin.js
```diff
@@ -0,0 +1,148 @@
...
+        done()
+      })
+  })
+
+  it('add (recursively by default)', (done) => {
+    console.log('cli pin add starting')
+    ipfs(`pin add ${keys.root}`).then((out) => {
+      console.log('cli pin add callback')
+      console.log(out)
+//      expect(out).to.eql(`pinned ${keys.root} recursively\n`)
```
diasdavid: So it seems that it is returning earlier (before actually running the pins) correct?

Sounds like a race condition. Are you successful running pin tests in core? If running pin tests in core is successful, then the issue must be in the cli/http-api endpoint implementation.

src/core/components/pin.js
```diff
@@ -0,0 +1,589 @@
...
+      if (Buffer.isBuffer(hashes) || typeof hashes.forEach !== 'function') {
+        hashes = [hashes]
+      }
+      const normalized = {
+        hashes: [],
+        update: (multihash) => {
+          multihashes.validate(multihash)
+          normalized.hashes.push(multihash)
+          if (hashes.length === normalized.hashes.length) {
+            return callback(null, normalized.hashes)
```
diasdavid: You shouldn't do a callback within a try statement, it means that any error that gets thrown after the callback will be captured by this try catch block

src/core/components/pin.js
```diff
@@ -0,0 +1,589 @@
...
+          }
+        }
+      }
+      hashes.forEach((hash) => {
+        if (typeof hash === 'string') {
+          // example: '/ipfs/QmRootHash/links/by/name'
+          const matched = hash.match(/^(?:\/ipfs\/)?([^\/]+(?:\/[^\/]+)*)\/?$/)
+          if (!matched) {
+            return callback(new Error('invalid ipfs ref path'))
```
diasdavid: this is calling a callback for each hash that is incorrect. Consider using async.each to test all the hashes and then return in case of error: https://caolan.github.io/async/docs.html#each

src/core/components/pin.js
```diff
@@ -0,0 +1,589 @@
...
+
+    add: promisify((hashes, options, callback) => {
+      // callback (err, pinset)
+      let recursive = true  // default
+      if (typeof options === 'function') {
+        callback = options
+      }
+      else if (options && options.recursive === false) {
+        recursive = false
+      }
```
diasdavid: Line 157 to 163 can be replaced by

`const recursive = options.recursive === false ? false : true`

src/core/components/pin.js
```diff
@@ -0,0 +1,589 @@
...
+              directPins.add(multihash)
+              result.update({
+                hash: keyString(multihash),
+                type: pin.types.direct
+              })
+            })
+          }
+        })
+      })
+    }),
```
diasdavid: I think I know the crux of the problem, it is the mix between callbacks and promises and lack of calling the callbacks only once.

I believe if you change your use of promises to callbacks in the CLI code and HTTP API you will see how your code doesn't finish.

src/core/components/pin.js
```diff
@@ -0,0 +1,589 @@
...
+                    // skip for now
+                    console.log('repo was closed within add callbacks')
+                    return callback(null, root)
+                  }
+                })
+              })
+            })
+          })
+        })
+      })
```
diasdavid: Please use async.series or async.waterfall for this code

https://caolan.github.io/async/docs.html#

src/core/components/pin.js
```diff
@@ -0,0 +1,589 @@
...
+                  }
+                )
+              }
+            )
+
+          })
+        })
+      })
+    })
+  }
```
diasdavid: Same here, async series or async waterfall https://caolan.github.io/async/docs.html#

src/core/components/pin.js
dignifiedquire: or `recursive = options.recursive !== false`

AdamStone:
@diasdavid Thanks for going through in detail, I'll look into these changes and reply inline.

package.json
AdamStone: This is the hash function used by the [go implementation](https://github.com/ipfs/go-ipfs/blob/master/pin/set.go#L31) in serializing the pin set. I just tried to replicate the same process as closely as I could.

package.json
AdamStone: To `get` or `put` in the datastore, the first argument should be a datastore Key object. Based on the example [here](https://github.com/ipfs/interface-datastore) it seems like I need to import this like `const Key = require('interface-datastore').Key`

src/http-api/resources/index.js
AdamStone: I haven't tried to update the http tests yet so I left it commented. Several other tests are also temporarily commented as I didn't see a better way to get just the pin tests to run without the rest.

test/core/bootstrap.spec.js
AdamStone: Ah, thanks, this is what I needed.

test/cli/pin.js
AdamStone: I added timestamps to these logs to make sure the sequence was clear. Here's what I'm seeing
```
   42.851 : pin add test starts (test/cli/pin.js)
   43.875 : ipfs.add called with callback (cli/commands/pin/add.js)
  *43.877 : cleanup called, which closes repo (cli/bin.js)
   44.008 : repo is closed when accessing datastore (core/components/pin.js)
   44.008 : ipfs.add callback ran (cli/commands/pin/add.js)
   44.015 : test callback ran (test/cli/pin.js)
```
It still looks to me like the timing/callbacks themselves are working within the tests and the pin code, but cleanup being called immediately rather than waiting for the callbacks finish. 

I'll focus on making the suggested changes first and then see if this is still a problem at that point.

src/core/components/pin.js
AdamStone: It may be, I'll look at incorporating these changes and see if it helps.

test/cli/pin.js
AdamStone: For the core tests, they pass locally for me if I make the changes discussed [here](https://github.com/ipfs/interface-ipfs-core/pull/63).

src/core/components/pin.js
mitra42: I've noticed the mix of callbacks and promises, in particular while a bunch of the libraries seem to have been updated to promises, for some reason Block is still callback, which means it has to be "promisified".

AdamStone:
@diasdavid I cleaned up the commented tests and tried to address the issues you indicated above. There are still several remaining issues:

**Interface/Core tests**
Fixes are still pending in [this PR](https://github.com/ipfs/interface-ipfs-core/pull/63). Locally they're passing if I make those changes.

**Http API tests**
Passing, but only if I use a [different datastore key](https://github.com/ipfs/js-ipfs/pull/107/files#diff-d96275251fef125ec785643a9facfab4R52) than the go repo. The tests load a datastore from `tests/go-ipfs-repo` which throws an error ("Groups are not supported") if I try to deserialize the existing pin set using the usual key. I think this is a problem with the test repo rather than the pin.load function, but it needs further investigation.

**CLI tests**
"On" tests are failing - for some reason `results` in the [callback within the handler functions](https://github.com/ipfs/js-ipfs/pull/107/files#diff-eca0da567e56ac217ce95c81b76ae6a0R20) is always empty, specifically just in the "on" tests, even though there's no error passed either. I suspect some error is happening within the pin code that's somehow failing silently, but maybe I'm misunderstanding something about the way the CLI-daemon-core interaction is supposed to work.

"Off" tests - There's still a timing problem where the repo is closed while the callbacks are still processing. I can get them to pass, but only if I [detect when the repo is closed and manually open the datastore](https://github.com/ipfs/js-ipfs/pull/107/files#diff-d96275251fef125ec785643a9facfab4R522) within the pinning code. It seems like this is a more general problem than something about my implementation, because I tried putting timestamp log statements in the `object.put` test path and saw a similar result:

```
  45.891  starting put
  45.892  calling parseBuffer
* 45.893  calling cleanup in bin.js
  45.895  calling ipldResolver.put
  45.924  calling object.get
  45.926  put callback
```
Not sure what to do about this.

src/core/components/pin.js
```diff
@@ -66,11 +66,10 @@ module.exports = function pin (self) {
       update: (multihash, cb) => {
         try {
           multihashes.validate(multihash)
-          normalized.hashes.push(multihash)
-          cb()
-        } catch (ex) {
-          cb(ex)
-        }
+        } catch (ex) { return cb(ex) }
```
diasdavid: `s/ex/err` so that any standard linter catches if this exception is not handled.

AdamStone: Sorry, can you explain what you mean by `s/ex/err`?

diasdavid: Ah, sorry, was using [`sed`](http://www.grymoire.com/Unix/Sed.html) notation.

I meant to replace the word `ex` by the word `err`, so that the [Standard JavaScript Linter](https://standardjs.com/) recognizes it as an error that needs to be taken care of.

AdamStone: Ah, ok, no problem.

diasdavid:
> Http API tests
Passing, but only if I use a different datastore key than the go repo. The tests load a datastore from tests/go-ipfs-repo which throws an error ("Groups are not supported") if I try to deserialize the existing pin set using the usual key. I think this is a problem with the test repo rather than the pin.load function, but it needs further investigation.

I think your diagnostic is correct, the test repo is from a old go version and go-ipfs as since migrated repos a couple of times. You can use the tool https://github.com/ipfs/fs-repo-migrations to migrate the test repo.

> "On" tests are failing because they go through js-ipfs-api, and the pin methods implemented there are anticipating results in a slightly different format than my implementation is returning. 

Both js-ipfs and js-ipfs-api must expose exactly the same interface that must pass exactly the same compliance tests (the ones in interface-ipfs-core). The Pinning API hasn't been set in stone and so you have more freedom and less friction to update the API. You can do that by creating the PRs necessary to update spec, tests, implementation in js-ipfs-api and complete this one.

> It seems like this is a more general problem than something about my implementation

I will review this again and in depth. It would help if the remaining (above) is finished so that I get to debug the final code.

package.json
diasdavid: Sounds good. I actually don't know why is this specific hashing function used, but that is something probably @Kubuxu can fill in?

package.json
diasdavid: Got it.

@dignifiedquire is there any reason why the .get and .put couldn't take a string or a buffer? I understand that `Key` has some utility methods, but the cast could happen internally as well.

src/cli/commands/pin/ls.js
```diff
@@ -0,0 +1,46 @@
...
+          let line = item.hash
+          if (!quiet) {
+            line += ` ${item.type}`
+            if (item.through) {
+              line += ` through ${item.through}`
+            }
+          }
+          console.log(line)
+        })
+      })
```
diasdavid: Either `.catch` a possible error or use the callback API so that you don't forget :)

src/core/components/files.js
```diff
@@ -119,6 +119,7 @@ function prepareFile (self, file, callback) {
 
   waterfall([
     (cb) => self.object.get(file.multihash, cb),
+    (node, cb) => self.pin.add(file.multihash, cb),
```
diasdavid: Why does pin happen here?

src/core/components/pin-set.js
```diff
@@ -0,0 +1,275 @@
...
+  // from go-ipfs/pin/internal/pb/header.proto
+  'message Set { ' +
+    // 1 for now
+    'optional uint32 version = 1; ' +
+    // how many of the links are subtrees
+    'optional uint32 fanout = 2; ' +
+    // hash seed for subtree selection, a random number
+    'optional fixed32 seed = 3; ' +
+  '}'
+)
```
diasdavid: Factor this into its own file

src/core/components/pin-set.js
```diff
@@ -0,0 +1,275 @@
...
+      }
+      root.links.forEach((link) => {
+        const bs58link = multihashes.toB58String(link.multihash)
+        if (bs58link === childhash) {
+          callback.fired = true
+          return callback(null, true)
+        }
+        dag.get(new CID(link.multihash), (err, res) => {
+          if (err) {
+            callback.fired = true
```
diasdavid: Where is this callback.fired used? You can always use https://www.npmjs.com/package/once to ensure it only gets called once.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,275 @@
...
+          }
+        }
+      }
+      if (!subwalkCount) {
+        return callback()
+      }
+    }
+  }
+  return pinSet
+}
```
diasdavid: There is a ton of good work! 👌🏽

At the current moment, seems a bit complex, might be good to factor it out into separate files and individually test each.

src/core/components/pin.js
```diff
@@ -0,0 +1,563 @@
...
+    return Object.keys(this.keys).map((hash) => {
+      return this.keys[hash]
+    })
+  }
+  this.toStringArray = () => {
+    return Object.keys(this.keys)
+  }
+  keys = keys || []
+  keys.forEach(this.add)
+}
```
diasdavid: Also, factor KeySet into its own file and test it individually

src/core/components/pin.js
```diff
@@ -0,0 +1,563 @@
...
+        cb()
+      }
+    }
+    each(hashes, (hash, cb) => {
+      if (typeof hash === 'string') {
+        // example: '/ipfs/QmRootHash/links/by/name'
+        const matched = hash.match(/^(?:\/ipfs\/)?([^/]+(?:\/[^/]+)*)\/?$/)
+        if (!matched) {
+          return cb(new Error('invalid ipfs ref path'))
+        }
```
diasdavid: You can use http://npmjs.org/is-ipfs for this one.

src/core/components/pin.js
```diff
@@ -0,0 +1,563 @@
...
+      const newInternalPins = new KeySet()
+      const logInternalKey = (mh) => newInternalPins.add(mh)
+      const handle = {
+        put: (k, v, cb) => {
+          handle[k] = v
+          cb()
+        }
+      }
+      waterfall([
+        (cb) => repo.closed ? repo.datastore.open(cb) : cb(null, null),  // temp
```
diasdavid: What's this step doing with `null, null`?

package.json
Kubuxu: The set serialization structure is very old. I was not here when it was designed, in future we plan migrating to same structure used buy directory sharding or whatever sharded array/map/set the IPLD provides.

From what I know it is same structure that is used by some DB software. FNV1a is generic fast hash function for sharding in this case.

src/core/components/pin.js
```diff
@@ -0,0 +1,563 @@
...
+        (_, cb) => repo.datastore.has(pinDataStoreKey, cb),
+        (has, cb) => has ? cb() : cb('break'),
```
diasdavid: If the first element is passed, it is assumed that is an error, if it is an error, it should be an Error instance.

src/core/components/pin.js
```diff
@@ -0,0 +1,563 @@
...
+        (cb) => repo.datastore.get(pinDataStoreKey, cb),
+        (serialized, cb) => dagPB.util.deserialize(serialized, cb),
```
diasdavid: The PinSet object is a dag Protobuf? I believe this should be `go-ipfs/pin/internal/pb/header.proto`

diasdavid:
@AdamStone quick status update?

AdamStone:
@diasdavid I had family visiting over the last few days so I'm just looking at your latest comments now. I'm still sort of unclear on which API's result format I should adhere to.

The main website [HTTP API reference](https://ipfs.io/docs/api/) says the following

> This is the HTTP API specification for IPFS.
> 
> IPFS HTTP API is an RPC API which **should work across different IPFS implementations**. The most feature-complete of those implementations, and current reference for this specification is go-ipfs.

`js-ipfs-api` is identified as "A client library for the IPFS HTTP API, implemented in JavaScript." So it seems correct that it should follow the spec above, as the current implementation does. But it also states "This client library implements the interface-ipfs-core" which has a [different API](https://github.com/ipfs/interface-ipfs-core#api).

From your last comment it sounds like I should update js-ipfs-api to match the interface-ipfs-core API, but this would mean it's no longer consistent with the go HTTP API spec, are you sure that's what we want? Or maybe I misunderstood?

Sorry to belabor the point.

src/core/components/pin.js
AdamStone: I was also using the regex to capture and split the part of the path after "ipfs/" to get the individual links in the path, is there an import I can use for that?

src/core/components/pin.js
AdamStone: Just trying to do a no-op that passes consistent arguments to the next step as the case where it opens the repo. There's probably a better way to do that... In any case, this line was a temporary workaround for the repo being closed, I assume the final implementation shouldn't try to open the datastore here?

src/core/components/pin.js
AdamStone: I wanted to short-circuit the rest of the waterfall and just skip to the end here, is there a good way to do that? It's not really an error case, but the error path seemed like the only simple way to do that.

diasdavid:
Hi @AdamStone, shame on us for making it sound so complicated.

There is the two API as you said, the [HTTP](https://github.com/ipfs/http-api-docs) and the [CORE](https://github.com/ipfs/interface-ipfs-core) API, the first gets implemented by a running daemon, the second is implemented at the instance level for js-ipfs and js-ipfs-api.

Users of js-ipfs and js-ipfs-api only see the CORE API, but js-ipfs-api uses the HTTP-API internally to contact the daemon.

Did this help make things more clear?

AdamStone:
@diasdavid Thanks, yes, I think that helps. So js-ipfs-api and the core section of js-ipfs should both follow the CORE API, and I suppose the http-api section of js-ipfs should follow the HTTP API. I think that should resolve the issue with the CLI on-and-off tests getting different result formats. I'll take a look tonight.

AdamStone:
@diasdavid I guess I spoke too soon about the CLI on and off tests. I was thinking the "on" tests were just going through js-ipfs-api, but it looks like they also go through the http-api layer of js-ipfs. So even if I make js-ipfs-api follow the CORE spec, if I also make the js-ipfs http-api results follow the HTTP spec, the "on" tests still get a different result format than the "off" tests. So... should the http-api section of js-ipfs also follow the CORE spec? Or should the CLI tests try to accommodate both formats, or split the "on" and "off" parts into separate 'describes'?

diasdavid:
@AdamStone to demystify:

- https://github.com/ipfs/js-ipfs/tree/master/src/core exposes the interface-ipfs-core 
- https://github.com/ipfs/js-ipfs-api exposes the interface-ipfs-core interface, internally talks with a js-ipfs or go-ipfs node using http. It is a client library, see this illustration https://github.com/ipfs/js-ipfs-api#--
- https://github.com/ipfs/js-ipfs/tree/master/src/http/api implements the http-api, which clients like __js-ipfs-api__ can use to interact with a node remotely
- CLI offline talks directly with js-ipfs core
- CLI online talks with js-ipfs http-api, using js-ipfs-api the client. See this illustration https://github.com/ipfs/js-ipfs#code-architecture-and-folder-structure

Does this help? Happy to jump on a call if you feel some sync minutes would help :)

AdamStone:
@diasdavid Sorry to be so dense with this, I appreciate the offer to get on a call about it. I found the source of my confusion, though. From what I was seeing, it looked like the result I returned in http-api was getting passed directly into the CLI handler without going through js-ipfs-api. The problem was that I had linked js-ipfs-api with interface-ipfs-core and js-ipfs with interface-ipfs-core, but failed to link js-ipfs with js-ipfs-api. Once I did that, with my changes to js-ipfs-api I am able to get a consistent result format in the CLI handlers. So hopefully that will be the last of these kinds of questions.

diasdavid:
No worries @AdamStone :) 

>  Once I did that, with my changes to js-ipfs-api I am able to get a consistent result format in the CLI handlers. So hopefully that will be the last of these kinds of questions.

Exciting! Let me know if I can help more! :)

AdamStone:
@diasdavid After forking the latest js-ipfs-api, its tests seem to bypass my link to interface-ipfs-core. That is to say, when I do `npm test` it's not running against my changes there, even though when I check node_modules I see the interface-ipfs-core folder does link to my local repo. Actually even if I delete interface-ipfs-core entirely from node_modules, it doesn't seem to matter, it still runs the same tests. Another case of [the test architecture changing underneath me](https://github.com/ipfs/js-ipfs-api/commit/9ffe84803733eeab82cca2b8bec96c2ba1021d45)? Any ideas?

diasdavid:
All these years using npm have taught me that when in doubt, start from the top.

```
> rm -r all the node_modules
> rm all the package-lock.json if you have any (in case of using npm 5)
> cd interface-ipfs-core
> npm install && npm link
> cd js-ipfs-api
> npm install && npm link interface-ipfs-core
> npm test # tests should pass
> cd js-ipfs
> npm install && npm link interface-ipfs-core && npm link ipfs-api
> npm test # tests should pass
```

Let me know if this works for you.

AdamStone:
Still not sure why, but even with those steps I just couldn't get the command-line linking to get all the links right. Eventually I just went through and copied the links manually into the node_modules folders, and that seems to work. Although the js-ipfs-api tests are still running a bunch of other tests in addition to what I mark with `describe.only`. Still need to make some of the other changes, then I'll push an update.

src/core/components/files.js
AdamStone: @diasdavid I saw that adding a file should also pin it, I just put it there to get something working so I could debug the tests, but that's probably not the right place to do it... Open to suggestion on this.

AdamStone:
Migrating the go test repo from v5 to v6 breaks http tests where something expects to find v5. I couldn't tell from this stack trace where that expectation is coming from.
image

diasdavid:
It was announced [yesterday on all hands](https://github.com/ipfs/pm/issues/501) that the IPFS Repo had been updated. It was an update also for me.

That said, the file structure didn't change. You just need to update the test to look for v6 :)

AdamStone:
@diasdavid I rebased against master just now, but I still see the go test repo as version 5. I also haven't been able to find where in the tests or configuration it's deciding that it expects to find a particular version. So I'm not seeing what it would mean exactly to update the test to look for v6?

AdamStone:
I rebased what I have so far against current master here and in these related PRs for [interface-ipfs-core](https://github.com/ipfs/interface-ipfs-core/pull/63) and [js-ifps-api](https://github.com/ipfs/js-ipfs-api/pull/602). With these repos linked locally and with the changes in these PRs, all the pin tests are locally passing for me. However, there are placeholder workarounds in the code for two main blocking issues:

1. The pin [datastore key](https://github.com/ipfs/js-ipfs/pull/107/files#diff-d96275251fef125ec785643a9facfab4R27) is `/local/pins/js` because using the key from go-ifps `'/local/pins` causes the tests to try to load pins from go-ipfs-repo, which doesn't work. Still unclear if the repo just needs to be migrated, since trying to do so causes version mismatch errors.
2. The datastore is closed during the CLI core tests when trying to [save](https://github.com/ipfs/js-ipfs/pull/107/files#diff-d96275251fef125ec785643a9facfab4R487) and [load](https://github.com/ipfs/js-ipfs/pull/107/files#diff-d96275251fef125ec785643a9facfab4R507) pins due to some timing problem - the tests only pass if I manually open the datastore within the pin code.

@diasdavid I hate to leave this PR hanging again after getting this far, I was hoping this would be the final push to get all this wrapped up while I had the time to spend on it, but sorry to say I'm basically out of time now and I don't expect I'll be able to take this any further. Hopefully someone else can find it a useful starting point.

PR #108, opened 2016-04-03
Fixed issues with some required modules

JGAntunes:
While trying to install and run ipfs I ran into some troubles, this should fix some of it.


JGAntunes:
It also seems to me that the current build is failing due to this - https://github.com/ipfs/js-ipfs-unixfs/pull/8

xicombd:
LGTM

src/cli/commands/config/replace.js
```diff
@@ -3,7 +3,7 @@ const debug = require('debug')
 const path = require('path')
 const log = debug('cli:config')
 log.error = debug('cli:config:error')
-const utils = require('utils')
+const utils = require('../../utils')
```
diasdavid: how was this not failing before? Could you add a test too @JGAntunes ? Thank you!

//cc @xicombd

package.json
```diff
@@ -79,6 +78,7 @@
     "boom": "^3.1.1",
     "bs58": "^3.0.0",
     "debug": "^2.2.0",
+    "fs-blob-store": "^5.2.1",
```
diasdavid: good catch!

diasdavid:
build is green now (for this branch as well), let me know if it is working for you.

src/cli/commands/config/replace.js
xicombd: No idea, I'll try to review the config commands later today, I think they're not working with the daemon running either :worried:

JGAntunes: @diasdavid sure thing, will look at it later today also, so maybe I can help you out @xicombd.

JGAntunes:
Seems good to me, I can now install and run the cli.

Issue #92, opened 2016-03-21
track: `jsipfs swarm`


diasdavid:
The swarm API enables IPFS users to learn about their connected peers and connect to other peers if desired
#### To learn more about the swarm API, you can:
- read about libp2p and how ipfs swarm features are implemented on top of libp2p - https://github.com/ipfs/specs/tree/master/libp2p
#### Implementation Roadmap
- [x] jsipfs daemon - turn own swarm when it boots up
- jsipfs swarm peers
  - [x] core
  - [x] http-api
  - [x] cli
- jsipfs swarm connect
  - [x] core
  - [x] http-api
  - [x] cli
- jsipfs swarm disconnect
  - [x] core
  - [x] http-api
  - [x] cli
- jsipfs swarm addrs
  - [x] core
  - [x] http-api
  - [x] cli
- jsipfs swarm addrs local
  - [x] core
  - [x] http-api
  - [x] cli

diasdavid:
follow on -> https://github.com/ipfs/js-ipfs/pull/101

diasdavid:
Update: Now we are only missing

- [ ] jsipfs swarm disconnect
- [ ] jsipfs swarm addrs
- [ ] jsipfs swarm addrs local

diasdavid:
//cc @VictorBjelkholm

diasdavid:
We are just missing the `disconnect` \o/, which is already implemented through the http-api and in core, so this should a very quick thing to implement.

https://github.com/ipfs/js-ipfs/blob/master/src/cli/commands/swarm/disconnect.js

diasdavid:
@nunofmn just PR'ed the last item 🎉

PR #151, opened 2016-04-17
Add swarm connect, peers and localAddrs

xicombd:
Tests seem to be failing on travis, still investigating why 😟

Targeting #101 and related to #92


diasdavid:
Thank you @xicombd. Seems fine on my machine too. Will merge this to the other branch and continue working from there.

PR #188, opened 2016-05-03
fix(cli): self host cmds listing

dignifiedquire:
Fixes Node.js 6 compat

This will make CircleCI green again (which is running stable = node 6)


diasdavid:
@dignifiedquire wanna report this to ronin and maybe( would be better to PR the fix for the ronin codebase? )

dignifiedquire:
@diasdavid as far as I can tell ronin is still not maintained, so we can do that in addition, but we need this fix here as well

PR #193, opened 2016-05-05
files add, cat, get core + cli offline

nginnever:


src/cli/commands/files/add.js
```diff
@@ -19,15 +24,72 @@ module.exports = Command.extend({
...
-      recursive: recursive
-    }, (err, stats) => {
+    let s
+    let rs
+
+    if (!path) {
+      throw new Error('Error: Argument \'path\' is required')
+    }
+
+    s = fs.statSync(path)
```
dignifiedquire: You could declare `s` here where it's needed instead of at the top

src/cli/commands/files/add.js
```diff
@@ -19,15 +24,72 @@ module.exports = Command.extend({
...
+          throw err
+        }
+        const i = ipfs.files.add()
+        i.on('data', (file) => {
+          console.log('added', bs58.encode(file.multihash).toString(), file.path)
+        })
+        if (res.length !== 0) {
+          const index = path.lastIndexOf('/')
+          async.eachLimit(res, 10, (element, callback) => {
+            rs = new Readable()
```
dignifiedquire: You can declare `rs` in here, no need for it at the outer scope

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,69 @@
...
+      if (err) {
+        throw err
+      }
+      ipfs.files.get(path, (err, data) => {
+        if (err) {
+          throw err
+        }
+        data.on('file', (data) => {
+          if (data.path.lastIndexOf('/') === -1) {
+            filepath = data.path
```
dignifiedquire: only needed in here, no need to be declared at the top level

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,69 @@
...
+            if (data.dir === false) {
+              ws = fs.createWriteStream(pathj.join(dir, data.path))
```
dignifiedquire: only needed in here, no need to be declared at the top level

src/core/index.js
```diff
@@ -403,10 +405,52 @@ function IPFS (repo) {
...
+      }
+
+      const i = new Importer(dagS)
+      const res = []
+
+      i.on('data', (info) => {
+        res.push(info)
+      })
+
+      i.on('end', () => {
```
dignifiedquire: use `once` to ensure `callback` is only called once

src/core/init.js
```diff
@@ -66,17 +70,48 @@ module.exports = (repo, opts, callback) => {
...
     const dag = new DagService(blocks)
 
     const initDocsPath = path.join(__dirname, '../init-files/init-docs')
 
-    importer.import(initDocsPath, dag, {
-      recursive: true
-    }, doneImport)
+    const i = new Importer(dag)
+    i.on('data', (file) => {
+    })
```
dignifiedquire: why this empty listener?

src/core/init.js
```diff
@@ -66,17 +70,48 @@ module.exports = (repo, opts, callback) => {
...
+
+    glob(path.join(initDocsPath, '/**/*'), (err, res) => {
+      if (err) {
+        throw err
+      }
+      const index = __dirname.lastIndexOf('/')
```
dignifiedquire: `__dirname` is this folder, are you sure that is what you want?

src/core/init.js
```diff
@@ -66,17 +70,48 @@ module.exports = (repo, opts, callback) => {
...
+      async.eachLimit(res, 10, (element, callback) => {
+        const addPath = element.substring(index + 1, element.length)
+        if (fs.statSync(element).isDirectory()) {
+          callback()
+        } else {
+          const buffered = fs.readFileSync(element)
+          var rs = new Readable()
```
dignifiedquire: let's use `const` here for consistency

src/core/init.js
```diff
@@ -66,17 +70,48 @@ module.exports = (repo, opts, callback) => {
...
+      }, (err) => {
+        if (err) {
+          throw err
+        }
+        i.end()
+        return
+      })
+    })
+
+    i.on('end', () => {
```
dignifiedquire: use `once` to ensure it's actually only called once

test/core-tests/test-files.js
```diff
@@ -0,0 +1,64 @@
...
+    ipfs = new IPFS(require('./repo-path'))
+    ipfs.load(done)
+  })
+
+  it('add', (done) => {
+    const buffered = new Buffer('some data')
+    var rs = new Readable()
+    rs.push(buffered)
+    rs.push(null)
+    var arr = []
```
dignifiedquire: `const`

test/core-tests/test-files.js
```diff
@@ -0,0 +1,64 @@
...
+  var ipfs
+
+  before((done) => {
+    ipfs = new IPFS(require('./repo-path'))
+    ipfs.load(done)
+  })
+
+  it('add', (done) => {
+    const buffered = new Buffer('some data')
+    var rs = new Readable()
```
dignifiedquire: `const`

test/core-tests/test-files.js
```diff
@@ -0,0 +1,64 @@
...
+
+const bl = require('bl')
+const expect = require('chai').expect
+const Readable = require('stream').Readable
+const bs58 = require('bs58')
+
+const IPFS = require('../../src/core')
+
+describe('files', () => {
+  var ipfs
```
dignifiedquire: `let`

src/core/init.js
nginnever: @noffle wrote that and made the note that it is not meant to run in the browser.

src/cli/commands/files/add.js
nginnever: 80:11  error    'rs' is not defined           no-undef
  83:11  error    'rs' is not defined           no-undef
  84:11  error    'rs' is not defined           no-undef
  85:49  error    'rs' is not defined

src/cli/commands/files/get.js
nginnever: linter:   
55:13  error  'filepath' used outside of binding context  block-scoped-var
57:44  error  'filepath' used outside of binding context  block-scoped-var

src/cli/commands/files/get.js
nginnever: 60:13  error  'ws' used outside of binding context        block-scoped-var
  61:30  error  'ws' used outside of binding context        block-scoped-var

src/core/index.js
nginnever: not sure what you mean by 'once'

test/core-tests/test-files.js
nginnever: how is an array that is going to be populated labelled a constant? usually languages don't offer dynamic array allocation... javascripts

test/core-tests/test-files.js
nginnever: this was borrowed from other tests as well

src/core/init.js
nginnever: not sure but it is timing out without it

src/cli/commands/files/add.js
dignifiedquire: yeah you will have to define it again on line 83 here, remember let and const are block scoped

src/cli/commands/files/get.js
dignifiedquire: one scope up sorry,

src/core/index.js
dignifiedquire: `.once` is like `.on` but it removes the listener after the first event

src/core/init.js
dignifiedquire: try `i.resume()` instead

src/core/init.js
dignifiedquire: it's still odd that it's checking the name of the folder the code is in

test/core-tests/test-files.js
dignifiedquire: `const` means immutable reference, not immutable object in JavaScript. This means the array is still very much mutable, but the you can not assign a different array to the identifier `arr`.

src/core/init.js
nginnever: i.resume() works! ty

src/core/index.js
nginnever: ahh i was having a hard time finding relevant once functions for event listeners, i see now

src/cli/commands/files/add.js
nginnever: doing that will give a lint error about defining twice

src/cli/commands/files/add.js
```diff
@@ -19,15 +24,72 @@ module.exports = Command.extend({
...
+    }
+    if (path === '.' && recursive === true) {
+      path = process.cwd()
+      s = fs.statSync(process.cwd())
+    } else if (path === '.' && recursive === false) {
+      s = fs.statSync(process.cwd())
+      if (s.isDirectory()) {
+        throw new Error('Error: ' + process.cwd() + ' is a directory, use the \'-r\' flag to specify directories')
+      }
+    }
```
diasdavid: What is the goal of this chunk of code? Can it be extracted to a function with a good name? At first glance seems madness to me :D

src/cli/commands/files/add.js
```diff
@@ -1,11 +1,16 @@
...
-const IPFS = require('../../../core')
+const utils = require('../../utils')
 const debug = require('debug')
 const log = debug('cli:version')
 log.error = debug('cli:version:error')
 const bs58 = require('bs58')
+const Readable = require('stream').Readable
+const fs = require('fs')
+const async = require('async')
+const pathj = require('path')
```
diasdavid: use `path` for the name of the variable that contains the module, and use another keyword for the argument path. So that we keep the usage of the module consistent

src/cli/commands/files/add.js
```diff
@@ -19,15 +24,72 @@ module.exports = Command.extend({
...
       }
-      console.log('added', bs58.encode(stats.Hash).toString(), stats.Name)
+      utils.getIPFS((err, ipfs) => {
+        if (err) {
+          throw err
+        }
+        const i = ipfs.files.add()
+        i.on('data', (file) => {
+          console.log('added', bs58.encode(file.multihash).toString(), file.path)
+        })
```
diasdavid: You never listen for the `end` event, how can you tell it is complete?

src/cli/commands/files/add.js
```diff
@@ -19,15 +24,72 @@ module.exports = Command.extend({
...
+              callback()
+            }
+          }, (err) => {
+            if (err) {
+              throw err
+            }
+            i.end()
+          })
+        } else {
+          rs = new Readable()
```
diasdavid: these 'forced' reuse of variables feels weird. A variable named should be reused if several points in the code will look at it and react to its change. If you instantiate a new object, but then have code that treats that object specifically, no need to reuse variables. (remember that in JS, each time you reassign a variable, you give V8 an headache :) )

src/cli/commands/files/add.js
```diff
@@ -19,15 +24,72 @@ module.exports = Command.extend({
...
+          const buffered = fs.readFileSync(path)
```
diasdavid: Why reading the file all together and shuve it on a readable stream, when you can `fs.createReadStream` and leverage the backpressured offered by the OS?

src/cli/commands/files/cat.js
```diff
@@ -0,0 +1,37 @@
...
+      }
+      ipfs.files.cat(path, (err, res) => {
+        if (err) {
+          throw (err)
+        }
+        if (res) {
+          res.on('file', (data) => {
+            data.stream.pipe(process.stdout)
+          })
+        }
```
diasdavid: What if not `err` nor `res`?

src/cli/commands/files/add.js
nginnever: It's to check for all of the inputs you could give to add command. variations i could think of are...

1) "." add the cwd but throw error for no recursion flag
2) "." -r add the cwd
3) "/some/path" but throw error for no recursion
4) "/some/path" -r 
5) No path, throw err
6) filename.type

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,69 @@
...
+      throw new Error("Argument 'path' is required")
+    }
+    if (!options) {
+      options = {}
+      dir = process.cwd()
+    } else {
+      if (options.slice(-1) !== '/') {
+        options += '/'
+      }
+      dir = options
```
diasdavid: if `options` is the outputPath, shouldn't it be called that?

src/cli/commands/files/add.js
nginnever: What would I need to do when it is complete?

src/core/index.js
```diff
@@ -403,10 +405,52 @@ function IPFS (repo) {
...
+      i.end()
+    },
+    cat: (hash, callback) => {
+      dagS.get(hash, (err, fetchedNode) => {
+        if (err) {
+          return callback(err, null)
+        }
+        const data = UnixFS.unmarshal(fetchedNode.data)
+        if (data.type === 'directory') {
+          callback('This dag node is a directory', null)
```
diasdavid: no need for `null`

src/cli/commands/files/add.js
diasdavid: Creating a function and adding that as a comment describing what it is doing would be considerably much better.

src/cli/commands/files/add.js
diasdavid: The process should exit gracefully

diasdavid:
CR'ed 👍 left some comments that can be applied through the code. Cool that things are going :D

src/cli/commands/files/cat.js
nginnever: it will fail silently? not sure what you want me to do here

nginnever: are return statements the proper way to exit gracefully?

src/cli/commands/files/add.js
nginnever: I don't fully understand what you mean. Will try to ponder on this a bit. The decelerations come in different parts of the logic. Or there shouldn't be a time when the if statement is satisfied for both conditions meaning that it is necessary instantiate rs in both logic choices. Let me know if I am missing something here.

PR #197, opened 2016-05-06
files add, cat, get core + cli

nginnever:


src/cli/commands/files/add.js
```diff
@@ -1,11 +1,45 @@
...
+  }
+
+  var s = fs.statSync(inPath)
+
+  if (s.isDirectory() && recursive === false) {
+    throw new Error('Error: ' + process.cwd() + ' is a directory, use the \'-r\' flag to specify directories')
+  }
+  if (inPath === '.' && recursive === true) {
+    inPath = process.cwd()
+    return inPath
```
diasdavid: Is there a need for this return?

src/cli/commands/files/add.js
```diff
@@ -18,16 +52,53 @@ module.exports = Command.extend({
...
+              throw err
+            }
+            i.end()
+          })
+        } else {
+          rs = fs.createReadStream(inPath)
+          inPath = inPath.substring(inPath.lastIndexOf('/') + 1, inPath.length)
+          filePair = {path: inPath, stream: rs}
+          i.write(filePair)
+          i.end()
```
diasdavid: what is all of this doing? Wouldn't glob do this for free?

src/cli/commands/files/cat.js
```diff
@@ -0,0 +1,38 @@
...
+      ipfs.files.cat(path, (err, res) => {
+        if (err) {
+          throw (err)
+        }
+        if (res) {
+          res.on('file', (data) => {
+            data.stream.pipe(process.stdout)
+          })
+        }
+        return
```
diasdavid: What are the chances of `res` not existing, if no error was thrown? Why fail silently?

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,67 @@
...
+log.error = debug('cli:files:error')
+var fs = require('fs')
+const pathj = require('path')
+
+module.exports = Command.extend({
+  desc: 'Download IPFS objects',
+
+  run: (path, outPath) => {
+    let dir
+    let filepath
```
diasdavid: filePath

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,67 @@
...
+    let ws
+
+    if (!path) {
+      throw new Error("Argument 'path' is required")
+    }
+    if (!outPath) {
+      outPath = {}
+      dir = process.cwd()
+    } else {
+      if (outPath.slice(-1) !== '/') {
```
diasdavid: you do this often, maybe create a func called "isRootPath"

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,67 @@
...
+
+    utils.getIPFS((err, ipfs) => {
+      if (err) {
+        throw err
+      }
+      ipfs.files.get(path, (err, data) => {
+        if (err) {
+          throw err
+        }
+        data.on('file', (data) => {
```
diasdavid: This is confusing, maybe `result.on('file', (file)`

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,67 @@
...
+              } catch (err) {
+                throw err
+              }
+            }
+          } else {
+            filepath = data.path.substring(0, data.path.lastIndexOf('/') + 1)
+            try {
+              fs.mkdirSync(pathj.join(dir, filepath))
+            } catch (err) {
+              throw err
```
diasdavid: if you catch to throw, better don't even catch

src/cli/commands/files/cat.js
nginnever: I was wondering the same logic when you asked on the other PR :D https://github.com/ipfs/js-ipfs/pull/193#discussion-diff-62258034

I will revert to my original conclusion and remove the return

src/cli/commands/files/add.js
nginnever: glob returns the full path, so I have to trim it to so we don't add unnecessary directories

src/cli/commands/files/get.js
nginnever: good catch.. ty!

src/cli/commands/files/add.js
nginnever: inPath is not declared in this scope and needs to be returned on line 58, unless I'm missing something.

src/cli/commands/files/get.js
nginnever: this needs a try as Jeromy pointed out in NY  when we were trying to get this to work.. or if the directory already exists it will try to write the directory multiple times and error. Unless we are missing something.

src/cli/commands/files/get.js
nginnever: agreed

src/cli/commands/files/get.js
nginnever: I am not familiar with another mechanism for ensuring that i don' receive file already exist errors that will break the command.

```
error	EEXIST: file already exists, mkdir '/home/n4th4n/Documents/test/QmTHwJEAryXDVmNWZYRvJBBdMjza7ZEgbcpUVwcRY5TsUi/'
```
If there is a better way let me know

src/cli/commands/files/get.js
nginnever: I do this once, and it is an attempt to be like go-ipfs. It is checking to see if you supply a path with either / on the end or without. If you don' t supply a / then one needs to appended to the supplied path. It's just so things don't break if for example you do this 

```
jsipfs files get QmTHwJEAryXDVmNWZYRvJBBdMjza7ZEgbcpUVwcRY5TsUi /home/n4th4n/Documents/test
```
instead of one with a / at the end like 

```
jsipfs files get QmTHwJEAryXDVmNWZYRvJBBdMjza7ZEgbcpUVwcRY5TsUi /home/n4th4n/Documents/test/
```

dignifiedquire: You could use `outPath.startsWith('/')` [Ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith)

src/cli/commands/files/get.js
dignifiedquire: I don't understand, you don't check for `EEXIST` you `throw` all errors from `mkdirSync`.
Also you really should use the non blocking `mkdir`

src/cli/commands/files/get.js
nginnever: ```outPath.slice(-1)``` is checking the end of the a given path not the beginning for '/' 

I see that you can specify a start index for startsWith, but the outPath will be variable length so I don't see how this is useful.

src/cli/commands/files/get.js
nginnever: I was not throwing the errors, this was only to make sure that an error did not break the command. I started throwing the error because of another comment or the linter saying that you shouldn't have a try catch without handling the error.

I agree that I should switch async mkdir.

src/cli/commands/files/get.js
dignifiedquire: nvm I can't read code..

src/cli/commands/files/get.js
dignifiedquire: https://www.npmjs.com/package/path-exists

src/cli/commands/files/get.js
dignifiedquire: there is also `.endsWith`

src/cli/commands/files/add.js
```diff
@@ -18,16 +52,53 @@ module.exports = Command.extend({
...
+          async.eachLimit(res, 10, (element, callback) => {
+            const addPath = element.substring(index + 1, element.length)
+            if (fs.statSync(element).isDirectory()) {
+              callback()
+            } else {
+              rs = fs.createReadStream(element)
+              filePair = {path: addPath, stream: rs}
+              i.write(filePair)
+              callback()
+            }
```
dignifiedquire: This could be simplified:

```js
if (!fs.statSync(element).isDirectory()) {
  i.write({path: addPath, stream: fs.createReadStrem(element})
}
callback()
```

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,107 @@
...
+      filePath = file.path
+      // Check to see if the result is a directory
+      if (file.dir === false) {
+        const ws = fs.createWriteStream(path.join(dir, file.path))
+        file.stream.pipe(ws)
+      } else {
+        // Check to see if the directory has already been created
+        pathExists(path.join(dir, file.path)).then(exists => {
+          if (!exists) {
+            fs.mkdir(path.join(dir, file.path), (err) => {
```
dignifiedquire: how about putting this into a function

```js
function ensureDir (dir) {
  pathExists(dir).then((exists) => {
    if (exists) {
      return
    }
    fs.mkdir(dir, (err) => {
      if (err) throw err
    })
  })
  .catch((err) => {
    throw err
  })
}

ensureDir(path.join(dir, file.path))
```

dignifiedquire: that way you can reuse it below

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,107 @@
...
+      pathExists(path.join(dir, filePath)).then(exists => {
+        // Create a directory for the incoming files
+        if (!exists) {
+          async.waterfall([
+            (cb) => {
+              fs.mkdir(path.join(dir, filePath), (err) => {
+                if (err) {
+                  cb(err)
+                }
+                cb(null)
```
dignifiedquire: no need for `null`, just `cb()` is enough

dignifiedquire: or even simpler:


```js
(cb) => fs.mkdir(path.join(dir, filePath), cb)
```

src/cli/commands/files/get.js
```diff
@@ -0,0 +1,107 @@
...
+          }
+        })
+      }
+    } else {
+      // Check to see if the directory has already been created
+      filePath = file.path.substring(0, file.path.lastIndexOf('/') + 1)
+      pathExists(path.join(dir, filePath)).then(exists => {
+        // Create a directory for the incoming files
+        if (!exists) {
+          async.waterfall([
```
dignifiedquire: as you are not actually passing values from one function to the next you use `async.series`

src/core/ipfs/init.js
```diff
@@ -67,15 +71,45 @@ module.exports = function init (self) {
...
+            callback()
+          } else {
+            const buffered = fs.readFileSync(element)
+            const rs = new Readable()
+            rs.push(buffered)
+            rs.push(null)
+            const filePair = {path: addPath, stream: rs}
+            i.write(filePair)
+            callback()
+          }
```
dignifiedquire: Can be simplified

```js
if (!fs.statSync(element).isDirectory()) {
  const rs = new Readable()
  rs.push(fs.readFileSync(element))
  rs.push(null)
  i.write({path: addPath, stream: res})
}
callback()
```

src/core/ipfs/init.js
```diff
@@ -67,15 +71,45 @@ module.exports = function init (self) {
...
+        }, (err) => {
+          if (err) {
+            throw err
+          }
+          i.end()
+          return
```
dignifiedquire: not needed

src/core/ipfs/libp2p.js
```diff
@@ -20,7 +20,11 @@ module.exports = function libp2p (self) {
       })
     },
     stop: (callback) => {
-      self._libp2pNode.swarm.close(callback)
+      try {
+        self._libp2pNode.swarm.close(callback)
+      } catch (err) {
+        console.log('It is fine :)')
```
dignifiedquire: lol

test/cli-tests/test-block.js
```diff
@@ -71,6 +71,7 @@ describe('block', () => {
   describe('api running', () => {
     let httpAPI
     before((done) => {
+      console.log('repoPath ->', repoPath)
```
dignifiedquire: this can be removed probably

src/cli/commands/files/add.js
diasdavid: I mean, since there is the return of inPath in the end of the func and this assignment of inPath is within a if/else clause, you just need the return once

src/cli/commands/files/add.js
diasdavid: Isn't inPath being converted to the filename? Instead of trimming the full path till the `cwd`?

Ran an experiment and I can only get the filename, the path value should be the total path relative to where the process was executed.

```
> var path = '/a/b/c/d'
undefined
> path.substring(path.lastIndexOf('/') + 1, path.length)
'd'
```

src/cli/commands/files/cat.js
```diff
@@ -0,0 +1,37 @@
...
+      }
+      ipfs.files.cat(path, (err, res) => {
+        if (err) {
+          throw (err)
+        }
+        if (res) {
+          res.on('file', (data) => {
+            data.stream.pipe(process.stdout)
+          })
+        }
```
diasdavid: Do we need the `if (res)`, in what situation do we do not get an `err` and no `res`?

diasdavid:
@nginnever CR'ed :)

src/core/ipfs/libp2p.js
diasdavid: Let's remove the `try catch`, rebase master onto this branch, now it is really fine :)

dignifiedquire:
Merged in #204

PR #227, opened 2016-05-17
WIP http cat

nginnever:
This is currently using the unixfs-engine event emitter but will be switching to a full duplex stream in the ```unixfs-engine.exporter``` soon


dignifiedquire:
What I thought when I read `...cat..` in my notifications

giphy 3

nginnever:
Haha let's hope for none of this. 

</img>

diasdavid:
Awesomesauce @nginnever :)

Just to confirm, you are top of converting the `exporter` in unixfs to a duplex stream, right?

src/http-api/resources/files.js
```diff
@@ -0,0 +1,50 @@
...
+
+const bs58 = require('bs58')
+const debug = require('debug')
+const log = debug('http-api:files')
+log.error = debug('http-api:files:error')
+
+exports = module.exports
+
+// common pre request handler that parses the args and returns `key` which is assigned to `request.pre.args`
+exports.parseKey = (request, reply) => {
```
noffle: Is this used anywhere?

nginnever:
@diasdavid yeah so this PR will update with the exporter.

test/http-api-tests/test-files.js
```diff
@@ -0,0 +1,85 @@
...
+        })
+
+        it('returns error for request with invalid argument', (done) => {
+          ctl.cat('invalid', (err, result) => {
+            expect(err).to.exist
+            done()
+          })
+        })
+
+        it('returns a stream', (done) => {
```
dignifiedquire: `Buffer` is not a `stream`

src/http-api/resources/files.js
```diff
@@ -0,0 +1,50 @@
...
+    request.server.app.ipfs.files.cat(key, (err, stream) => {
+      if (err) {
+        log.error(err)
+        return reply({
+          Message: 'Failed to cat file: ' + err,
+          Code: 0
+        }).code(500)
+      }
+      stream.on('data', (data) => {
+        return reply(data.stream)
```
dignifiedquire: This could lead to strange issues when multiple events are emitted from the stream.

nginnever:
this will need <a href="https://github.com/ipfs/js-ipfs-unixfs-engine/pull/42">PR</a> to be merged

src/cli/commands/files/cat.js
```diff
@@ -22,12 +22,22 @@ module.exports = Command.extend({
...
       ipfs.files.cat(path, (err, res) => {
         if (err) {
           throw (err)
         }
+        console.log(res)
         if (res) {
-          res.on('file', (data) => {
+          res.on('data', (data) => {
             data.stream.pipe(process.stdout)
           })
```
diasdavid: I believe this console.log should not be here and the if (res) is not necessary.

noffle:
I think @dignifiedquire had an unresolved line comment on this?

src/http-api/resources/files.js
nginnever: I think this will be okay. The exporter stream in object mode only emits one ```data``` event per file. And cat can only request one file to my knowledge. If we allow at some point for cat to take multiple hashs and return multiple file objects then we could have a problem with this. Thoughts?

noffle: It *probably* will be okay, but I think @dignifiedquire's point was about approaching the mechanism defensively: just because today we can reason out that this should never happen doesn't mean future changes by future authors will know to hold this requirement. "It's a stream after all, and you can have many `data` events on streams.."

dignifiedquire:
@noffle it will be fine as soon as #253 is merged

PR #268, opened 2016-05-25
account peer in peer-mux-established event

diasdavid:


dignifiedquire:
No tests?

diasdavid:
@dignifiedquire will step up the networks today and tomorrow, we need to have a broader coverage. For now, this change is the key enabler for Orbit, so I'm proposing to merge it right away so that @haadcode is not blocked. 

But again, always 100% in favour of more tests :)

PR #275, opened 2016-05-26
reorg tests

diasdavid:
This PR is a restructure of how the file for the core tests were layered. Helps isolate batches of tests. Useful for the network tests.

Pushing it now so that other parallel PR avoid complex rebases


test/core/both/index.js
```diff
@@ -0,0 +1,18 @@
...
+
+describe('--both', () => {
+  const tests = fs.readdirSync(__dirname)
+
+  tests.filter((file) => {
+    if (file === 'index.js') {
+      return false
+    } else {
+      return true
+    }
```
dignifiedquire: this could be much simpler

```js
tests.filter((file) => file !== 'index.js')
```

test/core/browser-only/index.js
```diff
@@ -0,0 +1,17 @@
...
+const fs = require('fs')
+
+describe('--browser only', () => {
+  const tests = fs.readdirSync(__dirname)
+  tests.filter((file) => {
+    if (file === 'index.js') {
+      return false
+    } else {
+      return true
+    }
```
dignifiedquire: same as above

dignifiedquire:
@diasdavid lgtm, just one note on cleaner code above

test/core/both/index.js
diasdavid: adding that to the network tests :)

PR #283, opened 2016-05-27
update deps

diasdavid:


diasdavid:
@dignifiedquire still same bug. It is because browser nodes are getting websockets addr, which they shouldn't. This also should not be a issue cause the transport for websockets is just added after the listen, but this errors comes with the fact that you goOnline and goOffline several times. 

See: https://github.com/ipfs/js-libp2p-ipfs-browser/blob/master/src/index.js#L36-L44

Adding ws multiaddrs to browser nodes is confusing, it already misguided @haadcode twice and will mislead our users (remember that those websockets addrs get transferred through identify). Would you mind if we change that from the tests?

dignifiedquire:
> Would you mind if we change that from the tests?

No, just make sure they are added when running in node where we need them

PR #322, opened 2016-06-15
Promisify ipfs.files.get

noffle:
This is needed for interface-ipfs-core consistency.

Depends on https://github.com/ipfs/js-ipfs-unixfs-engine/pull/52


diasdavid:
Thank you @noffle . I believe that it doesn't depend on the unixfs-engine to work, it just depends on the unixfs-engine to be fully compatible with interface-ipfs-core get spec. Let's get done today. Merging this one :)

PR #323, opened 2016-06-15
.add/files.add http-api endpoint

nginnever:
This continues <a href="https://github.com/ipfs/js-ipfs/pull/256">PR</a>

Currently the ```files add``` http resource is failing when the js-ipfs daemon is running do to this line in js-ipfs-api.

https://github.com/ipfs/js-ipfs-api/blob/master/src/get-dagnode.js#L30

@noffle might have some ideas on this. It seems that the js-ipfs-api ```object/data``` responses don't match when coming from js-ipfs and go-ipfs daemons.

PR #326, opened 2016-06-17
remove toString

nginnever:
I found that this was causing incorrect returns when writing the ```add``` http resource. JS-IPFS-API makes http requests to the js-ipfs ```object``` core now with the addition of the new ```add``` support for core <a href="https://github.com/ipfs/js-ipfs-api/blob/master/src/get-dagnode.js#L19">here</a>.

If the data is a string then the buffer is sent correctly, however if the data field is something else it seems to be wrapping the buffer with some extra maybe new line characters. 

@diasdavid @dignifiedquire 
Is there a specific reason for the data to be return as a string in the ```object/data``` http resource?

<a href="https://github.com/ipfs/js-ipfs/pull/323">PR 323</a> Is blocked by this issue.

### DIFF
--- a/src/http-api/resources/object.js
+++ b/src/http-api/resources/object.js
@@ -160,7 +160,7 @@ exports.data = {
         }).code(500)
       }
 
-      return reply(data.toString())
+      return reply(data)
     })
   }
 }


noffle:
Could you add a test case? I'm curious what data causes this to break.

nginnever:
@noffle there is a test case of it failing <a href="https://github.com/ipfs/js-ipfs/pull/323/commits/731d1896c746fe0efb9e648bb2e001437325e7aa#diff-acaee75ce8f82c0fd91ecffc2c199a5dR203">here</a> in the add http PR. 

The file ```otherconfig``` does not get the correct hash do to this issue. It's a plain text json doc while hello does get the correct data when toString() is applied. Could it be json plain text in a buffer being toString() that is the issue?

dignifiedquire:
We should make sure that this is always the same type of object (which `toString` did).

diasdavid:
LGTM :)

Issue #331, opened 2016-06-27
cli: doc: OPTIONS output not working


RichardLitt:
When I run the `--help` command in the CLI for js-ipfs commands, I get the following: 

```
$ node src/cli/bin.js id --help
Usage: bin.js id [OPTIONS]

  Shows IPFS Node ID info
```

What does `[OPTIONS]` mean? Shouldn't it show the options? Let's look at id.js: 

```js
module.exports = Command.extend({
  desc: 'Shows IPFS Node ID infos',

  options: {
    format: {
      alias: 'f',
      type: 'string'
    }
  },

  run: (name) => {
    ...
  }
})
```

Why isn't `[--format | -f]` output?

diasdavid:
Unfortunately that is not a given for free with [ronin](https://www.npmjs.com/package/ronin), the CLI toolkit we use.

dignifiedquire:
We should really fix this and probably switch away from ronin. Which is simply not being maintained anymore.

RichardLitt:
PRed a pretty easy fix for this. However, let's wait until we resolve https://github.com/ipfs/notes/issues/139.

diasdavid:
@dignifiedquire we can collaborate to make it better - https://github.com/vdemedes/ronin/issues/37

dignifiedquire:
Tbh it does not look like anything is happening on ronin. But there is [yargs](https://github.com/yargs/yargs), which looks really awesome and also allows for the command structure that we are after, see [here](https://github.com/yargs/yargs#example-command-hierarchy-using-commanddir) for an example.

PR #339, opened 2016-06-30
Update help function

RichardLitt:
This makes the OPTIONS output a bit better. See #331. Do not merge this, for now, as we may be replacing the js- and go-ipfs docs with an intermediary; storing this here as an example of how to implement help in [ronin](https://www.npmjs.com/package/ronin)

### DIFF
--- a/commands.md
+++ b/commands.md
@@ -1,0 +1,51 @@
+Usage: jsipfs COMMAND [OPTIONS]
+
+Available commands:
+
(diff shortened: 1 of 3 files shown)


dignifiedquire:
@RichardLitt given the new output in #390 is this still needed?

diasdavid:
@RichardLitt ping :)

RichardLitt:
Nope!

PR #376, opened 2016-07-27
Add security@ipfs.io GPG key

lgierth:
Not sure which other steps are involved here when updating init-docs.


noffle:
Is it worth mentioning what things the key is used for? As in, what data/documents/files users should expect to be able to check for signatures on?

diasdavid:
@lgierth finished this for you :)

lgierth:
diasdavid best david <3

PR #385, opened 2016-07-31
Module list

### LINKED ISSUE #271: roadmap / readme dont have listing of modules
the roadmap and readme have feature listings and links to issues, but there isnt a clear listing of all the sub module repos that make up js-ipfs. would be nice to have that list to quickly access those modules

dignifiedquire:
Based on the ideas in #271 and the list from @nginnever I made this package listing. 

Closes #271 

@diasdavid please review :)


diasdavid:
Thank you, this looks great! 👏 

I see some of the badges are broken:

image

And I would prefer that the libp2p are in libp2p (because they eventually will too and then we need to update two places, which is cumbersome)

Plus, do we need to have devdependencies badge? What about having the CI instead?

dignifiedquire:
> Plus, do we need to have devdependencies badge? What about having the CI instead?

We can do CI instead, or in addition. This is just based on cycle.js and babel which do this in this way already.

dignifiedquire:
> And I would prefer that the libp2p are in libp2p (because they eventually will too and then we need to update two places, which is cumbersome)

Sure we can add a link here to those packages so it's easy to see.

dignifiedquire:
> I see some of the badges are broken:

all good for me, probably your internetz

diasdavid:
@dignifiedquire try to refresh the badges, some of them point to `/diasdavid/` repos, which I think is the cause why I don't see them. As for the rest 👍

dignifiedquire:
@diasdavid moved libp2p https://github.com/libp2p/js-libp2p/pull/14/files

RichardLitt:
We could use an array of packages and do something similar to project-repos.ipfs.io with these, as well. That would remove the necessity to redo the work there, and would allow keeping it up to date easier, I think.

dignifiedquire:
@RichardLitt I don't see how we can get a markdown table from an array in here, could you explain?

RichardLitt:
I'm saying we could specify an array of modules, from anywhere, as things that we want to feed into a variable instance of project-repos.ipfs.io. That way, instead of just seeing IPFS repos, you could switch it to also see the list of js-ipfs submodules. We need to make ipfs/project-repos extensible, anyway, to make sure it works with multiformats and libp2p repositories. cc @harlantwood

diasdavid:
@RichardLitt could you open an issue in `ipfs/community` with that proposal, it looks that is something broader, that it will take more time and affect more projects.

@dignifiedquire let's move the "Packages" to right before "Contribute", the list is going to continue growing and we don't want to hide the architecture and descriptions. Also, `s/Packages/Modules`

Thank you!

dignifiedquire:
> Also, s/Packages/Modules

why?

dignifiedquire:
> let's move the "Packages" to right before "Contribute",

done

PR #421, opened 2016-08-16
update the `script` tag in an example provided by `README.md`

### LINKED ISSUE #418: there is no `assets` subdirectory
Currently `README.md` [recommends](https://github.com/ipfs/js-ipfs/blob/17a07363d6212c6fb2ebbac591d533900c1793ad/README.md#use-in-a-browser-using-a-script-tag) the following:

```js
<script type="text/javascript" src="assets/bundled.js"></script>
```

However,

* [HTML5 `script`](https://www.w3.org/TR/html5/scripting-1.html#the-script-element) defaults to `type="text/javascript"` and thus that attribute can be dropped from the example,

* there is no `assets` subdirectory in the results of `npm install ipfs` and thus the path `assets/bundled.js` should be replaced by `dist/index.min.js` if appropriate.

Mithgol:
Fixes #418.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,13 +84,27 @@ The CLI is available by using the command `jsipfs` in your terminal. This is ali
 
 The code published to npm that gets loaded on require is in fact a ES5 transpiled version with the right shims added. This means that you can require it and use with your favourite bundler without having to adjust the asset management process.
 
+```js
+var ipfs = require('ipfs');
+```
+
 ### Use in a browser using a script tag
 
-Simply include ipfs in your bundled javascript file.
+Loading this module in a browser (using a `<script>` tag) makes the `Ipfs` object available in the global namespace.
 
-```html
-<script type="text/javascript" src="assets/bundled.js"></script>
-```
+The last published version of the package becomes available on [npmcdn](https://npmcdn.com/) and thus you may use it as the source:
+
+* loading the minified version
+
+   ```html
+   <script src="https://npmcdn.com/ipfs/dist/index.min.js"></script>
+   ```
+
+* loading the human-readable (not minified) version
+
+   ```html
+   <script src="https://npmcdn.com/ipfs/dist/index.js"></script>
+   ```
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ var ipfs = require('ipfs');
 
 Loading this module in a browser (using a `<script>` tag) makes the `Ipfs` object available in the global namespace.
 
-The last published version of the package becomes available on [npmcdn](https://npmcdn.com/) and thus you may use it as the source:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,13 +84,27 @@ The CLI is available by using the command `jsipfs` in your terminal. This is ali
...
+
 ### Use in a browser using a script tag
 
-Simply include ipfs in your bundled javascript file.
+Loading this module in a browser (using a `<script>` tag) makes the `Ipfs` object available in the global namespace.
 
-```html
-<script type="text/javascript" src="assets/bundled.js"></script>
-```
+The last published version of the package becomes available on [npmcdn](https://npmcdn.com/) and thus you may use it as the source:
```
nginnever: you can link to the latest npmcdn dist version with this I believe ``[npmcdn](https://npmcdn.com/ipfs@*/dist/)``

diasdavid:
Other than @nginnever comment LGTM :)

README.md
Mithgol: The homepage of https://npmcdn.com/ currently says the following:

> You may also use a [tag](https://docs.npmjs.com/cli/dist-tag) or [version range](https://docs.npmjs.com/misc/semver) instead of a fixed version number, or omit the version/tag entirely to use the `latest` tag.

And that's why I've decided to omit the version/tag.

nginnever: Ahh cool, omitting the version tag entirely is an easy way to link to the latest dist

``` https://npmcdn.com/ipfs/dist/```

Mithgol: Just to confirm: do you suggest to link the word “npmcdn” to https://npmcdn.com/ipfs/dist/ instead of the npmcdn's homepage?

Originally I intended to use the examples (`<script src="https://npmcdn.com/ipfs/dist/index.min.js">` and `<script src="https://npmcdn.com/ipfs/dist/index.js">` below) to explain the script's location but I left the word “npmcdn” to tell the ignorant public what npmcdn is. That's how the README.md of `ipfs-api` [does it currently](https://github.com/ipfs/js-ipfs-api/blob/c5e9d27cbb898fa8836125e8364c93f0adf6a78f/README.md#in-the-browser-through-script-tag), for example.

nginnever: Yeah exactly, I was just suggesting the link could be directed towards the latest dist instead of the cdn homepage since sometimes I want to download a dist for requiring in my own project structures without relying on npm to be online.

Mithgol: I am still compelled to keep the hyperlink to https://npmcdn.com/ipfs/dist/ because it's educational and, in fact, that's how I've learnt of npmcdn myself.

That's why I've just added yet another hyperlink that leads to https://npmcdn.com/ipfs/dist/ and is titled “available for download” for people ready to download the dist manually.

diasdavid:
thank you :)

Issue #428, opened 2016-08-20
Use interface-ipfs-core tests over the http-api with ipfs-api


diasdavid:
Once https://github.com/ipfs/js-ipfs/pull/420 is done

diasdavid:
https://github.com/ipfs/js-ipfs/pull/432

PR #432, opened 2016-08-23
Test IPFS using ipfs-api with the same interface-ipfs-core tests

diasdavid:
> WIP, solves https://github.com/ipfs/js-ipfs/issues/428

TODO:

- [x] factory-http
- [x] test-block
  - [x] core
  - [x] http
- [x] test-object
- [ ] ~~test-files~~
- [x] test-config
- [ ] ~~test-swarm~~
  - [ ] core
  - [ ] http

Notes: 

Already finding problems just by reusing test-object, this was definitely a good idea :D


diasdavid:
Needs:

- [ ] https://github.com/ipfs/interface-ipfs-core/pull/64
- [ ] https://github.com/ipfs/js-ipfs-api/pull/356

diasdavid:
Needs (for swarm tests):

- [ ]  https://github.com/ipfs/js-libp2p-ipfs-browser/issues/83

diasdavid:
I want to go ahead and merge this one in, letting the `files` and `swarm` be inside of another PR, since they will require a bit more work across other modules, enabling me to focus back on the pull-stream endeavour.

@dignifiedquire would appreciate some review here. As said above, it is really great to reuse the same tests over js-ipfs-api, found a bunch of issues, this already made js-ipfs way more portable :)

diasdavid:
@dignifiedquire have you CR'ed this one yet? All good?

src/core/ipfs/block.js
```diff
@@ -1,24 +1,47 @@
...
+      }
+
+      self._blockS.deleteBlock(hash, callback)
     },
-    stat: (multihash, callback) => {
-      self._blockS.getBlock(multihash, (err, block) => {
+    stat: (hash, callback) => {
+      if (typeof hash === 'string') {
+        hash = multihash.fromB58String(hash)
+      }
```
dignifiedquire: this is a lot of duplication, would be nice to simplify this to something like this

```js
const ensureHash = (hash) => {
  if (typeof hash === 'string') return multihash.fromB58String(hash)
  return hash
}
```

src/core/ipfs/libp2p.js
```diff
@@ -64,22 +62,7 @@ module.exports = function libp2p (self) {
...
-          if (tuple[0] === 421) {
-            ipfsIdB58String = tuple[1]
-          }
-        })
-
-        const id = peerId.createFromB58String(ipfsIdB58String)
-
-        self._libp2pNode.dialByMultiaddr(maddr, (err) => {
-          callback(err, id)
-        })
```
dignifiedquire: Why can we drop all of this suddenly?

test/cli/test-block.js
```diff
@@ -68,7 +68,7 @@ describe('block', () => {
     })
   })
 
-  describe('api running', () => {
+  describe.skip('api running', () => {
```
dignifiedquire: why you skip me?

src/core/ipfs/libp2p.js
diasdavid: Because this was a check to make sure the multiaddr contained the ipfs multihash, but to be fully interop with go, it has to support discovering that post conn is established. Therefore https://github.com/ipfs/js-libp2p-ipfs/pull/16

dignifiedquire:
LGTM

PR #469, opened 2016-09-09
Upgrade the files branch (#323) to work with pull-streams

dignifiedquire:
Some tests are still breaking, need to figure out why.

Needs: 

- [x] https://github.com/ipfs/js-ipfs-unixfs-engine/pull/71

TODO:

- [x] rebased `add` onto `master`
- [x] started using pull-streams in the file operations
- [x] made `files get` actually stream the file content, using streams + size args
- fix tests:
  - [x] `http: files .cat streams a large file`
  - [x] `http: files .get a large file`
  - [x] `http: files .get directory`


dignifiedquire:
Quick update: I managed to find the issue, it is a double read of the first chunk of the file, which happens due to a race condition inside [`pull-file`](//github.com/pull-stream/pull-file). I am not certain how to best fix it though.

src/http-api/resources/files.js
```diff
@@ -33,18 +40,142 @@ exports.cat = {
...
+        pull.onEnd((err) => {
+          if (err) return handleError(err)
+
+          pack.finalize()
+          reply(pack).header('X-Stream-Output', '1')
+        })
+      )
+    })
+
+    function handleError (err) {
```
VictorBjelkholm: Might not be a great idea to create this function every time a request happens. And in general, `handleError` should probably be somewhere global for the lib instead of specific to get.

dignifiedquire:
Fix for pull-file is up: https://github.com/pull-stream/pull-file/pull/4

diasdavid:
@dignifiedquire awesome! Shall we use the fork meanwhile?

dignifiedquire:
@diasdavid still working out some issues :/

diasdavid:
@jbenet fixed all the things :D (except pull-file, that still requires the PR to be merged)

dignifiedquire:
Whoo a wild @jbenet appears and fixes my code 😹 

The other issue I fixed in https://github.com/ipfs/js-ipfs-repo/pull/85

diasdavid:
@dignifiedquire could you describe how we went from: 'pull-file has a race condition that emits the same chunk twice' to 'putting a lock that buffers the whole block is a fix'?

src/core/ipfs/files.js
```diff
@@ -53,6 +53,7 @@ module.exports = function files (self) {
       pull(
         pull.values([hash]),
         pull.asyncMap(self._dagS.get.bind(self._dagS)),
+        pull.take(1),
```
diasdavid: Can't these be replaced just with one call to dagService to get the DAGNode?

src/core/ipfs/files.js
```diff
@@ -81,6 +82,10 @@ module.exports = function files (self) {
           return file
```
diasdavid: We need to `pause` the stream? Would it otherwise start emitting without listeners?

src/http-api/resources/files.js
```diff
@@ -33,18 +40,152 @@ exports.cat = {
   // main route handler which is called after the above `parseArgs`, but only if the args were valid
   handler: (request, reply) => {
     const key = request.pre.args.key
+    const ipfs = request.server.app.ipfs
 
-    request.server.app.ipfs.files.cat(key, (err, stream) => {
+    ipfs.files.cat(key, (err, stream) => {
```
diasdavid: 👍

src/core/ipfs/files.js
dignifiedquire: what pause?

src/http-api/resources/files.js
```diff
@@ -33,18 +40,152 @@ exports.cat = {
...
       }
-      stream.on('data', (data) => {
-        return reply(data)
+
+      // hapi is not very clever and throws if no
+      // - _read method
+      // - _readableState object
+      // are there :(
+      stream._read = () => {}
+      stream._readableState = {}
```
diasdavid: This seems like a bug in Hapi //cc @sericaia

src/core/ipfs/files.js
dignifiedquire: it could, but I like it this way :P

dignifiedquire:
> @dignifiedquire could you describe how we went from: 'pull-file has a race condition that emits the same chunk twice' to 'putting a lock that buffers the whole block is a fix'?

I filed a PR on pull-file as mentioned above, where @domenictarr was so nice to point out that calling the file stream twice without waiting for the previous cb to finish was a bug in the sink stream, not in pull-file, as pull-streams should always wait for the previous one to be finished. So I started to dig in and found that the reason why the source stream was called twice, was that we were not properly locking gets to the same keys in the blockstore (we were doing this before we transitioned to pull-streams, I simply forgot to put it back when doing the refactor).

src/core/ipfs/files.js
diasdavid: Why would we have something in 3 lines when we can just use one? It adds unnecessary complexity, memory overhead and superfluous ops.

src/core/ipfs/files.js
diasdavid: Check the lines above

src/core/ipfs/files.js
dignifiedquire: no it doesn't add memory overhead, because the implementation of `.get` in the dagService is literally the same thing

src/core/ipfs/files.js
dignifiedquire: yeah, data can be lost if I don't pause it, oh my love for node streams is so great

src/http-api/resources/files.js
```diff
@@ -33,18 +40,152 @@ exports.cat = {
...
+}
+
+exports.add = {
+  handler: (request, reply) => {
+    if (!request.payload) {
+      return reply('Array, Buffer, or String is required.').code(400).takeover()
+    }
+
+    const ipfs = request.server.app.ipfs
+    // TODO: make pull-multipart
```
diasdavid: Track this in a issue. @xicombd you probably would like to get that working, since you did the first multipart parser :)

src/core/ipfs/files.js
dignifiedquire: sorry, nvm what I'm doing here is stupid -.-

src/http-api/routes/files.js
```diff
@@ -6,6 +6,7 @@ module.exports = (server) => {
   const api = server.select('API')
 
   api.route({
+    // TODO fix method
```
diasdavid: This is an http-api-spec thing

test/http-api/ipfs-api/test-files.js
```diff
@@ -1,38 +0,0 @@
...
-          done()
-        })
-      })
-    })
-
-    describe('.get', () => {}) // TODO
-
-    describe('.ls', () => {}) // TODO
-  })
-}
```
diasdavid: Delete the whole test? Why?

diasdavid:
@dignifiedquire interesting, so reading the same file twice causes an issue? That sounds weird.

Also, CR'ed this PR

src/http-api/routes/files.js
dignifiedquire: not sure who put that there, @VictorBjelkholm did you?

test/http-api/ipfs-api/test-files.js
dignifiedquire: because it is tested in interface-ipfs-core now, and the `returns a buffer` is not compliant with `interface-ipfs-core` anymore

dignifiedquire:
> @dignifiedquire interesting, so reading the same file twice causes an issue? That sounds weird.

No reusing the same stream to read the file can cause issues

src/http-api/routes/files.js
VictorBjelkholm: Yeah, I put it there. Mainly as a reminder that we should talk about setting the method of the endpoints in `http-api-spec` and change them accordingly here.

src/http-api/resources/files.js
dignifiedquire: https://github.com/ipfs/js-ipfs/issues/479

src/core/ipfs/files.js
diasdavid: It's weird to me that it starts flowing before a read on a 'data' event is attached.

dignifiedquire:
@diasdavid lets squash this into a single commit when merging please.

diasdavid:
@dignifiedquire I've squashed and reworded, but not all into a single commit, because that loses the contributions from the different participants.

That being said, thank you everyone @nginnever @noffle @VictorBjelkholm @dignifiedquire @jbenet that took part in this PR, js-ipfs is at least 40% more awesome and robust just for having all of these tests passing :D thank you! 👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏👏🎉👏

VictorBjelkholm:
!()[http://i.giphy.com/l46CvRFB1GqPYAOis.gif]

src/core/ipfs/files.js
jbenet: This is absurd. It's serious fancy abstraction creep. Careful

jbenet:
History is important too. Squash responsibly

Issue #286, opened 2016-05-28
jsipfs init works but no results


MasterJames:
when I run jsipfs init it does create the config so show will display the Identity.PeerId etc.
but it doesn't actually display what is shown in the demos as in results with the ipfs cat /ipfs/PeerId/readme
Maybe one needs some extra steps explained in the repo README to add files etc.?

When I add a with `files add NewFile.txt` it says 
`added ID NewFile.txt` as expected
if I `files get /ipfs/ID` it says
`error   EISDIR: illegal operation on a directory, open 'C:\WORK_FOLDER\IPFS\ipfs'`

If I put "files get /ifps/ID/NewFile.txt" is says
`error   Non-base58 character`

`jsipfs ls ID` results in a blank line, while the `files get` shows the directory in the `EISDIR: illegal operation on a directory` error and if I look there is nothing there (after adding from outside). If I move the file in from the parent folder it was added from I see no change.

Now If I `files add` from within the target folder in will create a directory with the same hash and the files is still there. Putting the file inside the folder named from the hash makes no difference to ls /ipfs/ID as it will always display blank.

MasterJames:
https://github.com/ipfs/js-ipfs/issues/285

MasterJames:
If I'm running the daemon straight or with my own mount points it won't allow `files add NewFile.txt` 
`TypeError: ipfs.files.add is not a function`
With call stack...
```

    at npm\node_modules\ipfs\src\cli\commands\files\add.js:62:30
    at Object.exports.getIPFS (npm\node_modules\ipfs\src\cli\utils.js:46:3)
    at npm\node_modules\ipfs\src\cli\commands\files\add.js:58:13
    at f (npm\node_modules\ipfs\node_modules\once\once.js:17:25)
    at Glob.<anonymous> (npm\node_modules\ipfs\node_modules\glob\glob.js:146:7)
    at emitOne (events.js:77:13)
    at Glob.emit (events.js:169:7)
    at Glob._finish (npm\node_modules\ipfs\node_modules\glob\glob.js:194:8)
    at done (npm\node_modules\ipfs\node_modules\glob\glob.js:179:14)
    at Glob._processGlobStar2 (npm\node_modules\ipfs\node_modules\glob\glob.js:631:12)

```

diasdavid:
>  ipfs cat /ipfs/PeerId/readme

It shouldn't be your peerId, but actually the folder hash that contains the Readme

[...]

PR #475, opened 2016-09-10
fix(cli): add output for cli init

VictorBjelkholm:
Shows exactly same information as go-ipfs except for change of the cmd +
cat so `ipfs cat` turns into `jsipfs files cat` instead. Ref: issue #286

### DIFF
--- a/src/cli/utils.js
+++ b/src/cli/utils.js
@@ -49,3 +49,18 @@ exports.getIPFS = (callback) => {
 exports.getRepoPath = () => {
   return process.env.IPFS_PATH || os.homedir() + '/.ipfs'
 }
+
+exports.createLogger = (visible) => {
+  return (msg, newline) => {
+    if (newline === undefined) {
+      newline = true
+    }
+    if (visible) {
+      if (msg === undefined) {
+        msg = ''
+      }
+      msg = newline ? msg + '\n' : msg
+      process.stdout.write(msg)
+    }
+  }
+}
--- a/src/core/ipfs/init.js
+++ b/src/core/ipfs/init.js
@@ -9,6 +9,7 @@ const fs = require('fs')
 const importer = require('ipfs-unixfs-engine').importer
 const pull = require('pull-stream')
 const file = require('pull-file')
+const mh = require('multihashes')
 
 module.exports = function init (self) {
   return (opts, callback) => {
@@ -19,6 +20,7 @@ module.exports = function init (self) {
 
     opts.emptyRepo = opts.emptyRepo || false
     opts.bits = opts.bits || 2048
+    opts.log = opts.log || function () {}
 
     let config
     // Pre-set config values.
@@ -47,6 +49,7 @@ module.exports = function init (self) {
 
     // Generate peer identity keypair + transform to desired format + add to config.
     function generateAndSetKeypair () {
+      opts.log(`generating ${opts.bits}-bit RSA keypair...`, false)
       var keys = peerId.create({
         bits: opts.bits
       })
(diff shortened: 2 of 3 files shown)


VictorBjelkholm:
This is the output of `init` with this change:

```
$ export IPFS_PATH=~/.ipfs-js
$ node src/cli/bin.js init 
initializing ipfs node at /home/victor/.ipfs-js
generating 2048-bit RSA keypair...done
peer identity: QmP7HvVg8wJV1YvbnmFsnxAP5ULyZaeCxhJx12kizyybfL
to get started, enter:

	 jsipfs files cat /ipfs/QmegvLXxpVKiZ4b57Xs1syfBVRd8CbucVHAp7KpLQdGieC/readme

```

VictorBjelkholm:
Not sure why Travis is failing here, working fine in CircleCI

diasdavid:
@VictorBjelkholm while you are this, can you alias `ipfs files {cat, add, get, ls}` to `ipfs {cat, add, get, ls}`, so that we can use the Sharness and Benchmark tests?

src/cli/utils.js
```diff
@@ -49,3 +49,15 @@ exports.getIPFS = (callback) => {
...
+  return (msg, newline = true) => {
+    if (visible) {
+      if (msg === undefined) {
+        msg = ''
+      }
+      msg = newline ? msg + '\n' : msg
+      process.stdout.write(msg)
+    }
+  }
+}
```
diasdavid: Isn't this just a console.log? All invisible logs should be through `debug` (http://npmjs.org/debug)

VictorBjelkholm: Almost, it's console log with the ability of deciding if next message should go on a new line or not. Necessary for the generating keys message, which prints "done" after it's done.

VictorBjelkholm:
@diasdavid Sure, I'll create the aliases in a different PR though.

src/cli/utils.js
```diff
@@ -49,3 +49,15 @@ exports.getIPFS = (callback) => {
 exports.getRepoPath = () => {
   return process.env.IPFS_PATH || os.homedir() + '/.ipfs'
 }
+
+exports.createLogger = (visible) => {
```
dignifiedquire: Why do you need to `create` the logger? It does not hold any state. You could just have

```js
```

src/cli/utils.js
```diff
@@ -49,3 +49,15 @@ exports.getIPFS = (callback) => {
...
+  return (msg, newline = true) => {
```
dignifiedquire: This will fail on node 4 and 5,  (default arguments are only supported in node 6)

src/core/ipfs/init.js
```diff
@@ -108,6 +113,16 @@ module.exports = function init (self) {
         }),
         pull.filter(Boolean),
         importer(dag),
+        pull.map((el) => {
```
dignifiedquire: No need for `map` you are not modifying the values. Use `pull.through` instead.

src/cli/utils.js
VictorBjelkholm: Oh, I was sure that js-ipfs was compiled, even when used from node. Is that not correct?

src/core/ipfs/init.js
VictorBjelkholm: Thanks, that was what I was looking for...

src/cli/utils.js
dignifiedquire: We do provide a compiled version when publishing do npm, but the tests run on node against the original source atm.

VictorBjelkholm: Shouldn't the tests run against the compiled version that we distribute rather than the source we develop with? Anyways, I'll fix this but worth thinking about.

dignifiedquire: If you npm install it, you can use both versions.

VictorBjelkholm: Ah, but `require('ipfs')` would use the compiled version right?

dignifiedquire: Yes. And the tests in the browser do run against the compiled version

Issue #476, opened 2016-09-10
`jsipfs files cat $HASH` shows no output


VictorBjelkholm:
To reproduce, run the following in js-ipfs repository

```
$ HASH=$(node src/cli/bin.js files add package.json | cut -f2 -d ' ')
$ node src/cli/bin.js files cat $HASH
```

No output, but should show the contents of package.json

VictorBjelkholm:
Still missing, seems like sub-files from a hash not working correctly either.

Example:

`node src/cli/bin.js files cat /ipfs/QmegvLXxpVKiZ4b57Xs1syfBVRd8CbucVHAp7KpLQdGieC/readme`

Shows no output atm.

diasdavid:
@VictorBjelkholm correct, that is a feature that isn't currently supported, although there is no technical reason why not to, all the pieces are there :)

VictorBjelkholm:
@diasdavid I'll add a new PR for that

PR #477, opened 2016-09-10
fix(cli): use right argument for cli .cat

VictorBjelkholm:
Add tests for making sure .cat shows right output and fix that test by
using the right argument from cli. Ref: issue #476

### DIFF
--- a/test/cli/test-files.js
+++ b/test/cli/test-files.js
@@ -17,6 +17,7 @@ describe('files', () => {
         .run((err, stdout, exitcode) => {
           expect(err).to.not.exist
           expect(exitcode).to.equal(0)
+          expect(stdout[0]).to.equal('hello world')
           done()
         })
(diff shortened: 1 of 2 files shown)


test/cli/test-files.js
```diff
@@ -9,14 +9,15 @@ const _ = require('lodash')
 
 describe('files', () => {
   const env = _.clone(process.env)
-  env.IPFS_PATH = repoPath
+  process.env.IPFS_PATH = repoPath
```
dignifiedquire: This should not be modified. As `env` is passed in below into the command and this would modify the whole process.

VictorBjelkholm: You're absolutely right, silly me. Fixed!

diasdavid:
👍 @VictorBjelkholm

PR #482, opened 2016-09-12
fix(cli): pipe content to the cli from cat it is a stream

dignifiedquire:
Found while testing the cli :)

### DIFF
--- a/src/cli/commands/files/add.js
+++ b/src/cli/commands/files/add.js
@@ -55,11 +55,17 @@ module.exports = {
     const index = inPath.lastIndexOf('/') + 1
 
     utils.getIPFS((err, ipfs) => {
-      if (err) throw err
+      if (err) {
+        throw err
+      }
 
       glob(path.join(inPath, '/**/*'), (err, list) => {
-        if (err) throw err
-        if (list.length === 0) list = [inPath]
+        if (err) {
+          throw err
+        }
+        if (list.length === 0) {
+          list = [inPath]
+        }
 
         pull(
           zip(
--- a/src/cli/commands/files/cat.js
+++ b/src/cli/commands/files/cat.js
@@ -14,19 +14,23 @@ module.exports = {
 
   handler (argv) {
     const path = argv['ipfs-path']
-
     utils.getIPFS((err, ipfs) => {
       if (err) {
         throw err
       }
 
-      ipfs.files.cat(path, (err, file) => {
-        if (err) {
-          throw (err)
-        }
+      if (utils.isDaemonOn()) {
+        return ipfs.cat(path, onFile)
+      }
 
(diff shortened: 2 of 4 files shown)


src/cli/commands/files/cat.js
```diff
@@ -23,10 +23,12 @@ module.exports = {
...
-          console.log(res.toString())
+
+          res.pipe(process.stdout)
         })
         return
       }
+
       ipfs.files.cat(path, (err, file) => {
         if (err) {
           throw (err)
```
diasdavid: This can be simplified now, since it doesn't matter anymore if the daemon is on, js-ipfs-api also has `ipfs.files.cat` and it is the same API

diasdavid:
We can now remove the `ipfs.isDaemonOn` checks, as APIs are the same

src/cli/commands/files/cat.js
dignifiedquire: oh cool, didn't know `ipfs.files.cat` was a thing on js-ipfs-api

dignifiedquire: done

dignifiedquire:
@diasdavid I'm afraid you are mistaken `ipfs.files.cat` is not a thing: https://github.com/ipfs/js-ipfs-api/blob/master/src/api/files.js

diasdavid:
@dignifiedquire They are aliased here: https://github.com/ipfs/js-ipfs-api/blob/master/src/load-commands.js#L35-L43

dignifiedquire:
@diasdavid there still is no `cat`

diasdavid:
@dignifiedquire got it, we missed that one!

src/cli/commands/files/cat.js
```diff
@@ -18,21 +18,16 @@ module.exports = {
...
-      })
+
+      ipfs.files.cat(path, onFile)
     })
   }
 }
+
+function onFile (err, file) {
+  if (err) {
+    console.log(err)
```
dignifiedquire: please remove :)

src/cli/commands/files/add.js
```diff
@@ -55,11 +55,17 @@ module.exports = {
...
 
       glob(path.join(inPath, '/**/*'), (err, list) => {
-        if (err) throw err
-        if (list.length === 0) list = [inPath]
+        if (err) {
+          throw err
+        }
+        if (list.length === 0) {
+          list = [inPath]
+        }
```
dignifiedquire: Please don't do style changes in a bug fix commit :)

diasdavid: `ack`, will avoid in future PR. Reverting just this and do another PR doesn't seem necessary.

Issue #439, opened 2016-08-25
Use interface-ipfs-core tests for the Swarm API


diasdavid:
Uncomment https://github.com/ipfs/js-ipfs/blob/master/test/http-api/interface-ipfs-core-over-ipfs-api/test-swarm.js#L5-L23 and make it work.

Needs:

- [x]  https://github.com/ipfs/js-libp2p-ipfs-browser/issues/83

diasdavid:
Ok, since https://github.com/ipfs/go-ipfs/issues/3218, what we needs is to fix the test swarm.connect in interface-ipfs-core to add the peer-id to the end of the multiaddr

dignifiedquire:
@VictorBjelkholm doing this now

VictorBjelkholm:
@dignifiedquire :+1:

diasdavid:
Thank you both :)

PR #486, opened 2016-09-13
feat(swarm): interface-ipfs-core swarm compatibility

diasdavid:
This contributes to https://github.com/ipfs/js-ipfs/issues/439

Only missing:
- [x] implement swarm.addrs

### DIFF
--- a/src/core/ipfs/libp2p.js
+++ b/src/core/ipfs/libp2p.js
@@ -2,6 +2,7 @@
 
 const multiaddr = require('multiaddr')
 const Libp2pNode = require('libp2p-ipfs').Node
+const promisify = require('promisify-es6')
 
 const OFFLINE_ERROR = require('../utils').OFFLINE_ERROR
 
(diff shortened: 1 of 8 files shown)


src/core/ipfs/libp2p.js
```diff
@@ -31,29 +32,50 @@ module.exports = function libp2p (self) {
...
+        const peers = self._libp2pNode.peerBook.getAll()
+        const mas = []
+        Object
+           .keys(peers)
+           .forEach((b58Id) => {
+             peers[b58Id].multiaddrs.forEach((ma) => {
+               // TODO this should only print the addr we are using
+               mas.push(ma)
+             })
+           })
```
dignifiedquire: Please map, instead of forEach and push

src/core/ipfs/libp2p.js
```diff
@@ -31,29 +32,50 @@ module.exports = function libp2p (self) {
...
-      localAddrs: (callback) => {
+        const peers = self._libp2pNode.peerBook.getAll()
+        const mas = []
+        Object
+           .keys(peers)
+           .forEach((b58Id) => {
+             peers[b58Id].multiaddrs.forEach((ma) => {
+               // TODO this should only print the addr we are using
+               mas.push(ma)
+             })
```
dignifiedquire: same as above

src/core/ipfs/libp2p.js
diasdavid: There is two forEach, it would require a flat step then

dignifiedquire: Then use _.flatten or pull.flatten :)

diasdavid: Is requiring a dependency to remove 0 lines (because the extra map to flatten step would take the exact same lines) worth it?

dignifiedquire: I think so, because it makes the code much easier to read and understand

dignifiedquire:
not a big fan of merging things that are still being discussed :/

PR #490, opened 2016-09-14
refactor: bubble swarm up and make a module just for it, preprepare libp2p to be used directly from libp2p instance, take off cat alias, make ping into its own module

diasdavid:


src/core/index.js
```diff
@@ -43,21 +46,24 @@ function IPFS (repoInstance) {
...
   this.config = config(this)
   this.block = block(this)
   this.object = object(this)
   this.libp2p = libp2p(this)
-  this.swarm = this.libp2p.swarm // for interface-ipfs-core sake
+  this.swarm = swarm(this)
   this.files = files(this)
-  this.cat = files(this).cat // Alias for js-ipfs-api cat
   this.bitswap = bitswap(this)
+  this.ping = ping(this)
```
diasdavid: This should make it more obvious for folks to understand why this pieces are all attached here.

src/core/modules/libp2p.js
```diff
@@ -0,0 +1,31 @@
...
+      })
+      self._libp2pNode.swarm.on('peer-mux-established', (peerInfo) => {
+        self._libp2pNode.peerBook.put(peerInfo)
+      })
+    }),
+    stop: promisify((callback) => {
+      self._libp2pNode.stop(callback)
+    })
+  }
+}
```
diasdavid: Goal: after DEVCON, come up with interface-libp2p and use libp2p instances here directly :)

src/core/modules/ping.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+const promisify = require('promisify-es6')
+
+module.exports = function ping (self) {
+  return promisify((callback) => {
+    callback(new Error('Not implemented'))
+  })
+}
```
diasdavid: @xicombd this is all yours :D

diasdavid:
request: @dignifiedquire || @VictorBjelkholm for a quick CR :)

src/core/index.js
diasdavid: Note for the future: We should use a more explicit DI than just through an object with all the things. It will make it easier for people that want to have their light ipfs nodes to understand what are the dependencies of each component that IPFS uses

src/core/index.js
```diff
@@ -7,21 +7,23 @@ const PeerBook = require('peer-book')
...
-const repo = require('./ipfs/repo')
-const init = require('./ipfs/init')
-const bootstrap = require('./ipfs/bootstrap')
-const config = require('./ipfs/config')
-const block = require('./ipfs/block')
-const object = require('./ipfs/object')
-const libp2p = require('./ipfs/libp2p')
-const files = require('./ipfs/files')
-const bitswap = require('./ipfs/bitswap')
+const goOnline = require('./modules/go-online')
```
dignifiedquire: Why the renaming to `modules`. I agree `ipfs` isn't a great name, but `modules` isn't describing it much better. Maybe `commands`?

src/core/index.js
dignifiedquire: Strongly agree with the `Note for the future`

src/core/modules/libp2p.js
dignifiedquire: It's funny, I was close to doing the exact same move, i.e. moving the swarm commands into their own file when working on #486

src/core/index.js
diasdavid: what about 'components'?

diasdavid: changed to `components`

Issue #455, opened 2016-09-07
Benchmarks


dignifiedquire:
In order to know how fast things are, and in which area we can improve we want to add benchmarks to js-ipfs and its modules. 

For the first round I suggest adding the following benchmarks
### js-ipfs
- [ ] {offline|online} file add {file|folder}
  - files: count {1|5|10} - size {1KB|10KB|100KB|1MB}
  - folders: count {1|5|10}
- [ ] {offline|online} file get {file|folder} 
  - files: count {1|5|10} - size {1KB|10KB|100KB|1MB}
  - folders: count {1|5|10}
- [ ] {offline|online} object put
  - count {1|5|10} - size {1KB|10KB|100KB|1MB}
- [ ] {offline|online} object get
  - count {1|5|10} - size {1KB|10KB|100KB|1MB}
- [ ] block distribution between {2|5|10} nodes (these should be different processes)
  - block count {10|100} - size {1KB|10KB}
### js-libp2p-secio
- [ ] throughput on a connection over {pull-pair|tcp|websockets}
  - messsage: size {100Bytes|10KB|100KB} - count {100|1000}

diasdavid:
Let's make sure these benchmarks present valid measures by adding the quantity of operations and size of files they will have to perform.

Kubuxu:
Take a look at https://github.com/whyrusleeping/ipfs-whatever which is very simple and naive benchmarks we have used for go-ipfs over HTTP API.

It would be best if we could reuse some/most benchmarks for both js-ipfs and go-ipfs not to double up the efforts.

dignifiedquire:
> Let's make sure these benchmarks present valid measures by adding the quantity of operations and size of files they will have to perform.

Added some numbers, feel free to suggest better ones, these are wild guesses currently.

dignifiedquire:
> It would be best if we could reuse some/most benchmarks for both js-ipfs and go-ipfs not to double up the efforts.

I would love if I can just run this against js-ipfs. That would be a great start and save us some time.

dignifiedquire:
I just tested to run ipfs-whatever against js-ipfs. If we have `add` and `cat` at the same place as go we can actually use it! Which I very much prefer over having a separate test suite for the first round.

dignifiedquire:
First results for things that worked out of the box. Before is js-ipfs, After is go-ipfs

```
Results          Before  After   % Change
PatchOpsPerSec   208.54  681.61  226.85%
[...]

PR #493, opened 2016-09-15
fix(cli): alias add, cat and get to top-level cli

VictorBjelkholm:
Aliases add, cat and get to be in the top-level namespace of the cli.

Reference: https://github.com/ipfs/js-ipfs/issues/455#issuecomment-245939074

Left to do:
- [x] Add tests for aliases

### DIFF
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -12,10 +12,23 @@ updateNotifier({
   updateCheckInterval: 1000 * 60 * 60 * 24 * 7 // 1 week
 }).notify()
 
-yargs
+const cli = yargs
   .commandDir('commands')
   .demand(1)
-  .help()
+
+// NOTE: This creates an alias of
+// `jsipfs files {add, get, cat}` to `jsipfs {add, get, cat}`.
+// This will stay until https://github.com/ipfs/specs/issues/98 is resolved.
+const addCmd = require('./commands/files/add')
+const catCmd = require('./commands/files/cat')
+const getCmd = require('./commands/files/get')
+const aliases = [addCmd, catCmd, getCmd]
+aliases.forEach((alias) => {
+  cli.command(alias.command, alias.describe, alias.builder, alias.handler)
+})
+
+// finalize cli setup
+cli.help()
   .strict()
   .completion()
-  .argv
+.argv
--- a/test/cli/test-files.js
+++ b/test/cli/test-files.js
@@ -16,6 +16,12 @@ describe('files', () => {
       })
     })
 
+    it('cat alias', () => {
+      return ipfs('cat QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o').then((out) => {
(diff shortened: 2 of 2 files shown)


dignifiedquire:
We also need this for the http api. Also please add some sanity tests to check these actually exist and work.

VictorBjelkholm:
@dignifiedquire oh, wasn't aware of that. All right, adding a TODO in PR.

src/cli/bin.js
```diff
@@ -15,6 +15,8 @@ updateNotifier({
 yargs
   .commandDir('commands')
   .demand(1)
+  // Temporary alias of {add,cat,get}
+  .commandDir('commands/files')
```
diasdavid: It is more the other way around. `{add, cat, get}` are temporary alias of `{ fileas add, files cat, files get}`

diasdavid:
> Add aliases in API as well

Which API? The core API should follow the interface-ipfs-core spec, which only defines files `{add, cat, get}`.

src/cli/bin.js
VictorBjelkholm: @diasdavid Hm, probably have to change the wording. My understanding is that `files add`, `files cat` and `files get` is here to stay. `add`, `cat` and `get` is what is the aliases and the code here is aliasing add/cat/get to files add/files cat/files get but the comment doesn't seem to be informative enough.

Have any ideas on how to improve it?

Kubuxu:
I think the http-api.

VictorBjelkholm:
@diasdavid can you please take a look at my comment: https://github.com/ipfs/js-ipfs/pull/493#r79014500

And answer if this has to be for the http-api as well

src/cli/bin.js
diasdavid: What you just said is correct :)

You can replace the comment to

```
Note: These creates an alias of `jsipfs files {add, get, cat}` to `jsipfs {add, get, cat}`. This will stay until https://github.com/ipfs/specs/issues/98 is resolved.
```

diasdavid:
@VictorBjelkholm answered :)

VictorBjelkholm:
@diasdavid no, I'll get this ready asap

VictorBjelkholm:
@diasdavid done with adding tests etc, ready to merge as soon as travis finishes.

dignifiedquire:
@VictorBjelkholm I also need those on the http-api

diasdavid:
@VictorBjelkholm it is missing a `get alias` test

diasdavid:
Hey! Seems that this one got lost in time, can we get it shipped soon?

VictorBjelkholm:
@diasdavid yes, absolutely, just missing one test and then it's ready. I dropped the ball on this and I'll jump right on it again!

VictorBjelkholm:
Added test, so this is ready to merge now @diasdavid

diasdavid:
Sweeeet 🍰 ! :D

PR #494, opened 2016-09-15
Less skipped tests

dignifiedquire:
I found some tests that were skipped and resulted in actual broken code. This is fixed now and things are more http-api and interface-ipfs-core compliant than before :)

### DIFF
--- a/src/cli/commands/config.js
+++ b/src/cli/commands/config.js
@@ -1,8 +1,6 @@
 'use strict'
 
 const debug = require('debug')
-const get = require('lodash.get')
-const set = require('lodash.set')
 const log = debug('cli:config')
 log.error = debug('cli:config:error')
 const utils = require('../utils')
@@ -43,26 +41,17 @@ module.exports = {
 
       if (!value) {
         // Get the value of a given key
-
-        if (utils.isDaemonOn()) {
-          return ipfs.config.get(key, (err, config) => {
-            if (err) {
-              log.error(err)
-              throw new Error('failed to read the config')
-            }
-
-            console.log(config.Value)
-          })
-        }
-
-        ipfs.config.get((err, config) => {
+        ipfs.config.get(key, (err, value) => {
           if (err) {
             log.error(err)
             throw new Error('failed to read the config')
           }
 
-          const value = get(config, key)
-          console.log(value)
+          if (typeof value === 'object') {
+            console.log(JSON.stringify(value, null, 2))
+          } else {
+            console.log(value)
+          }
         })
       } else {
         // Set the new value of a given key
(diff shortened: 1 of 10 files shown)


test/http-api/ipfs-api/test-config.js
```diff
@@ -60,14 +60,15 @@ module.exports = (ctl) => {
     it('.get updatedConfig', (done) => {
       ctl.config.get((err, config) => {
         expect(err).not.to.exist
-        expect(config).to.deep.equal(updatedConfig())
+        console.log(config, updatedConfig())
```
VictorBjelkholm: probably remove this

test/http-api/inject/test-version.js
```diff
@@ -17,9 +17,9 @@ module.exports = (http) => {
         method: 'GET',
         url: '/api/v0/version'
       }, (res) => {
-        expect(res.result.version).to.equal(pkgversion)
-        expect(res.result).to.have.a.property('commit')
-        expect(res.result).to.have.a.property('repo')
+        expect(res.result).to.have.a.property('Version', pkgversion)
```
VictorBjelkholm: Should not actually check the version rather than just making sure *something* is there?

VictorBjelkholm: Suggestion would be to read version from somewhere, then assert with that, rather than hard coding the version

src/cli/commands/version.js
```diff
@@ -37,11 +37,7 @@ module.exports = {
           throw err
         }
 
-        if (typeof version === 'object') { // js-ipfs-api output
-          version = version.Version
-        }
-
-        console.log(`js-ipfs version: ${version}`)
+        console.log(`js-ipfs version: ${version.version}`)
```
VictorBjelkholm: Should probably use the logger that I've created before, instead of using `console.log`

src/cli/commands/config.js
```diff
@@ -43,26 +41,17 @@ module.exports = {
...
             throw new Error('failed to read the config')
           }
 
-          const value = get(config, key)
-          console.log(value)
+          if (typeof value === 'object') {
+            console.log(JSON.stringify(value, null, 2))
+          } else {
+            console.log(value)
+          }
```
VictorBjelkholm: Same here as below, we should probably try using the logger rather than console.log

test/http-api/ipfs-api/test-config.js
dignifiedquire: why? ;)

VictorBjelkholm: Is a console.log in the tests?

test/http-api/inject/test-version.js
dignifiedquire: that's exactly what is happening here, `pkgversion` is read from `package.json` which is the version that js-ipfs uses as baseline when answering the version command.

VictorBjelkholm: Oh, crap, you're right, as always. I'll never get used to Chai/Expect syntax...

src/cli/commands/version.js
dignifiedquire: Let's open an issue for that, and do the conversion, if we do it at once, rather than here and there.

test/http-api/ipfs-api/test-config.js
dignifiedquire: It was a joke, that's why I wrote the `;)`. Already removed

VictorBjelkholm: Ah, sorry, went way above my head :crying_cat_face:

src/cli/commands/version.js
VictorBjelkholm: Sounds good to me.

test/http-api/ipfs-api/test-config.js
dignifiedquire: No worries 😸 jokes are somtimes hard to understand on the internetz

src/cli/commands/config.js
diasdavid: we can make that in a single PR and change all of them, no?

diasdavid:
5 🌟

Issue #498, opened 2016-09-15
bitswap is 'requesting one block at a time'


diasdavid:
In an experiment to fetch orbit (700Mb), js-ipfs hung at 650KiB, checking bitswap wantlist showed that bitswap is only requesting a block at a time.

![image](https://cloud.githubusercontent.com/assets/1211152/18570253/5fc48fe0-7b76-11e6-8446-0585f8019117.png)

dignifiedquire:
Working on this, including some fixes need to be made to `files.get`

dignifiedquire:
Short update while I am head down in this:

- Parallelised the block requests in ipfs-unixfs-engine, fixing the original issue mentioned here, that only a single block would be requested at once.
- While testing found a bug in js-ipfs-api, for `get` which resulted in not properly handling more than one or two files. I was able to fix this as well.
- Fixed the code in `cli/commands/files/get.js` to actually store files and directories to disk as expected.
- Now working on actual issues inside bitswap. The current problem is that after receiving some blocks things come to a halt, and it never finishes.

dignifiedquire:
Found some more details. 

- Sometimes js-ipfs will request a hash from go-ipfs that go-ipfs does not know about, even though the whole requested object was added on go-ipfs, so where are those coming from?

dignifiedquire:
Most of the time, it will just stop, with one or two entries in the wantlist and never receive them from go-ipfs, even though the entry is in the wantlist from go

dignifiedquire:
Not a bug in bitswap anymore. For details on interop transfer issues see: https://github.com/libp2p/js-libp2p-spdy/issues/48

PR #499, opened 2016-09-16
Fix/files get

dignifiedquire:
- Refactor cli tests to reduce code duplication and increase covered test cases. Also better failure mode, showing stderr and stdin when a cli tests fails.
- Fix bugs revealed due to better cli tests in
  - `files get`
  - `files cat`
  - `block stat`
  - `bootstrap add`
  - `bootstrap rm`

Ref https://github.com/ipfs/js-ipfs/issues/498

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "transform-loader": "^0.2.3"
   },
   "dependencies": {
+    "async": "^2.0.1",
     "babel-runtime": "^6.11.6",
     "bl": "^1.1.2",
     "boom": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "aegir": "^8.0.1",
     "buffer-loader": "0.0.1",
     "chai": "^3.5.0",
+    "execa": "^0.4.0",
     "expose-loader": "^0.7.1",
     "form-data": "^2.0.0",
     "fs-pull-blob-store": "^0.3.0",
--- a/src/cli/commands/bitswap/stat.js
+++ b/src/cli/commands/bitswap/stat.js
@@ -23,8 +23,7 @@ module.exports = {
         stats.Wantlist = stats.Wantlist || []
         stats.Peers = stats.Peers || []
 
-        console.log(`
-bitswap status
+        console.log(`bitswap status
   blocks received: ${stats.BlocksReceived}
   dup blocks received: ${stats.DupBlksReceived}
   dup data received: ${stats.DupDataReceived}B
--- a/src/cli/commands/block/stat.js
+++ b/src/cli/commands/block/stat.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const utils = require('../../utils')
-const bs58 = require('bs58')
 const debug = require('debug')
 const log = debug('cli:block')
 log.error = debug('cli:block:error')
@@ -24,7 +23,7 @@ module.exports = {
           throw err
         }
 
-        console.log('Key:', bs58.encode(stats.key).toString())
+        console.log('Key:', stats.key)
         console.log('Size:', stats.size)
       })
     })
--- a/src/cli/commands/files/cat.js
+++ b/src/cli/commands/files/cat.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const waterfall = require('run-waterfall')
 const debug = require('debug')
 const utils = require('../../utils')
 const log = debug('cli:files')
@@ -14,19 +15,16 @@ module.exports = {
 
   handler (argv) {
     const path = argv['ipfs-path']
-    utils.getIPFS((err, ipfs) => {
+
+    waterfall([
+      (cb) => utils.getIPFS(cb),
+      (ipfs, cb) => ipfs.files.cat(path, cb)
+    ], (err, file) => {
       if (err) {
         throw err
       }
 
-      ipfs.files.cat(path, onFile)
+      file.pipe(process.stdout)
     })
   }
-}
-
-function onFile (err, file) {
-  if (err) {
-    throw (err)
-  }
-  file.pipe(process.stdout)
 }
--- a/src/cli/commands/files/cat.js
+++ b/src/cli/commands/files/cat.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const waterfall = require('run-waterfall')
+const waterfall = require('async/waterfall')
 const debug = require('debug')
 const utils = require('../../utils')
 const log = debug('cli:files')
--- a/src/cli/commands/files/get.js
+++ b/src/cli/commands/files/get.js
@@ -65,6 +65,7 @@ function fileHandler (dir) {
           file.content.pipe(fs.createWriteStream(target))
             .once('error', cb)
             .once('end', cb)
+          return
         }
         cb()
       })
--- a/src/cli/commands/files/get.js
+++ b/src/cli/commands/files/get.js
@@ -7,6 +7,8 @@ log.error = debug('cli:files:error')
 var fs = require('fs')
 const path = require('path')
 const pathExists = require('path-exists')
+const pull = require('pull-stream')
+const toPull = require('stream-to-pull-stream')
 
 function checkArgs (hash, outPath) {
   // format the output directory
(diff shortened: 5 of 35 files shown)


dignifiedquire:
This can be reviewed for now, no need to wait until we fix the bitswap issues.

diasdavid:
@dignifiedquire can you add more description. It says that it fixes `files.get`. What was found, what was achieved?

dignifiedquire:
@diasdavid added a description on what happened^^

diasdavid:
thank you @dignifiedquire, left a bunch of comments :)

dignifiedquire:
> thank you @dignifiedquire, left a bunch of comments :)

not seeing any in this PR

diasdavid:
@dignifiedquire, my bad, I used the review tool and forgot to click submit

package.json
```diff
@@ -43,6 +43,7 @@
     "aegir": "^8.0.1",
     "buffer-loader": "0.0.1",
     "chai": "^3.5.0",
+    "execa": "^0.4.0",
```
dignifiedquire: To spawn the cli in a consistent and nice way.

src/cli/commands/bitswap/stat.js
```diff
@@ -23,8 +23,7 @@ module.exports = {
         stats.Wantlist = stats.Wantlist || []
         stats.Peers = stats.Peers || []
 
-        console.log(`
-bitswap status
+        console.log(`bitswap status
```
dignifiedquire: As I mentioned in another PR I would rather do that change as a single PR on its own, there are already too many things mixed in here.

src/cli/commands/block/stat.js
```diff
@@ -24,7 +23,7 @@ module.exports = {
           throw err
         }
 
-        console.log('Key:', bs58.encode(stats.key).toString())
+        console.log('Key:', stats.key)
```
dignifiedquire: Not currently

src/cli/commands/files/cat.js
```diff
@@ -14,19 +15,16 @@ module.exports = {
...
+    waterfall([
+      (cb) => utils.getIPFS(cb),
+      (ipfs, cb) => ipfs.files.cat(path, cb)
+    ], (err, file) => {
       if (err) {
         throw err
       }
 
-      ipfs.files.cat(path, onFile)
+      file.pipe(process.stdout)
```
dignifiedquire: I don't remember making that decision, but happy to move all these `run-*` to modular async.

dignifiedquire: > These are not changes to get

As mentioned in the description, fixes all over including `files cat` ;)

src/http-api/resources/block.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const bs58 = require('bs58')
+const mh = require('multihashes')
```
dignifiedquire: Why, we have been using `mh` for the multihashes module all over the place.

src/http-api/routes/bootstrap.js
```diff
@@ -17,14 +16,11 @@ module.exports = (server) => {
   api.route({
     method: '*',
     path: '/api/v0/bootstrap/add',
-    handler: resources.bootstrap.add,
     config: {
-      validate: {
-        query: {
-          arg: Joi.string().required(), // multiaddr to add
-          default: Joi.boolean()
-        }
```
dignifiedquire: - It is the only place in the http api where we have validation.
- It was incorrect as it threw errors on the global `stream` header.

test/cli/test-bitswap.js
```diff
@@ -2,34 +2,32 @@
...
 const expect = require('chai').expect
-const nexpect = require('nexpect')
 const Block = require('ipfs-block')
-const _ = require('lodash')
 const bs58 = require('bs58')
 
 const HttpAPI = require('../../src/http-api')
 const createTempNode = require('../utils/temp-node')
 const repoPath = require('./index').repoPath
+const ipfs = require('../utils/ipfs')(repoPath)
```
dignifiedquire: I couldn't come up with one, happy to hear suggestions.

test/cli/test-bitswap.js
```diff
@@ -41,42 +39,32 @@ describe('bitswap', function () {
...
-            '  dup data received: 0B',
-            '  wantlist [1 keys]',
-            `    ${key}`,
-            '  partners [0]',
-            '    '
-          ])
-          done()
-        })
+    it('stat', () => {
+      return ipfs('bitswap stat').then((out) => {
```
dignifiedquire: If you look into `test/utils/ipfs.js` you see it asserts that the exitcode is `0` and that stderr is ``. It returns a promise and will reject that promise when an error happens which will in turn display the failed things like exitcode and stderr.

src/cli/commands/block/stat.js
diasdavid: Right https://github.com/ipfs/interface-ipfs-core/tree/master/API/block#stat :)

test/cli/test-bitswap.js
```diff
@@ -41,42 +39,32 @@ describe('bitswap', function () {
...
-          expect(err).to.not.exist
-          expect(exitcode).to.equal(0)
-          expect(stdout).to.be.eql([
-            key
-          ])
-          done()
-        })
+      return ipfs('bitswap wantlist').then((out) => {
+        expect(out).to.be.eql(key)
+      })
```
dignifiedquire: Look closely, the `inject` call is just so we get the right response, but the actual command shouldn't happen inside, otherwise it would hang forever.

test/cli/test-bitswap.js
diasdavid: `ipfs-exec` or `ipfs-run`?

test/cli/test-bitswap.js
diasdavid: `stderr` might not always be ''.

dignifiedquire: If we want the command to success it will, for when you want it to fail, there is `ipfs.fail('my cmd')` which expects failure and allows you to assert agains `stderr`

test/cli/test-bitswap.js
dignifiedquire: I like `ipfs-exec`

src/http-api/resources/files.js
```diff
@@ -87,25 +89,27 @@ exports.get = {
...
             header.size = file.size
-            toStream.source(file.content)
-              .pipe(pack.entry(header, cb))
+            const packStream = pack.entry(header, cb)
+            if (!packStream) {
+              // this happens if the request is aborted
+              // we just skip things then
+              log('other side hung up')
+              return cb()
+            }
```
diasdavid: @dignifiedquire

src/http-api/routes/bootstrap.js
diasdavid: > It is the only place in the http api where we have validation

It was a start, a good practice introduced by @xicombd. 

> It was incorrect as it threw errors on the global stream header.

Why exactly? That sounds like a Joi bug, this is only checking the query param

test/cli/test-bitswap.js
diasdavid: sweet! 

Let's document all of that in that `ipfs-exec` module :)

src/http-api/resources/files.js
dignifiedquire: the `const` has nothing to do with it, `pack.entry` can return `undefined` if the packStream was errored or canceled before.

src/http-api/routes/bootstrap.js
dignifiedquire: I don't know why, but it wouldn't allow for any additional query parameters to be passed along except for the ones that were part of this validation.

test/cli/test-bitswap.js
dignifiedquire: done

test/cli/test-bitswap.js
dignifiedquire: done

src/cli/commands/files/get.js
```diff
@@ -33,30 +35,38 @@ function ensureDir (dir, cb) {
...
-        file.content.pipe(fs.createWriteStream(dirPath))
+        if (file.content) {
+          const filename = file.path.substring(lastSlash)
+          const target = path.join(dirPath, filename)
+
+          file.content.pipe(fs.createWriteStream(target))
+            .once('error', cb)
+            .once('end', cb)
+        }
+        cb()
```
dignifiedquire: fixed

src/http-api/routes/bootstrap.js
diasdavid: IRC log
```
11:04 <@daviddias> I don't believe we can break the user API in multihashing, have been thinking about it and the best option ( given that having both API would multiply deps) is to have multihashing-async module
11:04 <@daviddias> And recommend devs to use it in multihashing, explaining why
11:14 <@daviddias> On the first PR, my last comment is to Lee Joi validation, if there are more optional Args, then so be it :)
11:14 <@daviddias> s/Lee/lets keep
11:17 <@daviddias> @dignifiedquire: ^ :)
11:18 <dignifiedquire> so you mean creating a new module for multihashing?
11:20 <dignifiedquire> if we recommend devs to use the new version, we can just do a major version bump, with a changelog, that way everyone who wants to not refactor their code can stay on the old version, and we don't have to maintain to different versions
11:21 <dignifiedquire> daviddias: re Joi validation as I said in my comment I wasn't able to get that working, Joi would always throw if js-ipfs-api passed it something it didn't know
11:26 <dignifiedquire> daviddias: ^^
12:17 <@daviddias> dignifiedquire: ACK, what I meant is: the args that will be passed are defined on the spec
12:18 <@daviddias> so we can list them 
```

dignifiedquire: Actually, the offending parameter is not documented in the HTTP API: http://docs.ipfs.apiary.io/#reference/bootstrap/add/add does not list `stream-channels`, which is always set to true by js-ipfs-api: https://github.com/ipfs/js-ipfs-api/blob/master/src/request-api.js#L92

diasdavid: Let's add that as an issue in the http spec and add that rule here

diasdavid:
Seems that Joi is the last bit fo make this happen:

- [ ] https://github.com/ipfs/js-ipfs/pull/499#discussion_r81191055

dignifiedquire:
> Also, want to squash commits?

When CI is green and you gave your thumbs up

diasdavid:
@dignifiedquire rebased, tests pass on CI, but fail locally:

```
  1) HTTP API ## interface-ipfs-core tests over ipfs-api .files callback API .add add a nested dir as array:

      Uncaught AssertionError: expected 'test-folder/files' to equal 'test-folder'
      + expected - actual

      -test-folder/files
      +test-folder

      at node_modules/interface-ipfs-core/lib/files.js:152:34
```

dignifiedquire:
@diasdavid try running them again, they pass for me

diasdavid:
Got them running (had to do 2 fresh npm installs)

Issue #412, opened 2016-08-11
Sharness tests


diasdavid:
In order to improve the quality and robustness of js-ipfs, it is pretty important to reuse the SHarness testing infrastructure against the js-ipfs http-api.

@chriscool could you help us with that? I think the best first step is to have the go-ipfs cli point at the js-ipfs http-api and test the endpoints that js-ipfs already implements (i.e: object, block, bitswap, add, cat, get). We can do that by pulling `go-ipfs` as a devdependency and run it in CI

chriscool:
@diasdavid yeah I started to take a look at that during the Lisbon workshop, but there were a lot of problems, startin with `jsipfs version`.
I will take another look soon.

diasdavid:
I was more thinking in the lines of using go-ipfs cli to point to js-ipfs http api (ipfs --api='') so that we can tell that both behave the same way, avoiding to have to write the entire CLI.

diasdavid:
@chriscool any update about this?

chriscool:
@diasdavid I had a vacation and I am at the LinuxCon in Toronto now until next Monday.
I will try to take a look at it soon. Or maybe we should plan to discuss it, maybe on IRC, at the end of next week.

diasdavid:
@chriscool how are things? Can we look into this again? jsipfs is now interopable with go-ipfs (so we can have tests that use both at the same time) and the CLI is getting the same features + interface :)

chriscool:
@diasdavid great that it's now interoperable! Could we discuss it and try to do something tomorrow morning at around 9 am CEST time or are you available now?

diasdavid:
@chriscool I'm currently in NYC, so 9am CEST would be 4am for me. Can we make it in your afternoon?

chriscool:
@diasdavid then I think 9pm CEST is better for me. That would be 4pm for you.
Is it ok for you?

chriscool:
@diasdavid actually 9pm CEST is not possible today for me. I forgot that I had planned to meet friends. But it's possible tomorrow (Thursday September 15). Would that work for you?

diasdavid:
@chriscool Thank you for handling this, I see you have opened the PR's, let's follow on those ones:

- https://github.com/ipfs/js-ipfs/pull/522
- https://github.com/ipfs/js-ipfs/pull/521
- https://github.com/ipfs/js-ipfs/pull/520
- https://github.com/ipfs/js-ipfs/pull/519
[...]

PR #504, opened 2016-09-22
Add basic sharness tests

chriscool:
This is to start testing js-ipfs in the same way as go-ipfs.
See issue #412 (Sharness tests).


diasdavid:
\o/ Thank you @chriscool :D 

Do all of these run for you? Found any issues? Can I get a status of this endeavour?

I know you have been trying to reach me through IRC, unfortunately, with DEVCON2 I wasn't the most available. I'll have good wifi when the 29th of September comes.

chriscool:
There are a few that don't pass. For example I think the first to fail is the 5st or the 6st because `jsipfs version --all` doesn't work properly.
If you want I can change the last commit to use test_expect_failure for those that don't pass.
But I think it can still be interesting to commit this as it makes it possible for other developers and CI tools to try to run the tests and see what fails.

The Makefile also implements `make help` that provides a bit of help to run the tests.

dignifiedquire:
I think it would be good to leave the failing as failing for now so we can make a list and check things off :)

diasdavid:
Yeah, add all the tests @chriscool, if they fail, the better, more stuff for us to fix :D 

@chriscool idealy, we want:

- A PR that sets up the sharding tests (the ones that pass already) and then
- PR for each of the subcommands, with all the tests, highlighting the ones that fail

chriscool:
@diasdavid ok, I removed failing tests from t0010, so now everything should pass and you can merge this.
I am preparing other PRs, some that add passing tests, some that add failing tests.
I will send them when this PR will be merged, as they are based on it.

chriscool:
By the way, tests should either be run using `make` or `make deps` should be used before running the tests, and the path to 'js-ipfs/test/bin' should be in your PATH.

This is because a symbolic link from 'js-ipfs/test/bin/ipfs' to 'js-ipfs/src/cli/bin.js' is used to make it possible for the tests to run without having to change them much.

chriscool:
This last commit adds a check and improves the error message.

test/ipfs-test-lib.sh
```diff
@@ -0,0 +1,107 @@
+# Generic test functions for go-ipfs
```
diasdavid: I guess we can start calling it the IPFS cli :)

test/sharness/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+../bin/ipfs:
+	mkdir -p ../bin
+	cd ../bin && ln -s ../../src/cli/bin.js ipfs
+
+curl:
+	@which curl >/dev/null || (echo "Please install curl!" && false)
+
+.PHONY: all help clean clean-test-results $(T) aggregate deps sharness
+
```
diasdavid: can we bring the Makefile to the root, so we can run it on CI by adding a line with `make` to:

https://github.com/ipfs/js-ipfs/blob/master/.travis.yml#L14

test/sharness/lib/test-lib-hashes.sh
```diff
@@ -0,0 +1,7 @@
+# this file defines several useful hashes used across the test codebase.
+# thus they can be defined + changed in one place
+
+HASH_WELCOME_DOCS="QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG"
+HASH_GATEWAY_ASSETS="QmXB7PLRWH6bCiwrGh2MrBBjNkLv3mY3JdYXCikYZSwLED"
+HASH_HELP_PAGE="QmY5heUM5qgRubMDD1og9fhCPA6QdkMp3QCwd4s7gJsyE7"
+HASH_EMPTY_DIR="QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn"
```
diasdavid: Apparently, these change often (//cc @lgierth) how does go-ipfs handle it? Changing it manually everytime?

test/ipfs-test-lib.sh
chriscool: Year, sure I will change that.

diasdavid:
\o/ woot! 

There is a lot of bash magic, but they do work for me, so awesome! :D 

- [ ] small request https://github.com/ipfs/js-ipfs/pull/504/files#r81572723

Once we have CI running these, I'm happy to merge. Then we can see all the remaining PR's for each subcommand coming in :)

Super excited about this, getting the same level of robustness as the Go implementation 🌟

test/sharness/Makefile
chriscool: In go-ipfs there is also a Makefile at the root, but it is different.
You use `make test` at the root and `make` when you are in test/sharness.
I can do something like that but it could go into another PR I think as it is related to the whole project.

test/sharness/lib/test-lib-hashes.sh
chriscool: It doesn't change so often, so I think manually is fine for now.
Last time they changed was in January.

chriscool:
@diasdavid just pushed with the following changes:

- added a Makefile at the root as in the last commit,
- changed the first line of test/ipfs-test-lib.sh

test/ipfs-test-lib.sh
chriscool: Changed to: `# Generic test functions for ipfs cli tests`

test/sharness/Makefile
diasdavid: @chriscool what I'm looking for is to be able to add these sharness tests to CI before this merge.

chriscool: Yes, with the last commit I added you can just run `make test` at the root.
I will add that to the '.travis.yml'.

chriscool: I added that and it works.

test/sharness/lib/test-lib-hashes.sh
whyrusleeping: We update them manually here every time we update them in go-ipfs

chriscool:

diasdavid:
@chriscool done, thank you :) 

Can you start opening the PR for each of the subcommands now? If possible, listing which tests fail, so that we can fix them :)

chriscool:
@diasdavid ok I will do that now.

PR #508, opened 2016-10-02
Fixed path for module importing

nikuda:
This fixes `ipfs` package import. 

Also the npm package at `0.0.2` is very outdated . It would be useful if a more recent copy was published.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "bin": {
     "jsipfs": "src/cli/bin.js"
   },
-  "main": "lib/core/index.js",
+  "main": "src/core/index.js",
   "jsnext:main": "src/core/index.js",
   "scripts": {
     "lint": "aegir-lint",


dignifiedquire:
> Also the npm package at 0.0.2 is very outdated . It would be useful if a more recent copy was published.

https://www.npmjs.com/package/ipfs shows for me as latest version `0.16.0`

nikuda:
>https://www.npmjs.com/package/ipfs shows for me as latest version 0.16.0

Oh great, I was looking at https://www.npmjs.com/package/js-ipfs, perhaps that can updated to point to the above to prevent future confusion.

diasdavid:
Thank you @nikuda, just deprecated that one to avoid confusion and also released a new version to make it very obvious

https://www.npmjs.com/package/js-ipfs

PR #518, opened 2016-10-16
sharness: t0010- jsipfs version

chriscool:
This add some failing tests.

The first failure is:

```
expecting success: 
        ipfs version --all > version_all.txt &&
        grep "go-ipfs version" version_all.txt &&
        grep "Repo version" version_all.txt &&
        grep "System version" version_all.txt &&
        grep "Golang version" version_all.txt

not ok 5 - ipfs version --all has all required fields
```

This is because `jsipfs version --all` returns "js-ipfs version: 0.17.0" while:

```
$ ipfs version --all
go-ipfs version: 0.4.4-dev-8830aae
Repo version: 4
System version: amd64/linux
Golang version: go1.5.3
```

I understand that jsipfs cannot give a go-ipfs version or a golang version, but it could give equivalent information about the js-ipfs version and the js engine.

In go-ipfs reports we ask users to copy paste the output of `ipfs version --all` when filling an issue. If js-ipfs also implements that you could also ask users to do the same thing.

### DIFF
--- a/test/sharness/t0010-basic-commands.sh
+++ b/test/sharness/t0010-basic-commands.sh
@@ -26,4 +26,50 @@ test_expect_success "ipfs version output looks good" '
 	test_fsh cat version.txt
 '
 
+test_expect_success "ipfs version --all has all required fields" '
+	ipfs version --all > version_all.txt &&
+	grep "go-ipfs version" version_all.txt &&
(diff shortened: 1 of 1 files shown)


chriscool:
cc @diasdavid

PR #519, opened 2016-10-16
sharness: add t0020-init.sh

chriscool:
This adds 5 go-ipfs tests that pass.


chriscool:

PR #520, opened 2016-10-16
sharness: add t0022-init-default.sh

chriscool:
This adds 6 go-ipfs tests that pass.


chriscool:

PR #534, opened 2016-10-21
Add a few examples to README.md

SidHarder:


README.md
```diff
@@ -110,8 +112,40 @@ The last published version of the package become [available for download](htt
 
 ### Examples
 
-> **Will come soon**
+```js
+var fs = require('fs');
+var promisify = require('promisify-es6')
```
Mithgol: The variable `promisify` does not seem to be used anywhere below in your examples.

(Should you decide to delete this line?)

README.md
```diff
@@ -110,8 +112,40 @@ The last published version of the package become [available for download](htt
...
+        console.log(hash);
+    });    
+}
+
+function catFile() {
+    node.files.cat('HashOfExistingFile').then(function(stream) {
+        stream.pipe(process.stdout, { end : false });
+    });
+}
+```
```
dignifiedquire: please change the code style to be consistent with our code base, i.e [standard](https://github.com/feross/standard)

README.md
```diff
@@ -110,8 +112,40 @@ The last published version of the package become [available for download](htt
...
+}
+
+function initializeRepo() {
+    node.init(function(err){
+        console.log(err);
+    })    
+}
+
+function startDaemon() {
+    node.goOnline(function(msg) {});
```
dignifiedquire: This should probably call `addFile` in the callback to trigger the interaction once the node is online.

README.md
SidHarder: I will remove that line.  Thank you.

README.md
SidHarder: I will fix this up.

README.md
SidHarder: Thanks for your feedback.  I will read the standard and apply.

README.md
SidHarder: Is the idea that you would go online do some work and then go offline.

README.md
dignifiedquire: Thank you :)

README.md
dignifiedquire: That was my thought at least, did you have something else in mind?

SidHarder:
I'll will make it dance.

SidHarder:
I'm not able to get Node.id() to work.  Am I missing something?  The promise doesn't ever return, It appears to hang.

node.id().then(function(data) {
      conosle.log(data)
})

SidHarder:
Let me know what you guys think about this example!

VictorBjelkholm:
@SidHarder looks great! Thanks a lot for adding this. Got two minor points.

* Probably want to change the default path to the repo, instead of using ~/.ipfs
* Seems you're mixing callbacks and promises. We should just use callbacks instead.

SidHarder:
@VictorBjelkholm I would also prefer to use callbacks but for example when I use this:

node.files.add(readStream, function(data) {
          console.log(data)
      })

There is no data.  I don't seem to get the hash from the call back.

When I do:

node.files.add(readStream).then(function (data) {
        console.log(data)
      })

I get the hash back and can proceed. I am missing something?  This is what led me to mix callback with promises.  

Thanks for your help.

Mithgol:
@SidHarder Most Node.js callbacks have the signature `(error, something)` — you're getting “no error” as your function's first argument.

SidHarder:
@Mithgol I forgot, error first.  Thanks.

SidHarder:
@VictorBjelkholm I'm afraid I don't get your point on this "Probably want to change the default path to the repo, instead of using ~/.ipfs".

SidHarder:
I see that someone made some changes to the readme.md and  did a much better job.  This pr can be closed. Thanks for everyone's help.

diasdavid:
Oh, apologies @SidHarder, should have had continued on top of your PR, but ended up just completing the Readme as I was updating it with the awesome IPLD PR.

Nevertheless, your example has way more information, could you PR the example as working code into the examples folder?

VictorBjelkholm:
Closing in favor of https://github.com/ipfs/js-ipfs/pull/576

PR #583, opened 2016-11-11
Bugfix/unhandlederror

haadcode:
This PR will make sure libp2p.start() errors get handled when IPFS node starts libp2p.


haadcode:
Oops, ignore this PR, will create a new one

PR #595, opened 2016-11-15
Configuration for labels-and-milestones sync tool

flyingzumwalt:
This adds the labels-and-milestones sync tool with configs for the go-ipfs Q4 roadmap.

I'm not convinced this is the right place for these configs.

For more info, see ipfs/pm#257

### DIFF
--- a/tools/labels-and-milestones/.gitignore
+++ b/tools/labels-and-milestones/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/tools/labels-and-milestones/README.md
+++ b/tools/labels-and-milestones/README.md
@@ -1,0 +1,39 @@
+# labels-and-milestones Sync for js-ipfs
+
+> Sync labels and milestones across Github repositories
+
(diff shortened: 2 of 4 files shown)


diasdavid:
same as https://github.com/ipfs/js-ipfs/pull/596#issuecomment-261273089

flyingzumwalt:
@diasdavid please download this code and run it once. That will add the
milestones to all of your remaining repos so your team can associate issues
with them. Most of the repos have already been synced but I still don't
have access to the multiformats ones.

I can follow up on figuring out where these configs should live for the
long term and how to make CI run them.

diasdavid:
Where is this tool living currently? Is there a repo? Is it safe to close the PR?

diasdavid:
Closing this PR in favour of tracking the dev of the tool at https://github.com/ipfs/pm/issues/237

flyingzumwalt:
We should revisit to confirm that we still want to proceed with this approach -- it was all motivated by the [pm process document](https://github.com/ipfs/pm/blob/master/drafts/project-management-process.md), which we never ratified, instead leaving it as just a draft. I suspect that we will want to adopt a modified version of the stuff in that doc (possibly simplified) and will still have use for labels-and-milestones for as long as we use github (won't need it if/when we migrate to gitlab)

PR #600, opened 2016-11-17
feat(swarm): update swarm.peers to new api

dignifiedquire:
Ref https://github.com/ipfs/js-ipfs-api/pull/427

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "form-data": "^2.1.2",
     "fs-pull-blob-store": "^0.4.1",
     "gulp": "^3.9.1",
-    "interface-ipfs-core": "^0.21.0",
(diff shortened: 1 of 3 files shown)


dignifiedquire:
@diasdavid I don't understand this, even though we do a connect before calling `swarm.peers` it seems the peerbook is still empty. Could you check that out what's happening there, as that does not seem to be related to the changes I made.

diasdavid:
@dignifiedquire got it. you were calling swarm.peers on peerB, it is true that peerA dialed to peerB, but peerB needs time to finish the identify, because until that is finished, it has no idea of has called it. Updating the test :)

diasdavid:
thanks for updating this @dignifiedquire going to merge into feat/awesome-dag-pb because both depend on an updated interface-ipld-core

PR #613, opened 2016-11-21
fix: sharness self check test

diasdavid:
@chriscool sharness was falling with:

```
not ok 4 - We expect this to fail # TODO known breakage
# still have 1 known breakage(s)
```

Looking at the code we have:

```
test_expect_failure "We expect this to fail" "
    test 1 = 2
"
```

Apparently it is not failing as it should? Is this a problem in the sharness library? 

I've commented that one, so that CI is happy

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
     "release": "gulp release",
     "release-minor": "gulp release --type minor",
     "release-major": "gulp release --type major",
-    "coverage-publish": "aegir-coverage publish"
+    "coverage-publish": "aegir-coverage publish",
+    "sharness": "make test"
   },
   "pre-commit": [
     "lint",
--- a/test/sharness/t0000-sharness.sh
+++ b/test/sharness/t0000-sharness.sh
@@ -23,9 +23,9 @@ test_expect_success "You can test for a specific exit code" "
     test_expect_code 42 return_42
 "
 
-test_expect_failure "We expect this to fail" "
-    test 1 = 2
-"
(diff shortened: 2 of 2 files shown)


diasdavid:
Now linting is failing, not because of linting, just `aegir-lint` gives up, seems to be a problem specific with js-ipfs, tried again in js-ipfs-api and all is good there.

chriscool:
This is a Sharness self test to check that `test_expect_failure` works.
This is not preventing any other tests from running or passing.
And it should not make the CI unhappy.
So there is nothing to fix here and I think it is good to keep that test.

If it really annoys you, then the best is perhaps to remove all the Sharness self tests.

If the CI unhappy, perhaps something else is going on?
Could you post a link to the CI log?

diasdavid:
@chriscool see here: https://travis-ci.org/ipfs/js-ipfs/jobs/177579581#L3292

diasdavid:
Ok, confirmed that linting failing (https://github.com/ipfs/js-ipfs/pull/613#issuecomment-261868813) was not introduced with this PR, it is a problem also in Master

chriscool:
Yeah, the linting is failing the build, not the sharness tests.

At the end of the log, there is something saying that "make test" exited with code 0, so the sharness tests didn't fail.

diasdavid:
@chriscool oh right, big confusion, I saw the linting failing after the sharness and I read it after different things. nvm then :)

PR #631, opened 2016-11-28
Deamon init msg

VictorBjelkholm:
Just output a similar message as go-ipfs if you run `ipfs daemon` before having init a repo.

PR #633, opened 2016-11-29
fix(cli): Tell user to init repo if not initialized when starting daemon

VictorBjelkholm:
From https://github.com/ipfs/js-ipfs/pull/631

Had to create PR with branch from ipfs/js-ipfs instead of my fork, since running Sauce Labs would fail otherwise. See https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

### DIFF
--- a/src/cli/commands/daemon.js
+++ b/src/cli/commands/daemon.js
@@ -16,6 +16,11 @@ module.exports = {
     console.log('Initializing daemon...')
     httpAPI = new HttpAPI(process.env.IPFS_PATH)
     httpAPI.start((err) => {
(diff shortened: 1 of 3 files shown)


VictorBjelkholm:
@diasdavid tests were failing because 1) I set the timeout to one second while Travis is a lot slower to run than that, so removed my manual timeout to use the default one instead, 2) Sauce Labs can't authenticate if the branch comes from a fork, so I had to push the branch to ipfs/js-ipfs instead

So now everything should work, please review again :)

diasdavid:
Thank you <3

PR #637, opened 2016-12-04
docs: update architecture overview and core architecture image, add more info to the readme

diasdavid:
@pgte hope this is useful to you :)


README.md
```diff
@@ -251,27 +251,70 @@ Every IPFS instance also exposes the libp2p API at `ipfs.libp2p`. The formal int
...
+
+##### Source code
+
+```Bash
+> tree src -L 2
+src                 # Main source code folder
+├── cli             # Implementation of the IPFS CLI
+│   └── ...
+├── http-api        # The HTTP-API implementation of IPFS as defined by http-api-spec
+├── core            # IPFS implementation, the core (what gets loaded in browser)
```
dignifiedquire: @diasdavid these really should be three different modules at this point

diasdavid: Almost entirely agree with the caveats of:

- This forces us to run the 5 sets of tests every time
- It gets a way to people install js-ipfs as they are used to do it with go-ipfs (install one thing, get it all)

dignifiedquire: My idea was to have `js-ipfs` (`ipfs` npm module) be a tiny wrapper that exposes the same functionality as it does now for end users. But code and tests would move into

- js-ipfs-core
- js-ipfs-cli
- js-ipfs-http

This way you get the same convenience but if you only want to use js-ipfs in the browser all you need is to install `js-ipfs-core`without all the cli + http baggage.

diasdavid: I understand, let's get to js-ipfs 1.0.0 first, then we factor out :)

PR #652, opened 2016-12-10
CLI - add `block put` flags and `block get` CID support

kumavis:
This adds flags to the `jsipfs block put <hash>` command
* `format` (`f`)
* `mhtype`
* `mhlen`

these commands and their descriptions are based on go-ipfs https://github.com/ipfs/go-ipfs/blob/2aded6785dca9b8ce27595a9698d38ce993cbc5e/core/commands/block.go#L118-L121

This also adds support for CIDs in `ipfs block get <cid>` and fixes the output of binary data


src/cli/commands/block/put.js
```diff
@@ -1,52 +1,92 @@
...
+      default: 'v0'
+    },
+    mhtype: {
+      describe: 'multihash hash function',
+      default: 'sha2-256'
+    },
+    mhlen: {
+      describe: 'multihash hash length',
+      default: undefined
+    }
```
diasdavid: @whyrusleeping in what case is mhlength used ?

https://github.com/ipfs/go-ipfs/blob/2aded6785dca9b8ce27595a9698d38ce993cbc5e/core/commands/block.go#L121

src/cli/commands/block/put.js
```diff
@@ -1,52 +1,92 @@
...
+    if (opts.format === 'v0') {
+      block.key(done)
+    } else {
+      block.key(opts.mhtype, done)
+    }
+    function done (err, _mhash) {
+      if (err) return cb(err)
+      mhash = _mhash
+      cb()
+    }
```
diasdavid: since you are doing waterfall, you can pass the `cb` directly to the block.key function and capture the hash in the waterfall sequence

kumavis: right - i previously needed the reference elsewhere, but that got factored out. will clean up.

PR #670, opened 2016-12-15
fix: update sharness tests to use right cli

diasdavid:
@chriscool I've updated the sharness tests to point to `jsipfs` instead of `ipfs`. Everything seems ok to me, just wanted your confirmation.

### DIFF
--- a/test/sharness/t0010-basic-commands.sh
+++ b/test/sharness/t0010-basic-commands.sh
@@ -12,11 +12,11 @@ test_expect_success "current dir is writable" '
 	echo "It works!" >test.txt
 '
 
-test_expect_success "ipfs version succeeds" '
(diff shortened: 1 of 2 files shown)


chriscool:
I don't think that is necessary. The Makefile already does:

```
../bin/ipfs:
	mkdir -p ../bin
	cd ../bin && ln -s ../../src/cli/bin.js ipfs
```

This means that when running tests "ipfs" will be in fact "jsipfs".

And with this we don't need to change tests which mean that hopefully in the end we will be able to have exactly the same test suite for jsipfs and for go-ipfs.

diasdavid:
sweet! :D Thank you for clarifying @chriscool, closing then :)

PR #682, opened 2016-12-23
Use new bitswap

diasdavid:


dignifiedquire:
:tada:

diasdavid:
Not so fast :(

Having trouble with the cli command wantlist, the http-api should be issuing a a list of CID base58 encoded, but it is sending that is a weird buffer (where it should be a wantlist entry). The interop tests are also not fully passing, but that might be just http-api related.

dignifiedquire:
right, I know what fails. Internally the wantlists are stored indexed by buffer.toString as that is kucb more efficient than base encoding, so the core command has to be updated to use the values cid and base encoded instead of reusing the keys

diasdavid:
everything is passing locally now, let's wait for CI :)

PR #687, opened 2016-12-26
Fix the error encounter when running node

yxliang01:
Done by updating the peer-info package. Please merge this quickly and make a new release to fix the problem that I have.


diasdavid:
@yxliang01 happy to update and patch release, but do note that since this is a patch update, a fresh npm install should solve your issue

PR #688, opened 2016-12-26
feat: dag get cli

kumavis:
Here is the skeleton for the `dag` cli command, and an implementation of `dag get <ref>` where `ref` is { `<cid>` or `<cid>/<path>` }.

Not sure what the formatting for output should be, just logging the result as-in for now.

Was unable to run the test due to some unrelated bug on test start, but should be correct.

### DIFF
--- a/test/cli/test-dag.js
+++ b/test/cli/test-dag.js
@@ -1,0 +1,18 @@
+/* eslint-env mocha */
+'use strict'
+
+const expect = require('chai').expect
+const repoPath = require('./index').repoPath
+const describeOnlineAndOffline = require('../utils/on-and-off')
+const ipfs = require('../utils/ipfs-exec')(repoPath)
+
+describe('dag', () => {
+  describeOnlineAndOffline(repoPath, () => {
+    it('get', () => {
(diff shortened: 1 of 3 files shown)


test/cli/test-dag.js
```diff
@@ -0,0 +1,18 @@
...
+'use strict'
+
+const expect = require('chai').expect
+const repoPath = require('./index').repoPath
+const describeOnlineAndOffline = require('../utils/on-and-off')
+const ipfs = require('../utils/ipfs-exec')(repoPath)
+
+describe('dag', () => {
+  describeOnlineAndOffline(repoPath, () => {
+    it('get', () => {
```
VictorBjelkholm: Should probably have a test-case as well when we try to resolve a hash that does not exists.

kumavis: Yeah I wanted to write more extensive tests but I couldn't even get them to run bc of some issue with a networking module

diasdavid:
@kumavis I was rebasing master on this branch to then realise this PR was made to a old branch and not against master.

Mind rebasing  and PR'ing against master?

kumavis:
replaced by https://github.com/ipfs/js-ipfs/pull/753

Issue #693, opened 2017-01-01
No routes for HTTP Gateway


harshjv:
Currently, HTTP gateway has no routes so even if `daemon` is listening to `9090` isn't useful.

    Gateway (readonly) is listening on: http://127.0.0.1:9090

Is there any progress on this part? I would like to help in implementing gateway functionality for `js-ipfs`

diasdavid:
Hi @harshjv, we haven't made any development with regards to the IPFS gateway in js-ipfs, however, a lot of the groundwork is done for the HTTP API, which shares a lot of the same endpoints with the gateway.

Having the HTTP Gateway in the JS implementation is not a core priority, so it might take a while until we take a stab at it, however, if you are available, that would be awesome and I'm happy to provide all the feedback you might need :)

harshjv:
Hey @diasdavid I can help in this issue as currently, I am using `js-ipfs` for [Buffet (Cross platform IPFS application with built-in managed IPFS-node)](https://github.com/harshjv/buffet) and in that, I've already implemented file `cat/get` gateway functionality with directory node browsing and mime-type support.

It can also add files with `wrap-with-directory` option (using empty buffer for parent directory)

It would be great if you can review these two files (and related other project files), and let me know if everything looks good to you so that I can port it to `js-ipfs`.

https://github.com/harshjv/buffet/blob/master/app/gateway/routes.js
https://github.com/harshjv/buffet/blob/master/app/utils/ipfs.js

Cheers 🍻

diasdavid:
Overall LGTM, but it will get more review as it will be easier once there is a PR against js-ipfs :)

harshjv:
Cool then. I'll send a PR soon :)

harshjv:
@diasdavid Hey, I've sent a PR #698 for this. :)

diasdavid:
Let's continue the discussion on your PR #698, @harshjv :)

PR #698, opened 2017-01-03
Add gateway route /ipfs/*

harshjv:
## Add gateway route `/ipfs/*`

* Stream content with proper `content-type` when file name is available
* Allow directory listing with the same style as `https://ipfs.io/ipfs/<some-dir-hash>`

### Preview

screen shot 2017-01-03 at 7 15 02 pm

### Related issue
#693


kumavis:
test fails bc of some gulp-coverage breakage

harshjv:
@kumavis Fixed indentation and removed TODO comment

kumavis:
Dont mind the security warning, its a known situation and is in an outdated copy of the build tool

kumavis:
I guess I'd also expect go-ipfs and js-ipfs to just serve the same app, so it can be iterated on as a separate project
> @diasdavid 
thoughts?

otherwise LGTM

diasdavid:
I won't have the time to review this soon. Nevertheless one thing that caught my eye is that there aren't any tests. Also, if we are going to push forward and add the gateway to js-ipfs, I would prefer to:

- rename the folder http-api to just http
- create two folders, one being api and the other being gateway
- move https://github.com/ipfs/js-ipfs/blob/master/src/http-api/index.js out to the http folder, so that it spawns the daemon and loads both api and gateway.

dignifiedquire:
One other note, the current code base does not use promises, and we would like to keep it that way for now.

kumavis:
> One other note, the current code base does not use promises, and we would like to keep it that way for now.

I think this is another point for breaking the ui out into its own project

diasdavid:
Yeah, there is the webui and this 'file viewer' thing. Both can be their own projects I believe, transforming the file viewer to a NWA instead of a server side rendered thing, so that it can be used easily by go and js-ipfs.

VictorBjelkholm:
@diasdavid bit unclear what the decision is for this PR. Should we focus on merging this or should we focus on splitting it out and integrate as external module?

diasdavid:
@VictorBjelkholm

harshjv:
@diasdavid Something like this? https://github.com/harshjv/ipfb
NPM: https://www.npmjs.com/package/ipfb

diasdavid:
@harshjv completed missed the last comment. https://github.com/harshjv/ipfb looks great, that file browser should be something that both go and js can use and not only js.

Could you handle these tasks first:

- [ ] https://github.com/ipfs/js-ipfs/pull/698#issuecomment-280313223
- [ ] https://github.com/ipfs/js-ipfs/pull/698#issuecomment-280337849
- [ ] Tests.

I would be comfortable merging this if these tasks were completed. We can figure out how to have 'one file-browser' for both implementations next. It is very probable that by having the Gateway working on js-ipfs, that we will develop an even better file browser since it will enable more JavaScripters to play with it :)

ya7ya:
@diasdavid @harshjv Hi there, Are there any plans to fix the minor conflict and merge this PR ??

diasdavid:
@ya7ya see above, there are more things to do than merge conflicts. Wanna help?

ya7ya:
@diasdavid Yeah, I'm currently swamped with work trying to figure out the problem with #946 which is desperately needed for a project i'm working on, but I can try to help by finishing the tasks mentioned in your [previous comment](https://github.com/ipfs/js-ipfs/pull/698#issuecomment-313717846)

ya7ya:
hey @diasdavid , I been working on the points you mentioned above to get this thing done, But there is an issue I thought you might be able to clarify.

The `go-ipfs` implementation [throws an error when you `cat` a directory](https://github.com/ipfs/go-ipfs/blob/e8477b50c907c1544f8a7e15334d6fb595baac19/core/corehttp/gateway_handler.go#L186), which is useful to figure out whether a path should be served or render the ui for the index. in `js-ipfs` this error is never triggered so figuring out the directory isn't straightforward.

Does `js-ipfs` cat implmentation has that `ErrIsDir` Error ?? or do I need to figure out another way to find that out ?

diasdavid:
Merging this PR into a new target branch `feat/gateway` so that @ya7ya and @harshjv can collaborate together on it (without having to PR the PRs)

PR #703, opened 2017-01-10
http-api - parseArgs - allow lookup by CID

kumavis:


diasdavid:
just rebased master onto this branch.

PR #704, opened 2017-01-10
http-api - block - get - return value raw

kumavis:
value may be a binary buffer, it should be returned as is instead of coerced into unicode

note: if request is sent from the browser, the requestor can set the `responseType` to get the result as an ArrayBuffer in order to accurately get the binary data
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data

### DIFF
--- a/src/http-api/resources/block.js
+++ b/src/http-api/resources/block.js
@@ -46,7 +46,7 @@ exports.get = {
         }).code(500)
       }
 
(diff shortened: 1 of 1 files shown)


diasdavid:
just rebased master onto this branch

PR #707, opened 2017-01-11
support trickle DAG builder on CLI and API

pgte:
Depends on https://github.com/ipfs/js-ipfs-api/pull/505


diasdavid:
Once there are tests added to interface-ipfs-core, the http-api code will need to understand that there are options now to be passed, other than adding that to the mix, this PR LGTM 👍 :)

diasdavid:
Updated the released deps. Waiting for CI to flash green :)

diasdavid:
@lindybrits once this one gets merged and release, it means that you can add big files through js-ipfs just fine like go-ipfs does :)

diasdavid:
🎉🎉🎉🎉🎉 right on!

Issue #711, opened 2017-01-16
Browser + Node.js + Go Interop Examples/Tutorials


haadcode:
As part of sprint https://github.com/ipfs/pm/issues/310, we will create examples for go-ipfs and js-ipfs interop.

The goal is to demonstrate the interop functionality by showing files and data moving between the two implementations.

This issue is to track the development and progress of the interop examples.

haadcode:
Gonna drop some design notes for these.

## Example 1 - Cat a file 

Required components:
- go-ipfs cli, js-ipfs + ipfs-daemon in the browser

Steps:
- have one browser peer open and a go-ipfs daemon running
- 'ipfs add another-file.txt'
- copy file hash
- paste file hash in browser peer
- observe file contents appear

Notes:
- I want this to be s single .html file with minimal html, CSS and JS library overhead
- It contains minimal amount of external dependencies
- It should be easy to read what happens in the code
- It should not use any external tools, so that running and packaging the example is effortless 
- The focus is on robust code and functionality, not on visuals
- The example code should be documented and presented as a tutorial

haadcode:
## Example 2 - File Manager 

This one is more advanced example and as per the current design, will also use orbit-db. We may or may not end up using it.

Demonstrates:
- ipfs.files.add, ipfs.files.cat, ipfs.files.ls
- ipfs.object.put, ipfs.object.get
- ipfs.pubsub

Required components:
- go-ipfs cli, js-ipfs + ipfs-daemon, orbit-db

Steps:
- have two browser peers open and a go-ipfs daemon running
- add a file via the first browser peer
- observe file appear in the second browser peer
- copy file hash
- 'ipfs cat <file hash>'
- 'ipfs add another-file.txt'
- copy file hash
- 'node addfile.js peerID another-file.txt'
- observe file appear in browser peers
- close browser peers
- open a new browser peer
- observe all files appear

Notes:
- Idea here is to demonstrate more IPFS functionality via orbit-db
- Can't add directories in the browser. That might change the expected UX.
- Same notes as for Example1 apply here in terms of what the focus is in the implementation (clean, simple code, well documented, few deps, etc.)

haadcode:
Example2 might be too advanced for our purposes for this sprint. It'd be really cool though...

VictorBjelkholm:
> Can't add directories in the browser. That might change the expected UX.

Should be possible to iterate through all the files and directories, so this one shouldn't be impossible to do. 

[...]

PR #716, opened 2017-01-18
tutorial: File Exchange - Transfer files between desktop and browser using go-ipfs, js-ipfs and orbit-db

haadcode:
This is a WIP PR for the second interop example. A File Manager in the browser, which eventually will be able to display files added in go-ipfs. It currently works only between browser js-ipfs nodes.

No need to review it yet, still working on it.


diasdavid:
Not getting this example to work:
image

diasdavid:
This example is looking very similar to PaperHub that @dignifiedquire made a while ago, it is interesting, but it the main goal is to have a file manager that can open directories and traverse to them, these directories should be added through go-ipfs.

I know you are still heavily working on this one and that you have the intent of making it the actual File Browser, so I will review it again when it is fulfilling that goal.

diasdavid:
Ah I added a file on firefox and it showed up in chrome:
image

And then, I added an extra one to Chrome, and the 3 files appeared in Firefox too:

image

diasdavid:
I believe that this 'injection' of orbit into the example is creating a distraction to the main goal, which is why developing it [incrementally, as suggested](https://github.com/ipfs/js-ipfs/issues/711#issuecomment-272811870), would have been a big win.

diasdavid:
@haadcode updated the name of this PR to reflect the decision made through IRC chat. One task that needs to also happen is:

```
12:01 <haad> daviddias: eating lunch, will get back to you in a few. re. go-ipfs aprt of the tutorial for file feed: I was planning on writing a small .js file that spawns a go-ipfs daemon and adds the file to ipfs+orbit-db.
12:02 <haad> daviddias: ok so re that ^, should object.get/object.put and pubsub interop work with go-ipfs<-->js-ipfs?
12:02 <haad> because if they won't, then we can't do that and we're better off doing the file browser
12:05 <@daviddias> haad: I would say pubsub interop: yes it works, but won't officially say it unless there is a test in interop for that - https://github.com/ipfs/js-ipfs/tree/master/test/interop, it only thing I can say is: It used to work, now something might have changed, we can't tell.
12:05 <@daviddias> That interop test should be part of the list of tasks for that interop example (File Feed)
```

haadcode:
The browserify-zlib-next dependecy problems have been solved. What is still needed is to merge this: https://github.com/ipfs/browserify-zlib-next/pull/23. cc @diasdavid @dignifiedquire

haadcode:
I'm fixing and updating everything right now to make sure this can be finished by someone.

dignifiedquire:
`browserify-zlib-next@1.0.1` was just released with the fix

haadcode:
Thank you @dignifiedquire 🎉

haadcode:
Can verify that everything works now with browserify-zlib-next 1.0.1. Hooray \o/

diasdavid:
TODOs left by @haadcode 

## TODO

- Need to merge https://github.com/ipfs/browserify-zlib-next/pull/23 and use the updated module. Otherwise the UI will error on zlib._transform.
- Make sure go-ipfs and the UI connect to each other (last time I checked they didn't)
- Use new signal server address everywhere (src/App.js, ipfe-daemon.js, ipfe-add.js)
- ipfe-add currently uses js-ipf(node.js) daemon. But in order to add files from go-ipfs daemon, it would need to use js-ipfd-ctl. A quick way would be to use ipfs-daemon (I would *highly* recommend to do so), otherwise need to add the init dance for js-ipfsd-ctl much like `src/ipfs.js`.

diasdavid:
- [ ] The cli tool to add files is pointing at local signalling server
- [ ] still uses ipfs-daemon

diasdavid:
Found that it works in the browser with no signalling server at all because the browser is mounting always on top of the same repo.

image

diasdavid:
image

Got the Desktop example to work with DNS

examples/file-feed/complete/src/browser-app/DataStore.js
```diff
@@ -0,0 +1,104 @@
...
+class DataStore extends EventEmitter {
+  constructor (options) {
+    super()
+    spawnNode(options, (err, node) => {
+      if (err) {
+        console.log(err)
+      }
+      this.ipfs = node
+      this.emit('ready')
+    })
```
diasdavid: Async operations in a constructor are an anti-pattern, a good source of errors

diasdavid:
@haadcode how did you manage to use WebPack@2? I'm missing to find any note and ipfs doesn't really work if bundled with WebPack@1 and as @dignifiedquire just mentioned, WebPack@2 will only come in the next version of create-react-app.

examples/file-feed/complete/src/browser-app/DataStore.js
diasdavid: Getting issues from this

image

diasdavid: ah got it, this always happens if we open without being on a feed which is controlled by the top level domain.

diasdavid:
Clicking the "number of peers" throws

image

diasdavid:
If a peer joins after files where added, it stays on 'Loading database' forever

image

examples/file-feed/.gitignore
```diff
@@ -0,0 +1,18 @@
...
+# production
+build
+
+# misc
+.DS_Store
+.env
+npm-debug.log
+
+# orbit-db
+orbit-db
```
diasdavid: Why do we need to ignore a folder named `orbit-db`?

examples/file-feed/README.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+You should get something like this:
+
+`TODO: Insert final screenshot here`
+
+## Step-by-step instructions
+
+`TODO`
+
```
diasdavid: @RichardLitt is on track to do this once the example is completely done. @RichardLitt I'll notify you when that happens ❤️

examples/file-feed/complete/config/webpack.config.dev.js
```diff
@@ -0,0 +1,134 @@
...
+    new webpack.HotModuleReplacementPlugin(),
+    new CaseSensitivePathsPlugin(),
+    new WatchMissingNodeModulesPlugin(paths.appNodeModules)
+  ],
+  node: {
+    fs: 'empty',
+    net: 'empty',
+    tls: 'empty'
+  }
+}
```
diasdavid: Can we validate that we actually need all of this configuration?

examples/file-feed/complete/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "file-feed",
```
diasdavid: I've renamed it to `file-feed` as I'm afraid of stamping something with a "InterPlanetary X" name that is not a project (i.e example/tutorial) that we support and make it have the same level of activity of a product. 

Pointing people to the 'file-feed tutorial' has a completely different tone than 'InterPlanetary FileExchange'

@haadcode hope you understand, we can chat about it and reach an agreement.

examples/file-feed/complete/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "0.0.0",
+  "bin": "src/ff-cli/bin.js",
```
diasdavid: I've also put the CLI under the bin path, for ease of use.

examples/file-feed/complete/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "devDependencies": {
+    "buffer": "^5.0.2",
```
diasdavid: I'm not 100% sure why we need to install `Buffer`, is this a issue?

examples/file-feed/complete/public/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link href="https://fonts.googleapis.com/css?family=Anonymous+Pro|Source+Code+Pro" rel="stylesheet">
```
diasdavid: Can we make all the CSS, including fonts, be loaded from CSS files?

examples/file-feed/complete/src/browser-app/App.js
```diff
@@ -0,0 +1,177 @@
...
+    this.setStatus('Starting IPFS')
+
+    // Initialize our DataStore, ie. start IPFS
+    dataStore = DataStore.init({
+      path: '/ipfs/' + new Date().toString(),
+      signalAddr: '/dns4/star-signal.cloud.ipfs.team'
+    })
+
+    dataStore.on('error', (e) => console.error(e))
+    dataStore.on('ready', () => {
```
diasdavid: `on` or `once`?

examples/file-feed/complete/src/browser-app/DataStore.js
```diff
@@ -0,0 +1,104 @@
...
+    this.feed.events.on('history', () => this.emit('update'))
+    this.feed.events.on('data', () => this.emit('update'))
+    if (this.timer) clearInterval(this.timer)
+    this.timer = setInterval(() => this._updatePeers(name), 1000)
+  }
+
+  getFile (hash) {
+    return new Promise((resolve, reject) => {
+      // Cat the contents of a multihash
+      this.ipfs.files.cat(hash)
```
diasdavid: cat doesn't look like the right operation, as the user might start dropping images or video.

examples/file-feed/complete/src/browser-app/DataStore.js
```diff
@@ -0,0 +1,104 @@
...
+    this.feed = this.orbitdb.feed(name, { cachePath: '/ipfd/ipfd2.db' })
+    this.feed.events.on('ready', () => this.emit('feed'))
+    this.feed.events.on('history', () => this.emit('update'))
+    this.feed.events.on('data', () => this.emit('update'))
+    if (this.timer) clearInterval(this.timer)
+    this.timer = setInterval(() => this._updatePeers(name), 1000)
+  }
+
+  getFile (hash) {
+    return new Promise((resolve, reject) => {
```
diasdavid: Back in the January meetings, we talked about using Promises in examples and we agreed at the table that we would use only callbacks for our examples. Is the use of Promises strictly necessary here?

examples/file-feed/complete/src/browser-app/DataStore.js
```diff
@@ -0,0 +1,104 @@
...
+          stream.on('error', (e) => console.error(e))
+          stream.on('data', (d) => {
+            bytes += d.length
+            buffer.push(d)
+            this.emit('load', hash, bytes)
+          })
+          stream.on('end', () => {
+            this.emit('load', hash, bytes)
+            resolve(buffer)
+          })
```
diasdavid: Use `bl` or `concat` for this

examples/file-feed/complete/src/browser-app/DataStore.js
```diff
@@ -0,0 +1,104 @@
...
+        })
+        .catch(reject)
+    })
+  }
+
+  addFiles (file) {
+    const addToIpfs = (name, content) => {
+      return this.ipfs.files.add([{
```
diasdavid: IPFS might not exist in this context due to the async operation in the constructor.

examples/file-feed/complete/src/browser-app/DataStore.js
```diff
@@ -0,0 +1,104 @@
...
+        path: name,
+        content: new Buffer(content)
+      }])
+    }
+
+    const addToOrbitDB = (file, type) => {
+      return this.feed.add({
```
diasdavid: `this.feed` doens't exist if we aren't in a path.

examples/file-feed/complete/src/ff-cli/daemon.js
```diff
@@ -0,0 +1,67 @@
...
+        output += files.join('\n') + '\n'
+        output += `------------------------------\n`
+        console.log(output)
+      }
+    }
+  }
+
+  setInterval(checkForPeers, 1000)
+  setInterval(checkForFiles, 1000)
+})
```
diasdavid: The `daemon` and the `cli` tool share a ton of code, do we really need too files for this?

diasdavid:
**feature request:** A very cool features would be to 'watch' a folder, and for every file to gets added to the folder, it syncs to the browser and for every file that is added to the browser, it syncs back to the folder.

examples/file-feed/complete/public/index.html
VictorBjelkholm: Probably we should just not use third-party fonts and set `font-family: sans-serif`, it's usually good enough.

examples/file-feed/README.md
```diff
@@ -0,0 +1,149 @@
...
+6. Once you see the output above, it means the file was added successfully!
+
+7. Now go back to the browser app and observe the file in the list of files.
+
+8. Click the file to open it.
+
+9. You have successfully added a file in go-ipfs and downloaded it to the browser.
+
+You can also add files to the browser app by dragging and dropping them. Once you do so, you should see the updated file list in the terminal running `ipfe-daemon`.
```
VictorBjelkholm: `ipfs-daemon`, not `ipfe-daemon` I'm assuming

examples/file-feed/README.md
```diff
@@ -0,0 +1,149 @@
...
+8. You should now see the two IPFS instances connected. You can verify it by observing the Peer view in the browser app and running the following command in the terminal:
+```
+ipfs swarm peers
+```
+
+### Add files to InterPlanetary File Exchange
+
+1. Add a file by running:
+```
+node ipfe-add hello-world ipfe-add.js
```
VictorBjelkholm: `ipfe` > `ipfs`

haadcode:
I'm back from vacation and will work to finish this today. Thanks for the feedback @diasdavid! Will respond to individual comments when I start the work.

haadcode:
I just messed a rebase. Nothing was lost, fortunately. Will fix everything as I work on this.

examples/file-feed/complete/src/browser-app/DataStore.js
haadcode: Indeed, this is not good. Will change it.

examples/file-feed/.gitignore
haadcode: OrbitDB writes its cache file to that directory when run with go-ipfs or js-ipfs from the CLI. This is per user and we shouldn't add it to the repo.

examples/file-feed/complete/config/webpack.config.dev.js
haadcode: Yeah, we need all this configuration. See my long comment below as to why.

Short: the webpack config file here is the create-react-app/react-script config file with the only difference that we add the zlib shim to it.

examples/file-feed/complete/package.json
haadcode: Good! I was thinking we could yargsify the CLI tool so that we can run 
`ff-cli daemon hello-world` and `ff-cli add hello-world file.txt` instead of having two separate programs. This would make the code sharing between the two also a lot easier. Unless you're against yargsifying it, I'll go ahead and do that today.

examples/file-feed/complete/package.json
haadcode: I'm not sure 100% either anymore. It was needed to force webpack to use the latest version (instead 3.x) to fix some of the errors we were seeing with browserify-zlib, but those might be gone now with the updated, correct browserify-zlib-next. I'll verify if this is still needed and remove it if not.

examples/file-feed/complete/public/index.html
haadcode: Using default fonts will result in different look on different platforms. Using a consistent look (and a nicer-than-system-default) font is desirable. Is there a strong reason to **not** use 3rd party fonts?

@diasdavid yeah, I'll move the fonts to the CSS file.

examples/file-feed/complete/src/browser-app/App.js
haadcode: `on` is fine in this case, DataStore is a singleton (ie. there's only one instance of it throughout the app).

examples/file-feed/complete/src/browser-app/DataStore.js
haadcode: I don't remember us talking about Promises, but sure, we can make these callbacks too.

examples/file-feed/complete/src/browser-app/DataStore.js
haadcode: Can you elaborate here why you don't think it's the right operation? It was working quite nicely with cat before I went on vacation and it works perfectly fine with images, videos, text, binary, etc.

examples/file-feed/complete/src/browser-app/DataStore.js
haadcode: I would prefer not adding a library to abstract the underlaying functionality of the stream. After all, this is an example and people will have different approaches to how they use streams, some will use `bl` or `concat`, some won't., so keeping it clear what ipfs returns is clearer in this case imo.

examples/file-feed/complete/src/browser-app/DataStore.js
haadcode: Indeed. Will fix the constructor problem.

examples/file-feed/complete/src/browser-app/DataStore.js
haadcode: You're right, feed is locally scoped. Will address this.

examples/file-feed/complete/src/ff-cli/daemon.js
haadcode: See my previous comment re. yargisyfying and combining the two program into one. Yes, we should share the code :)

examples/file-feed/README.md
haadcode: `ipfe` is correct (InterPlanetary File Exchange). However, we'll change the name, so this will need to be updated in the readme.

examples/file-feed/README.md
haadcode: Same as above.

haadcode:
Thanks for the thorough review @diasdavid and @VictorBjelkholm! I've commented on the individual comments above and for the rest, I'll comment here:

> Currently, the App isn't in a state where it is easy to use it, as it shows several bugs during its usage. 

Many of these problems were due to missing webpack config file. Let me explain:

There was a "postinstall" step in `package.json` that copied `config/webpack.config.dev.js` to `node_modules/react-scripts/config/webpack.config.dev.js`. react-script (which come from create-react-app) don't support custom webpack configs by its design but due to our need to shim `zlib` for the browser with `browserify-zlib-next`, I needed to extract the config file, add `'zlib': 'browserify-zlib-next'` (this is the only difference to the default react-script config file) and copy it to the `node_modules/...` directory. This is a hack to bypass the limitations, but I think it's the simplest solution for us. This should change when create-react-app supports Webpack 2 and we can add `"browser: { "zlib": "browserify-zlib-next" }` to `package.json` where Webpack should be able to pick it up (Webpack 1 or create-react-app doesn't seem to do that). If you have a better idea for doing this, let me know.

That said, I added the "postinstall" step back to `package.json` and it'll fix most of the problems you were experiencing. More details below.

> @haadcode how did you manage to use WebPack@2? I'm missing to find any note and ipfs doesn't really work if bundled with WebPack@1 and as @dignifiedquire just mentioned, WebPack@2 will only come in the next version of create-react-app.

create-react-app uses Webpack 1 and I couldn't get it to work with Webpack 2. However, it seems to work fine with create-react-app w/ Webpack1, at least I didn't come across any problems. If we see some problems, let's figure them out. I'm pretty sure we're fine.

> If there was alternative, shall we eject and update the react deps?

Let's hold off from ejecting for now. What 'eject' does in create-react-app is it creates all the config files, etc. and it'll add a bunch of configs and imo it gets really confusing (I tried that when I was trying to fix the browserify-zlib problem the other week).

> Clicking the "number of peers" throws

This was due to the Webpack config problems (aforementioned missing config file).

> If a peer joins after files where added, it stays on 'Loading database' forever

What is happening here is that once you've added some files to the feed, orbitdb writes its cache file to `/ipfd/ipfd2.db` (https://github.com/ipfs/js-ipfs/pull/716/files#diff-b5b9ff646dc8a4b3b1b9b3770d417401R25), the contents of that file is `{ <feed-name>: <latest-hash>`. Upon init, orbitdb will load the cache file and tries to load the contents of that hash (with `ipfs.object.get`). However, since the ipfs repo path gets set based on the timestamp or randomized upon every reload (https://github.com/ipfs/js-ipfs/pull/716/files#diff-4ffb5432ceeae06cd55f19d89f429ba2R32 or https://github.com/ipfs/js-ipfs/pull/716/files#diff-fde09c1a5971560dcb7c427fed1fff44R11) that content doesn't exist in the blockstore --> `ipfs.object.get` never returns (I quickly checked this with a debugger the other week but didn't dig into why it's not throwing an error).

My idea was that the feeds are persistent locally, ie. you can reload a feed and see it in a state it was when you last time closed it, so using the same ipfs repo path was what I used to achieve this. I'll change the code in the upcoming commits to use the same repo upon every load of the app.

> feature request: A very cool features would be to 'watch' a folder, and for every file to gets added to the folder, it syncs to the browser and for every file that is added to the browser, it syncs back to the folder.

This would be a cool feature indeed but out of scope for this demo imo. Let's create an "help-wanted" issue for this and a nice contribution possibility for someone in the community?

examples/file-feed/complete/package.json
haadcode: @diasdavid I'm totally fine changing the name! However, let's catch up on IRC today to discuss the name as I think we can find something more descriptive and less "loaded" word than "feed" (people will make assumption about what a fedd is that I think we can't provide in this example).

haadcode:
Pushed a bunch of changes.

TODO:
- yargsify CLI programs
  - ...and share code as much as possible between all the programs
- verify if Buffer is still needed as dep

examples/file-feed/complete/src/browser-app/DataStore.js
haadcode: This was fixed in https://github.com/ipfs/js-ipfs/pull/716/commits/a7b60493ed907594ea8cf0f2c011fe23310c4a0c. Constructor is not async anymore.

examples/file-feed/complete/src/browser-app/index.css
```diff
@@ -0,0 +1,24 @@
+@import url('https://fonts.googleapis.com/css?family=Anonymous+Pro|Source+Code+Pro');
```
VictorBjelkholm: Either we should include the fonts so it won't need to fetch them externally, or we go with the system fonts.

haadcode: I'd like to understand your reasoning here, can you elaborate? Why should we not fetch the fonts externally?

I'll add a fallback to the correct system fonts, though.

VictorBjelkholm: Sure. My thinking is when the connection to `googleapis.com` is bad or completely missing and the UI would be designed for those fonts to exists. When they fail to load, the page would default to using some other font and the layout wouldn't work the same way anymore, could lead to unexpected shuffling around the elements.

examples/file-feed/README.md
diasdavid: exactly, @VictorBjelkholm if you see the code, it is the CLI tool

examples/file-feed/complete/package.json
diasdavid: I'll be around for the rest of the day

examples/file-feed/complete/package.json
diasdavid: 👍 not against at all :)

diasdavid:
@haadcode re: post install script - sounds good, please add that as one of the caveats to the README tutorial

haadcode:
Last mile!

I'm now at a point where the browser app and CLI app are working together nicely with js-ipfs. The CLI app works between js-ipfs and go-ipfs, but browser <--> go-ipfs is not :/

I have go-ipfs daemon listening at `/ip4/127.0.0.1/tcp/9999/ws` (defined in the config file. When I try to connect to it from the browser (with `ipfs.swarm.connect`), they don't connect. However, running go-ipfs with logging on, I can see the connection request and see some "new stream" related log messages:
```
09:21:00.614 DEBUG       conn: listener <Listener <peer.ID ZGH8Ge> /ip4/0.0.0.0/tcp/9999/ws> got connection: /ip4/127.0.0.1/tcp/9999/ws <---> /ip4/0.0.0.0/tcp/9999/ws listen.go:170
09:21:00.631 DEBUG       conn: newSingleConn 0xc42028d900: <peer.ID ZGH8Ge> to <peer.ID > conn.go:47
09:21:00.635 DEBUG      secio: 1.1 Identify: <peer.ID ZGH8Ge> Remote Peer Identified as <peer.ID S6mzgh> protocol.go:176
09:21:00.717 DEBUG     swarm2: [<peer.ID ZGH8Ge>] network opening stream to peer [<peer.ID S6mzgh>] swarm_net.go:141
09:21:00.717 DEBUG     swarm2: Swarm: NewStreamWithPeer... swarm.go:282
09:21:00.717 DEBUG     swarm2: [<peer.ID ZGH8Ge>] network opening stream to peer [<peer.ID S6mzgh>] swarm_net.go:141
09:21:00.717 DEBUG     swarm2: Swarm: NewStreamWithPeer... swarm.go:282
09:21:00.717 DEBUG     swarm2: [<peer.ID ZGH8Ge>] network opening stream to peer [<peer.ID S6mzgh>] swarm_net.go:141
09:21:00.717 DEBUG     swarm2: Swarm: NewStreamWithPeer... swarm.go:282
09:21:05.134 DEBUG reprovider: Failed to provide key: failed to find any peer in table reprovide.go:59
```

However, `ipfs swarm peers` shows nothing on either side. I'm running go-ipfs and js-ipfs both from master.

@diasdavid @whyrusleeping any suggestions?

diasdavid:
@haadcode similar to the other example, could you make this one to have a box with the list of peers connected + a box to "connect to" multiaddrs?

haadcode:
> could you make this one to have a box with the list of peers connected + a box to "connect to" multiaddrs?

There is: click on the "searching for peers" and you're there.

whyrusleeping:
@diasdavid have you tested browser to go-ipfs with websockets? does this work normally?

diasdavid:
This is what I'm getting after fetching the last commit:

image

@whyrusleeping I will answer that with a 'no', since we don't have interop tests that check for that.

haadcode:
@diasdavid that's the Buffer 4.x error. I solved that by installing Buffer 5.x in js-ipfs/ (root).

diasdavid:
@haadcode can you make that be part of the PR?

haadcode:
>  can you make that be part of the PR?

Absolutely.

diasdavid:
@haadcode can you fresh install in your machine? I'm seeing this issue after a fresh install

image

diasdavid:
Got this when running the CLI

```
» node src/ff-cli/bin.js listen 'hello-world'
Swarm listening on /libp2p-webrtc-star/dns4/star-signal.cloud.ipfs.team/wss/ipfs/QmPvSs68vWoEQmjK4eBej5SkVKxMYZtWfx46g6ekE2JVwG
Database ready.

hello-world is empty

/Users/ground-control/code/js-ipfs/examples/file-feed/complete/src/ff-cli/receive-files.js:25
        if (err) { throw err }
                   ^

Error: connect ECONNREFUSED 0.0.0.0:9999
    at Object.exports._errnoException (util.js:1022:11)
    at exports._exceptionWithHostPort (util.js:1045:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1087:14)

```

diasdavid:
Got this when trying to add a file:

```
» node src/ff-cli/bin.js add hello-world package.json
Database ready.

Searching for peers...
(node:93623) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: connect ECONNREFUSED 127.0.0.1:5001
{ Error: connect ECONNREFUSED 127.0.0.1:5001
    at Object.exports._errnoException (util.js:1022:11)
    at exports._exceptionWithHostPort (util.js:1045:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1087:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5001 }
{ Error: connect ECONNREFUSED 127.0.0.1:5001
    at Object.exports._errnoException (util.js:1022:11)
    at exports._exceptionWithHostPort (util.js:1045:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1087:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5001 }
{ Error: connect ECONNREFUSED 127.0.0.1:5001
    at Object.exports._errnoException (util.js:1022:11)
    at exports._exceptionWithHostPort (util.js:1045:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1087:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5001 }
{ Error: connect ECONNREFUSED 127.0.0.1:5001
    at Object.exports._errnoException (util.js:1022:11)
    at exports._exceptionWithHostPort (util.js:1045:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1087:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5001 }

```

haadcode:
@diasdavid for the errors above, see the README for the instructions (at the top). You need to change the code atm. to connect to your given go-ipfs address (they're hardcoded in the CLI)

diasdavid:
@haadcode please make it so that it spawns a go-ipfs daemon if it needs it, we need to keep these examples/tutorials as smooth as possible.

haadcode:
>  please make it so that it spawns a go-ipfs daemon if it needs it, we need to keep these examples/tutorials as smooth as possible.

I won't have time to do that unless I can use ipfs-daemon.

haadcode:
Btw. once we're done and ready to merge, we should squash all these commits to one with a nice commit message as per our guidelines.

diasdavid:
> Btw. once we're done and ready to merge, we should squash all these commits to one with a nice commit message as per our guidelines.

Could you squash now? If we get into the habit of pushing clean commit messages, it will be easier to keep it consistent

examples/file-feed/complete/src/browser-app/DataStore.js
diasdavid: Thank you

diasdavid:
Hey @pgte it seems that this example froze in time, however, it is pretty valuable.

Could you take a look and make it work just with IPFS primitives?

pgte:
@diasdavid sure, I'll look at this once I'm done with [ipfs-level](https://github.com/pgte/ipfs-level).

diasdavid:
Closing this PR as the example was never developed further and since then there were a ton of advancements in js-ipfs and orbit

PR #726, opened 2017-01-26
First pass looking at interop example tutorial

RichardLitt:
Started rewriting it. 

@diasdavid I'm having issues with npm start - it doesn't work at `8080`, but at `http://127.0.0.1:12345`, and I'm not sure I am running the right thing. Is that the right thing? Should I be in `/js-ipfs/examples/transfer-files/complete`?

### DIFF
--- a/examples/transfer-files/README.md
+++ b/examples/transfer-files/README.md
@@ -51,7 +51,7 @@ TODO: Insert final screenshot here
 
 ## Step-by-step instructions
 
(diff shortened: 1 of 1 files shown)


diasdavid:
It does run at 12345 now, I changed it, simply because 8080 is a poor port choice, it is the default that go-ipfs uses.

Yes, that is the example you should be running :)

PR #732, opened 2017-01-28
fix: ipfs.id does not double append ipfs/<id> anymore

diasdavid:


src/core/components/id.js
```diff
@@ -1,30 +1,27 @@
...
-        }).sort(),
-        agentVersion: 'js-ipfs',
-        protocolVersion: '9000'
-      })
-    }
+    setImmediate(() => callback(null, {
+      id: self._peerInfo.id.toB58String(),
+      publicKey: self._peerInfo.id.pubKey.bytes.toString('base64'),
+      addresses: self._peerInfo.multiaddrs.map((ma) => {
+        if (mafmt.IPFS.matches(ma)) {
```
VictorBjelkholm: Hm, we should've extract this logic and put it under test as well, just to be sure... Sorry for the late notice 📦

PR #733, opened 2017-01-28
test: refactor for test sanity

diasdavid:


package.json
```diff
@@ -10,7 +10,7 @@
     "libp2p-ipfs-nodejs": "libp2p-ipfs-browser",
     "./src/core/default-repo.js": "./src/core/default-repo-browser.js",
     "./src/core/components/init-assets.js": false,
-    "./test/utils/temp-repo.js": "./test/utils/temp-repo-browser.js",
+    "./test/utils/create-repo-node.js": "./test/utils/create-repo-browser.js",
```
VictorBjelkholm: can't this just be create-repo rather?

test/utils/create-repo-node.js
```diff
@@ -14,7 +14,7 @@ function createTempRepo (repoPath) {
 
   repo.teardown = (done) => {
     clean(repoPath)
-    done()
+    setImmediate(() => done())
```
VictorBjelkholm: Why is this needed here now?

VictorBjelkholm: Should probably explain in a code-comment with reason

diasdavid: It is just a safe guard, a best practice

package.json
diasdavid: it could, I just standardize it because we always tag `node` and `browser` everywhere to distinguish the two, if distinction needs to happen

VictorBjelkholm: Fair enough!

test/utils/create-repo-node.js
VictorBjelkholm: 👍

PR #735, opened 2017-01-29
feat: bootstrap as an option

diasdavid:
needs: 
- https://github.com/ipfs/js-libp2p-ipfs-nodejs/pull/64
- https://github.com/ipfs/js-libp2p-ipfs-browser/pull/145

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,7 @@ Commands:
 - default repo location: `~/.jsipfs` (can be changed with env variable `IPFS_PATH`)
 - default swarm port: `4002`
 - default API port: `5002`
+- default Bootstrap is off, to enable it set `IPFS_BOOTSTRAP=1`
 
 ### HTTP-API
 
--- a/src/cli/commands/swarm/peers.js
+++ b/src/cli/commands/swarm/peers.js
@@ -2,6 +2,8 @@
 
 const utils = require('../../utils')
 const debug = require('debug')
+const mafmt = require('mafmt')
+const multiaddr = require('multiaddr')
 const log = debug('cli:object')
 log.error = debug('cli:object:error')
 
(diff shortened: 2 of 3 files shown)


VictorBjelkholm:
I'm a bit worried about using `process.env` in the browser bundle, with risk of confusing people. Historically, browser javascript has not cared about environment variables, my only known example of environment variables in the wild is React which uses `NODE_ENV` to determine if it should do the type checking of props and other things, only if it's not set to `production`. 

In webpack it's relatively simple to inject `process.env`, and without bundler you could simply create a global object before loading the script. I think we just have to make sure we document how you can change them in both cases, with bundler and without.

diasdavid:
Understood, as noted in IRC:

```
16:46 <@victorbjelkholm> daviddias: ah, that's cool
16:47 <@victorbjelkholm> daviddias: process.env in the browser bundle, not sure about that, thinking we maybe should try to figure out an alternative way
16:47 <@daviddias> suggestions?
16:47 <@victorbjelkholm> experimental config object as part of the config or the options?
16:48 <@victorbjelkholm> not sure really but thinking it's gonna confuse people
16:49 <@daviddias> The thing with that is that it creates the pattern of passing down special config options everywhere for experimental things
16:51 <@victorbjelkholm> yeah, and surely that's bad. Depends on what we think is worse, passing down config options as feature flags, or environment variable which browsers doesn't really used but can be hacked in by creating global variables
16:52 <@daviddias> think I got the bug, somewhere we are using a old version of the cid package
16:52 <@daviddias> victorbjelkholm: I believe that `IPFS_BOOTSTRAP=1 npm run build` will pick it up
16:52 <@daviddias> and since most browser apps are always being bundled
16:52 <@daviddias> it means it always picks up that
```

Note that this is intended to be a quick way to try out Bootstrap, until we have a full compatible muxer, after that it should be always on.

Nevertheless, we should figure out a way to have experimental features that are enabled by some sort of 'env variables' everywhere.

diasdavid:
For exploration - https://github.com/ipfs/js-ipfs/issues/738

PR #746, opened 2017-01-31
feat(dag): basics (get, put)

diasdavid:
Needs:
- https://github.com/ipfs/interface-ipfs-core/pull/112


diasdavid:
Going to remove the `rm` call, as that is handled by pinning.

Issue #556, opened 2016-10-29
Improving init (aka, how do I get a js-ipfs node instance!)


diasdavid:
There has been a lot of feedback when it comes to the API for creating an IPFS node instance, some feedback from myself included. It is true, the process is a bit cumbersome, instead of having something beautiful as

``` JavaScript
// # Option 1
const options = {
  init: true
  online: true
}
const node = new IPFS(options)
```

We have:

``` JavaScript
const repo = <IPFS Repo instance or repo path>
const ipfs = new IPFS(repo)
ipfs.init({ emptyRepo: true, bits: 512 }, (err) => {
   if (err) { throw err }
   ipfs.load((err) => {
     if (err) { throw err }
     ipfs.goOnline((err) => {
       if (err) { throw err }
       // finally something you can work with!
   })
})
```

Yes, it is not nice, but why is it this way right now? **Why can't it be `sync`**, well for some reasons:
- `init` needs to get the repo started if it doesn't exist, since IPFS Repo use pull-blob-store adapters and these are async (because a Repo can be a network storage), this call needs to be async
- `load` is when we take a repo and load all the config values. This also can't be sync because we only store the Private Key on the config and the generation of the PublicKey needs to be async (Thanks WebCrypto! see more at: https://github.com/ipfs/js-ipfs/pull/485))
- `goOnline` this is where we turn on bitswap, it means that we are ready to fetch and serve blocks from and to other nodes.

This can be way more nicer, however, starting from documentation of this methods, examples and explaining what the functions are doing inside the readme.

Another thing to consider is making the IPFS instance an event emitter with a 'ready' event, so that `Option 1` can be achieved by loading everything as options and emitting a `ready` event.
### Other things to consider
- It would be sweet if the options passed to init or to new IPFS(options) would also accept values to overload the config with. Right now we always create the repo and then use the [config API](https://github.com/ipfs/interface-ipfs-core/tree/master/API/config), it works but this option would reduce the steps, making code easier to read.
- `goOnline` should be bitswap.on and `bitswap.off`, as we will need soon to be able to enable and disable things like the `dht`.

I would love to hear everyone's thoughts on this, I might be missing something that might get us to a even nicer start process, looking forward for feedback :) Thank you!

dignifiedquire:
I have seen apis where it is done in this way: 

```js
const ipfs = new IPFS(options)

ipfs.on('ready', () => {
  // We are ready to do things
})

ipfs.on('error', (err) => {
[...]

PR #749, opened 2017-02-01
feat(breaking change): experimental config options

diasdavid:
As recommended by @whyrusleeping, we need to pubsub behind a `EXPERIMENTAL` option in the config options.

Adding config options will cause a breaking cause in the API of creating a node, but it was a step we were going to make for #556 anyway.


src/core/components/index.js
```diff
@@ -0,0 +1,18 @@
+exports.goOnline = require('./go-online')
```
dignifiedquire: let's not duplicate `exports` everywhere

```js
module.exports = {
  goOnline: require('./go-online'),
  ...
}
```

README.md
```diff
@@ -200,7 +200,12 @@ The HTTP-API exposed by the js-ipfs daemon follows the [`http-api-spec`](https:/
 const repo = <IPFS Repo instance or repo path>
 
 // Create the IPFS node instance
-const node = new IPFS(repo)
+const node = new IPFS({
+  repo: repo,
+  EXPERIMENTAL: {
```
dignifiedquire: Do we have to upcase this? I think `experimental` is enough. What would be good is to print something to the console when this is enabled.

src/core/components/index.js
diasdavid: Yeah, that way we have to worry about those commas, the way it is right now doesn't

README.md
diasdavid: I like making things very OBVIOUS :D

dignifiedquire: BUT DO YOU REALLY NEED TO SCREAM AT ME?

src/core/components/index.js
dignifiedquire: Not sure what commas you mean

diasdavid: if it is inside an object, for every add/delete, we get the 'no dangling comma situation'

README.md
diasdavid: I'll add your log :)

diasdavid: added your log :)

src/core/components/index.js
dignifiedquire: we have a lot of mixed and unclean diffs, not sure this makes any **diff** ;)

PR #753, opened 2017-02-02
cli - add 'dag get' command

kumavis:


src/cli/commands/dag/get.js
```diff
@@ -0,0 +1,47 @@
...
+      const cidString = refParts[0]
+      const pathString = refParts.slice(1).join('/')
+      const cid = new CID(cidString)
+
+      let lookupFn
+      if (pathString) {
+        lookupFn = () => ipfs.dag.resolve(cid, pathString)
+      } else {
+        lookupFn = () => ipfs.dag.get(cid)
+      }
```
diasdavid: @whyrusleeping convinced me that we should merge this two funcs yesterday, will be writing something explaining what will happen in the next hour, but yeah, for now 👌🏽

PR #757, opened 2017-02-07
docs: add first release checklist

diasdavid:
@dignifiedquire @VictorBjelkholm @haadcode @RichardLitt what else should we include in the release check list?

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,21 @@
+# Release checklist
+
+> List of actions to perform before issuing a new js-ipfs release
+
+- Robustness and quality
+  - [ ] Ensure that all tests are passing, this includes:
+    - [ ] unit
+    - [ ] interop
+    - [ ] sharness
+  - [ ] Run tests of the following projects with the new release:
+    - [ ] orbitdb
+    - [ ] orbit-core
+    - [ ] ipfs-log
+- Documentation
+  - [ ] Ensure that README.md is up to date
+  - [ ] Ensure that all the examples run
+- Communication
+  - [ ] Create the release issue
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,14 @@
...
+> List of actions to perform before issuing a new js-ipfs release
+
+- [ ] Ensure that all tests are passing, this includes:
+  - [ ] unit
+  - [ ] interop
+  - [ ] sharness
+- [ ] Run tests of the following projects with the new release:
+  - [ ] orbitdb
+  - [ ] ipfs-log
+- [ ] Announcements (both pre-release and post-release)
```
dignifiedquire: This should include a standard message + account we are using.

RichardLitt:
Update README?

dignifiedquire:
> Update README?

with what changes?

RichardLitt:
Thought there was a version hardcoded in it. Doesn't look like it. Yeah, I can't think of anything to add.

VictorBjelkholm:
"Make sure changelog is updated" feels like something that is missing.

We should clearly define that "issuing a new js-ipfs release" means publishing to npm.

Once we have cross-platform workers in jenkins, we should also make sure that the tests passes on all the supported platforms and nodejs versions.

Maybe making sure the README.md is always up-to-date makes sense to put in the release checklist as well.

RELEASE.md
diasdavid: 👍

diasdavid:
- [ ] add orbit-core to the list

diasdavid:
There we go :)

PR #770, opened 2017-02-22
Allow IPFS object to be created without supplying configOpts

raksooo:
Allow IPFS object to be created without supplying configOpts to prevent the following code, from the readme, from failing:
```Javascript
const IPFS = require('ipfs')

const node = new IPFS()
```

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -11,6 +11,7 @@ const components = require('./components')
 
 class IPFS {
   constructor (configOpts) {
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thanks :octocat:

VictorBjelkholm:
We should probably have at least one test with the constructor without any options. This should have been failing tests before the publish.

dignifiedquire:
@VictorBjelkholm good point

PR #772, opened 2017-02-24
yarn.lock

dignifiedquire:
For those who want it, making installs a lot faster.


dignifiedquire:
CI breaks because of: https://github.com/ipfs/js-ipfs/issues/781

VictorBjelkholm:
@dignifiedquire is https://github.com/ipfs/js-ipfs/issues/781 blocking this from getting merged? Should not be compatible no matter if we have a yarn.lock or not.

dignifiedquire:
@VictorBjelkholm #781 blocks this because ci breaks in this PR because of it

diasdavid:
Going to close as it is already out of date. We will chat about yarn next week :)

PR #714, opened 2017-01-16
tutorial: Transfer Files 

haadcode:
This PR will add a basic interop example as per https://github.com/ipfs/js-ipfs/issues/711.

PR #774, opened 2017-02-24
Tutorial: transfer-files (previsously #714)

dignifiedquire:
Ref #714


test/cli/bitswap.js
```diff
@@ -13,7 +13,7 @@ describe('bitswap', () => runOn((thing) => {
     ipfs('block get ' + key)
       .then(() => {})
       .catch(() => {})
-    setTimeout(done, 100)
+    setTimeout(done, 800)
```
diasdavid: why the need to increase the timeout?

dignifiedquire: because it randomly fails on CI due to things being slow

dignifiedquire:
@diasdavid https://github.com/ipfs/js-ipfs/pull/772 but needs updating probably

diasdavid:
@dignifiedquire got it, could you remove the yarn.lock from here then?

VictorBjelkholm:
Worth to note is that this is **not** ready to be merged, the document at `examples/transfer-files/README.md` is still unfinished.

diasdavid:
Thanks for bringing that up @VictorBjelkholm 

# TODO

- [ ] Peers list have a comma between every peer that is not supposed to be there
- [ ] Multiaddr input is too wide and goes outside the containing box
- [ ] Refactor code + add valuable comments
- [x] Have .get actually work with just one file 
- [ ] Fix drag-and-drop not closing popup when ending up not dropping file in UI
- [ ] Finish tutorial in readme.md

VictorBjelkholm:
>  Multiaddr input is too wide and goes outside the containing box

Should not happen anymore, and I remember when Friedel demod multiplex by using this demo, it didn't happen anymore. But I'll verify to make sure. 

Seems otherwise to capture everything that is missing, thanks for that 👍

Issue #773, opened 2017-02-24
Browser: Reloading page with static repoName throws error


aphelionz:
Steps to reproduce:

Run this code in the browser to connect to a repo with a static name. In this case "pastebin"

```
const repoName = "pastebin";

let ipfs = new Ipfs({
  repo: repoName,
  EXPERIMENTAL: {
    pubsub: false
  }
});
 
const startApp = (err) => {
  if(err) throw err;
  console.log("good to go")
};

const goOnline = () => ipfs.goOnline(startApp);
const loadIpfs = () => ipfs.load(goOnline);

ipfs.init({ emptyRepo: true, bits: 2048 }, loadIpfs);
```

Expected Behavior: console should log "good to go" every time
Actual Behavior: the console logs "good to go" the first time, and throws the error the second time

This is related to IndexedDB. It seems ipfs.init doesn't like existing repos.

It also seems that emptyRepo has no effect, as this error occurs when it is true or false. Upon further inspection emptyRepo is used to "seed" the database with the initial blocks.

dignifiedquire:
This is somewhat expected, but I agree the api is currently not easy to understand.

- `ipfs.init` corresponds `ipfs init` on the cli, which has the same behaviour, it will error out if a repo already exists in its place. The concept here is, this should only be called once, and on subsequent calls it's not needed anymore. 
- `emptyRepo` as you wrote, controls if there are some seed files put into the repo, not if something should be overwritten. 

What to do: 
- Currently the expectation is that the application using ipfs, handles detection if there is already a repo initialized, and if not to call `init`
- There is an internal method `ipfs._repo.exists((err, exists) => {})` which can be used to check if a repo was initialized, it might be good to expose this properly and document it.

aphelionz:
Thanks! I will try `ipfs._repo.exists` and report back.

aphelionz:
@dignifiedquire that worked. How can I help getting it exposed and documented?

dignifiedquire:
How about exposing it to the main `ipfs` object in `src/core/index.js`, so it can be used like this

```js
ipfs.repoExists(callback)
```

diasdavid:
We kind of mimicked the behaviour of go-ipfs in this one, but I can the value of having a direct call to check for a boolean instead of checking for an error.

[...]

PR #776, opened 2017-02-24
Initial commit for repoExists

aphelionz:
I've created an extra utility function based on @dignifiedquire and my conversation in https://github.com/ipfs/js-ipfs/issues/773

Happy to fix any issues or write up any required documentation here.

Also, we should discuss what we should do with `utils.ifRepoExists` and `utils.repoExists` which are very similar but very differently behaving functions.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -17,7 +17,7 @@ exports.ifRepoExists = (repo, cb) => {
 exports.repoExists = (self) => {
   return (cb) => {
     self._repo.exists((err, exists) => {
-      if(err) {
+      if (err) {
         return cb(err)
       }
 
--- a/test/core/init.spec.js
+++ b/test/core/init.spec.js
@@ -89,6 +89,7 @@ describe('init', () => {
       expect(err).to.not.exist
 
       ipfs.repoExists((err, exists) => {
+        expect(err).to.not.exist
         expect(exists).to.be.true
         done()
       })
@@ -97,8 +98,9 @@ describe('init', () => {
 
   it('checks if nonexistant repo exists', (done) => {
     ipfs.repoExists((err, exists) => {
+      expect(err).to.not.exist
       expect(exists).to.be.false
       done()
-    });
+    })
   })
 })


src/core/utils.js
```diff
@@ -14,4 +14,16 @@ exports.ifRepoExists = (repo, cb) => {
...
+  return (cb) => {
+    self._repo.exists((err, exists) => {
+      if (err) {
+        return cb(err)
+      }
+
+      cb(err, exists)
+    })
+  }
+}
```
diasdavid: Let's converge repoExists and ifRepoExists and make the converged function `async` (with callbacks and a promisify wrapper), the reason why it needs to be async is because the repo might be over a wire (like a S3 bucket).

diasdavid:
@aphelionz could you also please follow the commit message format? https://github.com/ipfs/community/blob/master/js-project-guidelines.md#commit-message-format (I know this might feel annoying, but it helps us create awesome changelogs! :))

src/core/index.js
```diff
@@ -62,6 +63,7 @@ class IPFS {
     this.bitswap = components.bitswap(this)
     this.ping = components.ping(this)
     this.pubsub = components.pubsub(this)
+    this.repoExists = utils.repoExists(this)
```
VictorBjelkholm: Instead of adding a new utility function, we could just assign `this.repoExists` to `repoInstance.exists` and all should work. Would be less duplicated code + tests.

aphelionz: I'll give that a shot. The test should still run agains this.repoExists though, right?

VictorBjelkholm: Yeah, I would say we should still test it, to be sure. Better safe than sorry

src/core/utils.js
diasdavid: Hi @aphelionz , how is the development of this feature?

aphelionz: @diasdavid Gonna push some more commit(s) today.

diasdavid: Let me know if you need any help :)

diasdavid:
Is there still an interest in this feature?

PR #782, opened 2017-03-09
fix(files.add): error on invalid input

diasdavid:
There was a new test added for js-ipfs-api that needed a fix here too.


VictorBjelkholm:
If this is something that should have the same error in the same condition in both js-ipfs as in js-ipfs-api, it would make sense for us to put tests for this in `interface-ipfs-core`.

diasdavid:
@VictorBjelkholm correct thinking, it was in fact a test on interface-ipfs-core that caught it because that is where the test was added.

PR #783, opened 2017-03-09
More dag.api goodness

diasdavid:
Follows https://github.com/ipfs/interface-ipfs-core/pull/123

Needs:

- https://github.com/ipld/js-ipld-resolver/pull/80


src/core/components/dag.js
```diff
@@ -34,6 +36,42 @@ module.exports = function dag (self) {
...
+      }
+
+      pull(
+        self._ipldResolver.treeStream(cid, path, options),
+        pull.collect((err, results) => {
+          if (err) {
+            return callback(err)
+          }
+          callback(null, results)
+        })
```
dignifiedquire: you can just run `pull.collect(callback)` instead

diasdavid: good catch! (you know when you open a callback for testing? :P)

dignifiedquire: never :P

Issue #474, opened 2016-09-10
CLI should accept content from stdin


VictorBjelkholm:
The js-ipfs should, like the go-ipfs cli, accept input from stdin and add it correctly.
## go-ipfs

```
$ echo "hello" | IPFS_PATH=~/.ipfs-go ipfs add
added QmZULkCELmmk5XNfCgTnCyFgAVxBRBXyDHGGMVoLFLiXEN QmZULkCELmmk5XNfCgTnCyFgAVxBRBXyDHGGMVoLFLiXEN
```
## current js-ipfs

```
$ echo "hello" | IPFS_PATH=~/.ipfs-js node src/cli/bin.js files add
src/cli/bin.js files add <file>

Options:
  --help           Show help                                           [boolean]
  --recursive, -r                                     [boolean] [default: false]

Not enough non-option arguments: got 0, need at least 1
```

diasdavid:
Hey @VictorBjelkholm mind adding a test case for this showing the failure?

diasdavid:
@gpestana has proposed a fix here: https://github.com/ipfs/js-ipfs/pull/785

Let's follow up there.

PR #785, opened 2017-03-11
fix: CLI should accept content from stdin #474)

gpestana:
Adds support for unix-like piping in js-ipfs.

e.g. `echo "hello" | IPFS_PATH=~/.ipfs-go ipfs add` is handled as expected

### DIFF
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -2,6 +2,7 @@
 
 'use strict'
 
+require('pipe-args').load()
 const yargs = require('yargs')
 const updateNotifier = require('update-notifier')
 const readPkgUp = require('read-pkg-up')
--- a/test/cli/files.js
+++ b/test/cli/files.js
@@ -127,6 +127,23 @@ describe('files', () => runOnAndOff((thing) => {
       })
   })
 
+  it('add with piped argument', () => {
+    // echo 'src/init-files/init-docs/readme' | jsipfs files add
+    return ipfs('files add', { piped: 'src/init-files/init-docs/readme' })
+      .then((out) => {
+        expect(out)
+          .to.eql('added QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB readme')
+      })
+  })
+
+  it('add with extra piped argument', () => {
+    return ipfs('files add', { piped: 'src/init-files/init-docs/readme' })
+      .then((out) => {
+        expect(out)
+          .to.eql('added QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB readme')
+      })
+  })
+
   it('cat', () => {
     return ipfs('files cat QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB')
       .then((out) => {
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "peer-book": "~0.3.1",
     "peer-id": "~0.8.2",
     "peer-info": "~0.8.3",
+    "pipe-args": "^1.0.1",
     "promisify-es6": "^1.0.2",
     "pull-file": "^1.0.0",
     "pull-paramap": "^1.2.1",
--- a/test/utils/ipfs-exec.js
+++ b/test/utils/ipfs-exec.js
@@ -29,11 +29,24 @@ module.exports = (repoPath, opts) => {
 
   function ipfs () {
     let args = Array.from(arguments)
-    if (args.length === 1) {
+    let pipedArgs
+
+    _.map(args, arg => arg.piped ? (pipedArgs = arg.piped) : '')
+
+    if (args.length === 1 || args.length === 2 && pipedArgs) {
       args = args[0].split(' ')
     }
 
     const cp = exec(args)
+
+    // Passes content of pipedArgs to childProcess as if jsipfs had been called
+    // with piped arguments
+    if (pipedArgs) {
+      cp.stdin.setEncoding('utf-8')
+      cp.stdin.write(`${pipedArgs}\n`)
+      cp.stdin.end()
+    }
+
     const res = cp.then((res) => {
       // We can't escape the os.tmpdir warning due to:
       // https://github.com/shelljs/shelljs/blob/master/src/tempdir.js#L43


gpestana:
Failing in the module used to parse the piped arguments. Looking into it.

What do you guys think about the testing approach?

src/cli/bin.js
```diff
@@ -2,6 +2,7 @@
 
 'use strict'
 
+require('pipe-args').load()
```
diasdavid: Cool module!

It is doing a bit more than what we want, though. If enabled, every command will take stdin->args, which opens space for other errors.

VictorBjelkholm: Yeah, but it's not only add that supports stdin. You can find a list of the commands that supports stdin and which argument it applies to here: https://github.com/ipfs/go-ipfs/search?utf8=%E2%9C%93&q=EnableStdin

We should probably have something similar for js-ipfs.

gpestana: @diasdavid being /dev/stdin a symbolic link to /proc/self/fd/0 (and /proc/self a symlink to the running process), we want to grab whatever it is there and pass it as argument, right? The problem I see is `echo "file.sx file.json" | ipfs add` will not work since the input 'filesx file.json' will be passed as one string. Or I am missing something else?

diasdavid: there are commands like `ipfs swarm peers` that should not take input from stdin

gpestana: @VictorBjelkholm yep, the committed tests are just as proof of concept. Once the feature is polished, I will add for all the tests for commands that must support stdin redirection. The link is super helpful, thanks!!

gpestana:
@diasdavid added the necessary changes for the pipe-args to copy the stdin to process.argv in selected commands. It could also be done explicitly in the cli.js without using the pipe-args `opts` object.

There was a fix in pipe-args module related to the `Error: ENXIO, no such device or address '/dev/stdin'` error that pop up in the previous tests.

Let me know what you think so I can go ahead and add more tests or change what's needed :)

src/cli/bin.js
```diff
@@ -2,10 +2,19 @@
...
 
-const yargs = require('yargs')
+const pipe = require('pipe-args')
 const updateNotifier = require('update-notifier')
 const readPkgUp = require('read-pkg-up')
 
+const enableStdin = [
+  'files', 'path', 'object data', 'ref', 'domain-name', 'key', 'ipfs-path',
+  'name', 'address', 'data', 'peer', 'recursive', 'default-config', 'peer ID'
+]
```
diasdavid: what is domain name?

gpestana: https://github.com/ipfs/go-ipfs/blob/2cc5ce45b875e45fdbfc779e1a7c4bef99bb9877/core/commands/dns.go#L46

It should be renamed to 'dns'

gpestana:
Next coming up on this PR:

- [ ] master rebase
- [ ] domain-name -> dns
- [ ] jsipfs block put test
- [ ] improve test coverage

VictorBjelkholm:
@gpestana no need for "domain-name -> dns" since js-ipfs does not support dnslink resolution yet.

diasdavid:
@gpestana mind rebasing master into your branch?

@VictorBjelkholm could you provide the final review and recommendation to get this merged? 

Thank you both :)

gpestana:
@diasdavid Will do, thanks for reminding me :) 
@VictorBjelkholm, let's get this done! In meanwhile, if you come up with anything besides rebase, jsipfs block put test and improve test coverage let me know.

src/cli/bin.js
```diff
@@ -2,10 +2,19 @@
...
+
+pipe.load({ commands: enableStdin })
+
+const yargs = require('yargs')
```
VictorBjelkholm: Just put this at the top so all the dependencies can be seen easily

test/cli/files.js
```diff
@@ -127,6 +127,23 @@ describe('files', () => runOnAndOff((thing) => {
       })
   })
 
+  it('add with piped argument', () => {
+    // echo 'src/init-files/init-docs/readme' | jsipfs files add
+    return ipfs('files add', { piped: 'src/init-files/init-docs/readme' })
```
VictorBjelkholm: I'm unable to actually test this due to conflicts, if those could be resolved, would be 👍 

Also, judging by this test, the output of `echo 'src/init-files/init-docs/readme' | jsipfs files add` should place `src/init-files/init-docs/readme` as the content, without any filename, but the assertion seems to say that it just added a file called `readme` which is incorrect.

diasdavid:
@gpestana do you think we can get this done before #910 (say this week)

test/cli/files.js
gpestana: Good point, gonna take a look at that

gpestana:
@diasdavid yep, I think it's possible. Let's aim at getting this merged by end of the week.

test/cli/files.js
VictorBjelkholm: Yeah, just got a chance to try it again and this should work:

```
head -c 10 /dev/urandom | base64 | node src/cli/bin.js files add
```

but instead is giving me an error:

```
/Users/user/projects/ipfs/js-monorepo/packages/js-ipfs/node_modules/yargs/yargs.js:1079
      else throw err
           ^

Error: ENOENT: no such file or directory, stat 'Ksw1ReYL8xR2VA=='
    at Error (native)
    at Object.fs.statSync (fs.js:987:18)
    at checkPath (/Users/user/projects/ipfs/js-monorepo/packages/js-ipfs/src/cli/commands/files/add.js:32:10)
    at Object.handler (/Users/user/projects/ipfs/js-monorepo/packages/js-ipfs/src/cli/commands/files/add.js:72:20)
```

Because it's trying to find a file with that name, rather than actually using the data from stdin.

src/cli/bin.js
VictorBjelkholm: `add`, `get` and `cat` should also be here temporary as we have them as aliases. https://github.com/ipfs/js-ipfs/blob/ebaf9a07cacd8fb11062aca58b9532183ebefd1d/src/cli/bin.js#L26-L35

test/cli/files.js
gpestana: If I understood it correctly, it seems that it might be an issue with the CLI itself. The piped arguments work exactly the same way as if the arguments were passed directly:

```
⌁46% [home:~/dev/js-ipfs] cli_pipe_474(+1/-2)+ 2s 1 ± node src/cli/bin.js files add  $(head -c 10 /dev/urandom  | base64)
```

```
/Users/home/dev/js-ipfs/node_modules/yargs/yargs.js:1079
      else throw err
           ^

Error: ENOENT: no such file or directory, stat 're+g0ZVkYUUsKg=='
    at Object.fs.statSync (fs.js:967:11)
    at checkPath (/Users/home/dev/js-ipfs/src/cli/commands/files/add.js:32:10)
    at Object.handler (/Users/home/dev/js-ipfs/src/cli/commands/files/add.js:72:20)
    at Object.self.runCommand (/Users/home/dev/js-ipfs/node_modules/yargs/lib/command.js:233:22)
    at Object.Yargs.self._parseArgs (/Users/home/dev/js-ipfs/node_modules/yargs/yargs.js:990:30)
    at Object.self.runCommand (/Users/home/dev/js-ipfs/node_modules/yargs/lib/command.js:204:45)
    at Object.Yargs.self._parseArgs (/Users/home/dev/js-ipfs/node_modules/yargs/yargs.js:990:30)
    at Object.Yargs.self.parse (/Users/home/dev/js-ipfs/node_modules/yargs/yargs.js:532:23)
    at utils.getIPFS (/Users/home/dev/js-ipfs/src/cli/bin.js:67:6)
    at IPFS.node.once (/Users/home/dev/js-ipfs/src/cli/utils.js:64:5)
```

Or am I missing something?

VictorBjelkholm: No, using stdin and passing a argument should not work the same way. Passing a argument should use that argument as a path to a file. Using stdin to use whatever comes from stdin as the content to add, without any filename.

Compare it to how go-ipfs works, and maybe it'll be a bit easier to understand.

gpestana: Got it!

gpestana:
@VictorBjelkholm, the new commit should fix what you pointed out in the last comment: in the `files add` command with parameters coming from the stdin, the input is added directly (not pointing at a file). My suggestion was to create a separate pull stream for when the arguments are passed from the stdin.

Not sure if this would be possible to fork the flow in the pull stream itself, but I'm not that experienced with it so I didn't figure out other way. Any other suggestion?

There is still a problem though. When running the tests, I get the following error:

```
  1) cli files daemon on (through http-api) add with piped argument:
     Command failed: /Users/home/dev/js-ipfs/src/cli/bin.js files add
events.js:163
      throw er; // Unhandled 'error' event
      ^

TypeError: Invalid non-string/buffer chunk
    at validChunk (_stream_writable.js:209:10)
    at PassThrough.Writable.write (_stream_writable.js:239:21)
    at PassThrough.Writable.end (_stream_writable.js:475:10)
    at Multipart.addPart (/Users/home/dev/js-ipfs/node_modules/multipart-stream/index.js:59:14)
    at flatmap.forEach (/Users/home/dev/js-ipfs/node_modules/ipfs-api/src/get-files-stream.js:131:8)
    at Array.forEach (native)
    at getFilesStream (/Users/home/dev/js-ipfs/node_modules/ipfs-api/src/get-files-stream.js:130:6)
    at requestAPI (/Users/home/dev/js-ipfs/node_modules/ipfs-api/src/request-api.js:94:14)
    at send (/Users/home/dev/js-ipfs/node_modules/ipfs-api/src/request-api.js:188:12)
    at Function.send.andTransform (/Users/home/dev/js-ipfs/node_modules/ipfs-api/src/request-api.js:196:12)
```

This happens only with the `cli files daemon on (through http-api)` but the test set when the daemon is off works as expected. Does this case (using stdin to pass parameters) make sense with the daemon on?

src/cli/commands/files/add.js
```diff
@@ -94,26 +107,62 @@ module.exports = {
...
+      const dataStream = new stream.Readable()
+      dataStream.push(Buffer.from(data, 'utf8'))
+      dataStream.push(null)
+      createAddStream((err, addStream) => {
+        if(err)
         throw err
-      }
+        
+        addDataPipeline([dataStream], addStream)
+      })
```
diasdavid: Something doesn't look right here:

`createAddStream` is going to glob a dir and return the stream which has all the files resulting from that glob.

If you want to add one stream, you should not need to call that.

src/cli/commands/files/add.js
```diff
@@ -94,26 +107,62 @@ module.exports = {
...
     }
 
-    createAddStream((err, addStream) => {
-      if (err) {
+    // if there are piped arguments, input is data to publish instead of a file
+    // path
+    if(hasPipedArgs) {
+      const data = argv.file
+      const dataStream = new stream.Readable()
+      dataStream.push(Buffer.from(data, 'utf8'))
```
diasdavid: It might not be `utf8`, that option out, it doesn't really matter.

src/cli/commands/files/add.js
```diff
@@ -94,26 +107,62 @@ module.exports = {
       }
     }
 
-    createAddStream((err, addStream) => {
-      if (err) {
+    // if there are piped arguments, input is data to publish instead of a file
+    // path
+    if(hasPipedArgs) {
+      const data = argv.file
```
diasdavid: If I do `cat someFile.txt | ipfs add` does pipe-args buffer the whole file first? Well, then it is missing the point of pipes, right?

gpestana: Yes, the whole file is going to be buffered before the command is triggered. I assumed that since this is for a CLI interface, it would make sense to do so. So it's better to buffer the file using streams and make it async?

src/cli/commands/files/add.js
gpestana: Yep, it can use the default. I will remove that option parameter.

src/cli/commands/files/add.js
gpestana: I see. So the right solution would be to do something like

`
if (hasStdin) {
  ipfs.files.add({path:'', content: process.stdin}, (err, files)
} else { 
// as it is now
}
`

diasdavid: Yeah, see: https://github.com/ipfs/js-ipfs/pull/785#pullrequestreview-49607637

diasdavid:
PR now needs rebase from master

diasdavid:
@hacdias would you like to get this PR merged?

src/cli/bin.js
```diff
@@ -2,11 +2,19 @@
...
 
+const pipe = require('pipe-args')
 const yargs = require('yargs')
 const updateNotifier = require('update-notifier')
 const readPkgUp = require('read-pkg-up')
 const utils = require('./utils')
 
+const enableStdin = [
+  'data', 'path', 'object data', 'ref', 'key', 'ipfs-path', 'add', 'get', 'cat',
+  'name', 'address', 'files', 'peer', 'recursive', 'default-config', 'peer ID'
```
VictorBjelkholm: Seems to be missing a few commands. `config` and `pubsub` are two I remember from the top of my head, but I guess there is more.

VictorBjelkholm: Here is a Github search that shows which `go-ipfs` commands accepts stdin: https://github.com/ipfs/go-ipfs/search?utf8=%E2%9C%93&q=EnableStdin&type=

diasdavid:
@hacdias can you land this one?

PR #784, opened 2017-03-11
Feat/expose ipfs utils

JGAntunes:
Finishing the awesome work started by @interfect on #525.

I think I've addressed all your requests @diasdavid. Check it and let me know what you think.

PR #525, opened 2016-10-18
feat: Expose ipfs utils and add an example on how to run ipfs in the browser

interfect:
Trying to use IPFS in a browser, I ran into trouble actually invoking `ipfs.files.add`, because it wanted me to pass a `Buffer`, but no implementation of `Buffer` was available in the browser namespace. This lets the `Buffer` implementation used internally inside the `IPFS` class get exposed as a field on instances, so you can invoke `ipfs.files.add` given only an `IPFS` object and no other Node globals.

Also, I've included some examples demoing use in a browser, and featuring this new exported `Buffer` type. It took me a long time to work out the incantations to use, so I feel obliged to put them somewhere prominent.

If there was an official bootstrap `libp2p-webrtc-star` signaling server, I would use that instead of `127.0.0.1` in my example. If you don't have at least one address on your node, including a signaling server IP, you don't get the transport added to the node and then you can never dial out through any signaling server. So I figured I had to add it in the example, to not confuse people over why they couldn't use the transport.

Closes #523. Also should address #513. Also may help with #509, since I definitely have it loading up in a browser.

PR #786, opened 2017-03-12
Example: Add WebRTC star example

JGAntunes:
Add WebRTC star example done by @interfect, previously on #784 and #525.

### DIFF
--- a/examples/how-to-use-webrtc-star/README.md
+++ b/examples/how-to-use-webrtc-star/README.md
@@ -1,0 +1,14 @@
+# How to use WebRTC star
+
+There's still a bit of work required to start up an in-browser node in a robust way, so that it will work whether or not there is an existing initialized IPFS repo in the user's browser. If there isn't one, you need to call `init` as above, but if there is one, calling `init` will fail. Moreover, there's currently no good way to check if you need to call `init` or not.
+
+Also, an in-browser node isn't able to call up normal IPFS nodes over raw TCP; it can only communicate over Websockets and WebRTC. Currently, there are no Websockets or WebRTC bootstrap nodes run by the IPFS maintainers. You will probably want to set up a [libp2p-webrtc-star signaling server](https://github.com/libp2p/js-libp2p-webrtc-star) so nodes used in your application can find each other:
+
+```bash
+npm i libp2p-webrtc-star -g
+star-sig
+```
+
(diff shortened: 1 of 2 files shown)


examples/how-to-use-webrtc-star/README.md
```diff
@@ -0,0 +1,14 @@
+# How to use WebRTC star
+
+There's still a bit of work required to start up an in-browser node in a robust way, so that it will work whether or not there is an existing initialized IPFS repo in the user's browser. If there isn't one, you need to call `init` as above, but if there is one, calling `init` will fail. Moreover, there's currently no good way to check if you need to call `init` or not.
```
VictorBjelkholm: Not sure how that is related to "how to use webrtc-star", even if it's good information in general when using js-ipfs.

examples/how-to-use-webrtc-star/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Also, an in-browser node isn't able to call up normal IPFS nodes over raw TCP; it can only communicate over Websockets and WebRTC. Currently, there are no Websockets or WebRTC bootstrap nodes run by the IPFS maintainers. You will probably want to set up a [libp2p-webrtc-star signaling server](https://github.com/libp2p/js-libp2p-webrtc-star) so nodes used in your application can find each other:
```
VictorBjelkholm: We do run a publicly available signaling server which would better to point users to rather than everyone setting up their own. We should start with an explanation for connecting to the existing one and later also explain how to run their own if necessary.

VictorBjelkholm:
We could simplify this example by creating a new repository every time, by randomizing the path rather than having to figure out if it's already initialized or not. That way, you can open the same application in two different tabs.

diasdavid:
This example needs to be updated from latest js-ipfs master.

Now it is super simple to add webrtc start, check -- https://github.com/ipfs/js-ipfs#advanced-options-when-creating-an-ipfs-node -- on how to add custom addrs. For webrtc-star, all you need to add __now__ is:

`/libp2p-webrtc-star/dns4/star-signal.cloud.ipfs.team/wss`

@JGAntunes feeling like updating it?

diasdavid:
I believe this example is no longer necessary as the process of adding webrtc to the mix is quite simple now, plus transfer-files shows how to do it just well. 

Big thank you to everyone that keeps pushing stuff to make js-ipfs more awesome! The decision here is to keep the examples set small, so that it is easy to find what one needs.

PR #788, opened 2017-03-13
docs: block api calls

diasdavid:


diasdavid:
for https://github.com/ipfs/docs/issues/23

PR #793, opened 2017-03-15
feat: webrtc by default if supported

diasdavid:
I want to add this in, but cope with the scenarios where there is no webrtc-support. The current way does it, but also it makes it annoying to disable webrtc only, you'd have to swap all the multiaddrs at the same time to stop supporting webrtc. Will look into a better way.

### DIFF
--- a/src/core/components/init.js
+++ b/src/core/components/init.js
@@ -63,6 +63,20 @@ module.exports = function init (self) {
         opts.log('done')
         opts.log('peer identity: ' + config.Identity.PeerID)
 
+        const isWin = /^win/.test(process.platform)
+        const isLinux = /^linux/.test(process.platform)
+        const wrtcLinuxWindows = !process.env.IPFS_WRTC_LINUX_WINDOWS ||
+                                 self._options.EXPERIMENTAL.wrtcLinuxWindows
+
+        // For the lack of sane WebRTC support on Linux and Windows
+        if (wrtcLinuxWindows && (isWin || isLinux)) {
+          console.log('WARNING: Your platform does not have native WebRTC support, it won\' use any WebRTC transport')
+          const newAddrs = config.Addresses.Swarm.filter((addr) => {
+            return addr.indexOf('libp2p-webrtc-star') < 0
+          })
+          config.Addresses.Swarm = newAddrs
+        }
+
         self._repo.init(config, cb)
       },
       (_, cb) => self._repo.open(cb),
--- a/src/init-files/default-config-node.json
+++ b/src/init-files/default-config-node.json
@@ -2,7 +2,8 @@
   "Addresses": {
     "Swarm": [
       "/ip4/0.0.0.0/tcp/4002",
-      "/ip4/127.0.0.1/tcp/4003/ws"
+      "/ip4/127.0.0.1/tcp/4003/ws",
+      "/libp2p-webrtc-star/dns4/star-signal.cloud.ipfs.team/wss"
     ],
     "API": "/ip4/127.0.0.1/tcp/5002",
     "Gateway": "/ip4/127.0.0.1/tcp/9090"
--- a/README.md
+++ b/README.md
@@ -231,7 +231,8 @@ const node = new IPFS({
   // start: false,
   EXPERIMENTAL: { // enable experimental features
     pubsub: true,
-    sharding: true // enable dir sharding
+    sharding: true, // enable dir sharding
+    wrtcLinuxWindows: true // use unstable wrtc module on Linux or Windows with Node.js
   },
   config: { // overload the default config
     Addresses: {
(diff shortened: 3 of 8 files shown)


src/core/components/init.js
```diff
@@ -45,6 +46,12 @@ module.exports = function init (self) {
         opts.log('done')
         opts.log('peer identity: ' + config.Identity.PeerID)
 
+        if (webrtcSupport.support || isNode) {
```
dignifiedquire: isNode does not gurantee webrtc support

diasdavid: I take suggestions.

dignifiedquire: try {
  require('webrtc')
} catch () {
}

diasdavid: You mean `wrtc`?

dignifiedquire: yes we do

diasdavid:
Reviewing this, the hard case is when a user wants to say "no I don't want WebRTC".

On a second thought, we actually can just make it part of the default multiaddrs, swarm will know how to handle non existing transports for multiaddrs to listen.

diasdavid:
All right, this is ready :)

diasdavid:
Travis fails due to https://github.com/ipfs/js-ipfs/issues/844, not related with this PR

diasdavid:
Tested on Linux manually too:

image

image

diasdavid:
Ok, after experimenting a bit with it in Linux, I believe I found a more sane approach until we have a WebRTC transport that is rock solid in Linux and Windows.

@VictorBjelkholm and all the Linux users of js-ipfs, mind giving your feedback on:

https://github.com/ipfs/js-ipfs/pull/793/commits/568479f28c7033c434921e787340f609b97b8f1e

VictorBjelkholm:
Not a big fan of turning off webrtc for linux users, but since it's not stable enough, this is probably the best way to go about it.

However, if I specify a libp2p-webrtc-star address in my own swarm config, and don't set `USE_WRTC`, I wouldn't expect it to remove them for me, I can see how that would be very confusing since I'm explicitly setting it myself.

If I don't, I agree we shouldn't automatically add it.

VictorBjelkholm:
Ah, maybe include that you can use `USE_WRTC` to override, in the console.log

diasdavid:
From what I can read from your second point, it matches what is happening

`USE_WRTC` is only used during the `init`, this gives a way for every user to follow init -> daemon process without hitting any hiccup. For the more advanced user that modifies the IPFS config by hand, then that user will know better what they are doing and if they configured WRTC correctly, they will see it working.

Yeah, agree with you, wish there was a good WebRTC implementation.

VictorBjelkholm:
@diasdavid hm, you're thinking about the CLI, and I think in that case, it makes sense.

However, when using it in code, the usage becomes a lot harder if I can't just specify it in my swarm addresses and it sticks. If instead, I need to run through the normal startup, then once it's `ready`, I have to do `config.get`, modify and then `config.set`, it becomes a lot uglier.

I'm want this (https://github.com/VictorBjelkholm/js-ipfs-in-the-browser/blob/master/spawn-node.js#L14-L20) to still work the same way as before, even for Linux users, since the Swarm addresses are manually specified.

diasdavid:
@VictorBjelkholm 

In the browser, you won't have to do anything, since browser is not a linux or windows env

In Node.js, using it programmatically, you just need to set `process.env.USE_WRTC=1` before your `init` call, if and only if, you are in windows or linux. 

You will continue to be able to do -- https://github.com/VictorBjelkholm/js-ipfs-in-the-browser/blob/master/spawn-node.js#L14-L20 -- as before

VictorBjelkholm:
@diasdavid ay, of course! Duh. Still, if it was a script executed by nodejs, even if I explicitly set swarm address and not specify `USE_WRTC`, it'll be overwritten, correct?

diasdavid:
@VictorBjelkholm yes, but then you just set that environment flag and you are golden.

Another thing we can add is an EXPERIMENTAL flag for "WRTC_LINUX_WINDOWS" so that you can pass that instead of having to do the `process.env.USE_WRTC`

And yes, I agree if we add that experimental flag, it should be named the same. `WRTC_LINUX_WINDOWS` makes more sense so me now.

VictorBjelkholm:
@diasdavid 
> but then you just set that environment flag and you are golden

Yeah... My point is, if I explicitly set something in the config and run the script, I don't assume js-ipfs to make a choice for me. If I set those addresses explicitly, it's not my expectation that those will be removed automatically unless I set a env var.

However, if I don't specify them, it would make sense that we **don't add them**.

diasdavid:
I see a bunch of `if` for a lot of conditions that happen when you are a more savvy user. The goal to fulfill the expectations of our every day new users that want to spawn an instance in the browser without having to learn that it needs to add another multiaddr to see it getting connected to the other browser instances.

The case you talk about is for the user that needs to go through all the trouble of downloading and building wrtc by hand for Linux, at that state he will be more than familiar with the README and the case.

For all the other users, it will work.

Kubuxu:
In go we try to prefix IPFS flags with `IPFS_` so environment namespaces can't collide. You might want to consider it here.

src/init-files/default-config-node.json
```diff
@@ -2,7 +2,8 @@
   "Addresses": {
     "Swarm": [
       "/ip4/0.0.0.0/tcp/4002",
-      "/ip4/127.0.0.1/tcp/4003/ws"
+      "/ip4/127.0.0.1/tcp/4003/ws",
+      "/libp2p-webrtc-star/dns4/star-signal.cloud.ipfs.team/wss"
```
Kubuxu: It is swarm listen address list. Sure it should be here?

diasdavid: yep @Kubuxu, WebRTC references the signalling point, not a local server ;)

diasdavid:
good point @Kubuxu! changed the name of the env flag to be prefixed by `IPFS_`

PR #794, opened 2017-03-16
fix: give the daemon time to spawn

diasdavid:


diasdavid:
exactly :) Thank you!

PR #796, opened 2017-03-17
race on `jsipfs init`

diasdavid:
sometimes, travis fails with a repo that wasn't fully initiated on`jsipfs init` (just CLI)

I'm being unable to replicate it locally, so this PR is just me throwing things to see if I can figure it out :)


diasdavid:
Node.js stable passes on circle, but fails on travis.

src/http-api/index.js
```diff
@@ -21,113 +21,115 @@ function uriToMultiaddr (uri) {
...
+            pubsub: true
+          }
+        })
 
-        this.server.start((err) => {
+        this.node.once('error', (err) => {
+          err.code = 'ENOENT'
+          cb(err)
+        })
+        this.node.once('start', cb)
```
dignifiedquire: need to wrap the `cb` into a  `once` otherwise there could be a double calling

src/http-api/index.js
```diff
@@ -21,113 +21,115 @@ function uriToMultiaddr (uri) {
...
+
+          // Set default headers
+          setHeader(this.server,
+            'Access-Control-Allow-Headers',
+            'X-Stream-Output, X-Chunked-Output, X-Content-Length')
+          setHeader(this.server,
+            'Access-Control-Expose-Headers',
+            'X-Stream-Output, X-Chunked-Output, X-Content-Length')
+
+          this.server.start((err) => {
```
dignifiedquire: This can be just `this.server.start(cb)` and then the rest below be in the next block

src/http-api/index.js
```diff
@@ -21,113 +21,115 @@ function uriToMultiaddr (uri) {
...
+            if (err) {
+              return callback(err)
```
dignifiedquire: this should be `cb`

src/http-api/index.js
```diff
@@ -21,113 +21,115 @@ function uriToMultiaddr (uri) {
...
+            api.info.ma = uriToMultiaddr(api.info.uri)
+            gateway.info.ma = uriToMultiaddr(gateway.info.uri)
+
+            console.log('API is listening on: %s', api.info.ma)
+            console.log('Gateway (readonly) is listening on: %s', gateway.info.ma)
+
+            // for the CLI to know the where abouts of the API
+            fs.writeFileSync(apiFilePath, api.info.ma)
+
+            callback()
```
dignifiedquire: this should be `cb`

test/cli/daemon.js
```diff
@@ -10,14 +10,15 @@ describe('daemon', () => {
   let ipfs
 
   beforeEach(() => {
-    repoPath = '/tmp/ipfs-test-' + Math.random().toString().substring(2, 8)
+    repoPath = '/tmp/ipfs-test-not-found-' + Math.random().toString().substring(2, 8)
```
dignifiedquire: why `not-found`?

src/http-api/index.js
diasdavid: but there is a `once` listener, so it only does it `once`

test/cli/daemon.js
diasdavid: I was just making it obvious for my tests :)

src/http-api/index.js
dignifiedquire: but you register it twice, once for start and once for error

diasdavid: they are mutually exclusive, but I'll add it anyway just to be sure 👍

diasdavid:
before I squashed, this was passing:
image

Now it is just hogged because travis is not even starting the tests. Going to go ahead and merge :)

Issue #787, opened 2017-03-13
Awesome Datastore Endeavour


dignifiedquire:
As described in https://github.com/ipfs/js-ipfs-repo/issues/112 we need and want to upgrade our storage game. So here comes another awesome endeavour 🌟 

The general goal is to create a datastore interface and implementations to match the way go handles storage and matching features like `query` that we need down the line. The new library has been implemented in [interface-datastore](https://github.com/ipfs/interaface-datastore). 

### New Modules

| Name | Ready for Review | Merged | Published | 
|-------|-------------------|---------|-----------|
|[interface-datastore](https://github.com/ipfs/interface-datastore) | ✅ |✅ | ✅|
|[datastore-core](https://github.com/ipfs/js-datastore-core) | ✅ |✅ | ✅|
|[datastore-fs](https://github.com/ipfs/js-datastore-fs) | ✅ |✅ | ✅|
|[datastore-level](https://github.com/ipfs/js-datastore-level)  | ✅ |✅ | ✅|

### Modules that need updating

| Name | Ready for Review | Reviewed | Merged | Published | 
|-------|-------------------|---------|---------|-----------|
|[ipfs-repo](https://github.com/ipfs/js-ipfs-repo) | ✅ | ✅ | ✅ | ✅ |
|[ipfs-block](https://github.com/ipfs/js-ipfs-block) | ✅ | ✅ | ✅ | ✅ |
|[ipfs-block-service](https://github.com/ipfs/js-ipfs-block-service) | ✅ | ✅ | ✅ | ✅ |
|[ipfs-bitswap](https://github.com/ipfs/js-ipfs-bitswap) | ✅ | ✅ | ✅ | ✅ |
| ipld-dag-pb | ✅ | ✅ | ✅ | ✅ |
| ipld-dag-cbor | ✅ | ✅ | ✅ | ✅ |
| ipld-dag-eth-x | ✅ | ✅ | ✅ | ✅ |
| ipld-resolver | ✅ | ✅ | ✅ | ✅ |
| ipfs-unixfs-engine | ✅ | ✅ | ✅ | ✅ |
|[interface-ipfs-core](https://github.com/ipfs/interface-ipfs-core) | ✅ | ✅ | ✅ | ✅
|[ipfs-api](https://github.com/ipfs/js-ipfs-api) | ✅ | ✅ | |
|[ipfs](https://github.com/ipfs/js-ipfs) | ✅ | ✅ | | |

[...]

PR #797, opened 2017-03-17
Awesome datastore

### LINKED ISSUE #229: daemon leaves behind api file lock on error
When the daemon hits an error on load like https://github.com/ipfs/js-ipfs/issues/228, it will leave around the `~/.ipfs/api` file lock.

`go-ipfs` seems to be smart about detecting this and cleaning it up, but it will inhibit future invocations of the `js-ipfs` daemon.

dignifiedquire:
Closes https://github.com/ipfs/js-ipfs/issues/787
Closes #229

- [x] refactored cli to use a context ipfs daemon to ensure cleanup
- [x] update to new datastore dependencies
- [x] fix timeout issues and add two new events (`init` and `ready`) to handle the startup of a daemon more granularly (needs discussion probably)
- [x] upgrade to yargs@7


dignifiedquire:
I am afraid this is not ready yet. I've discovered that the new creation and startup api is the source of quite some race conditions, which are becoming more visible with this PR, as the locking of the repo and open and close are more enforced. 

We need to think through this very carefully as the options `init` and `start` combined with methods for doing those actions are nice for the user but create quite the tricky state management in the background. If we decide to keep the api we need to integrate a proper finite state machine to ensure all transitions are valid and very explicit descriptions on how things work. 

Tldr `setTimeout` usually means you have a race condition you are hiding..

src/cli/bin.js
```diff
@@ -27,9 +28,23 @@ aliases.forEach((alias) => {
   cli.command(alias.command, alias.describe, alias.builder, alias.handler)
 })
 
-// finalize cli setup
-cli // eslint-disable-line
-  .help()
-  .strict()
-  .completion()
-  .argv
+utils.getIPFS((err, ipfs, cleanup) => {
```
diasdavid: init and daemon don't use an `ipfs` instance to run, how does this cover those two cases?

diasdavid:
Note: We can add a `init` event to the set up, it should solve all of the issues you are seeing as the only issue was when a repo is init'ed by default and gets started 'later', this later is the source of trouble because there is no way to monitor when the init has finished.

The finite state machine approach is also interesting, probably even better to keep the simplicity. Wanna take a stab at it?

diasdavid:
See: https://github.com/ipfs/js-ipfs/compare/fix/race?expand=1

src/cli/bin.js
dignifiedquire: I will add an exception to those, I didn't realise that until later

diasdavid:
Saw now that you added more events.

- `init` is part of master now
- `ready` why the need for this one, exactly?

dignifiedquire:
> ready why the need for this one, exactly?

to detect the finishing things  like `new IPFS({init: false, start: false}` which are alway async now, due to the need to detect and open the repo

diasdavid:
@dignifiedquire in that case, there are no actions happening asynchronously.

dignifiedquire:
> @dignifiedquire in that case, there are no actions happening asynchronously.

but they are now, because repo checks are async

gulpfile.js
```diff
@@ -16,37 +15,25 @@ let nodes = []
...
+        `/ip4/127.0.0.1/tcp/10${num}`,
+        `/ip4/127.0.0.1/tcp/20${num}/ws`
+      ],
+      API: `/ip4/127.0.0.1/tcp/31${num}`,
+      Gateway: `/ip4/127.0.0.1/tcp/32${num}`
     },
-    start: false,
-    EXPERIMENTAL: {
-      pubsub: true
-    },
```
diasdavid: why disabling PubSub?

src/cli/bin.js
```diff
@@ -27,9 +28,30 @@ aliases.forEach((alias) => {
...
-  .help()
-  .strict()
-  .completion()
-  .argv
+const args = process.argv.slice(2)
+
+// Need to skip to avoid locking the daemon
+if (args[0] === 'daemon' || args[0] === 'init') {
+  return cli.help().strict(false).completion().parse(args)
+}
```
diasdavid: Make this more clear:

`// commands that don't require a daemon/instance`
`// commands that require a daemon/instance`

diasdavid: Also, use same indentation in both

src/core/components/block.js
```diff
@@ -11,43 +12,32 @@ module.exports = function block (self) {
...
-
-        callback(null, block)
-      })
+        (block, cb) => self._blockService.put(block, (err) => {
+          if (err) {
+            return cb(err)
+          }
+          cb(null, block)
+        })
+      ], callback)
```
diasdavid: This has changed the API, block needs to be able to:

```
» jsipfs block put --help
jsipfs block put [block]

Options:
  --help        Show help                                              [boolean]
  --format, -f  cid format for blocks to be created with.    [default: "dag-pb"]
  --mhtype      multihash hash function                    [default: "sha2-256"]
  --mhlen       multihash hash length
```

Before we passed that through CID, if you remove the CID from equation and always calculate the multihash of sha256, then you can't block put eth data anymore.

src/core/components/libp2p.js
```diff
@@ -14,9 +15,9 @@ module.exports = function libp2p (self) {
         }
 
         const options = {
-          mdns: config.Discovery.MDNS.Enabled,
-          webRTCStar: config.Discovery.webRTCStar.Enabled,
-          bootstrap: config.Bootstrap
+          mdns: get(config, 'Discovery.MDNS.Enabled'),
+          webRTCStar: get(config, 'Discovery.webRTCStar.Enabled'),
+          bootstrap: get(config, 'Bootstrap')
```
diasdavid: Requiring a dependency to pick one property of an object?

src/core/components/pre-start.js
```diff
@@ -31,7 +30,7 @@ module.exports = function preStart (self) {
 
           if (!mafmt.IPFS.matches(ma)) {
             ma = ma.encapsulate('/ipfs/' +
-              self._peerInfo.id.toB58String())
+                                self._peerInfo.id.toB58String())
```
diasdavid: strange identation

src/core/status.js
```diff
@@ -0,0 +1,12 @@
...
+// Possible IPFS status values:
+module.exports = {
+  new: 'new',
+  initializing: 'initializing',
+  initialized: 'initialized',
+  starting: 'starting',
+  running: 'running',
+  stopping: 'stopping',
+  stopped: 'stopped'
+}
```
diasdavid: Right now, in order to follow the sequence of booting a node, we need to open 4+ files. 

Let's put this into the state.js.

diasdavid: Rechecked, this is just used to populate a var `this._status` once and that const is not used anymore. ???

src/core/utils.js
```diff
@@ -1,17 +1,3 @@
...
-    }
-
-    if (exists === false) {
-      return cb(new Error('no ipfs repo found'))
-    }
-
-    cb()
-  })
-}
-
```
diasdavid: Where is this now? Breaking API change? If so, update README

src/http-api/index.js
```diff
@@ -22,56 +16,61 @@ function uriToMultiaddr (uri) {
...
+            init: init,
+            start: true,
+            config: config,
+            EXPERIMENTAL: {
+              pubsub: true
+            }
+          })
+        } catch (err) {
+          return cb(err)
+        }
```
diasdavid: What is throwing? Errors should come on the `error` event only.

test/go-ipfs-repo/datastore/LOG
```diff
@@ -0,0 +1,2 @@
+=============== Mar 18, 2017 (CET) ===============
+00:21:30.447664 log@legend F·NumFile S·FileSize N·Entry C·BadEntry B·BadBlock Ke·KeyError D·DroppedEntry L·Level Q·SeqNum T·TimeElapsed
```
diasdavid: .gitignore this

test/go-ipfs-repo/version
```diff
@@ -0,0 +1 @@
+1
```
diasdavid: is it really version 1?

gulpfile.js
dignifiedquire: It's not disabled, it is enabled in the httpserver as far as I understand

src/core/components/libp2p.js
dignifiedquire: This is so that undefined chains don't throw. This code explodes if you try to read a config from go-ipfs as it doesn't have a `webRTCStar` field in discovery.

src/core/components/pre-start.js
dignifiedquire: very strange

src/core/status.js
dignifiedquire: this is the rest from some experimentation, going to delete it

src/core/utils.js
dignifiedquire: I didn't know this was in the readme, this is not needed anymore internally that's why I removed it

src/http-api/index.js
dignifiedquire: this can throw if someone screwed up in the implementation, resulting in silent fail of the tests

test/go-ipfs-repo/version
dignifiedquire: hmm no not sure why that says version 1

src/core/utils.js
dignifiedquire: Can't find it in the readme, are you sure this was documented somewhere?

gulpfile.js
diasdavid: got it, true.

src/core/components/block.js
dignifiedquire: I don't understand your concern, if you want to specify your own things you pass a regular ipfs block  here which already defines the cid it wants, this way things are passed through as expected

dignifiedquire: the hashing is just a fallback if only a buffer is passed in.

src/core/utils.js
diasdavid: Ah, not on the readme, people are using it though - https://github.com/ipfs/js-ipfs/pull/776

Wanna finish that PR and merge it here?

test/go-ipfs-repo/version
dignifiedquire: on disk it says 5 for me

dignifiedquire: fixed, there were two

test/go-ipfs-repo/datastore/LOG
dignifiedquire: done

src/http-api/index.js
diasdavid: ok, let's add that comment, I want to avoid giving the idea that errors should be 'thrown' for contributors coming into the codebase.

src/core/utils.js
dignifiedquire: lets get this merged and then I can do that PR, I don't want to put even more things into here than there are already

src/http-api/index.js
dignifiedquire: done

src/core/components/block.js
diasdavid: Hm.. I see, you now expect everything in the block.

Can we also have the options object, we will need to support the hashAlg and format. But that can be a separate PR

diasdavid:
Seems that last thing is to rebase

diasdavid:
image

I need to leave now for up to 3 hours, will come back online later.

@dignifiedquire, let me know if you can check this.

dignifiedquire:
last missing piece https://github.com/ipfs/js-ipfs-repo/pull/128 (hopefully)

Issue #696, opened 2017-01-03
wrap-with-directory option approach


harshjv:
I was wondering if I were to implement `wrap-with-directory` option, is it a good approach to append a wrapper parent directory before adding it to preserve file name and implement `ipfs add --wrap-with-directory` option for `js-ipfs`?

### Example

    jsipfs add --wrap-with-directory /some/dir/example.jpg

### Approach

    node.files.add([
      {
        path: `/wrapper/example.jpg`,
        content: fs.createReadStream('/some/dir/example.jpg')
      }
    ], (err, data) => {
      if (err) throw err

      const directoryHash = data[0].hash
      const fileHash = data[1].hash

      console.log(`added ${fileHash} example.jpg`)
      console.log(`added ${directoryHash}`)
    })

### Output

    added QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH example.jpg
    added QmaG4FuMqEBnQNn3C8XJ5bpW8kLs7zq2ZXgHptJHbKDDVx

If this looks good, I can send a PR for `--wrap-with-directory`

diasdavid:
@harshjv it is indeed! I've no idea why it took me so much to review this issue. Go ahead with the PR :)

harshjv:
@diasdavid done. https://github.com/ipfs/js-ipfs/pull/798

diasdavid:
Thank you so much @harshjv :) I've already using it today!

PR #798, opened 2017-03-18
Wrap with directory option

harshjv:
Implements #696

### DIFF
--- a/src/cli/commands/files/add.js
+++ b/src/cli/commands/files/add.js
@@ -53,6 +53,11 @@ module.exports = {
       type: 'boolean',
       default: false,
       describe: 'Use the trickle DAG builder'
+    },
+    'wrap-with-directory': {
+      alias: 'w',
+      type: 'boolean',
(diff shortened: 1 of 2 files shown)


src/cli/commands/files/add.js
```diff
@@ -117,12 +124,16 @@ function addPipeline (index, addStream, list) {
     pull.filter((file) => !file.isDirectory),
     pull.map((file) => ({
       path: file.path.substring(index, file.path.length),
-      content: fs.createReadStream(file.path)
+      originalPath: file.path
+    })),
+    pull.map((file) => ({
+      path: wrapWithDirectory ? `${WRAPPER}${file.path}` : file.path,
```
dignifiedquire: this should use `path.join` to ensure correct path separators

harshjv:
@diasdavid ya. extra space in the output :/ Fixed it 👍 

@dignifiedquire Done 👍

PR #801, opened 2017-03-20
Avoid setTimeouts

diasdavid:


README.md
```diff
@@ -242,6 +242,13 @@ const node = new IPFS({
     }
   }
 })
+
+// Events
+
+node.on('init', () => {})     // Node has successfully finished initing the repo
```
dignifiedquire: We still need an event/callback for after `new IPFS({init: false, start: false})` as that can be async as well

src/core/components/init.js
```diff
@@ -70,6 +70,14 @@ module.exports = function init (self) {
           cb(null, true)
         })
       }
-    ], callback)
+    ], (err, success) => {
+      if (err) {
+        return callback(err)
+      }
+
+      setImmediate(() => self.emit('init'))
```
dignifiedquire: why the delay?

src/core/components/init.js
```diff
@@ -70,6 +70,14 @@ module.exports = function init (self) {
           cb(null, true)
         })
       }
-    ], callback)
+    ], (err, success) => {
+      if (err) {
+        return callback(err)
```
dignifiedquire: should emit error

diasdavid: It emits, but through the `series` look in core/index.js

src/core/components/init.js
diasdavid: no need, removing it :)

diasdavid: removed :)

src/core/components/init.js
dignifiedquire: 👍

diasdavid:
@dignifiedquire thanks for the CR :)

PR #808, opened 2017-03-23
Loosen up init requirements: accept already initialised repos

VictorBjelkholm:
This PR will make it so if you start a node twice with the same path, it won't crash when trying to initialize a new repo. Synonym with doing `ipfs daemon --init` with go-ipfs in terminal.

### DIFF
--- a/src/core/boot.js
+++ b/src/core/boot.js
@@ -23,22 +23,35 @@ module.exports = (self) => {
   // Will return callback with a bool indicating the existence
   // of the repo
   const maybeOpenRepo = (cb) => {
-    waterfall([
-      (cb) => self._repo.exists(cb),
-      (exists, cb) => {
-        if (exists && !repoOpen) {
-          return series([
-            (cb) => self._repo.open(cb),
-            (cb) => self.preStart(cb),
-            (cb) => {
-              self.state.initialized()
-              cb()
-            }
-          ], cb)
-        }
-        cb(null, exists)
+    // nothing to do
+    if (repoOpen) {
+      return cb(null, true)
+    }
+
+    series([
+      (cb) => self._repo.open(cb),
+      (cb) => self.preStart(cb),
+      (cb) => {
+        self.state.initialized()
+        cb(null, true)
       }
-    ], cb)
+    ], (err, res) => {
+      if (err) {
+        // If the error is that no repo exists,
+        // which happens when the version file is not found
+        // we just want to signal that no repo exist, not
+        // fail the whole process.
+        // TODO: improve datastore and ipfs-repo implemenations so this error is a bit more unified
+        if (err.message.match(/not found/) || // indexeddb
+            err.message.match(/ENOENT/) || // fs
+            err.message.match(/No value/) // memory
+           ) {
+          return cb(null, false)
+        }
+        return cb(err)
+      }
+      cb(null, res)
+    })
   }
 
   const done = (err) => {
--- a/src/core/boot.js
+++ b/src/core/boot.js
@@ -15,13 +15,17 @@ module.exports = (self) => {
   const repoOpen = !self._repo.closed
 
   const customInitOptions = typeof options.init === 'object' ? options.init : {}
-  const initOptions = Object.assign({ bits: 2048 }, customInitOptions)
+  const initOptions = Object.assign({
+    bits: 2048
+  }, customInitOptions)
 
+  // Checks if a repo exists, and if so opens it
+  // Will return callback with a bool indicating the existence
+  // of the repo
   const maybeOpenRepo = (cb) => {
     waterfall([
       (cb) => self._repo.exists(cb),
       (exists, cb) => {
-        self.log('boot:maybeOpen', exists, repoOpen)
         if (exists && !repoOpen) {
           return series([
             (cb) => self._repo.open(cb),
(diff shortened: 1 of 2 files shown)


VictorBjelkholm:
Currently failing with: 

```
  1) create node can start node twice without crash:
     Error: Not able to start from state: uninitalized
      at IPFS.start (src/core/components/start.js:23:19)
      at tasks.push (src/core/boot.js:106:33)
      at node_modules/async/internal/parallel.js:27:9
      at replenish (node_modules/async/internal/eachOfLimit.js:64:17)
      at iterateeCallback (node_modules/async/internal/eachOfLimit.js:49:17)
      at node_modules/async/internal/onlyOnce.js:12:16
      at node_modules/async/internal/parallel.js:32:13
      at apply (node_modules/lodash/_apply.js:15:25)
      at node_modules/lodash/_overRest.js:32:12
      at node_modules/async/internal/once.js:12:16
      at nextTask (node_modules/async/waterfall.js:15:29)
      at node_modules/async/waterfall.js:22:13
      at apply (node_modules/lodash/_apply.js:15:25)
      at node_modules/lodash/_overRest.js:32:12
```

src/core/boot.js
```diff
@@ -49,7 +49,21 @@ module.exports = (self) => {
 
   if (doInit) {
     self.log('boot:doInit')
-    tasks.push((cb) => self.init(initOptions, cb))
+    tasks.push((cb) => {
+      console.log('Doing init')
+      self._repo.exists((err, exists) => {
```
dignifiedquire: this doesn't make much sense as it's the same thing as in the else branch

src/core/boot.js
```diff
@@ -49,7 +49,21 @@ module.exports = (self) => {
...
+        if (err) cb(err)
+        console.log('exists', err, exists)
+        if (exists) {
+          cb(null, true)
```
dignifiedquire: you need to call maybeOpenRepo if you don't do init

src/core/boot.js
```diff
@@ -49,7 +49,16 @@ module.exports = (self) => {
 
   if (doInit) {
     self.log('boot:doInit')
-    tasks.push((cb) => self.init(initOptions, cb))
+    tasks.push((cb) => {
+      self._repo.exists((err, exists) => {
```
dignifiedquire: you are still doing the same call twice in the `if (doInit)`. you can just call `repo.exists` and do the if doInit check later down

src/core/boot.js
```diff
@@ -49,7 +49,16 @@ module.exports = (self) => {
...
+        if (err) cb(err)
```
dignifiedquire: needs to return and please on with braces on multiple lines

VictorBjelkholm:
Ready to be checked and eventually merged now :)

dignifiedquire:
@VictorBjelkholm did you see my comments?

src/core/boot.js
dignifiedquire: Something like this

```js
self._repo.exists((err, hasRepo) => {
  if (err) {
    return done(err)
  }
  if (doInit && !hasRepo) {
    tasks.push((cb) => self.init(initOptions, cb))
  }
  if (hasRepo) {
    tasks.push(maybeOpenRepo)
  } 
  if (setConfig) {
    // ...
})
```

src/core/boot.js
```diff
@@ -47,25 +47,25 @@ module.exports = (self) => {
...
+    if (err) {
+      return done(err)
+    }
+    if (doInit && !repoExists) {
+      tasks.push((cb) => self.init(initOptions, cb))
+    }
+    if (repoExists) {
+      tasks.push(maybeOpenRepo)
+    }
+    next(null, repoExists)
```
dignifiedquire: no need for the next function anymore :)

src/core/boot.js
```diff
@@ -47,25 +47,25 @@ module.exports = (self) => {
...
+  self._repo.exists((err, repoExists) => {
+    if (err) {
+      return done(err)
+    }
+    if (doInit && !repoExists) {
+      tasks.push((cb) => self.init(initOptions, cb))
+    }
+    if (repoExists) {
+      tasks.push(maybeOpenRepo)
+    }
```
dignifiedquire: There is still a bug, `next` needs to be called with `hasRepo` not `true`

VictorBjelkholm: @dignifiedquire I've tried that and then I get bunch of failures

```
WARNING, trying to set config on uninitialized repo, maybe forgot to set "init: true"
WARNING, trying to start ipfs node on uninitialized repo, maybe forgot to set "init: true"
```

src/core/boot.js
```diff
@@ -47,25 +47,25 @@ module.exports = (self) => {
...
-    self.log('boot:doInit')
-    tasks.push((cb) => self.init(initOptions, cb))
+  self._repo.exists((err, repoExists) => {
+    if (err) {
+      return done(err)
+    }
+    if (doInit && !repoExists) {
+      tasks.push((cb) => self.init(initOptions, cb))
+    }
+    if (repoExists) {
```
dignifiedquire: This needs to check for !doInit as well

src/core/boot.js
```diff
@@ -93,6 +92,7 @@ module.exports = (self) => {
       }
     }
 
+    // Do the actual boot sequence
     series(tasks, done)
-  }
+  })
 }
```
diasdavid: 👍 to the whole commit :)

PR #809, opened 2017-03-24
fix: pull-stream-to-stream replaced with duplex stream

pgte:
stream-to-duplex-stream ends the stream immediately when `classicStream.end` is called.
Since now the unifxfs addStream end happens asynchronously, this was ending the stream prematurely, before it had the chance to emit everything.


src/core/components/files.js
```diff
@@ -164,3 +178,28 @@ function normalizeContent (content) {
...
+  constructor (pullStream, push, options) {
+    super(Object.assign({}, { objectMode: true }, options))
+    this._pullStream = pullStream
+    this._pushable = push
+  }
+
+  _read () {
+    this._pullStream(null, (err, data) => {
+      if (err) {
+        if (err instanceof Error) {
```
dignifiedquire: what about `true` for end?

pgte: instead of calling it `err` I'm now calling it `end`.

PR #811, opened 2017-03-26
README: Fix new IPFS() example

magik6k:
Current example is not working:

```
$ node
> const IPFS = require('ipfs')
> const node = new IPFS()
TypeError: Cannot read property 'init' of undefined
    at IPFS (/home/magik6k/mess/js-ipfs/node_modules/ipfs/src/core/index.js:31:16)
    at repl:1:14
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at REPLServer.defaultEval (repl.js:340:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.onLine (repl.js:537:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:189:7)
    at REPLServer.Interface._onLine (readline.js:238:10)

```

PR #813, opened 2017-03-26
provide a default for options

thisconnect:
It seems new IPFS() isn't working at the moment, see this PR https://github.com/ipfs/js-ipfs/pull/811/files

cc https://github.com/ipfs/js-ipfs/pull/811

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -17,7 +17,7 @@ const boot = require('./boot')
 const components = require('./components')
 
 class IPFS extends EventEmitter {
-  constructor (options) {
+  constructor (options = {}) {
     super()
 
     this._options = {
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -26,7 +26,7 @@ class IPFS extends EventEmitter {
       EXPERIMENTAL: {}
     }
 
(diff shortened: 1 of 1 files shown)


thisconnect:
that didn't really work in node v4 :(

src/core/index.js
```diff
@@ -17,7 +17,7 @@ const boot = require('./boot')
 const components = require('./components')
 
 class IPFS extends EventEmitter {
-  constructor (options) {
+  constructor (options = {}) {
```
dignifiedquire: you can just do `options = options != null ? options : {}`

VictorBjelkholm:
We should add a test case to make sure all defaults are working as well, so this doesn't bite us again.

diasdavid:
@VictorBjelkholm defaults mean "connect to the real network"

VictorBjelkholm:
@diasdavid we're currently doing that anyways in the tests, for example: https://github.com/ipfs/js-ipfs/blob/master/test/core/create-node.spec.js

Should just have a additional test-case with no options provided.

diasdavid:
We don't interfere with user repo in tests.

PR #814, opened 2017-03-27
Fixed error message outputs/handling for help commands in CLI

rodkeys:
Starting jsipfs without any commands crashes it and gives a slightly unclear error message from yargs.

I changed this to have yargs display the original help message on starting jsipfs with no commands.

Also there is a fix for the --help command messages not being displayed at all.

### DIFF
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -17,7 +17,9 @@ const cli = yargs
   .commandDir('commands')
   .demandCommand(1)
   .fail((msg, err, yargs) => {
-    if (err) throw err // preserve stack
+    if (err) {
+      throw err // preserve stack
+    }
     console.error(yargs.help())
     process.exit(1)
   })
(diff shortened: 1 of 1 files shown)


diasdavid:
@rodkeys thank you! Great catch, mind updating the tests to reflect this?

src/cli/bin.js
```diff
@@ -16,6 +16,12 @@ updateNotifier({
 const cli = yargs
   .commandDir('commands')
   .demandCommand(1)
+  .fail((msg, err, yargs) => {
+    if (err) throw err // preserve stack
+    console.error(yargs.help())
+    process.exit(1)
+  })
+
```
nunofmn: The Travis tests are failing, because the linting is failing with this added empty line :)

src/cli/bin.js
```diff
@@ -16,6 +16,12 @@ updateNotifier({
 const cli = yargs
   .commandDir('commands')
   .demandCommand(1)
+  .fail((msg, err, yargs) => {
+    if (err) throw err // preserve stack
+    console.error(yargs.help())
```
nunofmn: Here maybe you could use the [showHelp()](https://github.com/yargs/yargs#showhelpconsolelevelerror) function instead.

src/cli/bin.js
rodkeys: Thank you for catching this, I will update my code

PR #817, opened 2017-03-28
test: add interop to circle

diasdavid:
I'm adding interop tests to circle, as I've found today that they broke with the latest unixfs-engine changes.

@pgte could you take a look? It seems that the objects that go-ipfs seems might be missing some fields that now js-ipfs expects.


diasdavid:
all green \o/ , thank you @pgte :)

PR #825, opened 2017-04-04
Add FOSSA reports/badges

xizhao:


diasdavid:
Did we enable this service?

PR #827, opened 2017-04-06
fix: make start an async event

diasdavid:


diasdavid:
@VictorBjelkholm wanna check if there is any other event?

src/core/components/start.js
```diff
@@ -3,19 +3,20 @@
...
+const setImmediate = require('async/setImmediate')
 
 module.exports = (self) => {
   return (callback) => {
     callback = callback || function noop () {}
 
     const done = (err) => {
       if (err) {
-        self.emit('error', err)
+        setImmediate(() => self.emit('error', err))
```
dignifiedquire: I suggest making the callbacks async as well to be consistent

diasdavid: Note that callbacks are already async, it just happens that when emitting and event, 'everything else' would happen before the callback if there was no setImmediate

Issue #758, opened 2017-02-07
Seeing a `zlib` error between browser nodes and node.js nodes?


diasdavid:
This is a known issue and we do have more than just a solution for it \o/

The problem is that the default shim for zlib that is available for webpack and browserify is simply incomplete, we've fixed this by upgrading that shim and implementing the remaining features, however, since the PR -- https://github.com/devongovett/browserify-zlib/pull/18 -- hasn't been merged for the course of almost a year, we have to shim it manually every time.

The complete zlib shim is available at: https://www.npmjs.com/package/browserify-zlib-next

Just add to your webpack config: `zlib: 'browserify-zlib-next',` or you can check an example on how to do it at: https://github.com/ipfs/js-ipfs/tree/master/examples

We are also working on getting this shim to be the default of both browserify and webpack, so that we don't have to wait for the PR above to be merged, see the developments here:

- https://github.com/webpack/node-libs-browser/issues/51
- https://github.com/substack/node-browserify/issues/1672

diasdavid:
This shouldn't be a problem anymore \o/ 

@dignifiedquire could you confirm that:

- [ ] Webpack config no longer has this substitution
- [ ] README has been updated to reflect this change
- [ ] Have examples been updated?

dignifiedquire:
@diasdavid I did not have time to do a new release for browserify-zlib I am afraid

dignifiedquire:
browserify-zlib@0.2.0 was just released :tada:

diasdavid:
Finished this with https://github.com/ipfs/js-ipfs/pull/882

PR #882, opened 2017-06-16
docs: browserify-zlib-next begone!

diasdavid:
closes https://github.com/ipfs/js-ipfs/issues/758

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,11 +142,7 @@ The CLI is available by using the command `jsipfs` in your terminal. This is ali
 
 ### Use in the browser with browserify, webpack or any bundler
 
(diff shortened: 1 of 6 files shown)


VictorBjelkholm:
I think this was a bit premature as the PRs to update zlib in webpack and browserify have yet to be merged.

PRs to watch for progress:

- https://github.com/webpack/node-libs-browser/pull/63
- https://github.com/substack/node-browserify/pull/1733

diasdavid:
It will work for [WebPack due to semver](https://github.com/webpack/node-libs-browser/blob/master/package.json#L12), not to [browserify](https://github.com/substack/node-browserify/blob/master/package.json#L27), unfortunately. 

That being said, users should not experience this as we now use multiplex by default, nevertheless, let's continue to push to get that updated in Browserify.

PR #893, opened 2017-06-30
fix(bootstrap:add): prevent duplicate inserts

### LINKED ISSUE #892: bootstrap:add prevent duplicate inserts
Right now it's possible to add the same address multiple times in bootstrap nodes list.

kenshyx:
Closes #892

### DIFF
--- a/test/http-api/over-ipfs-api/bootstrap.js
+++ b/test/http-api/over-ipfs-api/bootstrap.js
@@ -31,7 +31,11 @@ module.exports = (ctl) => {
       it('prevents duplicate inserts of bootstrap peers', () => {
         return ctl
           .bootstrap
-          .add(validIp4)
+          .rm(null, { all: true })
+          .then((res) => {
+            expect(res.Peers.length).to.equal(0)
+            return ctl.bootstrap.add(validIp4)
+          })
           .then(res => {
             expect(res).to.be.eql({ Peers: [validIp4] })
             return ctl.bootstrap.add(validIp4)
@@ -44,8 +48,7 @@ module.exports = (ctl) => {
             expect(res).to.exist()
             const insertPosition = res.Peers.indexOf(validIp4)
             expect(insertPosition).to.not.equal(-1)
-            const duplicatePosition = res.Peers.indexOf(validIp4, insertPosition + 1)
-            expect(duplicatePosition).to.equal(-1)
+            expect(res.Peers.length).to.equal(1)
           })
       })
 
--- a/test/http-api/over-ipfs-api/bootstrap.js
+++ b/test/http-api/over-ipfs-api/bootstrap.js
@@ -12,7 +12,7 @@ module.exports = (ctl) => {
     const validIp4 = '/ip4/101.236.176.52/tcp/4001/ipfs/QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z'
     let peers
 
-    describe.only('.add', () => {
+    describe('.add', () => {
       it('returns an error when called with an invalid arg', (done) => {
         ctl.bootstrap.add(invalidArg, (err) => {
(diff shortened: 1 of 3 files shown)


test/http-api/over-ipfs-api/bootstrap.js
```diff
@@ -28,6 +28,27 @@ module.exports = (ctl) => {
...
+          .then((res) => {
+            expect(res).to.be.eql({ Peers: [validIp4] })
+            return ctl.bootstrap.list()
+          })
+          .then((res) => {
+            expect(res).to.exist()
+            const insertPosition = res.Peers.indexOf(validIp4)
+            expect(insertPosition).to.not.equal(-1)
+            const duplicatePosition = res.Peers.indexOf(validIp4, insertPosition + 1)
+            expect(duplicatePosition).to.equal(-1)
```
VictorBjelkholm: Add a expectation about that the bootstrap list should now only have one element in it as well.

kenshyx: Line 46 verifies if the newly added peer was inserted. The bootstrap list contains the default peer list from https://github.com/ipfs/js-ipfs/tree/master/src/init-files + the ones added manually.

VictorBjelkholm: As far as I can tell, you're asserting that validIp4 is not missing from the list of peers, but never asserting that the amount of peers are correct in the test-case. Ideal, minimal test-case would start with a empty bootstrap list, add the same address twice and assert that the list only contain once element.

VictorBjelkholm: That way, this test would also be protected against future changes to the default bootstrap list.

kenshyx: Ok, how do I set the bootstrap list to be empty without calling rm for the default bootstrap list?

test/http-api/over-ipfs-api/bootstrap.js
```diff
@@ -28,6 +28,30 @@ module.exports = (ctl) => {
...
+            expect(res).to.be.eql({ Peers: [validIp4] })
+            return ctl.bootstrap.list()
+          })
+          .then((res) => {
+            expect(res).to.exist()
+            const insertPosition = res.Peers.indexOf(validIp4)
+            expect(insertPosition).to.not.equal(-1)
+            expect(res.Peers.length).to.equal(1)
+          })
+      })
```
diasdavid: Please write tests for `core` too. The ideal would be getting `interface-ipfs-core` tests for bootstrap - https://github.com/ipfs/interface-ipfs-core/tree/master/src so that we match expectations on js-ipfs-api and js-ipfs.

test/http-api/over-ipfs-api/bootstrap.js
diasdavid: I'm not getting your question @VictorBjelkholm, these test nodes have their bootstrap lists empty by default:

- https://github.com/ipfs/js-ipfs/blob/master/test/utils/ipfs-factory-instance/default-config.json#L39
- https://github.com/ipfs/js-ipfs/blob/master/test/utils/ipfs-factory-daemon/default-config.json#L48

VictorBjelkholm: Ah, that was what I thought... @kenshyx then you can remove the `.rm` call and just add the two addresses.

diasdavid:
I'll go ahead and merge this. Let's not forget about the standardization of API and tests though - https://github.com/ipfs/interface-ipfs-core/issues/97

test/http-api/over-ipfs-api/bootstrap.js
diasdavid: @kenshyx is doing `rm` because the previous tests add up to the bootstrap peers pool

PR #895, opened 2017-07-03
runtime specific things - bring libp2p bundles here

diasdavid:
Following: https://github.com/libp2p/js-libp2p/issues/92

[Moving bundles and tests from libp2p-ipfs-* to js-libp2p](https://github.com/libp2p/js-libp2p/pull/94)


src/core/components/libp2p.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const Node = require('libp2p-ipfs-nodejs')
+const Node = require('../runtime/libp2p-nodejs')
```
pgte: Perhaps add a comment before explaining that this will be switched to file x in the browser?

src/core/runtime/libp2p-nodejs.js
```diff
@@ -0,0 +1,80 @@
...
+  if (muxerPrefs && !muxers) {
+    return mapMuxers(muxerPrefs.split(','))
+  } else if (muxers) {
+    return mapMuxers(muxers)
+  } else {
+    return [multiplex, spdy]
+  }
+}
+
+class Node extends libp2p {
```
pgte: This file has a lot in common with it's browser version. Perhaps we could reduce this to a declarative config and not need to declare a class here?

test/http-api/over-ipfs-api/bootstrap.js
```diff
@@ -32,6 +32,7 @@ module.exports = (ctl) => {
         ctl.bootstrap.add({ default: true }, (err, res) => {
           expect(err).to.not.exist()
           peers = res.Peers
+          console.log(res)
```
pgte: letfover console.log?

src/core/runtime/libp2p-nodejs.js
diasdavid: True, at the same time it serves the purpose of being an awesome template for others to c&p

PR #901, opened 2017-07-07
swarm - move isConnected filter from addrs to peers

kumavis:
we sre accidentally filtering for isConnected on `swarm.addr` instead of `swarm.peers`
discussion here https://github.com/libp2p/js-libp2p-swarm/issues/221


VictorBjelkholm:
This should have included a fix to the tests as well, because this seems to not have been caught in any tests.

PR #903, opened 2017-07-07
Cli block get pipe fix

kumavis:
- disables execas's newline trimming for explicit newline testing
- changes the cli `block.get` to not append an additional newline
- adds an additional test for a block without a final newline


kumavis:
trifecta 👌

Issue #804, opened 2017-03-21
Deploying js-ipfs to the Galaxy


diasdavid:
Similar to go-ipfs, we want to run some js-ipfs nodes and monitor their behaviour. From a quick chat with @lgierth on IRC, here are some notes:

```
14:48 <@lgierth> daviddias: about running js-ipfs, would be best if we can have a dockerfile, and metrics exposed at :5001/debug/metrics/prometheus
14:49 <@lgierth> like, number of peers and such stuff
14:49 <@lgierth> datastore stats
14:50 <@daviddias> lgierth: thanks, that is what I needed to know :) 
14:50 <@daviddias> will let you know once we have metrics exposed, right now we have none
14:52 <@lgierth> memory/cpu usage and gc timings are welcome too but can at least mem/cpu can be pulled out of docker too
14:53 <@lgierth> relays?
15:00 <@daviddias> lgierth: is there a format that you expect coming from it?
15:06 <@lgierth> yeah prometheus text format
15:06 <@lgierth> let me see
15:07 <@lgierth> https://prometheus.io/docs/instrumenting/writing_exporters/
15:07 <@lgierth> or https://github.com/siimon/prom-client if it lgty
```

dignifiedquire:
![nasa-89125](https://cloud.githubusercontent.com/assets/790842/24155338/dbb1ad8e-0e53-11e7-9f2f-c59eae184812.jpg)

dryajov:
@lgierth @diasdavid 

I have a not there regarding passive and active relays:
> All peers are relays, but they can function in two modes:
> * active/proactive relays that dial into other nodes
>     * this are explicitly enabled by config options
> * passive relays, which only relay over known connections, no explicit dialing. If no connection exists or the requested peer, the relay is refused.

How are active relays gonna be handled, does it make sense to have them as dedicated nodes, maybe part of the bootstrap list?

https://github.com/libp2p/js-libp2p-circuit/issues/4

VictorBjelkholm:
# Metrics to expose

- Connected peers
- Memory usage
- CPU usage
- GC stats
- Datastore stats ( @diasdavid can you specify which? )

diasdavid:
@VictorBjelkholm is Datastore or Repo? Otherwise looks good :)

VictorBjelkholm:
@diasdavid 
> is Datastore or Repo? Otherwise looks good :)

Probably repository on second thought, went by the IRC log posted in the beginning.

[...]

PR #913, opened 2017-07-17
Daemon monitoring + Dockerfile

VictorBjelkholm:
Sets up collectors which exposes metrics at `$API/debug/metrics/prometheus` (as in go-ipfs) that can be consumed by Prometheus.

Ref: #804


src/http-api/routes/debug.js
```diff
@@ -0,0 +1,24 @@
+const register = require('prom-client').register
+const client = require('prom-client')
+
+// Endpoint for handling debug metrics
+module.exports = (server) => {
+  const api = server.select('API')
+  const gauge = new client.Gauge({ name: 'number_of_peers', help: 'the_number_of_currently_connected_peers' })
+
+  setInterval(() => {
```
VictorBjelkholm: Not sure about this approach. Guess we have two options, 1) (which is implemented here) is to once every X seconds, check the number of peers and save it or 2) when a request happens at `/debug/metrics/prometheus`, then check the number of peers and respond.

@diasdavid what you think?

VictorBjelkholm:
Missing: 

- [ ] Use webrtc-star by installing it globally
- [ ] Information about the endpoint in the documentation somewhere

src/http-api/index.js
```diff
@@ -23,6 +26,11 @@ function HttpApi (repo, config) {
   this.log = debug('jsipfs:http-api')
   this.log.error = debug('jsipfs:http-api:error')
 
+  // Setup debug metrics collection
+  const collectDefaultMetrics = prometheusClient.collectDefaultMetrics
+  collectDefaultMetrics({ timeout: 5000 })
+  prometheusGcStats(prometheusClient.register)()
```
diasdavid: Does this mean that every js-ipfs node will "phone home"?

src/http-api/routes/debug.js
diasdavid: let's calculate per request

src/http-api/index.js
VictorBjelkholm: @diasdavid "phone home" as in send the data somewhere? No, it'll only be in the local node, available to collect when hitting the endpoint. The data is not sent somewhere.

diasdavid: got it. Still, we should not add stuff that is not necessary by default. Can these be in a separate file that doesn't get loaded unless a env var is set?

VictorBjelkholm: Sure, that makes sense, although it makes the usage different from go-ipfs who exposes it by default. But that makes sense, I'll move the dependencies to optionalDependencies as well then.

diasdavid: Thank you! :)

diasdavid:
@VictorBjelkholm how close are we from this?

VictorBjelkholm:
@diasdavid just resolved a conflict, once CI is passing, is ready to be merged 👍

diasdavid:
rebase master onto this branch. it will solve the issues you are seeing.

diasdavid:
With the work of @harshjv and @ya7ya, we will have a gateway as well and from what I've discussed with Lars, we will be running some gateways as js-ipfs nodes for testing :)

PR #918, opened 2017-07-20
Fix linter errors

sktt:
- Use `safe-buffer` everywhere

I previously had issues with running `npm run lint`, which required me to additional eslint plugins. I can no longer reproduce that anymore. Strange..

e: Whoops, I happened to be using eslint-config-standard v10 instead of 7.. Version 7 does not yet check use of `new Buffer()`.. So this pr wont be relevant until aegir@next which will use a later version


sktt:
Closing because this is not relevant for now.. can reopen later

PR #925, opened 2017-07-23
fix: remove non existent commands

diasdavid:


diasdavid:
Removing this like `mount` from the list. Opening `ls`, `ping` and `dns` for contribution:

- https://github.com/ipfs/js-ipfs/issues/928
- https://github.com/ipfs/js-ipfs/issues/927
- https://github.com/ipfs/js-ipfs/issues/926

Issue #495, opened 2016-09-15
Migrate CLI output to Logger


VictorBjelkholm:
With the new files APIs, I've started on a logger that has many benefits compared to the current approach of using `console.log` for output.
- Intent is clear, `console.log` is often used to debugging, may not be clear it's actual output from CLI
- Easier to turn on/off if we want to add a --quiet flag that only outputs some things
- Abstract away the logging from the CLI itself
- Easier to build things that continues on the same line, commands that takes time to execute.

I don't think this is a urgent refactor we have to do right now, but maybe we can start thinking about it a little, and write new code with the logger, rather than migrating everything at once.

diasdavid:
👍 for having the ability to turn on/off. 

Since we use `debug` for dev logs (log and log.err), let's call the 'logger' that prints info to the user some other thing. Maybe 'printer'?

dignifiedquire:
`printer` sounds good, and the method should be called `print` :)

tarunbatra:
or maybe different methods, like `info`, `error` to log messages of corresponding type.

VictorBjelkholm:
@dignifiedquire agree that it should just be `print`, probably no need for many functions, just one called `print` and that would be about it.

@tarunbatra this is about printing output to the console and not logging things that can be controlled by for example log-level and such.

diasdavid:
@tarunbatra for log messages we use `debug`

diasdavid:
@VictorBjelkholm could you finish this since you got it started? :)

VictorBjelkholm:
@diasdavid I just started using it lightly. I think we shouldn't migrate everything at once but rather think about it when writing new/refactoring code (and when reviewing) so at one point, we'll be using it 100%. Does that make sense for you?

diasdavid:
Not really, it will confuse contributors and even ourselves at times as it will be always half done.

VictorBjelkholm:
@diasdavid hm, ok. Will have to push this forward then, won't have any bandwidth currently to work on it...

rasmuserik:
Hi, I might take a look at this, 
as an entry point to get to know 
the js-ipfs source code.

Feel free to assign it to me ;)

diasdavid:
@rasmuserik thank you! Let me know if you get questions along the way

rasmuserik:
Can this be closed now, or is anything else missing, after merge of #931 ?

diasdavid:
You have completed this! Thanks @rasmuserik 💖🍾

PR #931, opened 2017-07-23
Refactor printing in CLI

rasmuserik:
This changes the CLI to use a logger instead of console.out, as discussed in #495

There is a design decision about how we instantiate the logger: originally there was a logger instantiated in command/init, but I think that instantiating it in every command, will add too much boilerplate. Instead I just took the simplest approach, i.e. to instantiate an instance in the `utils` module, - and have a print function exported there. 

If we want to have a quiet option, it could then be parsed once in `bin.js`, and then the logger might be instantiated from there. Also if we intend always only have a single logger instance(?), we could remove `createLogger `, and just have the print function, plus the ability to configure whether to silence it or not.

### DIFF
--- a/src/cli/commands/bootstrap/rm.js
+++ b/src/cli/commands/bootstrap/rm.js
@@ -26,7 +26,7 @@ module.exports = {
         throw err
       }
 
-      list.Peers.forEach(peer => print(peer))
+      list.Peers.forEach((peer) => print(peer))
     })
   }
 }
--- a/src/cli/commands/bootstrap/rm.js
+++ b/src/cli/commands/bootstrap/rm.js
@@ -3,6 +3,8 @@
 const debug = require('debug')
 const log = debug('cli:bootstrap')
 log.error = debug('cli:bootstrap:error')
+const print = require('../../utils').print
+
 module.exports = {
   command: 'rm [<peer>]',
 
@@ -24,7 +26,7 @@ module.exports = {
         throw err
       }
 
-      list.Peers.forEach((l) => console.log(l))
+      list.Peers.forEach(peer => print(peer))
     })
   }
 }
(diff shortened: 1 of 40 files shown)


src/cli/commands/bootstrap/rm.js
```diff
@@ -24,7 +26,7 @@ module.exports = {
         throw err
       }
 
-      list.Peers.forEach((l) => console.log(l))
+      list.Peers.forEach(peer => print(peer))
```
diasdavid: small nitpick. Change to `(peer) => print(peer)`. Lint should have caught that.

src/cli/utils.js
```diff
@@ -83,3 +83,4 @@ exports.createLogger = (visible) => {
     }
   }
 }
+exports.print = exports.createLogger(true) // TODO refactor/remove createLogger?
```
diasdavid: Yeah, go for it.

rasmuserik:
> I like that, could you move forward with it?

Yep, I will do that :)

rasmuserik:
The `quiet` flag is now implemented :)

It is implemented as a verbosity level,
which was almost the same code, but more general,
and I  added a `verbose` flag at the same time,
as it might also be useful(?) - otherwise I'll just remove it.

src/cli/commands/bootstrap/rm.js
rasmuserik: Fixed in latest version of the pull request :)

src/cli/utils.js
rasmuserik: done.

diasdavid:
Mind adding tests for the new modes @rasmuserik ?

We just need:

- q
- regular
- verbose 

No need for multiple levels of 'verbosity'

rasmuserik:
I'll add a test for the quiet flag, 
&dash; and remove the verbose part.
The test of the regular mode, 
should be covered by existing tests :)

rasmuserik:
done :)

diasdavid:
Thank you @rasmuserik !! :D

PR #937, opened 2017-07-26
fix: stream issue, do not use isstream, use is-stream

diasdavid:


diasdavid:
Learned that [`isstream` uses instanceOf](https://github.com/rvagg/isstream/blob/master/isstream.js#L4-L6) while [`is-stream` doesn't](https://github.com/sindresorhus/is-stream/blob/master/index.js#L3-L6).

Issue #927, opened 2017-07-23
Implement the `jsipfs ls` command


diasdavid:
Following: https://github.com/ipfs/js-ipfs/pull/925/files

We want to support the same command that go-ipfs supports: `jsipfs ls`.

From go-ipfs:
```
» ipfs ls --help
USAGE
  ipfs ls <ipfs-path>... - List directory contents for Unix filesystem objects.

SYNOPSIS
  ipfs ls [--headers | -v] [--resolve-type=false] [--] <ipfs-path>...

ARGUMENTS

  <ipfs-path>... - The path to the IPFS object(s) to list links from.

OPTIONS

  -v,           --headers bool - Print table headers (Hash, Size, Name). Default: false.
  --resolve-type          bool - Resolve linked objects to find out their types. Default: true.

DESCRIPTION

  Displays the contents of an IPFS or IPNS object(s) at the given path, with
  the following format:

    <link base58 hash> <link size in bytes> <link name>

  The JSON output contains type information.
```

rasmuserik:
I will probably take a look at this next week.

rasmuserik:
I am starting to look at this now :)

Apparently, at first look, it seems like
directories are currently stored using
using the old Protocol Buffer format,
and not CBOR encoded IPLD?

Are we still only using the (legacy?) protocol buffers, 
or should we begin to use files/directories in IPLD?

---

(Adding a directory yields a block which definitely looks like a protocol buffer, and not CBOR. 
I found an example of what a directory in IPLD [looks like](https://github.com/ipld/specs/tree/master/ipld#a-directory), but the actual spec for [files](https://github.com/ipfs/specs/tree/master/files) / [unixfs](https://github.com/ipfs/specs/tree/master/unixfs) is empty..)

diasdavid:
@rasmuserik `dag-pb` is not legacy, it is just one of the many formats that IPLD supports. [`unixfs-engine`](https://github.com/ipfs/js-ipfs-unixfs-engine) uses `dag-pb` and will continue to :)

rasmuserik:
nice, thanks, that makes things a lot more clear :) 
(I thought they were legacy due to [this part](https://github.com/ipld/specs/blob/master/ipld/ipld-compat-protobuf.md) of the IPLD specs)

diasdavid:
Oh, thanks. We need to review all IPLD specs, added a note to the top of that one to make sure we don't misinform anyone else. Thanks for catching that :)

[...]

PR #940, opened 2017-08-02
WIP: ls

rasmuserik:
Implementation of `js-ipfs ls` (#927)

**Tasks:**

- [x] add `ls` command
- [x] implement initial working version
- [x] add `-v` / `--headers` support
- [x] indent/present results similar to the output of the go implementation
- [x] use ipfs-unixfs to check filetype, instead of just checking raw data
- [x] tests
- [ ] move implementation into core, and add http-api
- [ ] move the tests from <https://github.com/ipfs/js-ipfs-api/blob/master/test/ls.spec.js> to <https://github.com/ipfs/interface-ipfs-core>
- [ ] write ls spec on <https://github.com/ipfs/interface-ipfs-core/blob/master/src/files.js>
- [ ] make sure that ls tests pass in both js-ipfs-api and js-ipfs

This implements a basic working version of `js-ipfs ls`, supporting listing of directories, with similar output as `go-ipfs ls`. 

Notice: the progress will be slow, as I only have a few hours per week for this kind of projects.

### DIFF
--- a/src/cli/commands/ls.js
+++ b/src/cli/commands/ls.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const {print, rightpad} = require('../utils')
+const utils = require('../utils')
 const Unixfs = require('ipfs-unixfs')
 
 module.exports = {
@@ -51,8 +51,8 @@ module.exports = {
       const sizeWidth = Math.max.apply(null, links.map((file) => String(file.size).length))
 
       links.forEach((file) => {
-        print(rightpad(file.multihash, multihashWidth + 1) +
-          rightpad(file.size, sizeWidth + 1) +
+        utils.print(utils.rightpad(file.multihash, multihashWidth + 1) +
+          utils.rightpad(file.size, sizeWidth + 1) +
             file.name)
       })
     })
--- a/src/cli/commands/ls.js
+++ b/src/cli/commands/ls.js
@@ -1,0 +1,60 @@
+'use strict'
+
+const {print, rightpad} = require('../utils')
+const Unixfs = require('ipfs-unixfs')
+
+module.exports = {
+  command: 'ls <key>',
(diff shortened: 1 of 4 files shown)


src/cli/commands/ls.js
```diff
@@ -0,0 +1,60 @@
+'use strict'
+
+const {print, rightpad} = require('../utils')
```
diasdavid: We are not using destructuring just yet. Could you change it?

Why? We review the JS best practices every quarter or so and commit to a set of things, this helps us agree on codestyle and move forward faster :)

rasmuserik: Yes, that is changed now. Is there a place where the agreed code style is documented?
(The destructuring is not mentioned [js-project-guidelines.md](https://github.com/ipfs/community/blob/master/js-project-guidelines.md), nor in the [standard](https://github.com/standard/standard) code style)

rasmuserik:
I've added the tasks you mentioned to the list, and will look into it, when I have some time.

diasdavid:
Hi @rasmuserik how are things going? Any hopes of getting ls to see the light of day? Can we help?

rasmuserik:
Hi @diasdavid, I am busy with other projects at the moment, so unfortunately I do not have time to look into it for the time being (and no recent progress) &dash; it is on my backlog, but very very far down :(
It is also fine with me if somebody else takes over this issue (either build on the code I made here, or discard it and build it from scratch).

rasmuserik:
Closing this pull-request, as I am too busy with other projects.

diasdavid:
Let's keep it open and just signal that is open for contributions :)

pgte:
Being continued in https://github.com/ipfs/js-ipfs/pull/1073

PR #955, opened 2017-08-24
[WIP] feat: adding ability to specify a key type to use

ETeissonniere:
Depends on libp2p/js-peer-id#65

The following PR is enabling the possibility to specify the key type to use while running the `jsipfs init` command. Even if it isn't really useful at the moment, it open the path for a more modular implementation regarding key management.

### DIFF
--- a/src/cli/commands/init.js
+++ b/src/cli/commands/init.js
@@ -21,6 +21,12 @@ module.exports = {
       alias: 'e',
       type: 'boolean',
       describe: "Don't add and pin help files to the local storage"
+    },
+    type: {
+      alias: 't',
+      type: 'string',
+      default: 'RSA',
+      describe: 'Choose the key type used by your node'
     }
   },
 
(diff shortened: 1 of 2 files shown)


DeveloppSoft:
Hi @diasdavid thanks for the follow up! Yeah indeed I agree with you, let's wait.

DeveloppSoft:
Hi @diasdavid I finally have some spare time, so if you still want to lead me on this I am ready :wink:

diasdavid:
Awesome @DeveloppSoft ! Let's start with something of smaller scope and I say this because this multikey support is something also still being figured out and merged in go-ipfs and I would rather make sure that all decisions have been made than having to guide you through flux.

Wanna pick some of the issues available here https://github.com/ipfs/js-ipfs/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22 to get started and familiar with the codebase?

jacobheun:
Fixed with https://github.com/ipfs/js-ipfs/issues/2553

PR #968, opened 2017-08-29
feat: add gateway route to daemon

diasdavid:
@ya7ya @harshjv

PR #971, opened 2017-08-30
clean up, remove commented out lines.

ya7ya:
Code cleanup as requested in #968 

Changes
- [x] Delete commented out lines
- [ ] rename `http` to `daemon` . 
(In [`go-ipfs`](https://github.com/ipfs/go-ipfs/tree/master/core/corehttp) the module is still called `http` though.)
- [ ] Enable `interface` and `ipfs-over-api` Tests. 
(they're enabled in `tests/http-api` but not `test/gateway` so there won't be duplicate tests)

@diasdavid Please Let me know if the 2 remaining points are greenlit and i'll push the edits ASAP. Thanks.


diasdavid:
> (In go-ipfs the module is still called http though.)

Let's keep it as http then.

> Enable interface and ipfs-over-api Tests.

These tests are designed to work over the HTTP-API. Testing the gateway requires a new set of tests.

ya7ya:
Yes the gateway requires different tests. I removed the commented out blocks. and I will try to add more tests beside the ones in `tests/gateway/spec/gateway.js`

test/gateway/index.js
```diff
@@ -40,16 +40,4 @@ describe('HTTP GATEWAY', () => {
     fs.readdirSync(path.join(__dirname, '/spec'))
       .forEach((file) => require('./spec/' + file)(http))
   })
```
diasdavid: @ya7ya there are not "gateway spec tests" you probably just have "tests" :)

ya7ya: @diasdavid ma bad, I'll fix that,
I'm looking at the [`gateway_test.go`](https://github.com/ipfs/go-ipfs/blob/master/core/corehttp/gateway_test.go) to try to port these tests to `js-ipfs` .

diasdavid: awesome! That's perfect @ya7ya :)

diasdavid:
@ya7ya adding you to the JS team in IPFS so that we can collaborate on the same branch :) https://github.com/ipfs/js-ipfs/pull/968

PR #975, opened 2017-09-01
feat/bootstrapers gateway

diasdavid:
applying #973


diasdavid:
@VictorBjelkholm just discovered that the last PR introduced an issue for when multiaddrs are DNS. It seems that `connectedAddr` is coming as null and so any .toString on it fails.

This escaped because tests do not use DNS addrs.

PR #983, opened 2017-09-03
fix #952 : adding +50Mb files via createAddStream

### LINKED ISSUE #952: Adding 50Mb+ Files to a browser js-ipfs crashes Chrome
- **Version**: 0.25.1
- **Platform**: Ubuntu 16.04 4.4.0-42-generic , Chrome Version 57.0.2987.133 (64-bit)  and Win 10 (64-bit) Chrome Version 60.0.3112.90 (Official Build) (64-bit)
- **Subsystem**: `files.add`

- go to the `exchange-files-in-browser` example directory. 
- run `npm install`
- `npm start`
- follow along with the README.md , but upload a large file, around 90MB. 

Edit: Version fix.

ya7ya:
Hey , this is a fix for issue #952 
This uses `createAddStream` instead of add to add a fairly large file in the browser without crashing.

### DIFF
--- a/examples/exchange-files-in-browser/package.json
+++ b/examples/exchange-files-in-browser/package.json
@@ -2,10 +2,16 @@
   "name": "exchange-files-browser",
   "version": "0.0.0",
   "scripts": {
-    "start": "http-server -c-1 -p 12345 public"
+    "bundle": "browserify public/js/app.js > public/js/bundle.js",
+    "serve": "http-server -c-1 -p 12345 public",
+    "start": "npm run bundle && npm run serve"
   },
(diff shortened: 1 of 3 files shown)


diasdavid:
Although the solution should be really figuring out what is causing the mem leak, this sounds like a good interim solution.

@pgte it seems that files.add is somehow doing a memleak

ya7ya:
@diasdavid 👍 , Thanks. Yes this require alot more work to figure out the issue, but for use in an example it's important to keep code that doesn't crash. 

I think we should keep the issue open till we nail the main cause of the mem leak.

PR #988, opened 2017-09-04
docs(example): show folks how to use pull-streams instead

dignifiedquire:
Always stream incoming data through pull streams and use browserify-aes directly from master now. I was able to add a 300Mb sized file with this quite quickly.

Fixes most issues in https://github.com/ipfs/js-ipfs/issues/952  I believe

### DIFF
--- a/examples/exchange-files-in-browser/public/index.html
+++ b/examples/exchange-files-in-browser/public/index.html
@@ -62,7 +62,7 @@
     </div>
 
     <!-- The IPFS node module -->
-    <script src="//unpkg.com/ipfs/dist/index.min.js"></script>
+    <!-- <script src="//unpkg.com/ipfs/dist/index.min.js"></script> -->
     <!-- <script src="js/app.js"></script> -->
     <script src="js/bundle.js"></script>
   </body>
--- a/examples/exchange-files-in-browser/public/js/app.js
+++ b/examples/exchange-files-in-browser/public/js/app.js
@@ -1,5 +1,4 @@
 'use strict'
-/* global self */
 
 const $startButton = document.querySelector('#start')
 const $stopButton = document.querySelector('#stop')
@@ -20,7 +19,9 @@ const $details = document.querySelector('#details')
 const $allDisabledButtons = document.querySelectorAll('button:disabled')
 const $allDisabledInputs = document.querySelectorAll('input:disabled')
 const $filesList = document.querySelector('.file-list')
-const streamBuffers = require('stream-buffers')
+const Ipfs = require('../../../../src/core')
+const pullFilereader = require('pull-filereader')
+const pull = require('pull-stream')
 
 let node
 let peerInfo
@@ -33,7 +34,7 @@ function start () {
   if (!node) {
     updateView('starting', node)
 
-    node = new self.Ipfs({repo: 'ipfs-' + Math.random()})
+    node = new Ipfs({repo: 'ipfs-' + Math.random()})
 
     node.on('start', () => {
       node.id().then((id) => {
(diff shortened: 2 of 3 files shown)


examples/exchange-files-in-browser/public/index.html
```diff
@@ -62,7 +62,7 @@
     </div>
 
     <!-- The IPFS node module -->
-    <script src="//unpkg.com/ipfs/dist/index.min.js"></script>
+    <!-- <script src="//unpkg.com/ipfs/dist/index.min.js"></script> -->
```
diasdavid: This was supposed to be the example that showed that "no extra batteries required". Kind of sad we lost that.

examples/exchange-files-in-browser/public/js/app.js
```diff
@@ -119,93 +120,44 @@ function onDrop (event) {
...
-          })
+  pull(
+    pull.values(files),
+    pull.through((file) => console.log('Adding %s', file)),
+    pull.asyncMap((file, cb) => pull(
+      pull.values([{
+        path: file.name,
+        content: pullFilereader(file)
+      }]),
+      node.files.createAddPullStream(),
```
diasdavid: This is awesome, but we shouldn't have examples with undocumented (and already agreed to change) API

Needs this solved first. https://github.com/ipfs/interface-ipfs-core/issues/126

diasdavid:
Awesome job @dignifiedquire! I'll keep this PR open until https://github.com/ipfs/js-ipfs/pull/988#discussion_r136845093 is solved

examples/exchange-files-in-browser/public/index.html
VictorBjelkholm: @diasdavid unpkg doesn't allow linking to a branch? We could also publish latest master ourselves on IPFS and link that instead

VictorBjelkholm:
Hm, a bit worried about this as most people haven't been exposed to pull-streams and we're supposed to support the API that most people use as well. This seems to amount to that even though you can use js-ipfs without using pull-streams, it's really inefficient and you should probably use pull-streams.

As this is supposed to be a simple example, I'm wondering if it's not best to stick with the non-pullstreams version.

VictorBjelkholm:
Maybe we could show it's done with non-pullstreams in comments (risk of having code that is not being run though, might not actually work...) or make the pullstream into a separate example for better performance.

examples/exchange-files-in-browser/public/index.html
dignifiedquire: We don't have to keep this, with the new release that would be fine to use the unkpkg version again

dignifiedquire:
We are currently alienating people with the browser crashing, so I think having pull-streams is much better than crashing and being slow. node streams are really not meant to be used in the browser in most situations I don't think it's good to show examples with them, rather use non streaming methods by default and educate users about better options

diasdavid:
We will have both API and once they are properly documented, we can use them in the examples. This is one of the priorities for 0.27 (and probably the target thing to get it released)

diasdavid:
Actively working on https://github.com/ipfs/interface-ipfs-core/pull/162 this week to then finish up this PR

diasdavid:
Who was interested on this PR, please review https://github.com/ipfs/interface-ipfs-core/pull/162#issuecomment-341691928

diasdavid:
Redirecting the efforts here to this PR https://github.com/ipfs/js-ipfs/pull/1086

mitra42:
@diasdavid - it would be useful to leave issues open until they are solved, even if the solution is elsewhere.  I for one have been watching this thread in the hope of finding the docs implied by its title, " show folks how to use pull-streams instead  "  with the direction to #1086 , I can't find any answer to this question in that issue nor in ipfs/interface-ipfs-core#162, and I'm probably not the only one waiting on docs before converting app code to use pull-streams.

diasdavid:
@mitra42 valid point. I tend to keep issues and PR open until they are fixed or converge elsewhere. It is hard to keep track of multiple issues discussing the same things and that is why I always leave a pointer saying that "x discussion continues at issue y".

@mitra42, for your case, https://github.com/ipfs/interface-ipfs-core/pull/162 brought you something even better as you won't have to worry about pull-streams and just use ipfs.files.add with files and get hashes back. See the updated https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#add

See the updated example on https://github.com/ipfs/js-ipfs/pull/1086 that this PR was also updating (direct link here https://github.com/ipfs/js-ipfs/blob/f5e756106f5f3a4cd92cbea43863b35bf8d90a99/examples/exchange-files-in-browser/public/js/app.js#L102-L135),  as you will see there are no Readable or Pull Streams involved. I used this code files bigger than 750Mb, a significant improvement.

mitra42:
Email sent as this goes into multiple topics.

PR #989, opened 2017-09-04
Changes required for #968 feat/Gateway 

ya7ya:
These are some of the required Changes by @diasdavid to merge #968 

- [x] use `unix-fs` to detect directories
- [x] check for CID not hashes
- [x] use `js-ipfs-repo` instead of `go-ipfs-repo` for testing gateway
- [x] GATEWAY can be Gateway as it is not an acronym


test/gateway/index.js
```diff
@@ -10,8 +10,8 @@ const ncp = require('ncp').ncp
 const path = require('path')
 const clean = require('../utils/clean')
 
-describe('HTTP GATEWAY', () => {
-  const repoExample = path.join(__dirname, '../go-ipfs-repo')
+describe('HTTP Gateway', () => {
+  const repoExample = path.join(__dirname, '../js-ipfs-repo')
```
diasdavid: Hi @ya7ya. It seems I didn't explain myself well. What I mean is that there should be no "repo copying" either from a previously generated go-ipfs or js-ipfs node. When you spawn a daemon after doing init, it will have its own repo created, that is what should be used for testing.

ya7ya: @diasdavid oh, Ma bad, I'll fix that. Sorry.

diasdavid: No, my bad for not being clear. Thank you for pushing this :)

diasdavid:
Thank you @ya7ya :)

ya7ya:
@diasdavid 👍 :) Thank you for your review and comments, I appreciate it

PR #1003, opened 2017-09-07
required changes for the gateway

ya7ya:
These are the required changes requested by @dignifiedquire in #968 

**UDPATE**: I just noticed @VictorBjelkholm requested changes, I'll implement them and push them to this branch as well.

Changes:
- [x] use slice instead of splice
- [x] remove noop since promisify always defines a callback
- [x] use async reduce instead of eachOfSeries
- [x] no duplicate logging + returning error
- [x] smallCamelCase for non-constructors
- [x] fix indentations
- [x] use callbacks instead of promises.
- [x] use `pull.through` instead of `pull.drain` so response isn't completely held in memory before responding
- [x] use `switch` since there are too many conditions to check

- [ ] [use templating engine.](https://github.com/ipfs/js-ipfs/pull/968#pullrequestreview-61138346) (still in discussion, no clear decision yet. Let me know when you guys decide)
- [ ] [internal `cat` with pull streams for efficiency.](https://github.com/ipfs/js-ipfs/pull/968#pullrequestreview-61143881) (outside the scope of this feature. but a very cool idea that I think we should have)

Please Let me know if there are any other changes required. Thanks for your help 👍

### DIFF
--- a/src/http/gateway/resources/gateway.js
+++ b/src/http/gateway/resources/gateway.js
@@ -9,7 +9,7 @@ const fileType = require('file-type')
 const mime = require('mime-types')
 const Stream = require('readable-stream')
 
-const GatewayResolver = require('../resolver')
+const gatewayResolver = require('../resolver')
 const PathUtils = require('../utils/path')
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


diasdavid:
> use templating engine. (still in discussion, no clear decision yet. Let me know when you guys decide)

I'm fine with doing as it is for now and adding that to a separate issue.

> internal cat with pull streams for efficiency. (outside the scope of this feature. but a very cool idea that I think we should have)

This will be the focus of js-ipfs 0.27 -> https://github.com/ipfs/interface-ipfs-core/issues/126

src/http/gateway/resources/gateway.js
```diff
@@ -29,113 +29,116 @@ module.exports = {
...
+                    .header('Access-Control-Allow-Origin', '*')
+                    .header('Access-Control-Expose-Headers', 'X-Stream-Output, X-Chunked-Ouput')
+                    .send()
+                } else {
+                  response
+                   .header('Access-Control-Allow-Headers', 'X-Stream-Output, X-Chunked-Ouput')
+                   .header('Access-Control-Allow-Methods', 'GET')
+                   .header('Access-Control-Allow-Origin', '*')
+                   .header('Access-Control-Expose-Headers', 'X-Stream-Output, X-Chunked-Ouput')
+                   .send()
```
diasdavid: Some of these addrs are not required. I'll fix it in the main PR.

PR #1006, opened 2017-09-08
Fix failing gateway tests with this one simple trick!

ya7ya:
Fix for the failing tests in #968


test/gateway/index.js
```diff
@@ -24,12 +22,7 @@ describe('HTTP Gateway', () => {
   let gateway
 
   before((done) => {
-    const repoPath = path.join(
-      os.tmpdir(),
-      '/ipfs-' + Math.random().toString().substring(2, 8) + '-' + Date.now()
-    )
-
-    http.api = new API(repoPath, {
```
diasdavid: This is incorrect, explanation why here -> https://github.com/ipfs/js-ipfs/pull/968#issuecomment-328079053

ya7ya: @diasdavid okay, Got it, I'll make sure tests are self-contained and don't use the local repo. Sorry for misunderstanding.

diasdavid: No worries at all :) Thank you for working on the fix.

Issue #998, opened 2017-09-06
distribute js-ipfs binaries with pkg?


mkg20001:
There is a really cool tool out there (https://github.com/zeit/pkg) that allows nodejs apps to be packaged as binaries.
This would have the following benefits:
 - User would not have to download and install nodejs
 - Faster installation (npm is slow)
 - Newer nodejs version available

dryajov:
That would be super cool, really like pkg.

diasdavid:
If we were to do this, it would enable jsipfs and goipfs to be tested more easily against the same sharness tests in some CI (//cc @VictorBjelkholm).

We could also make it available through dist.ipfs.io, although that would be a lesser priority.

mkg20001:
I've tried it out (just run `pkg .`)

The cli has some missing commands if the binary is run outside of the source folder (all except add, cat, get, completion) and if a repo exists an error occurs because leveldown.node could not be found (pkg for now does not bundle .node files)

```
Kommandos:
  add <file>       Add a file to IPFS using the UnixFS data format
  cat <ipfs-path>  Fetch and cat an IPFS path referencing a file
  get <ipfs-path>  Fetch a file or directory with files references from an IPFS
                   Path
  completion       generate bash completion script

Optionen:
  --silent  Write no output                          [boolean] [Standard: false]
  --help    Hilfe anzeigen                                             [boolean]

```

```

/snapshot/js-ipfs/node_modules/levelup/lib/leveldown.js:31
    throw requireError(e)
    ^
LevelUPError: Failed to require LevelDOWN (Could not locate the bindings file. Tried:
 → /snapshot/js-ipfs/node_modules/leveldown/build/leveldown.node
 → /snapshot/js-ipfs/node_modules/leveldown/build/Debug/leveldown.node
 → /snapshot/js-ipfs/node_modules/leveldown/build/Release/leveldown.node
 → /snapshot/js-ipfs/node_modules/leveldown/out/Debug/leveldown.node
 → /snapshot/js-ipfs/node_modules/leveldown/Debug/leveldown.node
 → /snapshot/js-ipfs/node_modules/leveldown/out/Release/leveldown.node
 → /snapshot/js-ipfs/node_modules/leveldown/Release/leveldown.node
 → /snapshot/js-ipfs/node_modules/leveldown/build/default/leveldown.node
[...]

PR #1007, opened 2017-09-08
feat: pkg binaries

mkg20001:
Ref #998
With this pr it is possible to create binaries that embed the natives and commands (`npm run pkg`).
The module used to do that (https://github.com/mkg20001/pkg-natives) still isn't finished but the binaries seem to work.

Notes:
 - ~~Currently linux only~~
 - ~~Mac OSX may work (travis ci didn't run yet for mac)~~
 - ~~Windows probably not~~ (added thanks to appveyor)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 test/repo-tests*
 **/bundle.js
 
+/ipfs
+/ipfs.exe
+
 # Logs
 logs
 *.log
--- a/package.json
+++ b/package.json
@@ -6,6 +6,19 @@
     "jsipfs": "src/cli/bin.js"
   },
   "main": "src/core/index.js",
+  "pkg": {
+    "scripts": "src/cli/commands/**/*.js",
+    "assets": "natives/**"
+  },
+  "pkg-native": {
+    "scan": "all",
+    "mode": "bundle",
+    "modules": [
+      "node_modules/fs-ext",
+      "node_modules/sha3",
+      "node_modules/keccak/bindings.js"
+    ]
+  },
   "browser": {
     "./src/core/components/init-assets.js": false,
     "./src/core/runtime/config-nodejs.json": "./src/core/runtime/config-browser.json",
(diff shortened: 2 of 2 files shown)


dryajov:
I really like this, but we need tests to make sure that the generated package is working. As I ramp up on js-ipfs, I'm not entirely sure whats the best approach to test this is in the context of the package. Should we adapt ipfs-ctld to be able to work with go/js-ipfs/js-ipfs-pkg daemons, how do we run the sharnes tests against this, etc...

I'd love to hear some ideas from: @VictorBjelkholm @diasdavid @dignifiedquire 

Other than that 💯 !

VictorBjelkholm:
My first concern is adding another surface that we are gonna have to maintain for maybe not that big benefit. As I understand, the only benefit we get from packaging js-ipfs this way is that we get a binary that comes with nodejs, so it's not necessary to install the runtime.

Question regarding how packaging of modules happens as well, we're gonna have to have builds for windows, osx and multiple different linux dists, which I'm not sure we're ready to take on just yet.

For testing purposes, it might be better to use go-ipfs and js-ipfs as containers instead, as we already have infrastructure for both deployment and testing for it.

dryajov:
I agree with @VictorBjelkholm, this will definitely add to the maintenance burden for relatively little gain from the project perspective, but it is a really cool option if we ever want/need to distribute js-ipfs as a binary. Also, now that @mkg20001 did the legwork, it is an option for the community to use if such a need arises.

👍

mitra42:
I tend to agree, for our library we use browserify which bundles the parts it needs of js-ipfs along with other stuff into a single binary, so a pre-bundled binary doesnt help as much.

mkg20001:
> I tend to agree, for our library we use browserify which bundles the parts it needs of js-ipfs along with other stuff into a single binary, so a pre-bundled binary doesnt help as much.

@mitra42 What you are talking about is a bundle (aka just a bunch of js code in a single file). What this adds is a standalone executable similar to go-ipfs's `.exe` file that can be launched without having the nodejs runtime installed.
Btw, js-ipfs is already pre-bundled (just look into `node_modules/ipfs/dist`)

diasdavid:
Closing ref: https://github.com/ipfs/js-ipfs/issues/998

PR #1025, opened 2017-09-21
fix: Browser not adding additional transports

mkg20001:


dryajov:
I guess this makes sense if we want to allow consumers to specify additional transports and discovery mechanisms, I'll differ to others to approve it, but - LGTM.

mkg20001:
@diasdavid Fixed ci

diasdavid:
thanks @mkg20001 :)

PR #1027, opened 2017-09-22
docs: fix js dependencies in webpack example

dfguo:
example won't run unless these modules are installed.


examples/browser-webpack/package.json
```diff
@@ -11,13 +11,17 @@
     "babel-core": "^6.24.1",
     "babel-loader": "^7.0.0",
     "babel-preset-react": "^6.24.1",
+    "babel-preset-stage-0": "^6.24.1",
```
dryajov: Wha does the `babel-preset-stage-0` do and why do we need it here?

examples/browser-webpack/package.json
```diff
@@ -11,13 +11,17 @@
...
     "json-loader": "^0.5.4",
-    "react": "^15.5.4",
     "react-hot-loader": "^1.3.1",
     "webpack": "^2.5.1",
     "webpack-dev-server": "^2.4.5"
   },
   "dependencies": {
+    "ipfs": "^0.26.0",
```
dryajov: We still want to use IPFS from this repo -  it needs to be something like:

```
"dependencies": {
   "ipfs": "../../../src/core",
   ....
}
```

mkg20001: This one should be correct
```
"dependencies": {
   "ipfs": "file:../../",
}
```

examples/browser-webpack/package.json
dfguo: I'm not sure why it was included in the `.babelrc` https://github.com/ipfs/js-ipfs/blob/master/examples/browser-webpack/.babelrc

examples/browser-webpack/package.json
diasdavid: Oh, nice! Much better

examples/browser-webpack/src/components/app.js
```diff
@@ -2,8 +2,8 @@
 
 const React = require('react')
 const Buffer = require('safe-buffer').Buffer
-const IPFS = require('../../../../src/core') // replace this by line below
-// const IPFS = require('ipfs')
+// const IPFS = require('../../../../src/core') // replace this by line below
```
diasdavid: You can remove this line now.

diasdavid:
@dryajov can you investigate and test if these changes are necessary? I've already run the examples just fine.

dryajov:
looking into it!

PR #994, opened 2017-09-05
Progress bar for cli

bmordan:
**go-ipfs**

**js-ipfs**

This is the progress bar from [https://github.com/visionmedia/node-progress](https://github.com/visionmedia/node-progress) which is being called in `js-ipfs-unixfs-engine` in the [builder](https://github.com/ipfs/js-ipfs-unixfs-engine/blob/master/src/builder/builder.js#L96).

I have also done a little house keeping in the `src/files/add` as the callbacks were getting a little unwieldy.

PR #1036, opened 2017-10-10
 Feat/cli progress bar

dryajov:
this includes changes from #994


dryajov:
This PR changes the behavior of `files.add` considerably as it switches to a chunked response (stream of objects) that reports progress. One issue with this, is that the response is streamed and it's not possible to report an error with a standard 500 code, since it's already been initiated with a 200 code. My current solution is to report an error as part of the stream of objects and catch it in the `ProgressStream` through stream, if an error object is detected (`{Message: '', Code: #}`) the stream emits the error, which should propagate correctly. 

I'm not sure if this is the best approach and would be curious to know how Go handles it. Tagging @whyrusleeping and @Kubuxu for possible input.

whyrusleeping:
In go-ipfs we use an http trailer for errors

dryajov:
@whyrusleeping ah! Didn't know about those - but makes sense! Thanks!

dryajov:
Coverage has gone down due to https://github.com/ipfs/js-ipfs/pull/1036/files#diff-dcaf16a1df33e0e8f40195e527b9294dR244, but I don't see any tests that exercise that functionality directly. @diasdavid any ideas of how to tests this so that coverage goes up?

src/core/components/files.js
```diff
@@ -65,6 +65,14 @@ module.exports = function files (self) {
...
+      let prog = options.progress || (() => {})
+      const progress = (bytes) => {
+        total += bytes
+        prog(total)
+      }
+
+      options.progress = progress
       pull(
         pull.values(normalizeContent(data)),
         importer(self._ipldResolver, options),
```
dryajov: @diasdavid @VictorBjelkholm I noticed that there are two entry points for importer, here and above on line 26. The one above, gets called from the HTTP API, this one is called from the regular callback flow, they seem to be be doing exactly the same thing, perhaps we can replace this with a call to `createAddPullStream` and consolidate the progress accumulator logic there?

package.json
```diff
@@ -74,7 +74,7 @@
     "expose-loader": "^0.7.3",
     "form-data": "^2.3.1",
     "gulp": "^3.9.1",
-    "interface-ipfs-core": "~0.31.19",
+    "interface-ipfs-core": "^0.32.1",
```
diasdavid: @dryajov do not change ~ to ^, we use ~ until it gets to 1.0.0

diasdavid:
@dryajov I've tested this locally and the progress bar is not perfect. You can see it showing but it reaches quickly 100% and disappears and then it takes 5 or more seconds to actually get the hash of the file printed.

I'm going to go ahead and merge as this has an improvement, but it should not be considered complete.

PR #1043, opened 2017-10-20
feat: Unlink files locally

AquiGorka:
This change adds the method `unlink` to the `files` component.

I have read the discussiones ([here](https://github.com/ipfs/faq/issues/9) & [here](https://discuss.ipfs.io/t/can-i-delete-my-content-from-the-network/301)) and understand files are not deleted from the network (if another peer has pinned them, then the files can still be distributed).

We are working on a web project where we add files to the network and we wanted to remove the files to clear some space locally. With this method the files get `unlinked` locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "execa": "^0.8.0",
     "expose-loader": "^0.7.4",
     "form-data": "^2.3.1",
-    "gulp": "^3.9.1",
     "hat": "0.0.3",
     "interface-ipfs-core": "~0.36.7",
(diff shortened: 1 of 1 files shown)


src/core/components/files.js
```diff
@@ -120,6 +120,23 @@ module.exports = function files (self) {
     getPull: promisify((ipfsPath, callback) => {
       callback(null, exporter(ipfsPath, self._ipldResolver))
     })
```
chadoh: Looks like you forgot a comma at the end of this function. I think that's why the build failed.

PR #1045, opened 2017-10-22
feat: pin API

diasdavid:


src/core/components/pin-set.js
```diff
@@ -0,0 +1,277 @@
+'use strict'
+
+const multihashes = require('multihashes')
+const CID = require('cids')
+const protobuf = require('protocol-buffers')
```
diasdavid: should use the `protons` package instead

diasdavid:
@AdamStone stunning work with owning the Pin API endeavour on specs, js-ipfs-api and js-ipfs 👏🏽👏🏽👏🏽👏🏽👏🏽👏🏽

I went ahead and merged and released interface-ipfs-core + js-ipfs-api 🌟

As for merging this PR, since it is such a large code addition, I want to review it thoroughly one more time and I also want to invite other js-ipfs contributors to give their feedback. 

Let's shoot to merge this week :)

AdamStone:
@diasdavid Thanks very much, I didn't expect it was that close to ready for merge though. Were you able to resolve the issue with the pin datastore incompatibility with the go repo? As I left it, the JS implementation would save its pins under a different key than the go implementation.

Having finished some high priority things for work, I think I should have some free time again for another week or two, but after that I'll be full time taking care of a newborn.

test/cli/pin.js
```diff
@@ -0,0 +1,100 @@
...
+  it('ls (all)', () => {
+    return ipfs('pin ls').then((out) => {
+      expect(out).to.include(`${keys.leaf} direct\n`)
+      expect(out).to.include(`${keys.root} recursive\n`)
+      expect(out).to.include(`${keys.branch} indirect\n`)
+      expect(out).to.include(`${keys.subLeaf} indirect\n`)
+    })
+  })
+
+//  it('ls (quiet)', () => {
```
dryajov: is this still needed?

src/core/components/pin.js
```diff
@@ -0,0 +1,529 @@
...
+  const pinDataStoreKey = new Key('/local/pins/js')
+
+  const repo = self._repo
+  const dag = self.dag
+
+  function normalizeHashes (hashes, callback) {
+    // try to accept a variety of hash options including
+    // multihash Buffers, base58 strings, and ipfs path
+    // strings, either individually or as an array
+    if (Buffer.isBuffer(hashes) || typeof hashes.forEach !== 'function') {
```
dryajov: maybe use `Array.isArray()` here?

src/core/components/pin.js
```diff
@@ -0,0 +1,529 @@
...
+        } catch (err) { return cb(err) }
+
+        normalized.hashes.push(multihash)
+        cb()
+      }
+    }
+    each(hashes, (hash, cb) => {
+      if (typeof hash === 'string') {
+        // example: '/ipfs/QmRootHash/links/by/name'
+        const matched = hash.match(/^(?:\/ipfs\/)?([^/]+(?:\/[^/]+)*)\/?$/)
```
dryajov: would be nice to have utils around this :)

src/core/components/pin.js
```diff
@@ -0,0 +1,529 @@
...
+      }
+    }, (err) => {
+      if (err) { return callback(err) }
+      return callback(null, normalized.hashes)
+    })
+  }
+
+  function getRecursive (multihash, callback) {
+    // gets flat array of all DAGNodes in tree given by multihash
+    // (should this be part of dag.js API? it was in ipfs-merkle-dag)
```
dryajov: I've seen several of this questions, did we get this addressed - @diasdavid?

AdamStone:
Thanks for the code review, I'm just checking in for now to mention that my son was born about a week early, so it will be some time before I can follow up with this, probably at least a few weeks.

dryajov:
@AdamStone congrats on your newborn! Hope you're enjoying the moment - it flies by fast 😜

diasdavid:
@AdamStone just read your last message. Congratulations!!! 🎉🌟👶🌟🎉 I hope everything is great with the 3 of you :)

src/core/components/key-set.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
```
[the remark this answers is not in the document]
AdamStone: I did it this way to work around the problem mentioned in the comment
```
  // Buffers with identical data are still different objects, so
  // they need to be cast to strings to prevent duplicates in Sets
```
in order to have set-like storage for multihash buffers but indexed by hash string. But since it's straightforward to convert the hash string back to the buffer object, maybe it would be better to just use a Set to store the strings and use `multihashes.fromB58String()` where the object is needed.

src/core/components/pin.js
AdamStone: Hard to remember now, but I think I had used that originally and changed it after someone asked to avoid using it... Or maybe it had been `instanceof Array`. Another option is `hashes.constructor === Array`. I'm fine with any of these. Currently in the project I see three examples of `instanceof Array` and three of `Array.isArray()`.

test/cli/pin.js
AdamStone: I wasn't sure about this one. It's part of go-ipfs but wasn't mentioned in the js-ipfs specs, and I had some trouble getting it to work.

src/core/components/pin.js
AdamStone: I actually wonder if I should be doing this at all here, it seems like there should already be some built-in way to follow paths like this using the existing interface? All this is just for the sake of getting the cid of the thing at the end of the path. I could use `files.get` with the full path but I'm not sure how to get the cid from the result. @diasdavid any suggestion?

AdamStone:
@diasdavid @dryajov Thanks, we are doing well and getting used to our new life. I found some time to look into this and figured out why my saving and loading of the pin set wasn't working with the go test repo. I have it working now locally, but only if I comment out [this assert](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/dag-link/index.js#L10) to enable DAG links with size 0 -- it seems the go repo allows size 0 links and the pin set implementation includes them. Not sure what to do about that. The hashes I see for the same sets of pins pinned through the js interface are also different, so there must still be some discrepancy in the structure of the pin sets, but it doesn't seem to affect the end result.

AdamStone:
It seems like I can't push to this PR, so I pushed these changes to my [fork](https://github.com/AdamStone/js-ipfs/commit/c47f2af771aa1fbe808807ce17d98f05e8fd5c6f).

diasdavid:
@AdamStone just invited you to the JS team :) https://github.com/orgs/ipfs/teams/javascript-team/members

diasdavid:
@AdamStone note that there are merge conflicts due to recent changes in js-ipfs master, it might be wise to resolve them now so that you build on the current structure and avoid a big hairy rebase at the end.

AdamStone:
@diasdavid Thanks, I probably got dropped before from inactivity. Pushed an update. There were a couple of inline questions above about utility functions, any thoughts on those? Or the matter of zero-size links?

diasdavid:
@AdamStone re: zero size links - The reality is that the size value on a link can't really be trusted since there is no way to create a proof that the size is the real one other than checking manually. In JS land at least we make the obvious claim that if there is a link, at least it has to be bigger than 0 but it seems that Go didn't and ignores the size completely for Pinsets. Given that and that we want to be feature by feature (and sometimes bug by bug), we can remove that condition from ipld/dag-pb. Wanna submit that PR?

AdamStone:
I ended up moving getRecursive to the dag module because it just seems more sensible to have it there than buried in the pin code (since all it does is recursively traverse a dag node's links). And I moved normalizeHashes to utils.js for the same reason.

AdamStone:
I merged the latest master, updated the ipfs-api dependency to the latest version, but tests are still failing on "a link requires a size." The latest ipfs-api version requires an ipld-dag-pb version that should include the change merged above to allow empty links, so I'm not sure why the tests are still seeing this assert.

diasdavid:
@AdamStone Just ran a fresh npm install (note, it is important to remote the package-lock.json) and got 

```
» npm ls ipld-dag-pb
ipfs@0.27.5 /Users/imp/code/js-ipfs
├─┬ interface-ipfs-core@0.36.16
│ └── ipld-dag-pb@0.11.4
├─┬ ipfs-api@17.2.5
│ └── ipld-dag-pb@0.11.4  deduped
├─┬ ipfs-unixfs-engine@0.24.2
│ └── ipld-dag-pb@0.11.4  deduped
└─┬ ipld-resolver@0.14.1
  └── ipld-dag-pb@0.11.4  deduped
```

Which looks good. Running the tests also gave me:

```
      pin
        /pin/add
          ✓ pins object recursively by default (39ms)
        /pin/add (direct)
          ✓ pins object directly if specified
        /pin/ls (with path)
          ✓ finds specified pinned object
        /pin/ls (without path or type)
          ✓ finds all pinned objects
        /pin/rm (direct)
          ✓ unpins only directly pinned objects if specified
        /pin/rm
          ✓ unpins recursively by default
```

Perhaps you just didn't _fresh install_ your node modules? I've this alias on my zshrc to avoid any confusion: `alias fresh-modules="rm package-lock.json && rm -rf node_modules && npm install"` :)

src/cli/commands/pin/ls.js
```diff
@@ -0,0 +1,43 @@
...
+      describe: 'Write just hashes of objects.'
+    }
+  },
+
+  handler: (argv) => {
+    const paths = argv.path && argv.path.split(' ')
+    const type = argv.type
+    const quiet = argv.quiet
+    argv.ipfs.pin.ls(paths, { type }, (err, results) => {
+      if (err) { throw err }
```
diasdavid: Will this print a useful error message to the user?

src/cli/commands/pin/ls.js
```diff
@@ -0,0 +1,43 @@
...
+      default: false,
+      describe: 'Write just hashes of objects.'
+    }
+  },
+
+  handler: (argv) => {
+    const paths = argv.path && argv.path.split(' ')
+    const type = argv.type
+    const quiet = argv.quiet
+    argv.ipfs.pin.ls(paths, { type }, (err, results) => {
```
diasdavid: Mind being explicit and using `{ type: type }`?

src/cli/commands/pin/rm.js
```diff
@@ -0,0 +1,27 @@
...
+      default: true,
+      describe: 'Recursively unpin the objects linked to by the specified object(s).'
+    }
+  },
+
+  handler: (argv) => {
+    const paths = argv['ipfs-path'].split(' ')
+    const recursive = argv.recursive
+    argv.ipfs.pin.rm(paths, { recursive }, (err, results) => {
+      if (err) { throw err }
```
diasdavid: Same as above. Let's make sure that these error messages are printed in a way that it provides useful information to the user and it doesn't look like the whole program crashed

src/core/components/dag.js
```diff
@@ -73,6 +75,30 @@ module.exports = function dag (self) {
         self._ipldResolver.treeStream(cid, path, options),
         pull.collect(callback)
       )
+    }),
+
+    getRecursive: promisify((multihash, callback) => {
```
diasdavid: This getRecursive is only working for `ipld-dag-pb` right now. Either make it an internal method by prefixing with a `_` and add a comment saying just that or move it to https://github.com/ipld/js-ipld-resolver and leverage the `.tree` and `.isLink` API calls to make it agnostic to any kind of IPLD Format

src/core/components/pin-set.js
```diff
@@ -0,0 +1,283 @@
...
+          return fnv1a(data.toString('binary'))
+        }
+        // items will be distributed among `defaultFanout` bins
+        for (let i = 0; i < items.length; i++) {
+          let h = hashFn(seed, items[i].key) % defaultFanout
+          hashed[h] = hashed[h] || []
+          hashed[h].push(items[i])
+        }
+        const storeItemsCb = (err, child) => {
+          if (callback.called) { return }
```
diasdavid: You can use http://npmjs.com/once for this

src/core/components/pin-set.js
```diff
@@ -0,0 +1,283 @@
...
+              // all finished
+              DAGNode.create(rootData, rootLinks, (err, rootNode) => {
+                if (err) { return callback(err) }
+                return callback(null, rootNode)
+              })
+            }
+          })
+        }
+        _subcalls += Object.keys(hashed).length
+        for (let h in hashed) {
```
diasdavid: Mind replacing by a hashed.forEach instead?

src/core/components/pin-set.js
```diff
@@ -0,0 +1,283 @@
...
+        }
+        pinSet.walkItems(res.value, walkerFn, logInternalKey, (err) => {
+          if (err) { return callback(err) }
+          return callback(null, keys)
+        })
+      })
+    },
+
+    walkItems: (node, walkerFn, logInternalKey, callback) => {
+      // callback (err)
```
diasdavid: If not used, delete

src/core/components/pin.js
```diff
@@ -0,0 +1,431 @@
...
+          update: (hash) => {
+            result.payload.push({hash})
+            if (result.payload.length === mhs.length) {
+              pin.flush((err, root) => {
+                if (err) { return callback(err) }
+                return callback(null, result.payload)
+              })
+            }
+          }
+        }
```
diasdavid: This function seems a bit off. It is an async call that sets up `result`, which is going to be used below and when it gets called (result.update), it will be the call that calls the top-level callback.

If the section below needs that normalizeHashes to finish first, these steps should be wrapped into an `async.series` for better legibility and avoiding future errors (right now it is prone to race conditions)

src/core/components/pin.js
```diff
@@ -0,0 +1,431 @@
...
+              directPins.add(key)
+              result.update(key)
+            })
+          }
+        })
+      })
+    }),
+
+    rm: promisify((hashes, options, callback) => {
+      // callback (err)
```
diasdavid: remove if not used

src/core/components/pin.js
```diff
@@ -0,0 +1,431 @@
...
+          update: (hash) => {
+            result.payload.push({hash})
+            if (result.payload.length === mhs.length) {
+              pin.flush((err, root) => {
+                if (err) { return callback(err) }
+                return callback(null, result.payload)
+              })
+            }
+          }
+        }
```
diasdavid: Same as the function above. This whole method needs to be simplified.

src/core/components/pin.js
```diff
@@ -0,0 +1,431 @@
...
+          const result = {
+            // async result queue
+            payload: [],
+            update: (item) => {
+              result.payload.push(item)
+              if (result.payload.length === mhs.length) {
+                return callback(null, result.payload)
+              }
+            }
+          }
```
diasdavid: Same as above.

src/core/components/pin.js
```diff
@@ -0,0 +1,431 @@
...
+            return callback(null, result)
+          })
+        } else {
+          return callback(null, result)
+        }
+      }
+    }),
+
+    isPinned: (multihash, callback) => {
+      // callback (err, pinned, reason)
```
diasdavid: remove if not used

diasdavid: Also, it is preferable for the callback to be use `function (err, result)` to follow the same pattern as everything else

src/core/utils.js
```diff
@@ -1,3 +1,78 @@
...
+    }
+  }
+  const split = matched[1].split('/')
+  return {
+    root: split[0],
+    links: split.slice(1, split.length)
+  }
+}
+
+exports.normalizeHashes = function (ipfs, hashes, callback) {
```
diasdavid: Can you add tests specifically to this function?

src/core/utils.js
```diff
@@ -1,3 +1,78 @@
 'use strict'
 
+const multihashes = require('multihashes')
+const each = require('async/each')
+
 exports.OFFLINE_ERROR = 'This command must be run in online mode. Try running \'ipfs daemon\' first.'
+
+const splitIpfsPath = exports.splitIpfsPath = function (pathString) {
```
diasdavid: Can you add tests specifically to this function?

src/cli/commands/pin/ls.js
AdamStone: Sure, no problem

src/cli/commands/pin/ls.js
AdamStone: I suppose probably not. I just followed the same pattern as the other CLI modules.

src/cli/commands/pin/rm.js
AdamStone: Since it looks like all the CLI modules just throw the error, would that be better as a separate PR to replace all of them with some consistent response?

src/core/components/dag.js
AdamStone: I'll see if I can move it

src/core/components/pin-set.js
AdamStone: Oops, looks like I had onced the callbacks but missed removing some of these.

src/core/components/pin-set.js
AdamStone: Will do

src/core/components/pin-set.js
AdamStone: This is used within `loadSet` above.

src/core/components/pin.js
AdamStone: Agree it's convoluted, I'll work out a cleaner approach with async.

src/core/components/dag.js
AdamStone: @diasdavid Should the existing treeStream method already do what I'm trying to do here, if the recursive: true option is passed? I tried testing it with a CID for [this directory](https://ipfs.io/ipfs/QmUhUuiTKkkK8J6JZ9zmj8iNHPuNfGYcszgRumzhHBxEEU) but I'm getting an error in the recursive part. Specifically this [`new CID(...)`](https://github.com/ipld/js-ipld-resolver/blob/master/src/index.js#L314) errors because `paths` includes separate objects for link name, size and hash that all have `p.link` (so it tries to recurse all of them, when only the hash ones should be recursed). Is it a bug that the hashes array is treating these as separate links? Or should the recursion check specifically for the "hash" links?

```
[
   {
      "path":"Links/0/Name",
      "link":{
         "/":"about"
      }
   },
   {
      "path":"Links/0/Tsize",
      "link":{
         "/":1688
      }
   },
   {
      "path":"Links/0/Hash",
      "link":{
         "/":"QmZTR5bcpQD7cFgTorqxZDYaew1Wqgfbd2ud9QqGPAkK2V"
      }
   }
]
```

src/cli/commands/pin/rm.js
diasdavid: Sounds good 👍

src/core/components/dag.js
diasdavid: that definitely looks like a bug as it _brings more info than what it promises_. The method was created just to return all the paths. That `link` property is just a left over.

AdamStone: @diasdavid Based on [these](https://github.com/ipld/interface-ipld-format/pull/5) [discussions](https://github.com/ipld/js-ipld-dag-pb/pull/29) it looks like the presence of separate "paths" for name, size and hash for each link is correct/necessary, and the bug is that the isLink method (formerly isCID) was meant to identify specifically the "hash" paths but is actually more permissive. The tests pass because the [non-valid CID case](https://github.com/ipld/js-ipld-dag-pb/blob/master/test/resolver.spec.js#L230) is checking an [empty value](https://github.com/ipld/js-ipld-dag-pb/blob/master/test/resolver.spec.js#L25), which happens to return `false` correctly, whereas other non-empty non-CID values would not. I'll make another PR to address that and update the test.

diasdavid:
@AdamStone any updates here?

@hacdias would you like to help to get this PR merged?

AdamStone:
Started going through this but family situation became very busy again. Thinking I can finish this weekend.

hacdias:
@diasdavid before finishing this, I think it would be worth it to take a look at this issue on `go-ipfs`: https://github.com/ipfs/go-ipfs/issues/4586

diasdavid:
@hacdias valid point but that is going to take a while (decision + go-ipfs release) to get into place. Meanwhile the Pin API is a long time coming and we need to get it in.

@hacdias @AdamStone is this something you have the availability to continue working on?

hacdias:
@diasdavid @AdamStone what is missing here?

JonKrone:
I spent some time this afternoon looking and it seems pretty close to me. I was able to get it to pass the core, cli, and interface-core test suites essentially as-is. Reviews have requested some refactoring in [core/components/pin.js](https://github.com/ipfs/js-ipfs/pull/1045/files#diff-d96275251fef125ec785643a9facfab4) and a few more tests elsewhere.

This feature and those reviews touch on some new things for me; if y'all and @AdamStone are okay with it, I would be happy to try to move this towards master.

diasdavid:
@JonKrone that's great news! Thank you for the effort and pushing things forward.

Could you start with a PR to this PR for that first pass and then we will continue collab all in one PR?

AdamStone:
Hi everyone, since I am having trouble finding time to focus on this recently I thought it best to push what I have done locally to respond to the requested changes in the last review. I think most of them have been addressed. I had trouble getting the Dag getRecursive behavior moved, so I left it as an internal method in the Dag module to be moved later if someone else wants to try. I can't get the whole test suite to pass locally due to failures in files-sharding, but I'm not sure if that has anything to do with these changes. If one of you can take over getting this to the finish line that would be great, likewise with the pin label proposal.

JonKrone:
@AdamStone Awesome, thanks for pushing what you've done! It does resolve most of the comments. I'm going to add a few failure test cases to push the edges and some interop tests with go-ipfs to verify we're producing the same keysets. Thanks!

I hope you and your youngster are happy and healthy :)

JonKrone:
@daviddias I have a couple questions about the interop tests.

I see two ways to compare js/go daemon behavior:
  - spawn a daemon, pin some stuff, then kill that daemon and spawn the other daemon using the same repoPath and verify that they are compatible and produce the same output.
  - spawn independent go and js daemons, run the same commands on each, then compare their output via `ipfs pin ls` or similar.

Is it correct to use both? Spawning on top of each other verifies that they're flushing and loading correctly on the datastore while other commands can then be verified on their output after operations.

Second: the easiest way to check behavior is to compare the output of `ipfs pin ls`. Is this sufficient? Do we also need to inspect blocks using `ipfs object` or `ipfs block`?

diasdavid:
@JonKrone I'm seeing CI green. Let me know if it is time for review :)

JonKrone:
@diasdavid The only core code that remains are tests for `pin-set`. It also needs a final cleanup 
sweep and has a TODO or two but if you're willing to bear those, it's ready for you.

I had hoped to get those `pin-set` tests in this afternoon but have not found any time today and am instead aiming for tomorrow afternoon.

package.json
```diff
@@ -100,13 +100,15 @@
...
     "filesize": "^3.6.0",
+    "fnv1a": "^1.0.1",
     "fsm-event": "^2.1.0",
     "get-folder-size": "^1.0.1",
     "glob": "^7.1.2",
     "hapi": "^16.6.2",
     "hapi-set-header": "^1.0.2",
     "hoek": "^5.0.3",
     "human-to-milliseconds": "^1.0.0",
+    "interface-datastore": "^0.4.1",
```
diasdavid: @JonKrone Is this dependency necessary? If yes, for what?

src/core/components/dag.js
```diff
@@ -73,6 +75,31 @@ module.exports = function dag (self) {
...
+        links.forEach(link => {
+          self.dag._getRecursive(link.multihash, (err, subNodes) => {
+            if (err) { return callback(err) }
+            nodes.push(subNodes)
+            if (nodes.length === links.length + 1) {
+              return callback(null, _.flattenDeep(nodes))
+            }
+          })
+        })
+      })
```
diasdavid: @vmx can we get this API call on http://github.com/ipld/js-ipld?

src/core/components/files.js
```diff
@@ -72,6 +74,22 @@ function normalizeContent (content) {
...
+  const isTargetFile = !file.path.includes('/')
+  const shouldPin = pin && isTargetFile && !opts.onlyHash
+  if (shouldPin) {
+    self.pin.add(file.hash, (err) => {
+      cb(err, file)
+    })
+  } else {
+    cb(null, file)
+  }
+}
```
diasdavid: If I'm reading correctly, you are potentially calling `pin add --recursive` for every single file and directory that gets added in a single `ipfs files add`. This is excessive and will drag the process down.

src/core/components/init.js
```diff
@@ -84,16 +84,18 @@ module.exports = function init (self) {
...
 
-        self.log('adding assets')
         const tasks = [
           // add empty unixfs dir object (go-ipfs assumes this exists)
           (cb) => self.object.new('unixfs-dir', cb)
         ]
 
         if (typeof addDefaultAssets === 'function') {
+          // addDefaultAssets is undefined on browsers.
+          // See package.json browser config
```
diasdavid: Yeah, we don't add the default assets (initial docs) in the browser to reduce the js payload.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,269 @@
...
+  buf.writeUInt32LE(seed, 0)
+  const data = Buffer.concat([
+    buf, Buffer.from(toB58String(key))
+  ])
+  return fnv1a(data.toString('binary'))
+}
+
+exports = module.exports = function (dag) {
+  const pinSet = {
+    // should this be part of `object` API?
```
diasdavid: What's your suggestion here?

src/core/components/pin-set.js
```diff
@@ -0,0 +1,269 @@
...
+        const bs58Link = toB58String(link.multihash)
+        if (bs58Link === childhash) {
+          return callback(null, true)
+        }
+
+        // don't check the same links twice
+        if (bs58Link in _seen) { return }
+        _seen[bs58Link] = true
+
+        dag.get(new CID(link.multihash), (err, res) => {
```
diasdavid: Do you need to cast it to a CID yourself? I don't think so.

JonKrone: It turns out that some of our links' multihashes are buffers and `dag.get` does not know how to handle them. I've added a cast from buffer to CID in core/components/dag.js. Is that why you were thinking it should work without the cast?

https://github.com/ipfs/js-ipfs/pull/1045/files#diff-cb106b8602ebb9750632d0d9ce3c1848R38

src/core/components/pin-set.js
JonKrone: This was @AdamStone's suggestion so I can't speak to his thoughts on it. It does feel like this method is not in the domain of pinset but I think changing the impl to use an existing tree traversal like `ipld.treestream` might suffice. If we do find it valuable enough to separate, I'm not sure whether object or dag is the proper place.

I attempted an implementation with `ipld.treeStream` but surprisingly ran into a heap OOM during a light pin-set stress test below. Should we look further into that? I can provide a repro test case.

https://github.com/ipfs/js-ipfs/pull/1045/files#diff-fb2d6500d5a91459f37fa2dbc49d3908R101

diasdavid:
I'm a bit low on availability to fully review this PR. @achingbrain, @vmx, could one of you two (or both :)) help here?

src/core/components/pin-set.js
```diff
@@ -0,0 +1,271 @@
...
+      }
+      logInternalKey(emptyKey)
+
+      if (items.length <= maxItems) {
+        // the items will fit in a single root node
+        const itemLinks = []
+        const itemData = []
+        const indices = []
+
+        for (let i = 0; i < items.length; i++) {
```
vmx: I'm not sure about it, but wouldn't it be simpler to have a single array with the items being objects containing `link` and `data` properties? You could then sort them and in the and push them to the corresponding arrays.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,271 @@
...
+          const bs58Link = toB58String(link.multihash)
+          if (bs58Link in seen) return
+          if (bs58Link === childhash) {
+            return cb(null, true)
+          }
+
+          seen[bs58Link] = true
+
+          dag.get(link.multihash, (err, res) => {
+            if (err) return cb(err)
```
vmx: Nitpick: for consistency it would be cool to use brackets. This is not a blocker, but an easy thing to change if you do change anyway.

src/core/components/pin.js
```diff
@@ -0,0 +1,438 @@
...
+                  `${key} is pinned indirectly under ${reason}`
+                ))
+            }
+          })
+        }), (err, results) => {
+          if (err) { return callback(err) }
+
+          // update the pin sets in memory
+          results.forEach(key => {
+            if (recursive) {
```
vmx: Could be simplified to

```
if (recursive && recursivePins.has(key)) {
    recursivePins.delete(key)
} else {
    directPins.delete(key)
}
```
Though I'm not sure if that's easier to read/understand, so I leave it up to you if you if you want to leave it as it currently is or not.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,271 @@
...
+        const indices = []
+
+        for (let i = 0; i < items.length; i++) {
+          itemLinks.push(new DAGLink('', 1, items[i].key))
+          itemData.push(items[i].data || Buffer.alloc(0))
+          indices.push(i)
+        }
+        indices.sort((a, b) => {
+          const x = Buffer.compare(itemLinks[a].multihash, itemLinks[b].multihash)
+          if (x) { return x }
```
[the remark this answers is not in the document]
VictorBjelkholm: Do we need to sort the links? Imagine directories that have over a million links, can we reasonably sort them or should we ignore sorting like go-ipfs does?

test/core/pin.js
```diff
@@ -0,0 +1,328 @@
...
+      return pin.add(pins.root)
+    })
+
+    it('loads', function () {
+      pin.clear()
+      return pin.ls()
+        .then(ls => expect(ls.length).to.eql(0))
+        .then(() => pin.load())
+        .then(() => pin.ls())
+        .then(ls => expect(ls.length).to.be.gt(0))
```
achingbrain: Is it not possible to assert on a known value in this test?  Same thing for the `it flushes` test.

src/core/components/pin.js
```diff
@@ -0,0 +1,438 @@
...
+const once = require('once')
+
+function toB58String (hash) {
+  return new CID(hash).toBaseEncodedString()
+}
+
+module.exports = function pin (self) {
+  let directPins = new Set()
+  let recursivePins = new Set()
+  let internalPins = new Set()
```
achingbrain: What's the purpose of this Set?  From what I can see it's an in-memory cache of any hashes of pinned DAGNodes that we load or otherwise encounter but I can't see any reference to `internal` as a type in the [pin spec](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/PIN.md).

src/core/components/pin.js
```diff
@@ -0,0 +1,438 @@
...
+      const newInternalPins = new Set()
+      const logInternalKey = (mh) => newInternalPins.add(toB58String(mh))
+      const handle = {
+        put: (k, v, cb) => {
+          handle[k] = v
+          cb()
+        }
+      }
+      waterfall([
+        (cb) => repo.closed ? repo.datastore.open(cb) : cb(null, null), // hack for CLI tests
```
achingbrain: Has the case where no `~/.jsipfs` directory exists been handled?  I can't see anything obvious in the PR code. 
 `repo.datastore` will not exist until the user runs `jsipfs init`.

src/core/components/pin.js
```diff
@@ -0,0 +1,438 @@
...
+        (_, cb) => repo.datastore.has(pinDataStoreKey, cb),
```
achingbrain: [`interface-datastore`](https://github.com/ipfs/interface-datastore#api) has no `.has` method.  It might just be missing from the docs though.

src/core/components/pin.js
JonKrone: Oh good catch. If there is no ~/.jsipfs, it's impossible to do any datastore based operations such as `jsipfs object new`. We should have a common pattern for this, like `assertRepoExists` where appropriate. I think this falls under the umbrella of: https://github.com/ipfs/js-ipfs/issues/1180.

Code written just for tests is always something to try to avoid, though, and I hadn't looked at this line again so I will now.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,271 @@
...
+        childhash = toB58String(childhash)
+      }
+
+      function searchChildren (root, cb, _links, _checked) {
+        if (!root.links.length && _links === _checked) {
+          // all nodes have been checked
+          return cb(null, false)
+        }
+
+        some(root.links, (link, cb) => {
```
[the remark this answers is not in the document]
JonKrone: My understanding of `some` is that it does run tasks concurrently: https://caolan.github.io/async/docs.html#some. There is a `someSeries` function which does have sequential execution, though.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,271 @@
...
+  ])
+  return fnv1a(data.toString('binary'))
+}
+
+exports = module.exports = function (dag) {
+  const pinSet = {
+    // should this be part of `object` API?
+    hasChild: (root, childhash, callback) => {
+      let seen = {}
+      if (typeof childhash === 'object') {
```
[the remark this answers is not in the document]
JonKrone: Yep! Thanks

src/core/components/pin-set.js
JonKrone: Happy to; I'll look for other cases

src/core/components/pin-set.js
JonKrone: Yes I initially implemented that but reverted it in favor of keeping the original work. Because you've brought it up, though, I'll refactor it. I have seen some other patterns like this in the codebase so was not sure what others thought of it. I'll look into the sorting question first and then refactor this.

src/core/components/pin-set.js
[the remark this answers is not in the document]
JonKrone: Thanks guys, I will check go's handling and run tests then undo the sorting if we can get away with it :). Pin's performance isn't stellar so removing this if we can would be good.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,271 @@
...
+        childhash = toB58String(childhash)
+      }
+
+      function searchChildren (root, cb, _links, _checked) {
```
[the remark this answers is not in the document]
JonKrone: `_links` is used to carry information passed only on recursive calls. There is room to rename this. On it!

Aside: in js, a `_` prefix is typically used to indicate something like 'private'. To indicate an unused parameter, a plain `_` w/o anything before or behind it the most common pattern I've seen. I agree that we can easily avoid the underscore and potential confusion here.

src/core/components/files.js
```diff
@@ -72,6 +74,21 @@ function normalizeContent (content) {
   })
 }
 
+function pinFile (self, opts, file, cb) {
+  // Pin a file if it is the root dir of a recursive add or the single file
+  // of a direct add.
+  const pin = 'pin' in opts ? opts.pin : true
+  const isRootDir = !file.path.includes('/')
```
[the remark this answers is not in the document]
JonKrone: I believe so, this path is from `ipfs-unixfs-engine`'s export, which passes paths w/ a unix separator: https://github.com/ipfs/js-ipfs-unixfs-engine/blob/master/test/exporter.js#L302

test/core/pin.js
JonKrone: Yeah, I wasn't sure! My concern with hardcoded values is that it makes the tests fragile to changes in the default assets added during repo init, which these pin tests pick up.

Another approach would be to start each pin test with a clean slate by removing all pins. I did not go that route because of the added test overhead but looking back it's probably worth it. What do you think? Hardcode / naive `gt` checks / or cleaning the pins?

src/core/components/pin.js
JonKrone: Nope, it is easier! On it.

src/core/components/pin-set.js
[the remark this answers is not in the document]
achingbrain: Ah yes, you're right.

src/core/components/pin.js
achingbrain: It could just be missing from the documentation 😉

test/core/pin.js
achingbrain: I think a clean state is best in general, otherwise we could end up with tests that depend on the behaviour of other tests or the order in which they're run, etc.

I guess maybe try it and see how bad the overhead is?

src/core/components/pin-set.js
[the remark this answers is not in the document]
JonKrone: So go [does sort](https://github.com/JonKrone/go-ipfs/blob/master/pin/set.go#L97) the links for the DAGNode. I don't know why it's important though, do you know what impact sorting the links of a DAGNode could have? I can ask around irc as well.

It also turns out that our sorting never ran because the indices array was always already sorted (d'oh). Even without sorting, the pin interop tests show go and js produce the same pins.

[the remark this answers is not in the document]
JonKrone: @VictorBjelkholm The links are sorted so that the order of the pins doesn't affect the hash of the `DAGNode` we create and then `put` in the datastore. So storing pins `{A, B, C}` produces the same root node as storing pins `{B, A, C}` Added a comment to log this :).

src/core/components/pin-set.js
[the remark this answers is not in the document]
JonKrone: note for the future: `typeof Buffer('') === 'object'` so we also needed a `Buffer.isBuffer` check to catch multihashes.

test/core/pin.js
JonKrone: Turns out I was mistaken and in this case we can be explicit with the values because we are cleaning pins before/during tests. I was thinking of cli tests.

src/core/components/pin.js
JonKrone: @achingbrain Good question. I've only vaguely figured it was to prevent gc as it is not used except to keep a record of some multihashes. I've looked more into it and found:

Internal pins are only updated when flushing/loading pins to/from the datastore. The sources of our pins are:

1) A hash of a totally empty DAGNode. (kindof a zero-hash)
2) Pins are stored as two dags: the dag of recursive pins and dag of direct pins
    - *The hash of those pinset roots are added*
3) When we have > 8192 recursive or direct pins, we shard it into ~256 bins.
    - *The hash of each of those bins are added*
    - If any of those bins have more than 8192 items, we repeat the sharding

This means that when we have < 8192 pins, we have either 2 or 3 multihashes in `internalPins`. When we have more then 8192 pins, we likely have ~ 250. Then it's basically static until we have about 8192 * 256 pins, where it spikes to like 256^2 internal pins.

I think it's intended to keep a reference to prevent some gc during those high-pin count cases but I don't know enough about V8 gc to say. Looking through its uses, I don't think it keeps a ref to anything that isn't referenced somehow else. It might just be a [carry-over from go's impl](https://github.com/ipfs/go-ipfs/blob/master/pin/set.go#L148).

I've tested removing `internalPins` during a 8192 * 256 pin stress test but still hit OOM w/ 2Gb ram. Do you have suggestions what might be going on?

src/core/components/pin.js
JonKrone: :) https://github.com/ipfs/interface-datastore/pull/19

src/core/components/dag.js
JonKrone: @vmx Will selectors enable something like `<roothash>/**/*`?

vmx: @JonKrone Selectors don't really exist yet.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,269 @@
...
+        const opts = { cid: new CID(rootNode.multihash) }
+        dag.put(rootNode, opts, (err, cid) => {
+          if (err) { return callback(err) }
+          logInternalKey(rootNode.multihash)
+          callback(null, rootNode)
+        })
+      })
+    },
+
+    storeItems: (pins, logInternalKey, callback, _depth = 0, _binsToFill = 0, _binsFilled = 0) => {
```
vmx: I think I've left a similar comment before, but can't find it anymore. So the underscore means optional arguments? I haven't seen this convention throughout the JS projects yet. I normally use underscore for variables not being used.

I just want to have that said, feel free to keep it as it currently is.

diasdavid:
@JonKrone can you see what's up with CI and make it pass?

src/core/components/pin-set.js
JonKrone: Those `_`'d args are used by `storeItems` as it recursively builds a dag from a list of pinned hashes and so are not quite optional but more like private or internal. I've refactored it to make it clear that these are internal pieces of state.

JonKrone:
@diasdavid I don't think it is a timeout which is causing the problem here: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs/detail/PR-1045/16/pipeline

It seems to only happen with the "large" file browser tests on linux 9.2 machines. I've tried replicating locally with files up to 20x the size of those files but have not found this error. Pinning can be heavy on memory with large files but I'm not sure if that is causing this. Any suggestions how to replicate or figure it out?

@dryajov have you seen this error in other PRs or is it introduced by pin?

The last CI run did pass except for an EADDIRINUSE on macos: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs/detail/PR-1045/17/pipeline/15

vmx:
@JonKrone The big files tests failing sounds like https://github.com/ipfs/js-ipfs-api/issues/726

src/core/components/pin.js
JonKrone: @diasdavid The above is the open question I had about V8 garbage collection. I think that `internalPins` was used to prevent collection of pins in the go implementation and was carried over here but I don't know if it does anything in js-land.

JonKrone:
@vmx yes it does seem the same. I guess the additional overhead of pinning those files is making it fail more frequently.

src/core/components/pin-set.js
JonKrone: @vmx it turns out that `ipld.treeStream` on a root node created by `pinset.storeItems` ends up running until OOM, like there is a cyclical link or something else to cause an infinite. Counter to that, though, `dag._getRecursive` does not encounter such a problem https://github.com/ipfs/js-ipfs/pull/1045/files#diff-cb106b8602ebb9750632d0d9ce3c1848R83.

I've spent some time looking for a root cause and had to dig into ipld but haven't quite pinpointed it. Is it a serious concern I should spend more time looking into?

src/core/components/pin.js
JonKrone: Friendly ping @diasdavid @achingbrain. I'd like to remove `internalPins` but want another's thought first.

JonKrone:
This code is ready for final reviews. Two areas that may not have received attention:

- pin-set tests: https://github.com/ipfs/js-ipfs/pull/1045/files#diff-fb2d6500d5a91459f37fa2dbc49d3908
- core/utils: https://github.com/ipfs/js-ipfs/pull/1045/files#diff-47300e7ecd8989b6246221de88fc9a3c

I am looking into the large file failures right now but they shouldn't block review.

// cc @vmx @achingbrain

diasdavid:
@JonKrone are you still available to take this PR to the finish line? @achingbrain, @alanshaw can you help here?

JonKrone:
Hey there @diasdavid. I'm sorry about the absence. I would still like to bring this to completion and believe it is basically there. All tests pass, including those from https://github.com/ipfs/interop/pull/19 against the current master of Go-ipfs.

I spent some time over the past couple weekends developing a solution to enable non-default hashing algorithms but it turns out the feature was bugged (https://github.com/ipfs/js-ipfs/issues/1373) so I've [disabled pinning when it's used](https://github.com/ipfs/js-ipfs/pull/1045/files#diff-8a5275c36675ad91abac348c3166d456R97).

I believe the only remaining issue is: https://github.com/ipfs/js-ipfs/pull/1045#discussion_r178477118. I'm going to go ahead and work on removing `internalPins` today/tomorrow and if we decide to keep them, we can just revert it.

JonKrone:
Hey @alanshaw. Can you give an update on what the next steps are for completing this PR? Tests are passing and I don't believe there are any open comments. I would be happy to go over this feature and changeset with you on a call if it would help 😃.

We've got a lot of tests here that cover the majority of code paths and uses and https://github.com/ipfs/interop/pull/19 provides a suite of tests to compare the pin implementations of both Go and JS to give us confidence that we're producing compatible pinsets.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,267 @@
...
+          if (err) { return cb(err) }
+
+          const cid = new CID(child._multihash)
+          dag.put(child, { cid }, (err) => {
+            if (err) { return cb(err) }
+
+            fanoutLinks[bin] = new DAGLink('', child.size, child.multihash)
+            binsFilled++
+
+            if (binsFilled === binsToFill) {
```
alanshaw: I've found that keeping track of async tasks in this way is error prone and makes the code difficult to follow (and that's been my experience looking at this just now). The `async` library was designed to alleviate these problems.

I think if this was refactored to use `async.each` or `async.map` or something similar instead of `binKeys.forEach` then the readability and robustness of this code would improve considerably.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,267 @@
...
+                return cb(null, rootNode)
+              })
+            }
+          })
+        }
+      }
+    },
+
+    loadSet: (rootNode, name, callback) => {
+      callback = once(callback)
```
alanshaw: I don't think we need a `once` here unless it's possible for `pinSet.walkItems` to call the callback twice (which it shouldn't do)

src/core/components/pin-set.js
```diff
@@ -0,0 +1,267 @@
...
+    walkItems: (node, step, callback) => {
+      callback = once(callback)
+      let pbh
+      try {
+        pbh = readHeader(node)
+      } catch (err) {
+        return callback(err)
+      }
+      let subwalkCount = 0
+      let finishedCount = 0
```
alanshaw: Again, instead of using counts to determine when async tasks have been finished I'd defer to the async library.

src/core/components/pin.js
```diff
@@ -0,0 +1,412 @@
...
+        options = null
+      }
+      callback = once(callback)
+      const recursive = options ? options.recursive : true
+
+      resolvePath(self.object, paths, (err, mhs) => {
+        if (err) { return callback(err) }
+
+        // verify that each hash can be pinned
+        parallel(mhs.map(multihash => cb => {
```
alanshaw: Super minor, but `async.map` will save you creating an function for each multihash...

src/core/components/pin.js
```diff
@@ -0,0 +1,412 @@
...
+      } else if (options && options.recursive === false) {
+        recursive = false
+      }
+      callback = once(callback)
+
+      resolvePath(self.object, paths, (err, mhs) => {
+        if (err) { return callback(err) }
+
+        // verify that each hash can be unpinned
+        series(mhs.map(multihash => cb => {
```
alanshaw: Super minor https://caolan.github.io/async/docs.html#mapSeries ?

src/core/components/pin.js
```diff
@@ -0,0 +1,412 @@
...
+        callback = options
+      }
+      if (paths && paths.type) {
+        options = paths
+        paths = null
+      }
+      if (options && options.type) {
+        type = options.type.toLowerCase()
+      }
+      callback = once(callback)
```
alanshaw: Really needed?

src/core/components/pin.js
```diff
@@ -0,0 +1,412 @@
...
+      waterfall([
+        (cb) => repo.closed ? repo.datastore.open(cb) : cb(null, null), // hack for CLI tests
+        (_, cb) => repo.datastore.has(pinDataStoreKey, cb),
+        (has, cb) => has ? cb() : cb(new Error('No pins to load')),
+        (cb) => repo.datastore.get(pinDataStoreKey, cb),
+        (mh, cb) => dag.get(new CID(mh), cb),
+        (root, cb) => handle.put('root', root.value, cb),
+        (cb) => pin.set.loadSet(handle.root, pin.types.recursive, cb),
+        (rKeys, cb) => handle.put('rKeys', rKeys, cb),
+        (cb) => pin.set.loadSet(handle.root, pin.types.direct, cb)
```
alanshaw: Do these all need to be done sequentially? Consider using `async.auto` for these so sequential and parallel tasks can be orchestrated together 😄

src/core/utils.js
```diff
@@ -1,3 +1,113 @@
...
+
+  const matched = ipfsPath.match(/^(?:\/ipfs\/)?([^/]+(?:\/[^/]+)*)\/?$/)
+  if (!matched) {
+    throw invalidPathErr
+  }
+
+  const [hash, ...links] = matched[1].split('/')
+
+  try {
+    if (CID.isCID(new CID(hash))) {
```
alanshaw: By definition a `new CID` is a CID...

src/core/components/pin.js
```diff
@@ -0,0 +1,412 @@
...
+        if (dKeys) {
+          directPins = new Set(dKeys.map(toB58String))
+          recursivePins = new Set(handle.rKeys.map(toB58String))
+        }
+        self.log('Loaded pins from the datastore')
+        return callback()
+      })
+    })
+  }
+  return pin
```
alanshaw: There's a lot of properties on this object that aren't part of the spec'd public API. If they're not for public consumption we need to indicate that by prefixing them with `_` or better still pull them out into modules/functions that can only be used internally.

src/core/utils.js
```diff
@@ -1,3 +1,113 @@
...
+ * @return {Object}            { hash: base58 string, links: [string], ?err: Error }
+ * @throws on an invalid @param ipfsPath
+ */
+function parseIpfsPath (ipfsPath) {
+  const invalidPathErr = new Error('invalid ipfs ref path')
+
+  const matched = ipfsPath.match(/^(?:\/ipfs\/)?([^/]+(?:\/[^/]+)*)\/?$/)
+  if (!matched) {
+    throw invalidPathErr
+  }
```
alanshaw: Can we please use the `is-ipfs` module for this?

src/core/components/pin-set.js
```diff
@@ -0,0 +1,267 @@
...
+        const step = link => keys.push(link.multihash)
+        pinSet.walkItems(res.value, step, err => {
+          if (err) { return callback(err) }
+          return callback(null, keys)
+        })
+      })
+    },
+
+    walkItems: (node, step, callback) => {
+      callback = once(callback)
```
alanshaw: Again, `once` needed?

src/core/components/pin.js
```diff
@@ -0,0 +1,412 @@
...
+    }),
+
+    rm: promisify((paths, options, callback) => {
+      let recursive = true
+      if (typeof options === 'function') {
+        callback = options
+      } else if (options && options.recursive === false) {
+        recursive = false
+      }
+      callback = once(callback)
```
alanshaw: Really necessary?

src/core/components/files.js
```diff
@@ -87,6 +89,19 @@ function normalizeContent (opts, content) {
...
+  const shouldPin = pin && isRootDir && !opts.onlyHash && !opts.hashAlg
```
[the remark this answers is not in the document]
JonKrone: Yeah, this is related to https://github.com/ipfs/js-ipfs/issues/1373#event-1660703561. `hashAlg` creates an incorrect DAG and requires changes to IPLD before we can support it.

src/core/components/pin-set.js
```diff
@@ -0,0 +1,267 @@
...
+  const data = Buffer.concat([
+    buf, Buffer.from(toB58String(key))
+  ])
+  return fnv1a(data.toString('binary'))
+}
+
+exports = module.exports = function (dag) {
+  const pinSet = {
+    // should this be part of `object` API?
+    hasChild: (root, childhash, callback) => {
```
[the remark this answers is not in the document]
JonKrone: Looks like the `hasChild` name is taken from Go's impl: https://github.com/ipfs/go-ipfs/blob/master/pin/pin.go#L610.

It's not part of any spec but is still available as `core` commands; is the naming consistency or accuracy more important?

src/core/components/pin-set.js
```diff
@@ -0,0 +1,267 @@
...
+          callback(null, rootNode)
+        })
+      })
+    },
+
+    storeItems: (items, callback) => {
+      return storePins(items, callback)
+
+      function storePins (pins, cb, depth = 0, binsToFill = 0, binsFilled = 0) {
+        cb = once(cb)
```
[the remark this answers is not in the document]
JonKrone: 👍 I totally agree; I've removed them and refactored many of the async patterns here. I should have done it initially. Thanks!

src/core/components/files.js
```diff
@@ -32,7 +32,9 @@ function prepareFile (self, opts, file, callback) {
   }
 
   waterfall([
-    (cb) => opts.onlyHash ? cb(null, file) : self.object.get(file.multihash, opts, cb),
+    (cb) => opts.onlyHash
+      ? cb(null, file)
+      : self.object.get(file.multihash, opts, cb),
```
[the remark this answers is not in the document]
JonKrone: Agreed, this was a rebase conflict that I resolved to the previous formatting.

src/core/components/dag.js
```diff
@@ -73,6 +77,34 @@ module.exports = function dag (self) {
...
+        // branch case
+        links.forEach(link => {
+          self.dag._getRecursive(link.multihash, (err, subNodes) => {
+            if (err) { return callback(err) }
+            nodes.push(subNodes)
+            if (nodes.length === links.length + 1) {
+              return callback(null, _.flattenDeep(nodes))
+            }
+          })
+        })
```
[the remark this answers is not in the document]
JonKrone: Thank you :).

src/core/components/pin-set.js
```diff
@@ -0,0 +1,267 @@
...
+    // should this be part of `object` API?
+    hasChild: (root, childhash, callback) => {
+      const seen = {}
+      if (CID.isCID(childhash) || Buffer.isBuffer(childhash)) {
+        childhash = toB58String(childhash)
+      }
+
+      return searchChildren(root, callback, root.links.length, 0)
+
+      function searchChildren (root, cb, numToCheck, numChecked) {
```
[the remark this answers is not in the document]
JonKrone: 👍 refactored

src/core/components/pin-set.js
```diff
@@ -0,0 +1,267 @@
...
+          })
+        } else {
+          // If the array of pins is > maxItems, we:
+          //  - distribute the pins among `defaultFanout` bins
+          //    - create a DAGNode for each bin
+          //      - add each pin of that bin as a DAGLink
+          //  - create a root DAGNode
+          //    - store each bin as a DAGLink
+          //  - send that root DAGNode via `cb`
+          // (using go-ipfs' "wasteful but simple" approach for consistency)
```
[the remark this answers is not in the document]
JonKrone: I'm not exactly sure what this references as this line was in the original impl. I've added a link to the section of go's pin code that handles the fanout bins.

JonKrone:
Thanks so much for the comments, just a heads up I'm working on them as I have time. I hope to push the set of changes later tomorrow.

src/core/utils.js
JonKrone: Yeah, I agree it's super weird! What we're doing is just validating that a `CID` can be constructed from `hash` and the linter prevented some other techniques so I've reverted and told eslint to ignore the line.

src/core/components/pin.js
JonKrone: I've pulled some properties out entirely and `_`-spaced a few others. We now have:

`types`
`add`
`rm`
`ls`
`_load`
`_isPinnedWithType`

`_load` is used by ipfs core to initialize the pinset. `_isPinnedWithType` is not strictly necessary but seems to be a strong feature for developers working with the core modules.

src/core/components/pin.js
JonKrone: Oh woah, `async.auto` is pretty cool! I've refactored the async here but am using other parts of `async` as it's mostly sequential actions.

src/core/utils.js
JonKrone: I really wanted to as well, and did with an earlier version but reverted it when I tried to enable custom hash algorithms as the `is-ipfs.isMultihash` helper does not work for non sha2-256 hashes.

I just looked again at the library and noticed `isCID` should cover our bases here; I've implemented it and added a test for those non default hashes for when we change it later.

src/core/components/pin-set.js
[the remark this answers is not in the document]
JonKrone: note: I've changed it to `hasDescendant` as this is internal code

src/core/components/dag.js
```diff
@@ -73,6 +77,34 @@ module.exports = function dag (self) {
         self._ipld.treeStream(cid, path, options),
         pull.collect(callback)
       )
+    }),
+
+    // TODO - use IPLD selectors once they are implemented
+    _getRecursive: promisify((multihash, callback) => {
```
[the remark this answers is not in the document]
JonKrone: Hmm. Good question, I'm not sure.

My first thought is that yes we could hit a loop but when I look at `ipld`'s treeStream implementation (https://github.com/ipld/js-ipld/blob/master/src/index.js#L322), I don't see any checking that we're not traversing into a node we've seen before. Which I would expect ipld to have to watch for if nodes can link to an ancestor. I can also imagine that many problems would arise in serialization, hashing, etc.. if that were possible. Can nodes link to an ancestor?

I think the answer here also effects the `seen` check in `pinset.hasDescendant`: https://github.com/ipfs/js-ipfs/pull/1045/files#diff-cd1d776f0354636e2cbc1d609af7fddfR62

JonKrone:
Hey @alanshaw!

Thanks again for your comments - I loved a lot of the refactoring that you suggested.

It's ready for another round of review. I've addressed all your comments and believe most are closed but I want to make sure you look at https://github.com/ipfs/js-ipfs/pull/1045#discussion_r194561922 and https://github.com/ipfs/js-ipfs/pull/1045#discussion_r193713732 again, just to be sure that's sufficient.

All of the failing tests seem to be due to some external change that is effecting all PRs (noted in an issue https://github.com/ipfs/js-ipfs/issues/1397).

TODO:
- [ ] update interface-ipfs-core tests
  - Spec for pin.add says recursive is default false, should be true
  - Spec for pin.rm says recursive is default false, should be true

src/core/components/pin.js
alanshaw: 👍 

The pin types are constants that don't _need_ to be attached to an instance, I'd export them separately so that they can be used like:
`import { PinTypes } from 'ipfs'` or `const { PinTypes } = require('ipfs')`.

src/core/components/dag.js
```diff
@@ -33,6 +35,12 @@ module.exports = function dag (self) {
         } else {
           path = '/'
         }
+      } else if (Buffer.isBuffer(cid)) {
+        try {
+          cid = new CID(cid)
+        } catch (err) {
+          callback(err)
```
alanshaw: This is missing a `return` statement

src/core/components/dag.js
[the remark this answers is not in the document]
alanshaw: > Can nodes link to an ancestor?

No because D**A**G (acyclic).

src/core/components/files.js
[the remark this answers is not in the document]
alanshaw: If the `hashAlg` is the default value it would be ok? I think we need to call back with an error if `hashAlg` is passed, instead of silently not pinning.

src/core/components/pin.js
```diff
@@ -0,0 +1,392 @@
...
+            return callback()
+          } else {
+            return callback(err)
+          }
+        }
+
+        async.parallel([
+          cb => pinset.loadSet(pinRoot.value, types.recursive, cb),
+          cb => pinset.loadSet(pinRoot.value, types.direct, cb)
+        ], (err, [rKeys, dKeys]) => {
```
alanshaw: You can't destructure like this in callbacks, if an error occurred then the result will be null/undefined this will throw.

alanshaw: That is so rad 💯

src/core/components/files.js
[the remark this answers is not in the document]
JonKrone: No, I don't think it'd work. IIRC, it's more like 'IPLD can't handle CIDv1' and anytime `hash` is provided there is a cast to v1 [here](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/files.js#L31) (as it should and is documented in the command's help text).

Example: An object's `Hash` is CIDv0 while the key it was stored under is v1
image

I am happy to respond with an error though I must ask should we just revert the functionality?

src/core/components/pin.js
diasdavid: Yeaass! :D

diasdavid:
@alanshaw are the pin tests being run https://github.com/ipfs/js-ipfs/tree/master/test/core ? doens't seem like it.

alanshaw:
🤦‍♂️ oh no it doesn't

diasdavid:
Let's see what https://github.com/ipfs/js-ipfs/pull/1405 says :)

PR #1051, opened 2017-10-28
Fix 404 links to API docs

dasilvacontin:
Please check for other occurrences around the repo, might not be the only one.


dasilvacontin:
Np :)

Issue #1055, opened 2017-11-03
offline error loses stack


pgte:
- **Version**: 0.26.0
- **Platform**: Browser
- **Subsystem**: 

#### Type:

Bug

#### Severity:

Low

#### Description:

Offline error is created globally and, when thrown, does not contain stack.

#### Steps to reproduce the error:

Create an IPFS node in the browser while offline.
See the error, it's something like this:

```
utils.js:3 Uncaught (in promise) Error: This command must be run in online mode. Try running 'ipfs daemon' first.
    at Object.../peerpad-core/node_modules/ipfs/src/core/utils.js (utils.js:3)
    at __webpack_require__ (bootstrap 17bb53dc92d5561e6d85:698)
    at fn (bootstrap 17bb53dc92d5561e6d85:111)
    at Object.../peerpad-core/node_modules/ipfs/src/core/components/swarm.js (swarm.js:7)
    at __webpack_require__ (bootstrap 17bb53dc92d5561e6d85:698)
    at fn (bootstrap 17bb53dc92d5561e6d85:111)
    at Object.../peerpad-core/node_modules/ipfs/src/core/components/index.js (index.js:17)
    at __webpack_require__ (bootstrap 17bb53dc92d5561e6d85:698)
    at fn (bootstrap 17bb53dc92d5561e6d85:111)
    at Object.<anonymous> (index.js:16)
    at Object.../peerpad-core/node_modules/ipfs/src/core/index.js (index.js:120)
    at __webpack_require__ (bootstrap 17bb53dc92d5561e6d85:698)
    at fn (bootstrap 17bb53dc92d5561e6d85:111)
    at Object.../peerpad-core/src/backend/ipfs.js (ipfs.js:3)
    at __webpack_require__ (bootstrap 17bb53dc92d5561e6d85:698)
    at fn (bootstrap 17bb53dc92d5561e6d85:111)
    at Object.../peerpad-core/src/index.js (index.js:4)
    at __webpack_require__ (bootstrap 17bb53dc92d5561e6d85:698)
    at fn (bootstrap 17bb53dc92d5561e6d85:111)
    at <anonymous>
```

Notice that the stack stays at `utils.js:3`, where the error was created.

PR #1056, opened 2017-11-03
make offline error retain stack

pgte:
... by throwing a new error instead of a cached one.


pgte:
Should fix #1055

dryajov:
nice - thats much better. LGTM.

PR #1063, opened 2017-11-08
feat: Circuit Relay

dryajov:
Remaining tasks:
- [x] Fix Node.js tests
- [x] update ipfsd-ctl to support spawning js and go nodes
- [x] add ability to ipfsd-ctl to spawn remote js and go nodes to enable browser testing
- [x] Add a test for Relay On without having any transport attached. Caught a big hairy bug here https://github.com/ipfs/js-ipfs/pull/1072#issuecomment-343730826
- [x] Complete interop tests in https://github.com/ipfs/interop/pull/6
- [x] Add documentation on how to enable Relay


dryajov:
> - [ ] Consolidate the 3 ways in which a daemon can be spawned

I did a review of all the different ways we have to spawn go and js daemons as well as past conversations on IRC 

```
[2017-09-04 08:38:04] <daviddias> dryajov: the issue on ipfsd-ctl suggests that we were already unhappy with the fragmentation. Why not evolve ipfsd-ctl to support spawning js-ipfs nodes as well?
```

and it seems like, what we're looking for is actually extending the ipfsd-ctl daemon to spawn js nodes as well?

> - [ ] Create browser interop tests

It seems like the best course of action to get tests in browsers across the board is to port the code from `http-api/test/ipfs-factory` which allows spawning nodes remotely (using a `hapi` backend).

Given the above notes, I believe the tasks should be expanded to:

- [ ] Do the required modifications to ipfsd-ctl to support spawning both go and js nodes
- [ ] Extract and generalize the code under `http-api/test/ipfs-factory`. This will allow us to run our tests seamlessly in several environments  - node, browser (electron?), etc, as well as reusing them across other projects.

@diasdavid @VictorBjelkholm

diasdavid:
> upgrading ipfsd-ctl

Yes! Let's do that

> port the code from http-api/test/ipfs-factory

Learn how https://github.com/ipfs/js-ipfs-api/tree/master/test/ipfs-factory is doing it. Ideally this should be all ipfd-ctl

dryajov:
@diasdavid woot! Awesome, I'll do that.

richardschneider:
@dryajov I was just looking at `ipfs-ctl`.  As usual, it does not run on windows.  I'll be happy to get it running on windows, once you do your consolidatiob.

dryajov:
@richardschneider that would be great!

richardschneider:
Raised [windows interop](https://github.com/ipfs/js-ipfsd-ctl/issues/167) issue.

diasdavid:
@richardschneider ipfsd-ctl does run on Windows, in fact it was one of the first modules to get support :) -- https://ci.appveyor.com/project/diasdavid/js-ipfsd-ctl-a9ywu -- thanks to @thisconnect :)

richardschneider:
@diasdavid as usually you are correct.  I'll close the issue.

diasdavid:
## Compatability table

|test              |goD relay|jsD relay|jsB relay|
|------------|----------|---------|-------|
|goD<>goD  |✅|✅|❌|
|goD<>jsD   |✅|✅|❌|
|goD<>jsB   |✅| ✅|❌|
|jsD<>goD   |✅| ✅|❌|
|jsD<>jsD    |✅| ✅|❌|
|jsD<>jsB     |❌|✅|❌|
|jsB<->goD |❌|❌|❌|
|jsB<->jsD   |❌|❌|❌|
|jsB<->jsB   |✅|✅|❌|

diasdavid:
@dryajov brought back the compatibility table and added a column for when the browser acts as a Relay https://github.com/ipfs/js-ipfs/pull/1063#issuecomment-343645193

dryajov:
@diasdavid thanks!

diasdavid:
Make sure to add a test for Relay On without having any transport attached. Caught a big hairy bug here https://github.com/ipfs/js-ipfs/pull/1072#issuecomment-343730826

dryajov:
> Add a test for Relay On without having any transport attached. Caught a big hairy bug here #1072 (comment)

@diasdavid, I believe all we vant to do is to set the dafault flag for `EXPERIMENTAL.Swarm.DisableRelay` to true, I see that you changed the options and now they don't match Go's - https://github.com/ipfs/js-ipfs/blob/104ef1ef6cc64c09ec886f67c28e9b5d37bc9e66/src/core/components/libp2p.js#L24-L30

I also think that in addition to this, there should be a check to skip circuit if swarm doesn't have any transports registered (that is skip it in swarm proper).

diasdavid:
@dryajov EXPERIMENTAL flags are not the same as things in the CONFIG file. If you want to have that information in both places that is fine but for now, js-ipfs won't have Relay enabled by default until it is properly tested (i.e this PR finished)

dryajov:
@diasdavid I agree, we should disable circuit for now (i.e. `enabled: false`). But I don't think we should change the config flags, since they stop matching Go's at that point.

dryajov:
I think I know what you're saying now - _yes, we should keep the config flags and the EXPERIMENTAL flags the same_ and also disable relay by default with ` enabled: !get(config, 'EXPERIMENTAL.Swarm.DisableRelay', true), `.

dryajov:
@diasdavid this should address dialing when no transports are registered - https://github.com/libp2p/js-libp2p-swarm/pull/236

diasdavid:
@dryajov I know you are working on upgrading ipfsd-ctl, just wanted to suggest that it would be good to fix the issues to make the Circuit Relay tests in `test/core/circuit-relay.spec.js` pass first and the same goes for the Interop tests that are already there. The ipfsd-ctl upgrade is just needed for some of the tests.

dryajov:
@diasdavid I agree, ipfsd-ctl will mostly help covering the browser based cases. I'll fix the tests we currently have.

.aegir.js
```diff
@@ -18,7 +18,8 @@ function start (done) {
       (cb) => js([`${base}/10012`, `${base}/20012/ws`], true, 31012, 32012, cb),
       (cb) => js([`${base}/10013`, `${base}/20013/ws`], true, 31013, 32013, cb),
       (cb) => js([`${base}/10014`, `${base}/20014/ws`], true, 31014, 32014, cb),
-      (cb) => js([`${base}/10015`, `${base}/20015/ws`], true, 31015, 32015, cb)
+      (cb) => js([`${base}/10015`, `${base}/20015/ws`], true, 31015, 32015, cb),
+      (cb) => go([`${base}/10027`, `${base}/20027/ws`], true, 33027, 44027, cb), // we need this for circuit for now
```
diasdavid: Not true, you should move the interop test to the interop suite

test/core/circuit-relay.spec.js
```diff
@@ -102,9 +101,7 @@ describe.skip('circuit', function () {
 
   after((done) => factory.dismantle(done))
 
-  // TODO: 1) figure out why this test hangs randomly
-  // TODO: 2) move this test to the interop batch
-  it.skip('node1 <-> goRelay <-> node2', (done) => {
+  it('node1 <-> goRelay <-> node2', (done) => {
```
diasdavid: Please move this test to the interop suite. That was what was in the TODO

test/core/circuit-relay.spec.js
```diff
@@ -115,15 +112,19 @@ describe.skip('circuit', function () {
...
+        (buffer, cb) => {
+          expect(buffer).to.deep.equal(data)
+          cb()
+        }
       ], done)
     })
   })
 
-  it('node1 <-> jsRelay <-> node2', (done) => {
+  it.skip('node1 <-> jsRelay <-> node2', (done) => {
```
diasdavid: Why skip?

.aegir.js
dryajov: Interop tests don't run in browsers right now, I thought we wanted some browser coverage?

test/core/circuit-relay.spec.js
dryajov: I can move those to the interop tests, but we'll lose any sort of browser coverage until ipfsd-ctl is ready.

dryajov:
@diasdavid the tests in `test/core/circuit-relay.spec.js` are already part of the interop suit, the only reason those were added is to get some sort of browser coverage, since interop tests don't run in the browser.

diasdavid:
> since interop tests don't run in the browser.

You can run interop tests in the browser too, it uses `aegir` today, it follows the same rules

dryajov:
@diasdavid that's true, but we would have to rely on prestarted nodes as we do in `core`, which wasn't the pattern that the rest of the interop tests were following, when initially developed it didn't seem like the right way to go, because of above reason. Now with ipfsd-ctl being close to working in node and browsers we should be able to run this tests everywhere without relying on ad-hoc nodes.

diasdavid:
@dryajov please rebase master onto this branch :)

diasdavid:
@dryajov just an heads up that the interop tests have been ported to the interop test repo https://github.com/ipfs/interop, you will need to open a PR there.

RangerMauve:
Will circuit-relay remove the need for `websocket-star` or `webrtc-star` servers?

dryajov:
@RangerMauve yes, it should at least reduce the need for them, tho probably not eliminating them entirely.

dryajov:
Work to add full circuit coverage (https://github.com/ipfs/js-ipfs/pull/1063#issuecomment-343645193) across `go`, `js` and `browser` is in part being handled here - https://github.com/ipfs/interop/pull/5

diasdavid:
Is this PR still needed now that we have https://github.com/ipfs/interop/pull/6 ?

diasdavid:
I guess this answers my question:

image

Can you make sure to add an example to the examples folder?

RangerMauve:
Sorry, a little confused. Is circuit relay working between js-ipfs instances at the moment? 
Is it being automatically used when dialing into peers?
Is there still something pending for getting peers to advertise circuits?

dryajov:
@diasdavid I had to make a small adjustment to the table in https://github.com/ipfs/js-ipfs/pull/1063#issuecomment-343645193. The reason is that, we can't relay if both the relay and one other node are browser, since there is no way for the two browsers to dial each other. We discussed the scenario where all the nodes are browser, but it also applies to the scenario above.

diasdavid:
> there is no way for the two browsers to dial each other.

How come? One of the `*-star` transports should do.

dryajov:
@diasdavid ok, trying with `websockets-star`

dryajov:
@diasdavid is there any `*-star` implementation for go, otherwise I believe we could only test js when a `*-star` transport is involved?

diasdavid:
@dryajov the point on having Relay, is that we are not dependent on transport compatibility

you can do `browser <*-star> browser <websockets> goDaemon`

dryajov:
@diasdavid you're correct - it was too late yesterday when I was looking at it.

dryajov:
We're back at having all the browser cases covered (except  `browser <> browser <> browser`), thanks @diasdavid for clarifying how to connect those nodes!

dryajov:
@diasdavid how do we want to handle the config options here? Are we fine if they don't match Go options/configs, or do we want to make them match across both implementations? I say they _should_ match, but we reverted that some time ago and I'm not entirely sure why.

diasdavid:
@dryajov what config options, what is your proposal vs what go does?

dryajov:
@diasdavid these are the go config parameters I'm talking about:

screen shot 2018-02-06 at 5 04 09 pm

my question is, should we for now leave our config parameters and options as they are : 

```
      EXPERIMENTAL: {
        relay: {
          enabled: true,
          hop: {
            enabled: false
          }
        }
      }
```

or should we make them match Go's config parameters?

dryajov:
Hi @Mr0grog, @Admin-DataRoads and @JonKrone! 

I finally have a pretty solid draft of the tutorial in this branch, it would be awesome if you could run through it and leave some feedback. 

Ideally, you should be able to complete the tutorial and have the included example communicate over the established relay. I'd love to hear how hard/easy it was to complete it and if you were able to make it through to the end at all (hopefully you could!).

dryajov:
BTW, the tutorial is located under `examples/circuit-relaying` in this branch.

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,247 @@
...
+API is listening on: /ip4/127.0.0.1/tcp/5002
+Gateway (readonly) is listening on: /ip4/127.0.0.1/tcp/9090
+Daemon is ready
+```
+
+Look out for an address similar to `/ip4/127.0.0.1/tcp/4003/ws/ipfs/Qm...` note it down somewhere, and lets move on to the next step. 
+
+### Configure and run the bundled example
+
+Now that we have ipfs installed and initialized, lets set up the included example. This is a standard npm package, so the usual `npm install` should get us going.
```
JonKrone: Might want to be explicit and mention `cd`ing into this example dir

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+The bundled example is a simple chat app that uses another cool ipfs feature - `pubsub`. Lets open up a browser and paste the above url into the address bar. We should see something similar to the following image:
+
+![](./img/img1.png)
+
+### Connect the two browser nodes to the circuit relay
+
+In order for our browser nodes to be able to send messages to each other, we need to first get them connected, but for that we need to use the relay, remember, browser nodes can't be dialed directly because of lack of socket support, so the relay is here to solve that.
+
+Enter the `/ip4/127.0.0.1/tcp/4003/ws/ipfs/...` address noted above into the `Connect to Peer` field and hit the connect button:
```
JonKrone: After reading, I realize I don't know *how* circuit relay works. This section might be a good place to mention how these nodes are connecting.

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+Now that we have the two browsers connected, lets try the app out. Type a few words in one of the browser windows and you should see them appear in the other as well!
+
+![](./img/img7.png)
+
+Thats it!
+
+### Conclusion
+
+Lets recap what we accomplished in this tutorial. We where able to install a js and go ipfs node and configure them as circuit relays, we connected our browsers to the relay and were able to use the bundled chat app to send messages browser to browser.
```
JonKrone: typo: "We where" -> "We were"

JonKrone:
Hey! Your tutorial is great - really clear and descriptive. I feel like I learned about more of IPFS than just circuit relays. I added just a couple comments in-line and below I've listed some general questions that I have after reading. Maybe some can be mentioned in the tut.

- Is there/will there be a mechanism for discovering relays?
- Will future IPFS releases be configured as circuit relays by default?
- When a node marks itself as a relay, is there a supported way to filter which connections you accept? i.e. maybe a company wants to set up a dedicated relay node for only their apps?
    - Not sure how I feel about being able to do that; just curious. I think there's some ability to filter connections in other places.
- I think there is some more room to explain *how* circuit relays work internally. What does connecting two nodes mean, how does a relay actually forward messages?

I think this is a great piece for @Mr0grog to review and give some feedback on.

(I missed your initial ping, sorry)

examples/circuit-relaying/README.md
dryajov: > After reading, I realize I don't know how circuit relay works

Thats a very good point - although I'm not sure what is a reasonable amount of info to include here? There is the circuit [readme](https://github.com/libp2p/js-libp2p-circuit), should it be linked somewhere perhaps?

> This section might be a good place to mention how these nodes are connecting.

In this case `Connect to Peer` here, it will use the websockets transport to dial the relay node - which is your usual `ipfs swarm connect <QmPeer...>` operation. 

I'll add that in there so its clear what's going on 👍!

dryajov:
Great questions @JonKrone 

> Is there/will there be a mechanism for discovering relays?

We have discussed service discovery in several places, but we don't have anything of this sort in place just yet.  Currently, circuit will try to identify if any incoming node is also a relay, enabling a sort of "poor mans" relay discovery.

> Will future IPFS releases be configured as circuit relays by default?

There are actually two parts to circuit

- There is the dialer and listener parts that behave as a regular transport, these are enabled with `relay: {enabled: true}`, which allows your node to use other circuit relay nodes in the network. 

- There is also the actual `Relay` aka `HOP` which is the part that actually relays the connections. This is enabled with `{hop: {enabled: true}}`

I don't think we should have the `Relay` enabled by default, as that would mean that any node will be forwarding traffic on behalf other nodes, this can quickly overload low powered nodes such as phones and IoT devices, and browsers (browsers also relay ;)). But we should allow any node in the network to  use dial/listen for relayed connections.

> When a node marks itself as a relay, is there a supported way to filter which connections you accept? i.e. maybe a company wants to set up a dedicated relay node for only their apps?

Not right now, but a great idea indeed 👍 

> I think there is some more room to explain how circuit relays work internally. What does connecting two nodes mean, how does a relay actually forward messages?

Thats a great suggestion, I'll add more info to the tutorial. Those questions above are great talking points as well.

Awesome stuff! Thanks so much!

dryajov:
@JonKrone I've updated the tutorial with an explanation of how circuit works. Could you please take a look and let me know if it makes sense?

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+![](./img/img6.png)
+
+### Send data browser to browser.
+
+Now that we have the two browsers connected, lets try the app out. Type a few words in one of the browser windows and you should see them appear in the other as well!
+
+![](./img/img7.png)
+
+Thats it!
```
Mr0grog: I know this tutorial gets a little long, but it might still be helpful to call out what happened in the browser code that was important. I *think* that’s:

1. Set the right config options in `new IPFS(config)`: https://github.com/ipfs/js-ipfs/blob/feat/circuit-interop/examples/circuit-relaying/src/app.js#L31-L39

2. The `/p2p-circuit/...` addresses came from calling `ipfs.id`: https://github.com/ipfs/js-ipfs/blob/feat/circuit-interop/examples/circuit-relaying/src/app.js#L52-L59

3. We called `ipfs.swarm.connect(address)` to connect, which should be nothing special (all the magic is in the address itself!) https://github.com/ipfs/js-ipfs/blob/feat/circuit-interop/examples/circuit-relaying/src/app.js#L102-L107

4. Sending and receiving don’t actually require any addressing because we are [broadcasting to anyone listening on the room name](https://github.com/ipfs/js-ipfs/blob/feat/circuit-interop/examples/circuit-relaying/src/helpers.js#L42-L49) and, on the other side, just [listening on the room name](https://github.com/ipfs/js-ipfs/blob/feat/circuit-interop/examples/circuit-relaying/src/helpers.js#L32-L36).

Did I get that right?

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### Whats up with this `HOP` and `STOP` nonsense?
+
+Circuit relay consists of two logical parts - dialer/listener and relay (`HOP`). The listener is also known as the `STOP` node. Each of this - dial, listen and relay happen on a different node. If we use the nodes from the above example, it looks something like this:
+
+- The `dialer` knows how to dial a `relay` (`HOP`) - `Node A`
+- The `relay` (`HOP`) knows how to contact a destination node (`STOP`) and create a circuit - `Relay` node
+- The `listener` (`STOP`) knows how to process relay requests that come from the relay (`HOP`) node - `Node B`
+
+These are also the names used by circuit internally to identify the network message types.
```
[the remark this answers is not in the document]
dryajov: Ha! I struggled to word this when putting the sentence together - I mean `HOP` and `STOP` :) Not even sure if this is needed - but sounded like a good "fun fact":)

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### A few caveats (and features).
+
+There are a couple of caveats and features to be aware of
+
+- Currently a `Relay` will only work if it already has a connection to the node being relayed to
```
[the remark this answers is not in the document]
dryajov: Well both - relay (`HOP`) has two modes - `active` and `passive`

- `active` will dial the `STOP` node if there is no connection to it already
- `passive` will NOT dial the `STOP` node and fail if there is no connection

The thing is that, even tho it's implemented in both go and js, we're not advertising that just yet. But I still though that the point about having an active connection to be able to use the relay was important. I'll remove the currently - you're right, it's confusing 👍

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+After it finishes, we should be able to run the project with `npm start` and get a similar output to the bellow one:
+
+```
+npm run start
+Server running at http://localhost:1234
+```
+
+The bundled example is a simple chat app that uses another cool ipfs feature - [pubsub](https://github.com/libp2p/specs/tree/master/pubsub). Lets open up a browser and paste the above url into the address bar. We should see something similar to the following image:
```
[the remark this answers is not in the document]
dryajov: I've reworded the next sections to start with two browser windows - let me know how it looks.

examples/circuit-relaying/README.md
dryajov: I think this is a great idea - I'll add an explanation of what the browser is doing! 

And you got it right! 💯

dryajov:
> Side note: not sure what was going on, but this pegged my CPU in Firefox. Not quite as bad in Chrome, but still bad.

Hrm... it seems like parcel is the culprit here. It goes into a loop trying to reload css files (even when not modified) and eventually runs out of stack. I might scrap it in favor of good old browserify.

examples/circuit-relaying/README.md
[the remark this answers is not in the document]
Mr0grog: Gotcha. If we aren’t advertising that functionality, it makes sense not to mention.

examples/circuit-relaying/README.md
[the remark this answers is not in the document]
Mr0grog: Looks good.

dryajov:
@Mr0grog I changed the last section to kinda expand on what happens in the browser - please let me know if that makes any sense. 👍

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,336 @@
...
+This is what it looks like, in simplified steps:
+
+1. `Node A` tries to connect to `Node B` over one of its known addresses
+2. Connection fails because of firewall/NAT/incompatible transports/etc...
+3. Both `Node A` and `Node B` know of the same relay - `Relay`
+4. `Node A` falls back to dialing over `Relay` to `Node B` using its `'/p2p-circuit'` address, which involves:
+   1. `Node A` sends a `HOP` request to `Relay`
+   2. `Relay` extracts the destination address, figures out that a circuit to `Node B` is being requested
+   3. `Relay` sends a `STOP` request to `Node B`
+   4. `Node B` responds with a `SUCCESS` message
```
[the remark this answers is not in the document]
dryajov: > Does Node B send a SUCCESS message to Relay or to Node A?

`Node B` sends success to `Relay`.

> When a Relay has been established between two nodes, is the connection two-way or is another handshake needed if Node B wants to communicate with Node A?

It's two way

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,336 @@
...
+![](./img/img7.png)
+
+Thats it!
+
+### So what just happened?
+
+Good question! 
+
+- We used [js-ipfs](htpps://github.com/ipfs/js-ipfs) running in the browser with circuit relay enabled:
+  - _Notice the `EXPERIMENTAL.relay.enabled` bellow_
```
Mr0grog: Typo: “bellow” → “below”

We should probably also call out `EXPERIMENTAL.pubsub`, since that’s also a non-default. (right?)

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,336 @@
...
+
+```js
+const ipfs = new IPFS({
```
Mr0grog: I think it might be helpful to use a comment to note the file this is in so a reader can go see it in context, e.g:

```js
// in `src/app.js`
const ipfs = new IPFS({
```

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,336 @@
...
+  },
+  config: {
+    Bootstrap: []
+  }
+})
+```
+
+- We connected the browser nodes to an external node over its websocket transport using the `/ip4/127.0.0.1/tcp/4003/ws/ipfs/...` multiaddr. That external node happens to be `HOP` node, meaning that it can relay connections for our browsers (and other nodes) allowing our browsers to connect
+
+- And finally we connected the two browser nodes using the `/p2p-circuit/ipfs/...` multiaddr
```
Mr0grog: I know it’s not special or unique to what this tutorial is about, but I do still think it would be helpful to show that code here (it’s the same for both connections and we should say that).

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,336 @@
...
+
+Notice how there wasn't anything special we had to do to use the circuit once we had everything connected, all the magic is in the multiaddr! Multiaddrs are **AWESOME**!
```
Mr0grog: Finally, I do still think it would be helpful to also say that nothing special has to be done for pubsub because *any* nodes that are connected can see any pubsub messages. A reader might not be familiar with how it works. I don’t think a deep explanation is necessary, but at least noting that (because it’s critical to what happened in the end) is a good idea.

examples/circuit-relaying/README.md
dryajov: ack on typo. 

I think we can leave `pubsub` out (remove from options in the example), cause it doesn't have much to do with circuit - I think it would add to the confusion?

examples/circuit-relaying/README.md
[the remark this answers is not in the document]
Mr0grog: > What does a failure look like?

Ahhhhh, that’s a *really* good question that should be answered in the tutorial here.

Mr0grog:
@dryajov that helps a lot! Left some more notes inline.

examples/circuit-relaying/README.md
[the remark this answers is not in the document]
dryajov: From the perspective of `Node A`, the stream closes - we don't propagate the error. So the user should not care about handling circuit specific errors, if an error occurred establishing the circuit, it will look just as if the connection dropped.

dryajov:
@Mr0grog just made a few more adjustments based on your suggestions 👍 I didn't go all out explaning what pubsub is, but I linked to a few resources explaining it. I think its a good middle ground?

dryajov:
@Mr0grog and @JonKrone. Thanks so much for the great feedback, I've updated the tutorial with the latest suggestions from @Mr0grog and I believe I've covered all the requests, can we give this one more pass before I sumon @diasdavid to give it a final review? Thanks so much in advance!

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,349 @@
...
+```js
+    ipfs.swarm.connect(peer, (err) => {
+      if (err) {
+        return console.error(err)
+      }
+      $pAddrs.innerHTML += `<li>${peer.trim()}</li>`
+    })
+```
+
+The above code snippet, handles connecting to other nodes using `ipfs.swarm.connect`, notice how there wasn't anything special we had to do to use the circuit once we had everything connected, all the magic is in the multiaddr! [Multiaddrs](https://multiformats.io/multiaddr/) are **AWESOME**!
```
Mr0grog: The commas are a little odd in this first sentence. Maybe break it in two and use a semicolon before “all the magic”:

> The above code snippet handles connecting to other nodes using `ipfs.swarm.connect`. Notice how there wasn't anything special we had to do to use the circuit once we had everything connected; all the magic is in the multiaddr! [Multiaddrs](https://multiformats.io/multiaddr/) are **AWESOME**!

JonKrone:
The tutorial LGreatTM! Is there anywhere in our documentation that we can add links to this as reference documentation, perhaps in js-libp2p land? Like 'check out this example' for more information.

dryajov:
@JonKrone we usually have separate examples on how to enable this at the libp2p level.

README.md
```diff
@@ -205,7 +205,13 @@ const node = new IPFS({
...
     pubsub: true,
     sharding: true, // enable dir sharding
-    dht: true // enable KadDHT, currently not interopable with go-ipfs
+    dht: true, // enable KadDHT, currently not interopable with go-ipfs
+    relay: {
+      enabled: true, // enable circuit relay dialer and listener
+      hop: {
+        enabled: true // enable circuit relay HOP (make this node a relay) 
+      }
+    }
```
diasdavid: 👍

examples/circuit-relaying/.gitignore
```diff
@@ -0,0 +1,119 @@
...
+# *.iml
+# modules.xml
+# .idea/misc.xml
+# *.ipr
+
+# Sonarlint plugin
+.idea/sonarlint
+
+
+# End of https://www.gitignore.io/api/vim,macos,webstorm,visualstudiocode
```
diasdavid: Where did this file came from? Let's remove it before merging

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,349 @@
...
+   1. `Node A` sends a `HOP` request to `Relay`
+   2. `Relay` extracts the destination address, figures out that a circuit to `Node B` is being requested
+   3. `Relay` sends a `STOP` request to `Node B`
+   4. `Node B` responds with a `SUCCESS` message
+   5. `Relay` proceed to create a circuit over the two nodes
+5. `Node A` and `Node B` are now connected over `Relay`
+
+That's it!
+
+#### What's up with this `HOP` and `STOP` nonsense?
```
diasdavid: remove the word `nonsense`.

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,303 @@
...
+- 5. Dial the two browser nodes using a `/p2p-circuit` address
+- 6. Finally, send data from one browser using the bundled example!
+
+> We should end up with something similar to the bellow screenshot after we've gone through all the steps:
+
+![](./img/img7.png)
+
+Let's go.
+
+### 1. Set up
```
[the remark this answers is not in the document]
diasdavid: 👍

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+- And finally we connected the two browser nodes using the `/p2p-circuit/ipfs/...` multiaddr. Take a look at the code below in [src/app.js](src/app.js#L102...L107) - lines 102-107
+
+```js
+    ipfs.swarm.connect(peer, (err) => {
+      if (err) {
+        return console.error(err)
+      }
+      $pAddrs.innerHTML += `<li>${peer.trim()}</li>`
+    })
```
diasdavid: unnecessary indentation.

examples/circuit-relaying/README.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+The above code snippet handles connecting to other nodes using `ipfs.swarm.connect`. Notice how there wasn't anything special we had to do to use the circuit once we had everything connected; all the magic is in the multiaddr! [Multiaddrs](https://multiformats.io/multiaddr/) are **AWESOME**!
+
+I encourage the reader to take a look at the bundled app code to see how the browser nodes get setup, suffice to say nothing changes from the perspective of using an `IPFS` node in js code, apart from the new `EXPERIMENTAL` options.
+
+Finally, a side note on [pubsub](https://github.com/libp2p/specs/blob/master/pubsub/README.md). We've used the amazing [ipfs-pubsub-room](https://github.com/ipfs-shipyard/ipfs-pubsub-room) module, to enable the chat functionality. Make sure to take a look at the demo [video](https://www.youtube.com/watch?v=Nv_Teb--1zg) that explains how pubsub works and how it can be used to build other applications!
+
+Cheers!
```
diasdavid: This is a stellar tutorial, great work @dryajov and @Mr0grog 👏🏽👏🏽👏🏽

diasdavid:
I'm going ahead and merge this one. Let's save the fireworks for when https://github.com/ipfs/interop/pull/6 is merged though :) Anyways, great work! :D It took a while but it got done! :)

dryajov:
Weeeeee!!!!! ❤️

nezzard:
Why i get "Circuit not enabled" after ipfs.swarm.connect ?

PR #1064, opened 2017-11-08
Adding babel

### LINKED ISSUE #938: Does not work with UglifyJS version 2
- **Version**: 0.25.0
- **Platform**: JS
- **Subsystem**:

#### Description: 
Trying to build IPFS node in the browser, uglifyJS through webpack won't build the app. 

```
static/js/main.725ac8bb.js from UglifyJs
Unexpected token: name (CID) [./~/cids/src/index.js:23,0][static/js/main.725ac8bb.js:8442,6]
```

Hey guys, I'm trying to build an IPFS node in my browser. Building the app with `yarn run build` using create-react-app gives the error above. It looks like it doesn't like the type CID. I've installed CID as well, but no go. Anyone know what's going on here?

louisgv:
This will resolve #938 , https://github.com/ipld/js-cid/issues/38 and https://github.com/ipld/js-cid/issues/42

# TODO: 

Add this to the Travis pipeline and expose it as the final build for other module to import.


louisgv:
Actually the built lib generates this error:
```
TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
```

package.json
```diff
@@ -19,6 +19,7 @@
     "npm": ">=3.0.0"
   },
   "scripts": {
+		"babel": "babel dist -d lib",
```
diasdavid: Could this create a a file to `dist/index.es5.js` instead?

diasdavid:
Just want to point out that @fazo96 provided some good intel on how to work around the fact that uglify2 doesn't support ES6, check it here https://github.com/ipfs/js-ipfs/issues/938#issuecomment-367280091

PR #1072, opened 2017-11-12
Figure out why tests are failing

diasdavid:
Solved some issues on the sharding tsts, but still seems something is weird with that codebase, see:

image

### DIFF
--- a/test/core/files-sharding.spec.js
+++ b/test/core/files-sharding.spec.js
@@ -11,20 +11,25 @@ const pull = require('pull-stream')
 const IPFS = require('../../src/core')
 const createTempRepo = require('../utils/create-repo-nodejs.js')
 
-describe('files dir', () => {
-  const files = []
-  for (let i = 0; i < 1005; i++) {
-    files.push({
-      path: 'test-folder/' + i,
-      content: Buffer.from('some content ' + i)
-    })
+describe('files directory (sharding tests)', () => {
+  function createTestFiles () {
+    const files = []
+
+    for (let i = 0; i < 1005; i++) {
+      files.push({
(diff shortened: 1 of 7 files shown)


test/core/files-sharding.spec.js
```diff
@@ -79,22 +83,25 @@ describe('files dir', () => {
...
-        pull.values(files),
+        pull.values(createTestFiles()),
         ipfs.files.createAddPullStream(),
         pull.collect((err, results) => {
           expect(err).to.not.exist()
           const last = results[results.length - 1]
-          expect(last.path).to.be.eql('test-folder')
-          expect(last.hash).to.be.eql('QmY8TxNWtNViN7syd2DHazPqu21qWfSNjzCDe78e4YMsUD')
+          expect(last.path).to.eql('test-folder')
+          expect(last.hash).to.eql('Qmb3JNLq2KcvDTSGT23qNQkMrr4Y4fYMktHh6DtC7YatLa')
```
diasdavid: @pgte had to change this hash because now createTestFiles creates a clean array of files, resulting in a different hash. This means that js-ipfs-unixfs-engine mutates the files that are passed to it, this should be considered a bug.

diasdavid:
I keep seeing these errors coming out from anywhere:
image

diasdavid:
Just seeing 
image

now

diasdavid:
Found the issue. **Circuit Relay was always on by default**, causing IPFS to fail when it had no addresses to dial from, but it had relay on. It was trying to use relay without knowing how to use any other transport). This is a bug in the Circuit Relay code, @dryajov

Issue #1068, opened 2017-11-11
Impl the http endpoints and cli cmds for PubSub


diasdavid:
https://github.com/ipfs/js-ipfs/pull/610 never really added support for PubSub to the http-api and the cli, just for core. 

We do have tests and the impl is underway which should make things easier for someone to pick it up. To try it out you can  enable these tests:

- https://github.com/ipfs/js-ipfs/blob/master/test/http-api/spec/pubsub.js
- https://github.com/ipfs/js-ipfs/blob/master/test/http-api/interface/pubsub.js
- https://github.com/ipfs/js-ipfs/blob/master/test/cli/pubsub.js

richardschneider:
@diasdavid There are incompatibilities between `go` and `js` in the published message that is sent by `pubsub publish`.

- js has renamed the field `topicIDs` to `topicCIDs`
- js encodes the field `from` in base58 whereas `go` uses base64

See the [code](https://github.com/libp2p/js-libp2p-floodsub/blob/master/src/index.js#L262) in `libp2p-floodsub`.

richardschneider:
Note that none of the above tests check a published message.

Also the test for [pubsub ls](https://github.com/ipfs/js-ipfs/blob/master/test/http-api/spec/pubsub.js#L98) is skipped.

richardschneider:
Changes will have to be made
- [ ] [flood-sub](https://github.com/libp2p/js-libp2p-floodsub)
- [ ] [ipfs](https://github.com/ipfs/js-ipfs)

diasdavid:
@richardschneider great catch!

- https://github.com/libp2p/go-floodsub/blob/master/pb/rpc.proto
- https://github.com/libp2p/js-libp2p-floodsub/blob/master/src/message/rpc.proto.js

No idea how did this happen. This will be a breaking change. Wanna ship it in time for js-ipfs 0.27? An interop test would also be welcome

richardschneider:
Working on changes now.  when is js-ipfs 0.27 scheduled to ship.

diasdavid:
@richardschneider once I finish https://github.com/ipfs/interface-ipfs-core/pull/162 which should be this week (so releasing next week)

richardschneider:
No worries, will have changes in tonight/tomorrow.

Will the release include `jsipfs files ls ...`?

diasdavid:
@richardschneider it will :)

richardschneider:
[...]

PR #1077, opened 2017-11-14
Pubsub message fields

richardschneider:
- Rename topicCIDs to topicIDs
- `msg.from` is base64 encoded in go-ipfs

See the [background discussion](https://github.com/ipfs/js-ipfs/issues/1068).


richardschneider:
@diasdavid Needs PR https://github.com/libp2p/js-libp2p-floodsub/pull/49 and https://github.com/ipfs/interface-ipfs-core/pull/169 released.  Then update dependencies.

diasdavid:
@richardschneider deps updated. Could you please add some interop pubsub tests to make sure both communicate well?

richardschneider:
It's not an easy test to create; which is most likely why no-one has written it yet.  I'm solving other issues now.  Perhaps create an issue and assign it to me.

src/http/api/resources/pubsub.js
```diff
@@ -20,10 +21,10 @@ exports.subscribe = {
 
     const handler = (msg) => {
       res.write(JSON.stringify({
-        from: msg.from,
+        from: bs58.decode(msg.from).toString('base64'),
```
diasdavid: ah, missed this! This should continue as b58String. Will revert

diasdavid: https://github.com/ipfs/js-ipfs-api/commit/470777d00eb94906d6aca57b2034814d49e081f0#commitcomment-25652929

diasdavid: After all I didn't miss it, in js-ipfs-api it is correct. here is where it is wrong.

We want this value to be a b58String. The reason why the http-api sends base64 is just because how JSON serialization works in go-ipfs, the users should not have to worry about that.

diasdavid:
@richardschneider wrote the set up for you, all you have to do is `npm run test:interop` and whatever assertions you write there get checked. Wanna give it a try?

richardschneider:
will try to write interop test tomorrow

richardschneider:
@diasdavid Could you merge these changes.  Then I can starting testing from a solid base.  Cheers!

test/utils/interop-daemon-spawner/go.js
```diff
@@ -39,7 +40,7 @@ class GoDaemon {
         this.node = node
         this.node.setConfig('Bootstrap', '[]', cb)
       },
-      (res, cb) => this.node.startDaemon(cb),
+      (res, cb) => this.node.startDaemon(this.flags, cb),
```
diasdavid: @richardschneider this was causing all interop tests to fail.

PR #1082, opened 2017-11-17
Fix: Promisify stop

### LINKED ISSUE #1076: Promisify stop()?
We currently return from `stop()` with a [callback](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/stop.js#L5).

In order to stop IPFS properly, we write:
```javascript
ipfs.stop(() => {
// do other shutdown things
})
```

I find myself often preferring to use Promises and more recently await. As per IPFS api functions, we return a Promise (or callback) and I find it confusing that `stop()` is only callback.

If we returned a Promise, we would write:
```javascript
ipfs.stop().then(() => {
// do other shutdown things
})

// Or:
await ipfs.stop()
// do other shutdown things
```

**I'd like to propose we promisify the stop function.**

We would keep the original callback in addition to returning a Promise, same as we do for [API functions](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/files.js#L63) and the [start()](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/start.js#L10) function.

How does everyone feel about this?

I'd be happy to PR this, but this would be a great task for **first contributors** if someone wants to work on it!

haadcode:
This PR will promisify stop().

Closes #1076 

The effective changes are in stop.js and in create-node.spec.js. The changes to ls and files are purely to make linter happy.

### DIFF
--- a/src/cli/commands/ls.js
+++ b/src/cli/commands/ls.js
@@ -1,8 +1,6 @@
 'use strict'
 
 const utils = require('../utils')
-const Unixfs = require('ipfs-unixfs')
-const pull = require('pull-stream')
 
 module.exports = {
   command: 'ls <key>',
--- a/src/core/components/stop.js
+++ b/src/core/components/stop.js
@@ -1,9 +1,10 @@
 'use strict'
 
 const series = require('async/series')
+const promisify = require('promisify-es6')
 
 module.exports = (self) => {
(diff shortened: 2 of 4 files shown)


haadcode:
My pleasure! This is gonna be so much nicer to use now :) 👍 ❤️ 🎉

PR #1085, opened 2017-11-17
fix: Promisify .block (get, put, rm, stat)

haadcode:
This PR will promisify the functions in .block

Depends on https://github.com/ipfs/interface-ipfs-core/pull/170


haadcode:
👍

PR #1091, opened 2017-11-19
create snap package

mkg20001:
This adds a snapcraft file for building a jsipfs snap
ToDos:
 - [ ] Fix all the bugs
 - [ ] Store repo in `common` directory to avoid duplication
 - [ ] Setup automatic building via https://build.snapcraft.io
 - [ ] Release to stable channel

The current (not working) release can be tested with `snap install jsipfs --edge`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,11 @@ test/test-data/go-ipfs-repo/LOG.old
 
 # while testing npm5
 package-lock.json
+
+# snapcraft
+
+/parts
+/prime
+/stage
+/snap/.snapcraft
+*.snap
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,21 @@
+name: jsipfs
+version: 0.26.0
+summary: JavaScript implementation of the IPFS specification
+description: |
+  The JavaScript implementation of the IPFS protocol.
+
+grade: stable
+confinement: strict
+
+apps:
+  jsipfs:
+    command: jsipfs
+    plugs:
+      - network
+      - network-bind
+
+parts:
+  jsipfs:
+    plugin: nodejs
+    node-engine: 8.9.1
+    source: .


dryajov:
@elopio you might be interested in this :)

elopio:
oh, yes, this would be awesome!
@mkg20001 I see a problem with the command trying to call chown. Is that what's failing for you?

mkg20001:
@elopio No. For me when I run the command without arguments it works the first time I run it and then the second time it suddenly stops working (command just hangs - Ctrl+C does nothing)

diasdavid:
Hi all! We will continue support js-ipfs install through the npm package manager. I think making a snap package only makes sense for go-ipfs for now, which @elopio already created :) Closing for now, hope you agree.

elopio:
@diasdavid I think that the js snap would not hurt. It gives a one command install to all the ubuntu users, so plenty of testers without a lot of effort. That's the theory at least, because I haven't been able to focus on this to debug it :)
If it's a snapcraft bug, we'll need to fix it anyway, so I'll keep it in my backlog. I'll report any findings.

elopio:
oh, some info @mkg20001, when I ran jsipfs with strace I got:

`write(2, "need to run as root or suid", 27need to run as root or suid) = 27`

so, if I run `sudo jsipfs` it seems to work nicely. Now, to find what makes it require root on the second run...

VictorBjelkholm:
@elopio it's not just building the snap package we'll need to do if we should support snap but also testing for each release, which adds maintenance burden. I don't think it's worth supporting that way of distributing js-ipfs at this point.

mkg20001:
@VictorBjelkholm The snap could - for now - be left in the `edge` channel so people know it's not finished.
And at some point in the future a stable release could be made.

elopio:
Yeah @VictorBjelkholm. I mean, if you don't want the snap, that's of course understandable. But putting it on edge and getting people from the ubuntu community to test it requires to just push the button in build.snapcraft.io.

What's not so simple is to understand what's the problem here. The suid error was actually coming from strace, not from ipfs. Our security lead taught me how to properly strace this.

This is the log of the first execution of jsipfs --help, which works:
paste.ubuntu.com/26509187/

This is the log of the second execution, which gets stuck:
https://paste.ubuntu.com/26509173/

PR #1105, opened 2017-11-23
fix SIG tests

dryajov:
Tests are failing due to timeouts. These tests are awfully slow, we might want to consider skipping them on normal runs?

### DIFF
--- a/test/cli/daemon.js
+++ b/test/cli/daemon.js
@@ -71,9 +71,8 @@ describe('daemon', () => {
 
   afterEach(() => clean(repoPath))
 
-  // TODO: test fails
-  it.skip('do not crash if Addresses.Swarm is empty', function (done) {
-    this.timeout(20 * 1000)
+  it('do not crash if Addresses.Swarm is empty', function (done) {
+    this.timeout(100 * 1000)
 
     ipfs('init').then(() => {
       return ipfs('config', 'Addresses', JSON.stringify({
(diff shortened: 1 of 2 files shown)


test/cli/daemon.js
```diff
@@ -88,18 +87,16 @@ describe('daemon', () => {
...
     testSignal(ipfs, 'SIGINT').then(() => {
       checkLock(repoPath, done)
     }).catch(done)
   })
 
-  // TODO: test fails
-  it.skip('should handle SIGTERM gracefully', function (done) {
-    this.timeout(20 * 1000)
+  it('should handle SIGTERM gracefully', function (done) {
+    this.timeout(100 * 1000)
```
diasdavid: Why does this take so much time?

diasdavid:
Please rebase master onto this branch

test/cli/daemon.js
dryajov: It needs to start and init the repo for the IPFS node each time, which seems to be slow.

diasdavid: The other tests take 40s or less, why 100s in this one?

dryajov: I wasn't able to get it to work with anything below 60 on my box, 100 seems to be the most stable.

I'm guessing that it is because it has to:
- start the daemon
- init the repo
- send the signal and wait for daemon to clean up and exit

The 40 (in i.e. block tests) https://github.com/ipfs/js-ipfs/blob/ac95601fe88a25299dfb1eb00d4c8688f0a4af0e/test/cli/block.js#L15, are coupled with the 30 https://github.com/ipfs/js-ipfs/blob/104ef1ef6cc64c09ec886f67c28e9b5d37bc9e66/test/utils/on-and-off.js#L20 from the `before` hook that gets the daemon started and initialized.

So I think it currently is more like 70 + ~30 for graceful shutdown. I think we can tweak the 30 a little bit.

dryajov:
done rebasing.

test/cli/daemon.js
diasdavid: ok

diasdavid:
Some tests are failing but not due to this PR.

PR #1109, opened 2017-11-25
Make Travis Happy

diasdavid:


diasdavid:
Now I think Travis is just screwing with me:

image

PR #1117, opened 2017-11-28
fix: update *-star multiaddrs to explicity say that they need tcp and a port

diasdavid:
Depends on:

- [x] https://github.com/libp2p/js-libp2p-webrtc-star/issues/129
- [x] https://github.com/libp2p/js-libp2p-websocket-star/issues/37


diasdavid:
@pgte something weird is happening with the latest p2p-websocket-star, the example exchange files never starts the node (it was before).

pgte:
@diasdavid checking it out.

pgte:
@diasdavid it's yielding this error for me:

ipfs_-_transfer_files

pgte:
Will push a fix for this.

pgte:
@diasdavid pushed temporary fix for the latest problem, allowing websocket-star join with disabled challenge. Without that, for now, websocket-star doesn't seam to work..

mkg20001:
@diasdavid When the crypto challenge is disable strict multiaddr is activated. Maybe this is causing the issue?

diasdavid:
We want the crypto challenged enabled. seems that the version deployed is the wrong one. //cc @VictorBjelkholm

hacdias:
This also fixes the problem we were having when getting a file. @diasdavid and I were looking into that and David found out that the [Blob API](https://developer.mozilla.org/en-US/docs/Web/API/Blob) has changed. 😄 It works now!

Issue #1075, opened 2017-11-14
Tests on Windows


richardschneider:
The path `/tmp/...` does not exist on windows. Use `path.join(os.tmpdir(), '...')`.

See some [possible bad tests](https://github.com/ipfs/js-ipfs/search?q=tmp&type=Code&utf8=%E2%9C%93)

diasdavid:
@richardschneider Now that you pushed Windows support, do you want to share an update here on your plans to fix the tests to make sure Appveyor is happy?

richardschneider:
First, I'll change all tests not to use `/tmp/...`.

Next change is harder.  It seems that NPM likes to randomly throw EPERM every now and then.  I've even seen it once or twice on my local machine.

According to https://github.com/Medium/phantomjs/issues/19 it may be Windows Anti-Virus software; which I consider a virus in itself.

richardschneider:
Also https://github.com/npm/npm/issues/18380#issuecomment-326813399 suggests downgrading  NPM

richardschneider:
@VictorBjelkholm have you seen any of these EPERM errors on Jenkins Windows build?

VictorBjelkholm:
@richardschneider yes indeed I've seen those errors that happened every once in a while. The only reliable solution I've found is to turn off real time protection in Windows Defender. 

In Powershell, you can disable it like this:

https://github.com/ipfs/jenkins/blob/a73cb56893a80427ce0e87601ef1dcb8326222f0/worker/main.tf#L220

VictorBjelkholm:
For the record, ended up going with Yarn as it handles concurrent installs better than npm. But, got those errors with Yarn as well, unless turning off the real-time protection.

diasdavid:
Lot's of work happening on https://github.com/ipfs/js-ipfs/pull/1103

PR #1126, opened 2017-12-04
test: use os.tmpdir()

richardschneider:
See https://github.com/ipfs/js-ipfs/issues/1075

### DIFF
--- a/test/cli/daemon.js
+++ b/test/cli/daemon.js
@@ -8,6 +8,7 @@ const pull = require('pull-stream')
 const toPull = require('stream-to-pull-stream')
 const os = require('os')
 const path = require('path')
+const hat = require('hat')
 const fs = require('fs')
 
 const isWindows = os.platform() === 'win32'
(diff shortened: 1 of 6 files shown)


test/cli/daemon.js
```diff
@@ -65,7 +65,7 @@ describe('daemon', () => {
   let ipfs
 
   beforeEach(() => {
-    repoPath = '/tmp/ipfs-test-not-found-' + Math.random().toString().substring(2, 8)
+    repoPath = path.join(os.tmpdir(), 'ipfs-test-not-found-' + Math.random().toString().substring(2, 8))
```
diasdavid: If you don't mind here as well (ref https://github.com/ipfs/ipfs-performance-profiling/pull/12#discussion_r154862400)

test/core/create-node.spec.js
```diff
@@ -19,7 +21,7 @@ describe('create node', function () {
     this.timeout(80 * 1000)
 
     const node = new IPFS({
-      repo: '/tmp/ipfs-repo-' + Math.random(),
+      repo: path.join(os.tmpdir(), 'ipfs-repo-' + Math.random()),
```
diasdavid: one more (ref: https://github.com/ipfs/ipfs-performance-profiling/pull/12#discussion_r154862400)

richardschneider: @diasdavid Good idea.  I don't like seeing Math.random, because I always worry if a cryptographically secure P/RNG should be used.

diasdavid:
image


Seems like CI uses windows-386 and currently there is no go-ipfs build available due to: https://github.com/ipfs/distributions/issues/165

richardschneider:
@diasdavid Yes, 386 builds are failing, I added https://github.com/ipfs/js-ipfs/issues/1127

Issue #1145, opened 2017-12-12
Should specify which link didn't exist on ipfs.cat


daviddias:
https://github.com/ipfs/js-ipfs/blob/master/src/core/components/files.js#L140 should report the same error as js-ipfs-api. 

Example: https://github.com/ipfs/interface-ipfs-core/pull/183#discussion_r156408064

//cc @pgte

diasdavid:
The HTTP API didn't get updated

```
  1) HTTP API
       ## interface-ipfs-core over ipfs-api
         .files
           .cat
             errors on unknown path:
     AssertionError: expected 'Failed to cat file: Error: No such file' to be one of [ Array(2) ]
      at ipfs.files.cat.catch (node_modules/interface-ipfs-core/src/files.js:439:36)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

youngnicks:
I'm new to ipfs, but am willing to give this a shot. It looks like it is just changing the error message in **src/core/components/files.js** from 'No such file' to `no link named "${fileName}" under ${path}`. Is this correct?

youngnicks:
If the above is true, should the `ipfs/interface-ipfs-core` be updated to remove `No such file` as a valid scenario? The test also fails due to the test looking for the error message thrown in files.cat, however, `Failed to cat file:` is prepended to the error message. I see other tests use `expect(err.message).to.contain` instead of `expect(err.message).to.oneOf`.

```javascript
expect(err.message).to.contain(
  'Error: no link named "does-not-exist" under Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
)
```

niinpatel:
Is this still open? Can I look into this?

achingbrain:
Closing as this issue is very stale.

PR #1146, opened 2017-12-13
fix: fix bug introduced by 1143

diasdavid:


VictorBjelkholm:
No tests? :'(

diasdavid:
The tests are on https://github.com/ipfs/interface-ipfs-core/blob/master/src/files.js and are incorrect, tracked by this issue: https://github.com/ipfs/js-ipfs/issues/1145

PR #1151, opened 2017-12-15
Fix: use new bitswap stats

pgte:
Adapts to the new Bitswap stats API.


diasdavid:
@pgte I believe you forgot to update the bitswap dependency to latest.

PR #1155, opened 2017-12-18
fix(files.add): directory with odd name

richardschneider:
`glob` was interpreting the '[' as a special magic character.  According to its documentation, a caller should be able to escape it with `\`; but it does not work.

See the [glob issue](https://github.com/isaacs/node-glob/issues/277).


richardschneider:
@diasdavid typical browser failure on other tests on travis.  Circle CI passes everything.

richardschneider:
@diasdavid can you send me the git commands for `git rebasing` a branch.

PR #1103, opened 2017-11-22
Updating CI files

VictorBjelkholm:
This commit updates all CI scripts to the latest version

PR #1159, opened 2017-12-19
cli files on Windows

richardschneider:
Path separator issue on Windows.  This should make #1103 pass.


richardschneider:
@diasdavid RTM

PR #1181, opened 2018-01-22
feat: `ipfs version` flags + `ipfs repo version`

JonKrone:
Hi all,

First time looking at any of the ipfs codebases, wanted to contribute a little thing so I could feel my way around it.

- [X] add `api/v0/repo/version` http api
- [X] implement `ipfs repo version`
- [X] implement the `--number`, `--repo`, and `--all` flags for the `ipfs version` command
- [ ] implement the `--commit` flag to `ipfs version`
- [X] tests for above work

[**related js-ipfs-api PR**](https://github.com/ipfs/js-ipfs-api/pull/676)

Modeled `ipfs version` behavior off of [go-ipfs](https://github.com/ipfs/go-ipfs/blob/9db5471e72a3701c8708b9b286cd4c41996a8bd1/core/commands/version.go#L30).

The `--commit` flag is parsed but doesn't do much because I don't know how to get the commit hash of the current ipfs version. [In go-ipfs](https://github.com/ipfs/go-ipfs/blob/master/repo/config/version.go), it's made available at project build-time but I didn't see something like that for us. Don't think implementing it is necessary for this PR.


JonKrone:
CI is failing because this change requires the `/repo/version` API from [ipfs/js-ipfs-api#676](https://github.com/ipfs/js-ipfs-api/pull/676/files#diff-9314bebf539a650be3efc61e30b60697).

How do y'all usually get around this?

diasdavid:
Thanks for contributing and welcome to the project! Let's get the one in js-ipfs-api merged first and then come back to this one :)

diasdavid:
@JonKrone js-ipfs-api was released with your patch, you can update it as a dep of this PR so that tests pass :)

JonKrone:
Looks like there's a failing test related to interface-ipfs-core tests of pubsub (https://travis-ci.org/ipfs/js-ipfs/jobs/332979384#L4315) and a set of fails from `make test` and sharness. 

I don't quite have time to look into sharness today though it looks like updating expected output. I noticed in #1103 that the `make test` command is being removed from Travis CI; is that the plan?

diasdavid:
@JonKrone merging into a branch on the main repo so that I can fix sharding :)

src/core/components/version.js
```diff
@@ -3,17 +3,24 @@
...
+    self.repo.version((err, repoVersion) => {
+      if (err) {
+        throw err
+      }
+
+      callback(null, {
+        version: pkg.version,
+        repo: repoVersion,
+        commit: ''
+      })
```
diasdavid: @JonKrone found the issue. Here you are assuming that IPFS has always been init when it is not the case always. the Sharness test is checking the version with ipfs offline and without init.

JonKrone: Thanks! In order to avoid the init constraint, go-ipfs accesses the repo version directly from a constant exposed by [go-ipfs/repo](https://github.com/ipfs/go-ipfs/blob/master/core/commands/version.go#L40) so the repo version is available offline and without an initialized repo. Analogous to `const repoVersion = require('ipfs-repo').version`

Unfortunately, there's no similar field exposed by the `js-ipfs-repo` package. We could read the [`js-ipfs-repo/constants`](https://github.com/ipfs/js-ipfs-repo/blob/master/src/constants.js) file directly like we're doing with `pkg.version` above or add the version as an export of `js-ipfs-repo`.

If we need to be able to read the version w/o init, I think an export is the way to go and replace the dependence of [`repo.version`](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/repo.js#L10) on an initialized `_repo` with just `require('ipfs-repo').version`. Should we then add this version constant to the repo SPEC?

Feels like a big change to not use the repo API, what are your thoughts?

diasdavid: We can totally expose the constant, however, I do not think is the best approach. If we use the version from the module, then we might have a false positive where the module is signaling the latest repo version than the one that is inited.

I feel that the right way to do it is:

- if repo inited, then read the `version` from the repo
- if the repo is not inited, then read the `version` from the constant

JonKrone: Cool, I agree! I had also looked into what happens when the version is outdated but got a little caught up because starting the daemon with an outdated `~./jsipfs/version` throws during startup, making me think that we just bailed on mismatches. test: `echo "5" > ~/.jsipfs/version && jsipfs --help`

If that's expected behavior, I can't think of a case where you can run a cli command on an initialized but outdated repo, and we could safely report the constant `version`.

I think you have the right of it, though, and should aim for reporting the version of `~/.jsipfs/version` if the repo is initialized and maybe at some point in the future, change startup so that we delay opening the repo connection or make accessing the version independent of opening a repo?

either way: https://github.com/ipfs/js-ipfs-repo/pull/158

PR #1184, opened 2018-01-23
fix(cli): show help for subcommands

vmx:
If subcommands are called, e.g. `jsipfs dag` there was no output
at all. Now the help message (which you also get with `--help`
is displayed.


diasdavid:
nice! Thanks @vmx :)

PR #1186, opened 2018-01-23
chore: change from pre-commit to pre-push hooks

vmx:
Instead of running the linting and tests on pre-commit, do it
on pre-push. That makes it less frequent (and annoying) while
doing local development.


hacdias:
Thumbs up for this!!!

PR #1198, opened 2018-01-29
feat: stats API (stats.bitswap and stats.repo)

### LINKED ISSUE #1116: stats: document and exemplify stats
[Bitswap now has the capability](https://github.com/ipfs/js-ipfs-bitswap/pull/158) of producing some interesting stats, and also some moving averages.

It should be better documented / applied to the interface spec.
Also, I think we should add some stats to the existing examples or create new ones.

### LINKED ISSUE #1197: Implement Stats API
Track on #1198

hacdias:
Right now, `js-ipfs` doesn't implement any of the [STATS spec](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/STATS.md) and it is required to make IPFS Desktop work with js-ipfs.

## Prerequisites and related PRs

- [x] On **interface-ipfs-core**, merge (https://github.com/ipfs/interface-ipfs-core/pull/210):
	- [x] Move `stats.bitswap` to `bitswap.stat` and link `stats.bitswap` to `bitswap.stat`
	- [x] Move `stats.repo` to `repo.stat` and link `stats.repo` to `repo.stat`
- [x] On **js-ipfs-api** (https://github.com/ipfs/js-ipfs-api/pull/684)
	- [x] Make `bitswap.stat` **equal to** `stats.bitswap` 
- [x] On **js-ipfs-repo** (https://github.com/ipfs/js-ipfs-repo/pull/159):
	- [x] Make `stat` command.
- [x] Finish Stats SPEC (https://github.com/ipfs/interface-ipfs-core/pull/211)
	- [x] Actually finish
	- [x] Merge
	- [x] Replace on `package.json`
- [x] On **js-ipfs-api**:
	- [x] https://github.com/ipfs/js-ipfs-api/pull/686
	- [x] Release a new version and replace it on `package.json`
- [ ] Update the tests on **interface-ipfs-core** to run on `js-ipfs`

## ToDos

- [x] `bitswap.stat` must implement [the SPEC](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/STATS.md#bitswap)
	- [x] HTTP API
	- [x] Core
	- [x] CLI
- [x] `stats.bitswap` just points to `bitswap.stat`
	- [x] HTTP API
	- [x] Core
	- [x] CLI
- [x] `repo.stat` must implement the [SPEC](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/STATS.md#repo)
	- [x] HTTP API
	- [x] Core
	- [x] CLI
- [x] `stats.repo` just points to `repo.stat`
	- [x] HTTP API
	- [x] Core
	- [x] CLI

## Not Fundamental, but also done in this PR

- [x] `/api/v0/repo/version`
- [x] `ipfs repo version`
- [x] `ipfs.repo.version()`

Closes #1197
Closes #1116


hacdias:
Now I need to develop some kind of bandwidth metrics and I already know it is related to `libp2p`. I saw that with `go-ipfs` there is this [library](https://github.com/libp2p/go-libp2p-metrics) that was developed to take out metrics.

I was wondering if any of you (@diasdavid @dignifiedquire @pgte) could point me to the right direction since you are really into this project 😄

pgte:
Regarding bandwidth, you'll have to look into js-ipfs.bitswap.
There are two metrics: `dataReceived` and `dataSent` (bytes). But this only gives you a cumulative measure.

If you want per second, you can use a moving average. You have moving averages for 1, 5 and 15 minutes (this is configurable if you want to add a new interval).

There are no API docs (there is an issue about this somewhere), but here are the tests for this functionality: https://github.com/ipfs/js-ipfs-bitswap/blob/feat/benchmarks/test/bitswap-stats.js#L137-L145

Please tell me if you need more pointers.

hacdias:
Thanks @pgte! 

Another question: when we have [this](https://github.com/ipfs/js-ipfs-api/blob/master/src/stats/bw.js#L7-L18) on the API how should we do the response on the HTTP API? I tried this:

```
    const stream = Readable()
    reply(stream)

    stream.push(Buffer.from(JSON.stringify({
      TotalIn: stat.totalIn,
      TotalOut: stat.totalOut,
      RateIn: stat.rateIn,
      RateOut: stat.rateOut,
    })))
``` 

but it didn't work

hacdias:
@pgte could you take a look [here](https://github.com/ipfs/js-ipfs/pull/1198/files#diff-3c4270ffa4fb79adeb31c248f2d9d9f2)? It seems to always print 0 for every field. And how should I get the stats for specific peers and protocols?

Ref.: [Go implementation](https://github.com/libp2p/go-libp2p-metrics/blob/master/bw_stats.go)

pgte:
What is the client expecting? Is it a new-line delimited JSON stream?
If so, you will need to also output “\n” after each entry.

hacdias:
@pgte after digging into the source code I noticed we are using two funny special header `x-stream-output` and `x-chunked-output`. Didn't notice 😄

pgte:
@hacdias are you still getting problems with the stats?

Regarding peer-specific stats: currently, js-ipfs-bitswap doesn't keep stats per peer.
It should be easy to add it. I'd be extra careful in garbage collection (removing stats for peers that are no longer relevant to us)..

hacdias:
@pgte the stats always have zero values:

```javascript
bw: promisify((options, callback) => {
  if (typeof options === 'function') {
    callback = options
    options = {}
  }

  options = Object.assign({
    peer: '',
    proto: '',
    poll: false,
    interval: '1s'
  }, options)

  let stats

  if (options.peer) {
    // TODO: stats for a specific peer
  } else if (options.proto) {
    // TODO: stats for a specific proto
  } else {
    const stat = self._bitswap.stat()
    const snapshot = stat.snapshot
    const movingAverages = stat.movingAverages

    stats = {
      totalIn: snapshot.dataReceived,
      totalOut: snapshot.dataSent,
      rateIn: movingAverages.dataReceived[60000].movingAverage() / 60, // Bytes per second
      rateOut: movingAverages.dataSent[60000].movingAverage() / 60 // Bytes per second
    }
  }

  callback(null, stats)
})
```

That code always gives me a `0` on each key.

pgte:
@hacdias this may be a timing issue.
Stats updates are throttled to, by default, a max of 1 seconds. This is configurable, by the way, but the stats are always updated in bulk to minimize interference with particular operations.

I don't know if it helps, but, like [exemplified in the test](https://github.com/ipfs/js-ipfs-bitswap/blob/feat/benchmarks/test/bitswap-stats.js#L126), the stat object emits an `update` event once they're updated, so maybe you can subscribe to that before replying? I suggest making some experiments with that..

pgte:
and, of course, opening an issue on js-ipfs-bitswap (preferably with a reproducible problem if you can :) ) if you're not able to get values.

hacdias:
@pgte I just added this:

image

to the place where bitswap is initialized. It seems that it never gets called (at least waiting 10 mins). I'm running `js-ipfs` directly with `node src/cli/bin.js` daemom.

pgte:
@hacdias I think you'll have to [enable stats like this](https://github.com/ipfs/js-ipfs-bitswap/blob/feat/benchmarks/test/bitswap-stats.js#L89).

pgte:
... which I think may have to become a default setting in js-ipfs.
Or you could change the default in js-ipfs-bitswap if you don't want to add to the config clutter.

pgte:
I will also be needing per-peer stats because of https://github.com/ipfs/dynamic-data-and-capabilities/issues/3#issuecomment-361919002.

I opened bitswap issue here: https://github.com/ipfs/js-ipfs-bitswap/issues/165

@hacdias could you state your requirements there?

pgte:
@hacdias any luck enabling stats?

hacdias:
@pgte not really, they are always 0.

pgte:
@hacdias are the stats tests passing?:

```
$ mocha test/bitswap-stats.js
```

They are for me, and they test for > 0 values.

(also, keep in mind that values are Big: https://github.com/MikeMcl/big.js#readme

pgte:
Took a stab at documenting current stats API:
https://github.com/ipfs/js-ipfs-bitswap/pull/166/commits/4b27d3647465937641180ac251f7db3667017982

Hope this + tests help.

hacdias:
@diasdavid this just needs https://github.com/ipfs/interface-ipfs-core/pull/216 to be merged and published and a new version of `js-ipfs-api` with the latest changes. Otherwise, this is done. The `stats.bw` will be done in a separate PR.

diasdavid:
A new version of js-ipfs-api is blocked on https://github.com/ipfs/js-ipfs-api/issues/683 + https://github.com/ipfs/js-ipfs-api/pull/688

hacdias:
gotcha!

diasdavid:
@hacdias js-ipfs-api was released. You should be able to finish this PR now :)

hacdias:
@diasdavid there is still one thing missing: we need to merge https://github.com/ipfs/interface-ipfs-core/pull/216 in order to finish this PR.

diasdavid:
merging https://github.com/ipfs/js-ipfs/pull/1216 first and then back at this one :)

hacdias:
OKay, as soon as you merge it, I'll finish this PR, but I'll still need https://github.com/ipfs/interface-ipfs-core/pull/216 to be merged.

diasdavid:
@hacdias do it like this:

1. Rebase master onto this PR
2. Run tests locally and guarantee that everything passes
3. Ping me with everything ready so that I can merge and release and update the dep here in one go

README.md
```diff
@@ -256,11 +256,6 @@ A complete API definition is in the works. Meanwhile, you can learn how to you u
   - [`ipfs.block.put(block, cid, [callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/SPEC/BLOCK.md#put)
   - [`ipfs.block.stat(cid, [callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/SPEC/BLOCK.md#stat)
 
-- [repo](https://github.com/ipfs/interface-ipfs-core/tree/master/SPEC/)
-  - `ipfs.repo.init`
-  - `ipfs.repo.version`
-  - `ipfs.repo.gc` (not implemented, yet!)
```
diasdavid: Why move these to the Graph section??

diasdavid: I would agree if moved to the Node Management Section

src/cli/commands/bitswap/stat.js
```diff
@@ -16,22 +16,22 @@ module.exports = {
...
+  blocks received: ${stats.blocksReceived}
+  dup blocks received: ${stats.dupBlksReceived}
+  dup data received: ${stats.dupDataReceived}B
+  wantlist [${stats.wantlist.length} keys]
+    ${stats.wantlist.join('\n    ')}
+  partners [${stats.peers.length}]
+    ${stats.peers.join('\n    ')}`)
     })
   }
 }
```
diasdavid: After Rebase, enable the CLI Bitswap test `stat`

src/core/components/bitswap.js
```diff
@@ -16,16 +18,27 @@ module.exports = function bitswap (self) {
       const list = self._bitswap.getWantlist()
       return formatWantlist(list)
     },
-    stat: () => {
+
+    stat: promisify((callback) => {
       if (!self.isOnline()) {
-        throw new Error(OFFLINE_ERROR)
+        callback(new Error(OFFLINE_ERROR))
```
diasdavid: This needs two things:

- Be wrapped with a setImmediate from the `async` module
- Have a `return`, otherwise the rest of the function will be executed.

src/core/components/start.js
```diff
@@ -45,7 +45,9 @@ module.exports = (self) => {
       self._bitswap = new Bitswap(
         self._libp2pNode,
         self._repo.blocks,
-        self._peerInfoBook
+        Object.assign({
+          statsEnabled: true
+        }, self._peerInfoBook)
```
diasdavid: This is weird. Are you saying that in order to enable stats on Bitswap there is a variable on PeerBook that needs to be true? This Object Assign is also probably messing with the fact that PeerBook is a Object from Class and not an Object Literal. //cc @pgte

README.md
hacdias: It's on the Node Management section

diasdavid: image

diasdavid: nvm, I got fooled twice by GH interface. all good :)

src/core/components/start.js
hacdias: We need to have the stats enabled and this was the best way I found to do that.

src/cli/commands/bitswap/stat.js
hacdias: Done and it works!!

src/core/components/start.js
diasdavid: @pgte could you comment here given that you implemented stats on Bitswap

hacdias:
@diasdavid the related tests to this PR seem to be working here 😄

src/core/components/start.js
diasdavid: I see. the thing is that bitswap doesn't need access to the PeerInfoBook at all. You can just pass the object options.

Issue #1192, opened 2018-01-26
[proposal] Allow graceful programmatic shutdown


dryajov:
It's currently not possible to programmatically do a graceful shutdown of ipfs daemons on windows through the use of signals, the way it's done on Unix like systems - due to windows lack of signal support. From Node.js documentation - https://nodejs.org/api/process.html#process_signal_events. 

This is an issue for applications that try to interact with the daemons programmatically such as `ipfsd-ctl`, since sending a `SIGINT` to any process will not trigger a signal handler, but simply kill the process, which in turn leaves the repo locked and possibly in an inconsistent state.

One possible solution would be to add some sort of explicit shutdown command (`ipfsd shutdown`) - that would shutdown the daemon. So far, I think this can be implemented either as an additional HTTP endpoint or by the presence of a 0 byte file at the root of the repo:

i.e.

- `/api/v0/shutdown` resource - when calling the endpoint, the daemon responds with a `200` or `202` HTTP response code and beggins the shutdown of the daemon
- `.jsipfs/shutdown` file - the daemon monitors the repo for the presence of this file, if its there, it will delete the file and initiate the daemon shutdown

This is just a suggestion, and if there are better/cleaner ways of achieving this, I'd love to hear about them. 

Note that anything we do here, would also have to be done in the go implementation, since it affects both daemon types - // cc - @whyrusleeping @lgierth @Stebalien

richardschneider:
@dryajov The [docs](https://ipfs.io/docs/api/#apiv0shutdown) already mention the `shutdown` endpoint. 

IMHO, `js-ipfs` should implement it and then `ipfsd-ctl` can just call it, instead of using *nix signals.

dryajov:
@richardschneider thanks for finding that, however I don't see this endpoint anywhere in `js-ipfs`, perhaps it is implemented in go only - in which case we should simply implement it in js.

mkg20001:
Seems like on win10 `SIGINT` suddenly started to "just work" https://stackoverflow.com/a/45369603/3990041

dryajov:
@mkg20001 hehe - yes, node correctly emulates `SIGINT` when using the cmd.

From the docs:

`
SIGINT from the terminal is supported on all platforms, and can usually be generated with <Ctrl>+C (though this may be configurable). It is not generated when terminal raw mode is enabled.
`

It's when using `process.kill('SIGINT')` that the signal emulation degrades to killing the process immediately.

PR #1200, opened 2018-01-30
feat: ipfs shutdown

richardschneider:
See issue https://github.com/ipfs/js-ipfs/issues/1192


package.json
```diff
@@ -121,7 +121,7 @@
     "libp2p-circuit": "~0.1.4",
     "libp2p-floodsub": "~0.13.1",
     "libp2p-kad-dht": "~0.6.0",
-    "libp2p-mdns": "~0.9.1",
+    "libp2p-mdns": "^0.9.2",
```
diasdavid: keep the use of `~`

richardschneider: oops, that was `npm install` and poor review by me.  Will fix.

richardschneider:
Will rebase tomorrow.

diasdavid:
@richardschneider since this exists for go and for js ipfs, can you make tests on interface-ipfs-core and document the API call?

richardschneider:
Did documentation, from IRC
> <makaretu> daviddias: accidentally did commit to `master` https://github.com/ipfs/interface-ipfs-core/commit/9d91267ab7ab61e41c84eaa81c647cd630d83797
9:50 AM Sorry, it's no biggi just documentation update

README.md
```diff
@@ -325,6 +325,7 @@ A complete API definition is in the works. Meanwhile, you can learn how to you u
 - [miscellaneous operations](https://github.com/ipfs/interface-ipfs-core/tree/master/SPEC/MISCELLANEOUS.md)
   - [`ipfs.id([callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/SPEC/MISCELLANEOUS.md#id)
   - [`ipfs.version([callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/SPEC/MISCELLANEOUS.md#version)
+  - [`ipfs.shutdown([callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/SPEC/MISCELLANEOUS.md#shutdown)
```
diasdavid: I'm thinking more about this. `shutdown` is a c&p of `stop`. Can we use just stop instead?

diasdavid: https://github.com/ipfs/interface-ipfs-core/pull/215

richardschneider: Yes we could change it to `stop`.  It's not a c&p, it's even simpler: `this.shutdown = this.stop`.

However, the `/api/v0/shutdown` end-point is already defined by `go`, so we must support it.  

Also what do we do about the CLI; `jsipfs shutdown` or `jsipfs stop` or both?

diasdavid: @richardschneider let's do the `this.shutdown = this.stop` for now and keep the CLI as just shutdown.

Note, the HTTP API doesn't have to match entirely to the Core API which doesn't have to match the CLI API. It should enable the users to do the same operations, but we do have to have into account that they are used in different contexts. We will need to revisit all of the APIs later.

richardschneider: @diasdavid I'm very confused. Is this PR now okay?  What about https://github.com/ipfs/js-ipfs-api/pull/685?

diasdavid: Almost.

The spec should be `ipfs.stop` as it is on https://github.com/ipfs/interface-ipfs-core/pull/215/files and we can have an alias for `ipfs.shutdown` in both js-ipfs and js-ipfs-api just for the sake of convenience.

Similar to how we do alias on https://github.com/ipfs/js-ipfs-api/#files

richardschneider:
@diasdavid This is RTM.  Failures are in other tests.

diasdavid:
@RichardLitt Circle CI disagrees

The error comes from the code that changed here. However, why are MFS tests running against js-ipfs, @hacdias ?

image

RichardLitt:
Ping @richardschneider

richardschneider:
@RichardLitt you can email on makaretu (at) gmail

RichardLitt:
I, uh, don't need to. @diasdavid pinged the wrong richard. :)

Issue #928, opened 2017-07-23
Implement the `jsipfs ping` command


diasdavid:
Following: https://github.com/ipfs/js-ipfs/pull/925/files

We want to support the same command that go-ipfs supports: `jsipfs ping`.

From go-ipfs:
```
» ipfs ping --help
USAGE
  ipfs ping <peer ID>... - Send echo request packets to IPFS hosts.

SYNOPSIS
  ipfs ping [--count=<count> | -n] [--] <peer ID>...

ARGUMENTS

  <peer ID>... - ID of peer to be pinged.

OPTIONS

  -n, --count int - Number of ping messages to send. Default: 10.

DESCRIPTION

  'ipfs ping' is a tool to test sending data to other nodes. It finds nodes
  via the routing system, sends pings, waits for pongs, and prints out round-
  trip latency information.
```

trendsetter37:
@diasdavid I can take a stab at this

diasdavid:
Thank you @trendsetter37 🌟

melvin0008:
Hey @diasdavid,

Thanks a lot for the IPFS project. I really enjoyed reading about the project and have been blown away with the vision <3.
I would like to start contributing to this project.
  
I gave this issue a shot and got the ping implementing but had few implementation detail questions

1. In short I called the ping exposed by [libp2p](https://github.com/libp2p/js-libp2p#libp2ppingpeer--options-callback) . However I noticed although the documentation says it takes in an option. The underlying [implementation](https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L255) doesn't take in options. So would you like me to change the underlying implementation in and help it take options (--count)

2.  I would like to update the curent API for [ipfs.ping()](https://github.com/ipfs/js-ipfs#node-management) from not taking anything to take in key, option and callback. I see that we have two folders one called core and other commands. Since commands is where we would be logging the ping times and average latency. What is expected from the ipfs.ping() in core to return?
List of times for all ping operations?

Once, these doubts are cleared I would like to submit a PR if needed.

Thanks :)

diasdavid:
> Thanks a lot for the IPFS project. I really enjoyed reading about the project and have been blown away with the vision <3.

Thank you ❤️❤️❤️

> I would like to start contributing to this project.

YEAAS, Welcome! :)

> 1. In short I called the ping exposed by libp2p ...

You should not have to change the implementation. The implementation gives you a event emitter and a stop call. You can get the `--count` to function with just that.

[...]

PR #1202, opened 2018-02-02
feat: implement `ipfs ping` flags #928

melvin0008:
Attempt to solve #928 

- [ ] Add Tests
- [ ] Look for a better way to use the ping object to reduce HTTP calls
- [ ] Add fake delay between ping 
- [ ] If none of the peers are initialized it gives Circuit is not enabled error.
- [ ] Better error handling 

@diasdavid Let me know your thoughts on the same

### DIFF
--- a/src/cli/commands/ping.js
+++ b/src/cli/commands/ping.js
@@ -1,0 +1,44 @@
+'use strict'
+
+const print = require('../utils').print
+
+module.exports = {
+  command: 'ping <peerId>',
+
+  describe: 'Measure the latency of a connection',
+
+  builder: {
+    count: {
+      alias: 'n',
+      type: 'integer',
+      default: 10
+    }
+  },
+
+  handler (argv) {
+    const peerId = argv.peerId
+    const count = argv.count || 10
+
+    print('PING ' + peerId)
+
+    let noOfTimes = 0
+    let totalTime = 0
+
+    const pingCb = (err, p) => {
+      if (err) {
+        throw err
+      }
+      let time = p.Time
+      totalTime = totalTime + time
+      noOfTimes = noOfTimes + 1
+      print('Pong received: time=' + time + ' ms')
+      if (noOfTimes === count) {
+        print('Average latency: ' + totalTime / count + 'ms')
+      }
+    }
+
+    for (let i = 0; i < count; i++) {
+      argv.ipfs.ping(peerId, pingCb)
+    }
+  }
+}
--- a/src/core/components/ping.js
+++ b/src/core/components/ping.js
@@ -1,9 +1,35 @@
 'use strict'
 
 const promisify = require('promisify-es6')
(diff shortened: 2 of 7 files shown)


src/cli/commands/ping.js
```diff
@@ -0,0 +1,44 @@
...
+      noOfTimes = noOfTimes + 1
+      print('Pong received: time=' + time + ' ms')
+      if (noOfTimes === count) {
+        print('Average latency: ' + totalTime / count + 'ms')
+      }
+    }
+
+    for (let i = 0; i < count; i++) {
+      argv.ipfs.ping(peerId, pingCb)
+    }
```
diasdavid: A ping with count 10 is a ping with 10 packets, not 10 ping calls. You need to count the number of pings that come in p.

diasdavid:
@melvin0008 still looking into finishing this PR?

melvin0008:
@diasdavid Yes will do it the coming weekend. Sorry for the wait

src/cli/commands/ping.js
melvin0008: I was able to use the ping object and stream the times to the api.
But I have run into a problem. In [js-ipfs-api](https://github.com/ipfs/js-ipfs-api/blob/master/src/ping.js#L27) the response assumes there will be one ping request. 

So even if I send I stream multiple ping responses only the first one is returned.

My understanding of the flow when someone types jsipfs ping <peerId> is
1.  cli/commands/ping.js -> this calls the ipfs-api ping
2. node_modules/ipfs-api/src/ping.js -> this calls the http api ping
3. http/api/resources/ping.js -> this calls the component ping
4. core/components/ping.js which would then return the ping object
5. http/api/resources/ping.js would then stream the times to the ipfs-api and end the stream when number of pings are complete.

The problem is [js-ipfs-api](https://github.com/ipfs/js-ipfs-api/blob/master/src/ping.js#L27) is expecting only one response ping object. Is there a workaround I'm missing or should I create an issue inn js-ipfs-api ping and fix that. I just want to confirm its an issue. :)


Also, irrespective of the api I was able to get the curl outputs for both the go-ipfs http api and js-ipfs http api to be the same . I have attached images for the same. But again I find one inconsistency . Since we are using new Date() in [js-libp2p-ping](https://github.com/libp2p/js-libp2p-ping/blob/master/src/ping.js#L42) we are not getting the same unit as go-ipfs. go-ipfs returns in nanosecond and js-ipfs returns the time in milliseconds.  Let me know whether I should create a issue and fix that. I would be glad to do it. I just want to confirm its an issue. :)

Curl response to js-ipfs
screen shot 2018-02-17 at 6 37 38 pm

Curl response to go-ipfs
screen shot 2018-02-17 at 6 36 28 pm

diasdavid:
Let's coalesce the energies here on https://github.com/ipfs/js-ipfs/pull/1299

PR #1215, opened 2018-02-14
feat: use PubSub API directly from libp2p

diasdavid:


diasdavid:
Done

diasdavid:
Had to fix some linting issues that were merged in Master (somehow escaped the check, shame on me for not triple checking.)

diasdavid:
image

CI is green \o/ Merging as this solves also some issues in Master that are preventing other PRs

src/core/components/bootstrap.js
```diff
@@ -1,9 +1,24 @@
...
 const defaultNodes = require('../runtime/config-nodejs.json').Bootstrap
-const MultiAddr = require('multiaddr')
+const Multiaddr = require('multiaddr')
 const promisify = require('promisify-es6')
 
+function isValid (ma) {
+  if (typeof ma === 'string') {
+    try {
+      ma = new Multiaddr(ma)
+      return Boolean(ma)
```
vmx: This idea comes from type checking discussion. It would be cool if `isValid()` functions would return `false` if invalid and return a valid object whatever got checked (in this case a `Multiaddr`).

That way you could also use `isValid()` to make sure you always get a valid `Multiaddr` object, no matter if it was already one or of if it was a string.

src/core/components/libp2p.js
```diff
@@ -50,9 +51,7 @@ module.exports = function libp2p (self) {
         })
 
         self._libp2pNode.start((err) => {
-          if (err) {
-            return callback(err)
-          }
+          if (err) { return callback(err) }
```
vmx: Nitpick: I prefer not having unrelated formatting changes together with real code changes.

test/core/bitswap.spec.js
```diff
@@ -16,10 +16,9 @@ const isNode = require('detect-node')
 const multihashing = require('multihashing-async')
 const CID = require('cids')
 
-const DaemonFactory = require('ipfsd-ctl')
-const df = DaemonFactory.create({ type: 'js' })
-
-const dfProc = DaemonFactory.create({ type: 'proc' })
+const IPFSFactory = require('ipfsd-ctl')
+const fDaemon = IPFSFactory.create({ type: 'js' })
```
vmx: I find `fDaemon` a strange, hard to read, uncommon name. What about just `Deamon` or `jsDeamon`.

test/core/bitswap.spec.js
```diff
@@ -16,10 +16,9 @@ const isNode = require('detect-node')
...
+const fInProc = IPFSFactory.create({ type: 'proc' })
```
vmx: I find `fInProc` a strange, hard to read, meaningless, uncommon name. What about just `InProc` or `InProcDeamon`.

vmx:
Oh, I just saw that it got already merged :) Well, then see my comments as FYI and fill free to do a new PR if you think it's worth it.

test/core/bitswap.spec.js
diasdavid: It's a Factory. I could call it factoryDaemon if that makes more sense for you

src/core/components/bootstrap.js
diasdavid: This got upgraded in https://github.com/ipfs/js-ipfs/pull/1227

test/core/bitswap.spec.js
diasdavid: Same as https://github.com/ipfs/js-ipfs/pull/1215#discussion_r169066092. I can use the full word if that works better :)

test/core/bitswap.spec.js
vmx: Does it matter that's a factory? If you really want to know, I think it's apparent from `IPFSFactor.create()` that it is a factory.
This don't find this super important, so if you don't find it worth the change, feel free to leave it as it currently is.

Issue #1129, opened 2017-12-05
pubsub.subscribe option parameter not optional in 0.27


kevinsimper:
According to the docs, option is optional, but it is required in 0.27 as it will try to call the callback as the third parameter.
https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/PUBSUB.md#pubsubsubscribe

If only two parameters were given, both callback and handler would be undefined

https://github.com/ipfs/js-ipfs/blob/bd264f27ba2f435c3ae3388c7503d6118627bd73/src/core/components/pubsub.js#L16

diasdavid:
@kevinsimper Wanna submit a test that shows that incorrect behavior?

achingbrain:
If two arguments are passed, `callback` and `handler` would indeed be undefined.  The block in question reassigns `options` to `handler` if `options` is a function and then assigns `handler` to `callback` (to cover the case where you've passed three arguments).  Later on we guard on `callback` and return a promise if it's falsey.

So if there are two args it should be ok as it'll just result in a promise being returned from the function call, which is what happens [in this test case](https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/pubsub.js#L166).

@kevinsimper have you observed any unexpected behaviour?

kevinsimper:
It has been fixed in the newest version of ipfs `0.27.7` and I just tested again and got the error still in `0.27.0`, so good 👍 

https://jsbin.com/zuvuxijuma/edit?html,js,output
```
var id = 'ipfs-repo-' + String(Math.random())
console.log({id})
const node = new Ipfs({
  repo: id,
  EXPERIMENTAL: {
    pubsub: true
  },
  config: {
    Addresses: {
      Swarm: [
        '/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star'
      ]
    }
  }
})

node.on('ready', function() {
  // Your node is now ready to use \o/
  console.log('ready')
  // stopping a node

  let topic = 'fruit-of-the-daykevin'

  let receiveMsg = function(msg) {
    console.log(msg.data.toString(), Date.now())
  }

  node.pubsub.subscribe(topic, receiveMsg)
  setInterval(function() {
    node.pubsub.peers(topic, (err, peerIds) => {
      if (err) {
        throw err
      }
      console.log(peerIds)
    })
  }, 1000)
  setInterval(function() {
[...]

PR #1217, opened 2018-02-14
fix: Adds tests for programmatic pubsub

achingbrain:
Attempts to address #1129.

There were no tests that invoked the subscribe method with only two arguments so I added some.  I also added tests for other pubsub methods.

I found some [existing pubsub tests](https://github.com/ipfs/js-ipfs/blob/master/test/cli/pubsub.js#L81-L97) but these go via the CLI and from what I can see the subscribe/publish test only uses one node.

Happy to make whatever changes are required, please let me know if anything needs changing to be more idiomatic for the project or if you'd like them added to the existing cli pubsub test suite instead.

### DIFF
--- a/test/core/pubsub.spec.js
+++ b/test/core/pubsub.spec.js
@@ -1,0 +1,210 @@
+/* eslint-env mocha */
+'use strict'
+
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const expect = chai.expect
+chai.use(dirtyChai)
+const parallel = require('async/parallel')
+const times = require('async/times')
+const waterfall = require('async/waterfall')
+const isNode = require('detect-node')
+const DaemonFactory = require('ipfsd-ctl')
+const df = DaemonFactory.create({ type: 'js' })
+
+function connectNodes (localNode, remoteNode, callback) {
+  remoteNode.id((err, peerInfo) => {
+    expect(err).to.not.exist()
+
+    localNode.swarm.connect(peerInfo.addresses[0], (err) => {
+      callback(err)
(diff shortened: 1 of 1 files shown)


diasdavid:
We have a larger batch of tests coming from our interface-ipfs-core repo. They get loaded here -- https://github.com/ipfs/js-ipfs/blob/master/test/core/interface/pubsub.js -- and they are run against js-ipfs and js-ipfs-api to ensure that both have the same interface.

Could you diff your tests to that batch and add the ones that are different to that set? This way we are sure that both implementations comply.

Thanks @achingbrain :)

test/core/pubsub.spec.js
```diff
@@ -0,0 +1,210 @@
...
+        (next) => remoteNode.pubsub.publish(topic, Buffer.from(message, 'utf8'), next)
+      ], (err) => {
+        expect(err).to.not.exist()
+      })
+    })
+
+    it('should subscribe without options', (done) => {
+      const handler = (event) => {
+        expect(event.data.toString('utf8')).to.equal(message)
+        done()
```
diasdavid: Should unsubscribe before done to avoid side effects.

achingbrain:
Ah, nice.  I think the invocation I wanted to test is covered [here](https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/pubsub.js#L166) so this PR is sort of redundant.

diasdavid:
@achingbrain do you confirm that #1129 is solved then?

achingbrain:
Yes, I think that issue should be closed.

PR #1222, opened 2018-02-16
Recursive list directory

JonKrone:
Adds an -r flag to `ipfs ls`.

image

This is just a minor change with a little bit of cleanup. Moves the ls cli tests to their own suite and prints a deprecation notice on `ipfs file ls`.

I thought there might be useful information to show under an option like `ls -l` but I don't know of any information available that's important enough. I was at various times a little confused by the three ls commands, `file ls` `files ls` and `ls`, and how they differed and inherited so I hope I've consolidated them into a single impl.

Requires a small change to `js-ipfs-api`. Will post a ref when I've made that PR.

related issue: https://github.com/ipfs/interface-ipfs-core/issues/92


src/cli/commands/file/ls.js
```diff
@@ -11,6 +11,8 @@ module.exports = {
 
   handler (argv) {
     let path = argv.key
+    // `ipfs file ls` is deprecated. See https://ipfs.io/docs/commands/#ipfs-file-ls
+    print(`This functionality is deprecated, and will be removed in future versions. If possible, please use 'ipfs ls' instead.`)
```
JonKrone: We should be suggesting `files ls`, right? per https://github.com/ipfs/specs/issues/98#issuecomment-241261107
copied this from ipfs.io, will change.

diasdavid: Still just `ipfs ls` for now until both APIs are united. Currently, `ipfs ls !== ipfs files ls`

diasdavid:
@JonKrone could you rebase master onto this branch for a clean merge (CI green)?

src/cli/commands/file/ls.js
JonKrone: Yeah I was a bit confused here. Tracing code, I found that `ipfs ls` is an alias of `ipfs files ls` in js-ipfs/core here - https://github.com/ipfs/js-ipfs/blob/905bdc03edfba23d35aa6a5b8e356ab56f6f36b1/src/core/index.js#L118

There is also no `/files/ls` route handler in js-ipfs and a `jsipfs files ls` command with a daemon running triggers a js-ipfs-api '/ls' request. So there are a couple places where it seems `ipfs ls == ipfs files ls`. I'm sure there's something that I'm missing, let's go over it Wednesday.

diasdavid:
@JonKrone just released the js-ipfs-api update you needed. Once you update the deps here you should be all good :)

test/cli/ls.js
```diff
@@ -0,0 +1,92 @@
...
+          'QmdncfsVm2h5Kqq9hPmU7oAVX2zTSVP3L869tgTbPYnsha 1728 quick-start\n' +
+          'QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB 1102 readme\n' +
+          'QmTumTjvcYCAvRRwQ8sDRxh8ezmrcr88YFU7iYNroGGTBZ 1027 security-notes\n' +
+          'QmciSU8hfpAXKjvK5YLUSwApomGSWN5gFbP4EpDAEzu2Te 863  tour/\n'
+        )
+      })
+  })
+
+  it('recursively follows folders, -r', function () {
+    this.slow(2000)
```
diasdavid: @JonKrone what does this call do?

Issue #1053, opened 2017-10-28
Please add to browser example how to "put" Blobs with filename 


muksihs:
Please add to browser example how to  "put" several Javascript data Blobs with filename sequentially with a final hash that represents a folder containing all the Blobs put.
Thanks!

diasdavid:
@muksihs wanna try to give it a stab with the latest Files API? See docs here https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md

diasdavid:
@muksihs, @achingbrain just shipped an example with exactly what you were looking for :) Check it out at https://github.com/ipfs/js-ipfs/tree/master/examples/browser-add-readable-stream

PR #1223, opened 2018-02-16
docs: Add browser example for streaming files

achingbrain:
Addresses #1053 by adding an example of how to use the streaming interface to add files sequentially and obtain the hash of the directory containing the files.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -20,6 +20,7 @@ Let us know if you find any issue or if you want to contribute and add a new tut
 - [js-ipfs in the browser with WebPack](./browser-webpack)
 - [js-ipfs in the browser with a `<script>` tag](./browser-script-tag)
 - [js-ipfs in electron](./run-in-electron)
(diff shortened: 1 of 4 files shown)


diasdavid:
@achingbrain while at it, wanna try creating an example that shows how to use https://github.com/jhiesey/videostream to create a video stream from js-ipfs?

achingbrain:
Sure, maybe on a different PR though 😉

diasdavid:
@achingbrain yeah, absolutely. Just add this example to the index as suggested above and it is good to merge :)

mitra42:
Its something like that videostream example that we are looking for. What's missing is how to create a function such as in the https://github.com/jhiesey/videostream Readme that takes an opts structure with start and end, and retrieves those bytes from a IPFS stream.

achingbrain:
> Make sure to list it https://github.com/ipfs/js-ipfs/tree/master/examples#tutorials

Done!

diasdavid:
thank you :)

PR #1230, opened 2018-02-20
feat: .stats.bw* - Bandwidth Stats

hacdias:
It was decided on #1198 that the implementation of the Bandwidth Stats should be done in a different PR since it first needs https://github.com/libp2p/js-libp2p/issues/157 to get merged. This issue is used to track the progress on that.

## Prerequisites

- [x] Needs [libp2p/js-libp2p-switch#243](https://github.com/libp2p/js-libp2p-switch/pull/243)
- [x] Needs https://github.com/libp2p/js-libp2p/pull/169
- [x] https://github.com/ipfs/interface-ipfs-core/pull/225

## ToDos

- [x] Core
	- [x] `ipfs.stats.bw`
	- [x] `ipfs.stats.bwReadableStream`
	- [x] `ipfs.stats.bwPullStream`
- [x] HTTP `/api/v0/stats/bw`
- [x] CLI `ipfs stats bw`

### DIFF
--- a/bw
+++ b/bw
@@ -1,0 +1,1 @@
+{"TotalIn":"0","TotalOut":"0","RateIn":"0","RateOut":"0"}
--- a/bw
+++ b/bw
@@ -1,1 +1,0 @@
-{"TotalIn":"0","TotalOut":"0","RateIn":"0","RateOut":"0"}
--- a/package.json
+++ b/package.json
@@ -122,11 +122,7 @@
     "joi-browser": "^13.0.1",
     "joi-multiaddr": "^1.0.1",
     "libp2p": "~0.19.0",
-<<<<<<< HEAD
     "libp2p-circuit": "~0.1.5",
-=======
-    "libp2p-circuit": "~0.1.4",
->>>>>>> feat: .stats.bw*
     "libp2p-floodsub": "~0.14.1",
     "libp2p-kad-dht": "~0.9.0",
     "libp2p-keychain": "~0.3.1",
--- a/package.json
+++ b/package.json
@@ -74,9 +74,8 @@
     "form-data": "^2.3.2",
     "go-ipfs-dep": "^0.4.13",
     "hat": "0.0.3",
-    "interface-ipfs-core": "~0.56.5",
     "ipfsd-ctl": "~0.30.1",
-    "left-pad": "^1.2.0",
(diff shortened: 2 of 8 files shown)


hacdias:
@diasdavid do you think we need more tests than the ones from `interface-ipfs-core`?

diasdavid:
@hacdias https://github.com/libp2p/js-libp2p/pull/169 done :)

hacdias:
@diasdavid do you have any idea why is [this test](https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/swarm.js#L83) failing since I updated the libp2p? It seems that we need to change it to `expect(PeerId.isPeerId(peer.peer.id)).to.equal(true)` to pass. It doesn't seem to be correct according to https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/SWARM.md#peers.

hacdias:
@diasdavid just pushed https://github.com/ipfs/js-ipfs/pull/1230/commits/09b02e26f19549e05e468286d1d09e54dabb6c02 which fixes the issue.

hacdias:
~@diasdavid do you have any idea why is this test failing since I updated the libp2p? It seems that we need to change it to expect(PeerId.isPeerId(peer.peer.id)).to.equal(true) to pass. It doesn't seem to be correct according to ipfs/interface-ipfs-core:SPEC/SWARM.md@master#peers.~

@diasdavid just pushed 09b02e2 which fixes the issue.

/cc @alanshaw

hacdias:
So, my fix only made more stuff to fail. @alanshaw I see that you were the one who changed that code so I think you should know more about this than I do.

Anyway, this PR wasn't created to fix other issues than to create the .stats.bw* interface. That's done and the tests are passing. @diasdavid could you please merge https://github.com/ipfs/interface-ipfs-core/pull/225 and release a new version in order to update the interface-ipfs-core here and merge this, please?

diasdavid:
@hacdias rebased master onto this branch, it should all be fine. Please confirm.

bw
```diff
@@ -0,0 +1 @@
+{"TotalIn":"0","TotalOut":"0","RateIn":"0","RateOut":"0"}
```
diasdavid: @hacdias where does this file come from?

diasdavid:
image

Woot! Thanks @hacdias ❤️

hacdias:
Thanks 😄

PR #1232, opened 2018-02-21
feat: use reduces keysize

dryajov:


test/cli/ls.js
```diff
@@ -31,7 +31,7 @@ describe('ls', () => runOnAndOff((thing) => {
     // If the daemon is on, ls should search until it hits a timeout
     return Promise.race([
       ipfs.fail('ls QmYmW4HiZhotsoSqnv2o1oSssvkRM8b9RweBoH7ao5nki2'),
-      new Promise((res, rej) => setTimeout(res, 4000))
+      new Promise((resolve, reject) => setTimeout(resolve, 4000))
```
diasdavid: Now it is a real Promise 😁

dryajov: hehe, lint was madly yelling that its unacceptable :)

diasdavid:
ipfsd-ctl 0.29 released. Let's get the dep updated and see how much time we shave. Right now it is:

Circle:
image

Travis:
image

dryajov:
Success! Seems like we got around ~7 mins off

screen shot 2018-02-21 at 12 54 37 pm

diasdavid:
Roughly 5 minutes and half. Good, was hopping for something better.

image
image

@VictorBjelkholm it would be so great to parallelize the CLI, HTTP, Core and Browser tests.

PR #1234, opened 2018-02-23
feat: rework http-api tests to use ipfsd-ctl

dryajov:
I looked at the http-api tests and reworked those that made sense


dryajov:
windows tests are failing because of lack of graceful stop - https://github.com/ipfs/js-ipfsd-ctl/pull/205

PR #1239, opened 2018-02-28
feat: add config validation

### LINKED ISSUE #1193: Validate config and provide good error messages if something is wrong
- **Version**: 0.27.7
- **Platform**: 
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G1114

#### Description:
If an invalid protocol is specified in Ipfs' config constructor argument, it will be persisted and then pre-start keeps failing due to the invalid protocol – even after changing the config.

```
› node src/index.js

Swarm listening on /ip4/127.0.0.1/tcp/4001/ipfs/YYYYYYYYYY
I am ready!
Waiting for peers...
peers:  [ '/ip4/127.0.0.1/tcp/9999/ws/ipfs/XXXXXXXXXX' ]
```

Protocol used on `Addresses.Swarm` config list `/ip6/::/etcp/4001`
```
› node src/index.js

/Users/thiagodelgado/code/ipfs-playground/node_modules/multiaddr/src/protocols-table.js:17
    throw new Error('no protocol with name: ' + proto)
    ^

Error: no protocol with name: etcp
```

After fixing the config object
```
› node src/index.js

/Users/thiagodelgado/code/ipfs-playground/node_modules/multiaddr/src/protocols-table.js:17
    throw new Error('no protocol with name: ' + proto)
    ^

Error: no protocol with name: etcp
```

Same error even though not using that protocol anymore. Editing `.jsipfs/config` file removing the bad line solved it.

```
› node src/index.js

Swarm listening on /ip4/127.0.0.1/tcp/4001/ipfs/YYYYYYYYYY
I am ready!
Waiting for peers...
peers:  [ '/ip4/127.0.0.1/tcp/9999/ws/ipfs/XXXXXXXXXX' ]
```

alanshaw:
This PR validates config options passed to IPFS and throws immediately if they are not as expected.

This is a proposal to fix #1193.

We use `joi` to validate config as it's expressive and already included in the project. It's also the best validation library I've come across.

I've verified the config objects used in all the tests work but **I can't say for sure if this is a breaking change or not so you should probably regard it as one when releasing**.

The idea is that the validation should be flexible and shouldn't block new features landing in JS-IPFS. Validation of values that are objects **allows for unknown keys** so that features can land, and validation for the config can be backfilled at a later date if needs be.

I've also configured the `browser` field in `package.json` to use `joi-browser`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "is-stream": "^1.1.0",
     "joi": "^13.1.2",
     "joi-browser": "^13.0.1",
-    "joi-ipfs-config": "^1.0.2",
+    "joi-multiaddr": "^1.0.1",
     "libp2p": "~0.18.0",
     "libp2p-circuit": "~0.1.4",
     "libp2p-floodsub": "~0.14.1",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "./src/core/runtime/repo-nodejs.js": "./src/core/runtime/repo-browser.js",
     "./src/core/runtime/dns-nodejs.js": "./src/core/runtime/dns-browser.js",
     "./test/utils/create-repo-nodejs.js": "./test/utils/create-repo-browser.js",
-    "stream": "readable-stream"
+    "stream": "readable-stream",
+    "joi": "joi-browser"
   },
   "engines": {
     "node": ">=6.0.0",
@@ -119,6 +120,8 @@
     "is-ipfs": "^0.3.2",
     "is-stream": "^1.1.0",
     "joi": "^13.1.2",
+    "joi-browser": "^13.0.1",
+    "joi-ipfs-config": "^1.0.2",
     "libp2p": "~0.18.0",
     "libp2p-circuit": "~0.1.4",
     "libp2p-floodsub": "~0.14.1",
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -1,0 +1,41 @@
+const Joi = require('joi').extend(require('joi-multiaddr'))
+
+const schema = Joi.object().keys({
+    repo: Joi.alternatives().try(
+      Joi.object(), // TODO: schema for IPFS repo
+      Joi.string()
+    ).allow(null),
+    init: Joi.alternatives().try(
+      Joi.boolean(),
+      Joi.object().keys({ bits: Joi.number().integer() })
+    ).allow(null),
+    start: Joi.boolean(),
+    pass: Joi.string().allow(''),
+    EXPERIMENTAL: Joi.object().keys({
+      pubsub: Joi.boolean(),
+      sharding: Joi.boolean(),
+      dht: Joi.boolean()
+    }).allow(null),
+    config: Joi.object().keys({
+      Addresses: Joi.object().keys({
+        Swarm: Joi.array().items(Joi.multiaddr().options({ convert: false })),
+        API: Joi.multiaddr().options({ convert: false }),
+        Gateway: Joi.multiaddr().options({ convert: false })
+      }).allow(null),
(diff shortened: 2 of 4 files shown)


VictorBjelkholm:
Seems strange that this is in it's own config, doesn't it just apply for js-ipfs or this is a package that will be reused? In what way?

alanshaw:
I'm happy to move it into this project if deemed necessary. It could be used anywhere you'd like to validate config from an untrusted source without initializing an IPFS node.

I could use it for https://github.com/ipfs-shipyard/ipfs-companion/pull/395 to validate config on the options page before it is saved to storage. The background process picks up the new config and then attempts to restart the node. It's quicker and less awkward to catch config problems "client side" (on the options page).

package.json
```diff
@@ -119,6 +120,8 @@
     "is-ipfs": "^0.3.2",
     "is-stream": "^1.1.0",
     "joi": "^13.1.2",
+    "joi-browser": "^13.0.1",
+    "joi-ipfs-config": "^1.0.2",
```
diasdavid: @alanshaw are you proposing that we need to update an external module every time we want to tinker with the config? Sounds like a bit of too much overhead.

alanshaw: I'll move it into js-ipfs

src/core/index.js
```diff
@@ -11,8 +11,8 @@ const CID = require('cids')
 const debug = require('debug')
 const extend = require('deep-extend')
 const EventEmitter = require('events')
-const Joi = require('joi').extend(require('joi-ipfs-config'))
 
+const Config = require('./config')
```
diasdavid: `s/Config/config`

src/core/config.js
```diff
@@ -0,0 +1,43 @@
...
+      webRTCStar: Joi.object().keys({
+        Enabled: Joi.boolean()
+      }).allow(null)
+    }).allow(null),
+    Bootstrap: Joi.array().items(Joi.multiaddr().IPFS().options({ convert: false }))
+  }).allow(null),
+  libp2p: Joi.object().keys({
+    modules: Joi.object().allow(null) // TODO: schemas for libp2p modules?
+  }).allow(null)
+}).options({ allowUnknown: true })
```
JonKrone: Hey @alanshaw. Love this. Can you explain why we allow unknown options? Are we trying to be helpful for 
developers or open to external config options?

alanshaw: I don't really have a strong opinion either way on this one, I guess it's just to be flexible...

alanshaw: oh look, I forgot I already wrote a reason in the description!

> The idea is that the validation should be flexible and shouldn't block new features landing in JS-IPFS. Validation of values that are objects **allows for unknown keys** so that features can land, and validation for the config can be backfilled at a later date if needs be.

diasdavid: 👍

diasdavid:
Seems that Jenkins failed in Browser because the browser didn't have enough space to run:
image

this only happens sometimes and it might be because all the tests are using the same browser/host to run? @VictorBjelkholm ? Anyways, not a blocker for this PR.

Thank you @alanshaw ❤️

olizilla:
This PR: https://github.com/ipfs-shipyard/ipfs-companion/pull/395 to allow users to configure their js-ipfs node in ipfs-companion is blocked on this PR. Is this one good to go @diasdavid?

diasdavid:
@olizilla yes, thank you for the ping :)

@VictorBjelkholm can you look into the Quota issue in Jenkins?

diasdavid:
@olizilla going to start a patch release, bare with me as tests take a bit! :)

PR #1243, opened 2018-03-02
Less Crashy gateway. catch stream2 error

ya7ya:
This catches the following error. which results in a less crashy gateway as mentioned here https://github.com/ipfs/js-ipfs/issues/804#issuecomment-369912316

```bash
 Error: write after end
    at writeAfterEnd (/media/op/w3/js-ipfs/node_modules/readable-stream/lib/_stream_writable.js:288:12)
    at PassThrough.Writable.write (/media/op/w3/js-ipfs/node_modules/readable-stream/lib/_stream_writable.js:332:20)
    at pull.through (/media/op/w3/js-ipfs/src/http/gateway/resources/gateway.js:134:21)
    at /media/op/w3/js-ipfs/node_modules/pull-stream/throughs/through.js:17:24
    at drain (/media/op/w3/js-ipfs/node_modules/stream-to-pull-stream/index.js:141:18)
    at Stream.<anonymous> (/media/op/w3/js-ipfs/node_modules/stream-to-pull-stream/index.js:150:5)
    at Stream.emit (events.js:159:13)
    at next (/media/op/w3/js-ipfs/node_modules/pull-stream-to-stream/index.js:120:13)
    at /media/op/w3/js-ipfs/node_modules/pull-stream/throughs/map.js:19:9
    at /media/op/w3/js-ipfs/node_modules/pull-traverse/index.js:42:7
```
License: MIT
Signed-off-by: Yahya <ya7yaz@gmail.com>

### DIFF
--- a/src/http/gateway/resources/gateway.js
+++ b/src/http/gateway/resources/gateway.js
@@ -98,6 +98,10 @@ module.exports = {
         //  response.continue()
         let filetypeChecked = false
         let stream2 = new Stream.PassThrough({ highWaterMark: 1 })
+        stream2.on('error', (err) => {
+          log.error('stream2 err: ', err)
+        })
+
         let response = reply(stream2).hold()
 
         pull(


diasdavid:
@VictorBjelkholm Jenkins is not happy anymore:
image

VictorBjelkholm:
@diasdavid seems to be about LocalStorage (or similar) has reached the maximum amount of storage allowed.

diasdavid:
@VictorBjelkholm It is IndexedDB related. 

> The maximum browser storage space is dynamic — it is based on your hard drive size. The global limit is calculated as 50% of free disk space.

and

> here's also another limit called group limit — this is defined as 20% of the global limit, but it has a minimum of 10MB and a maximum of 2GB. 

Source: https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria

VictorBjelkholm:
I left this comment somewhere else (on the commit itself I think) but can't find it so leaving it here as well just in case.

While I agree that this patch fixes the particular issue with `write after end`, it doesn't feel like the completely right fix. We have more uncaught errors/promise rejections that are not solved by this. We need a real fix that catches all errors within the gateway + daemon and prints them, while leaving the gateway + daemon still running.

PR #1252, opened 2018-03-07
fix: changes peer prop in return value from swarm.peers to be a PeerId 

### LINKED ISSUE #1248: swarm.peers return type is different in js-ipfs and js-ipfs-api
`swarm.peers` returns an object with a `peer` property. In the [spec it says this will be a `PeerInfo` object](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/SWARM.md#peers), and it [is in js-ipfs](https://github.com/ipfs/js-ipfs/blob/5b40b41f26573856b206e6b21be89cfc67ef39b4/src/core/components/swarm.js#L27-L34), but if you call via js-ipfs-api you'll [get back a `PeerId`](https://github.com/ipfs/js-ipfs-api/blob/78ba320573ab73ff4224740f59e82e2e504d7881/src/swarm/peers.js#L48) (js-ipfs http [converts the PeerInfo to a PeerId](https://github.com/ipfs/js-ipfs/blob/5b40b41f26573856b206e6b21be89cfc67ef39b4/src/http/api/resources/swarm.js#L54)).

Just wondering if this is expected, and if not, what is the desired result?

alanshaw:
To be consistent with go-ipfs. See discussion here in #1248

depends on https://github.com/ipfs/interface-ipfs-core/pull/230
fixes #1248


diasdavid:
image

Mac OS still has some issues. But remaining CI shows it is good.

diasdavid:
With this merge, now the next release needs to be a __minor__

PR #1253, opened 2018-03-07
docs: added some small verbage to the README for "Resolve through IPLD grap…

datafatmunger:
datafatmunger vmx & daviddias i realized the folly of my earlier attempts of the ipld git tutorial was that the files in the git-objects directory of the example have already been decompressed via zlib ... i should have cat'd those files hours ago, but since they have hash names i assumed they came strait out of .git/objects :-P
daviddias datafatmunger: oh, that should be part of the README of the example. Wanna add that step?
vmx datafatmunger: do i understand correctly that git objects or compressed by default and you need to uncompress them before you use them with ipfs/ipld?
datafatmunger yes they are compressed by default ...
...
datafatmunger and need to be decompressed before the tutorial
datafatmunger the plugin method i used from magik6k, however ... does do the zlib step
datafatmunger ipfs dag put -f git --input-enc zlib .git/objects/7d/df2..
datafatmunger in any case: cat df25817f57c2090a9568cdb17106a76dad7d04 | zlib-flate -uncompress  gitobject
datafatmunger will give a text file format that works with your tutorial
...
datafatmunger daviddias i'll try to make a pr for the readme

### DIFF
--- a/examples/traverse-ipld-graphs/README.md
+++ b/examples/traverse-ipld-graphs/README.md
@@ -18,6 +18,9 @@ This tutorial goes through several operations over IPLD graphs using the [DAG AP
 ## [traverse through a slice of the ethereum blockchain](./eth.js)
 
 ## [traverse through a git repo](./git.js)
+The example objects contained in "git-objects" have already been decompressed with zlib.  An example of how to do this:
+
+    $ cat .git/objects/7d/df25817f57c2090a9568cdb17106a76dad7d04 | zlib-flate -uncompress > 7ddf25817f57c2090a9568cdb17106a76dad7d04 
 
 ## Video of the demos
 


examples/traverse-ipld-graphs/README.md
```diff
@@ -18,6 +18,9 @@ This tutorial goes through several operations over IPLD graphs using the [DAG AP
 ## [traverse through a slice of the ethereum blockchain](./eth.js)
 
 ## [traverse through a git repo](./git.js)
+The example objects contained in "git-objects" have already been decompressed with zlib.  An example of how to do this:
+
+    $ cat .git/objects/7d/df25817f57c2090a9568cdb17106a76dad7d04 | zlib-flate -uncompress > 7ddf25817f57c2090a9568cdb17106a76dad7d04
```
diasdavid: This should use the `jsipfs block put` with respective arguments. //cc @vmx

diasdavid:
@datafatmunger did you get my comment?

datafatmunger:
I did.  I just don't quite understand it.  :-)  Your tutorial included already decompressed git object files, so I was just including instructions on how someone could create similar files.

Following your comment ... I could, of course, put a block on ipfs ... but then I lose the ipld data structure, no?  Maybe jsipfs is now doing something like the go ipfs extension in mentioned in the irc thread pasted above with the "-f git" arguments, and maybe that's what you mean in your comment: "with respective arguments", but I don't know what those arguments are.

examples/traverse-ipld-graphs/README.md
vmx: @diasdavid Did you mean `jsipfs dag put`?

alanshaw:
@diasdavid can you elaborate here or approve so we can merge and address changes at a later date?

examples/traverse-ipld-graphs/README.md
diasdavid: I did really meant `block put` as what @datafatmunger is doing is putting an already serialized node from git into ipfs, therefore `block put` rather than `dag put` which would try to serialize it again.

PR #1255, opened 2018-03-08
fix bug: node.files.add do not accept object

tiagochen:
file: src\core\components\files.js
line: 199
api: add
problem: add api's param do not accept object, use Buffer instead


JonKrone:
@tiagochen Thanks for the PR! And on the ipfs-101 example no less.

Would you please also update the `README` with these changes?

diasdavid:
This was a regression, fix here: https://github.com/ipfs/js-ipfs/pull/1257. My apologies for the troubles.

PR #1257, opened 2018-03-09
fix: accept objects in file.add

dryajov:
Adding back ability to pass objects to `files.add` - `{path: <string>, data: Buffer.from('hello')}`. This regressed in 4b79066b924b8e8bf7f8a0fde0da40c3c1b84f31

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "hapi-set-header": "^1.0.2",
     "hoek": "^5.0.3",
     "human-to-milliseconds": "^1.0.0",
-    "ipfs-api": "^18.1.1",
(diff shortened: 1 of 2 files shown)


Mr0grog:
Seems like it might be good to add a regression test for that, too.

dryajov:
Adding test https://github.com/ipfs/interface-ipfs-core/pull/234

diasdavid:
Good catch! Can you check why CI isn't green too?

diasdavid:
@dryajov tested locally. It fails because js-ipfs-api has the same issue.

diasdavid:
verified locally and in some of the CI runs. hotfix incoming.

PR #1262, opened 2018-03-13
fix: match error if repo doesn't exist

dryajov:


JonKrone:
I believe that this is the same handling as is happening [in boot.js here](https://github.com/ipfs/js-ipfs/blob/master/src/core/boot.js#L43). Could we alternatively migrate this check to the `ipfs-repo._isInitialized` method and convert these errors into a unified `repo doesn't exist` error?

dryajov:
@JonKrone That already handles `IndexDb` errors with the `not found` string... This fix is specifically for the version call. I'm not sure if they can be unified.

Issue #1245, opened 2018-03-04
Cannot create a node with an initialized repo


realharry:
- **Version**: 0.28.0
- **Platform**: Ubuntu Desktop 17.10 (Linux 4.13.0-36-generic #40-Ubuntu SMP Fri Feb 16 20:07:48 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux)
- **Subsystem**: Node: 6.11.5

#### Description:

Attempting to create an IPFS node using an initialized Repo object throws "Callback was already called." error.

Not sure if I'm using the API correctly (practically, it's my first day of using js-ipfs), but if I try to pass the repo argument with the initialized Repo object in the IPFS constructor, I get the following exception from Async:

    /home/harry/ipfsjs/ipfs-repo/node_modules/async/internal/onlyOnce.js:9
            if (fn === null) throw new Error("Callback was already called.");
                         ^
    Error: Callback was already called.
        at /home/harry/ipfsjs/ipfs-repo/node_modules/async/internal/onlyOnce.js:9:32
        at /home/harry/ipfsjs/ipfs-repo/node_modules/async/internal/parallel.js:36:13
        at /home/harry/ipfsjs/ipfs-repo/node_modules/write-file-atomic/index.js:142:7
        at FSReqWrap.oncomplete (fs.js:123:15)

#### Steps to reproduce the error:

Here' a sample code snippet:

    const IPFS = require('ipfs');
    const IPFSRepo = require('ipfs-repo');
    const repo1 = new IPFSRepo('/tmp/ipfs-repo1');

    repo1.init({}, (err) => {
        if (err) { console.log(err); return; }
        console.log('IPFS repo 1 initialized.');

        console.log('Creating IPFS node 1.');
        let node1 = new IPFS({
          repo: repo1,
          init: false,
          start: true,
          EXPERIMENTAL: {
          },
        });
        console.log('IPFS node 1 created.');
        node1.on('start', () => {
          console.log('IPFS node 1 started.');
        });
      })

JonKrone:
Hey @realharry I spent a little time looking at this problem and was able narrow the problem but I'm not sure of expected behavior here so I wanted to share what I've found.

IPFS takes additional steps when initializing a new repo, adding a `Keychain` and `Identity` config property for local storage of a node's identity and private keys:

https://github.com/ipfs/js-ipfs/blob/d945fceeb09c1c4289a61edb40209271e8318d25/src/core/components/init.js#L44-L69

[...]

PR #1263, opened 2018-03-13
Describe when and why to use `init: false`

Mr0grog:
This adds some description to the "advanced options" documentation for the `IPFS` constructor detailing that `init: false` should really be used if a repo has already been initialized *by IPFS* and that simply calling `repoInstance.init()` doesn't init a repo with everything IPFS needs.

This partially handles #1245, but other parts of the experience (e.g. #1262) could also be improved.

Does this help clarify things? Or is it too confusing?

Is there anywhere else these options are documented that I need to update? (I couldn’t find anywhere, but I wouldn’t be surprised if I missed something.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,11 +194,8 @@ const repo = <IPFS Repo instance or repo path>
 const node = new IPFS({
   repo: repo,
   init: true, // default
-  // init: false, // If you have already initialized the repo via `new IPFS()`.
-  //              // Be careful using this on an IPFSRepo instance you have
-  //              // initialized yourself -- the IPFS constructor initializes
-  //              // repos with extra options, so calling `init()` on an
-  //              // IPFSRepo may not always be compatible with IPFS.
+  // init: false, // You will need to set init: false after time you start instantiate a node as
+  //              // the repo will be already initiated then.
   // init: {
   //   bits: 1024 // size of the RSA key generated
   // },
--- a/README.md
+++ b/README.md
@@ -194,7 +194,11 @@ const repo = <IPFS Repo instance or repo path>
 const node = new IPFS({
   repo: repo,
   init: true, // default
-  // init: false,
+  // init: false, // If you have already initialized the repo via `new IPFS()`.
+  //              // Be careful using this on an IPFSRepo instance you have
+  //              // initialized yourself -- the IPFS constructor initializes
+  //              // repos with extra options, so calling `init()` on an
+  //              // IPFSRepo may not always be compatible with IPFS.
   // init: {
   //   bits: 1024 // size of the RSA key generated
   // },


dryajov:
This looks good - thanks @Mr0grog!

dryajov:
Perhaps @diasdavid has more (historical) insight into that - but it might be worth adding that, even though `IPFS` takes a repo instance its not intended (IMHO) to be initialized outside of `IPFS` at all.

README.md
```diff
@@ -194,7 +194,11 @@ const repo = <IPFS Repo instance or repo path>
 const node = new IPFS({
   repo: repo,
   init: true, // default
-  // init: false,
+  // init: false, // If you have already initialized the repo via `new IPFS()`.
+  //              // Be careful using this on an IPFSRepo instance you have
+  //              // initialized yourself -- the IPFS constructor initializes
+  //              // repos with extra options, so calling `init()` on an
+  //              // IPFSRepo may not always be compatible with IPFS.
```
diasdavid: I'm not sure I don't understand what this comment is trying to prevent the user to do. Either a user knows what they are doing and they know supposed pitfall or they don't and the terms here don't give them any other help.

Mr0grog: OK. We clearly have users who don’t know what this is for as noted in the summary — this is a result of #1245.

What language would you suggest better clarifies when to and not to use `init: false`? What should I say here that helps someone like that know they can’t just:

```js
const repo1 = new IPFSRepo('/tmp/ipfs-repo1');
repo1.init({}, (err) => {
  if (err) { console.log(err); return; }

  const node1 = new IPFS({
    repo: repo1,
    init: false,
    start: true
  });

  node1.on('start', () => {
    console.log('IPFS node 1 started.');
  });
})
```

Something like:

> Use this if the repo has previously been initialized by passing it to `new IPFS()`. You should not attempt to call `init()` yourself on an IPFSRepo instance and then set `init: false` here.

?

Clearly this could go into a lot more detail, but because this is all inline in a code sample, there’s not a whole lot of room.

mitra42: I think part of the confusion with repo is how you right code so that it works both the first time its called, and following times.  init confused me a lot when I first saw it. Do I 
a) set init=true - in which case I would assume it will clear the repo every time, losing all saved data or 
b) set init=false in which case it will fail the first time. 

I think most distributed IPFS based apps are going to want to default to "initialize a new repo if it doesn't exist, or if its incompatible (because IPFS has changed version and broken old repo's) and otherwise leave it alone".  I still have no clue wither that is to set it to false, true or leave it missing (which is what we currently do. ) 

Some comments here about when to set it to true, when to false, and when to omit it would be useful.

README.md
```diff
@@ -194,7 +194,8 @@ const repo = <IPFS Repo instance or repo path>
 const node = new IPFS({
   repo: repo,
   init: true, // default
-  // init: false,
+  // init: false, // You will need to set init: false after time you start instantiate a node as
+  //              // the repo will be already initiated then.
```
diasdavid: @Mr0grog what about this? :)

PR #1233, opened 2018-02-22
feat: jsipfs add --only-hash

JonKrone:
This enables generating a would-be hash for content via http or cli: `jsips add --only-hash <file>`, `ipfs.add(<content>, { onlyHash: true })`

image

Turns out unix-fs-engine has been ready for this since https://github.com/ipfs/js-ipfs-unixfs-engine/pull/183.

### Depends on:
- [ ] https://github.com/ipfs/js-ipfs-api/pull/700

### Related issues:
https://github.com/ipfs/js-ipfs-api/issues/509
https://github.com/ipfs/js-ipfs/issues/1205

PR #1266, opened 2018-03-16
feat: jsipfs add --only-hash (#1233)

diasdavid:
Continuing the work from #1233

### DIFF
--- a/src/cli/commands/files/add.js
+++ b/src/cli/commands/files/add.js
@@ -186,10 +186,12 @@ module.exports = {
     const index = inPath.lastIndexOf('/') + 1
     const options = {
       strategy: argv.trickle ? 'trickle' : 'balanced',
-      shardSplitThreshold: argv.enableShardingExperiment ? argv.shardSplitThreshold : Infinity,
+      shardSplitThreshold: argv.enableShardingExperiment
+        ? argv.shardSplitThreshold
+        : Infinity,
       'cid-version': argv['cid-version'],
       'raw-leaves': argv['raw-leaves'],
-      onlyHash: argv['only-hash']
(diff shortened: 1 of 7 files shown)


src/cli/commands/files/add.js
```diff
@@ -182,7 +188,8 @@ module.exports = {
       strategy: argv.trickle ? 'trickle' : 'balanced',
       shardSplitThreshold: argv.enableShardingExperiment ? argv.shardSplitThreshold : Infinity,
       'cid-version': argv['cid-version'],
-      'raw-leaves': argv['raw-leaves']
+      'raw-leaves': argv['raw-leaves'],
+      onlyHash: argv['only-hash']
```
diasdavid: it should be `'only-hash'`

JonKrone: Saw you just pushed a change for this, thanks. You'll need to change all/most instances of `onlyHash`, including in http-api/resources/files.js. I converted it all to onlyHash as dot properties are easier to work with and read. Interestingly, `yargs` also automatically converts all hyphenated options to camelCase: this could also be `onlyhash: argv.onlyHash`

JonKrone: Actually, I suggest we keep it `onlyHash`. `cid-version` and `raw-leaves` are the outlier here. There are many hyphenated options we convert to camelCase at the entry points of cli/http-api such as the above `shardSplittingThreshold` and `dag --local-resolve`, `object --input-enc`.

I can change raw-leaves and cid-version instead

diasdavid: @JonKrone I like that. Go for it :)

PR #1268, opened 2018-03-17
docs: add a npm header in the README.md

qpakzk:
In the Table of Contents, there is a link of npm.
But it did't work, because npm wasn't a header.
It's necessary to add a npm header to work a link of it well.


diasdavid:
thanks @qpakzk :)

PR #1270, opened 2018-03-18
feat: show which link does not exist on ipfs.cat

### LINKED ISSUE #1145: Should specify which link didn't exist on ipfs.cat

youngnicks:
Closes #1145.

For `interface-ipfs-core` test to pass, the corresponding test there needs to be updated to be similar to below as one version of running tests has an error message prepended with `Failed to cat file: Error:`, causing the test to fail.

```javascript
expect(err.message).to.contain(
  'no link named "does-not-exist" under Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
)
```

### DIFF
--- a/src/core/components/files.js
+++ b/src/core/components/files.js
@@ -131,7 +131,7 @@ module.exports = function files (self) {
       throw new Error('You must supply an ipfsPath')
     }
 
-    let bestMatch = 0
+    let bestMatch = { depth: -1 }
 
     ipfsPath = normalizePath(ipfsPath)
     const pathComponents = ipfsPath.split('/')
@@ -139,10 +139,8 @@ module.exports = function files (self) {
     const restPath = normalizePath(pathComponents.slice(1).join('/'))
     const filterFile = (file) => {
       if (file.path === ipfsPath.substring(0, file.path.length)) {
-        const matchedNodes = file.path.split('/').length
-
-        if (matchedNodes > bestMatch) {
-          bestMatch = matchedNodes
+        if (file.depth > bestMatch.depth) {
+          bestMatch = file
         }
       }
 
@@ -159,9 +157,9 @@ module.exports = function files (self) {
           files = files.filter(filterFile)
         }
         if (!files || !files.length) {
-          const parent = pathComponents.slice(0, bestMatch).join('/')
-          const link = pathComponents.slice(bestMatch).join('/')
-          return d.abort(new Error(`no link named "${link}" under ${parent}`))
+          const hash = toB58String(bestMatch.hash)
+          const link = pathComponents[bestMatch.depth + 1]
+          return d.abort(new Error(`no link named "${link}" under ${hash}`))
         }
 
         const file = files[0]
--- a/src/core/components/files.js
+++ b/src/core/components/files.js
@@ -146,7 +146,7 @@ module.exports = function files (self) {
           files = files.filter(filterFile)
         }
         if (!files || !files.length) {
-          return d.abort(new Error('No such file'))
+          return d.abort(new Error(`no link named "${restPath}" under ${pathComponents[0]}`))
         }
 
         const file = files[0]
--- a/src/core/components/files.js
+++ b/src/core/components/files.js
@@ -138,6 +138,8 @@ module.exports = function files (self) {
     const prePath = normalizePath(pathComponents.slice(0, 1).join('/'))
     const restPath = normalizePath(pathComponents.slice(1).join('/'))
     const filterFile = (file) => {
+      // Save off best match to provide a better error message if file
+      // isn't found.
       if (file.path === ipfsPath.substring(0, file.path.length)) {
         if (file.depth > bestMatch.depth) {
           bestMatch = file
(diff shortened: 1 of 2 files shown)


youngnicks:
@diasdavid I added logic to determine which node in the path is missing and updated the error message. I want to get some feedback on how I did this as it could have performance hits on nodes containing lots of files. If this is acceptable, I will add a test case for the scenario when part of the path doesn't exist. If this isn't acceptable, I'm going to have to put logic in a higher function to determine where the missing path is.

test/cli/files.js
```diff
@@ -322,7 +322,11 @@ describe('files', () => runOnAndOff((thing) => {
     return ipfs('cat QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB/dummy')
       .then(() => expect.fail(0, 1, 'Should have thrown an error'))
       .catch((err) => {
+        const message = err.stderr.match(/^Error:(?: Failed to cat file: Error:)? (.*)$/m)[1]
         expect(err).to.exist()
+        expect(message).to.eql(
+          'no link named "dummy" under QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB'
+        )
```
diasdavid: This feature is still not complete. Please add tests for multiple levels of nestedness (i.e `/a/b/c/d`) and you will see that after the first level, this PR fails.

youngnicks: Thanks, I was more asking if this would be an acceptable location for this logic, knowing that it could possibly have a performance hit when there is a large amount of files on a node.

What should the error message show when a middle level is missing? I am thinking the message should be `no link named "c/d" under a/b` to provide more context to as to what the requested file is.

youngnicks:
@diasdavid I added an additional test for nestedness. Both tests are passing. Is the format of the nested error correct, or should it be saying something different?

youngnicks:
@diasdavid After playing around with `go-ipfs`, I noticed how it handles this situation and updated the code to provide the same error.

diasdavid:
@pgte mind reviewing this PR and help @youngnicks get it completed?

test/cli/files.js
```diff
@@ -322,7 +322,33 @@ describe('files', () => runOnAndOff((thing) => {
...
+      })
+  })
+
+  it('cat specifies missing directory in a nested link', () => {
+    return ipfs('cat QmYmW4HiZhotsoSqnv2o1oUusvkRM8b9RweBoH7ao5nki2/init-docs/docs/wrong-dir/dummy')
+      .then(() => expect.fail(0, 1, 'Should have thrown an error'))
+      .catch((err) => {
+        const message = err.stderr.match(/^Error:(?: Failed to cat file: Error:)? (.*)$/m)[1]
+        expect(message).to.eql(
+          'no link named "wrong-dir" under QmegvLXxpVKiZ4b57Xs1syfBVRd8CbucVHAp7KpLQdGieC'
```
pgte: I would expect the error to say something like either:

* `"no file named "wrong-dir" under QmYmW4HiZhotsoSqnv2o1oUusvkRM8b9RweBoH7ao5nki2/init-docs/docs"` or
* `"no file named "init-docs/docs/wrong-dir" under QmYmW4HiZhotsoSqnv2o1oUusvkRM8b9RweBoH7ao5nki2"` or

Would any of these be possible?

src/core/components/files.js
```diff
@@ -131,22 +131,35 @@ module.exports = function files (self) {
...
+
     ipfsPath = normalizePath(ipfsPath)
     const pathComponents = ipfsPath.split('/')
+    const prePath = normalizePath(pathComponents.slice(0, 1).join('/'))
     const restPath = normalizePath(pathComponents.slice(1).join('/'))
-    const filterFile = (file) => (restPath && file.path === restPath) || (file.path === ipfsPath)
+    const filterFile = (file) => {
+      if (file.path === ipfsPath.substring(0, file.path.length)) {
+        if (file.depth > bestMatch.depth) {
+          bestMatch = file
```
pgte: Instead of replacing, could we retain the complete path of the best match to support more relevant error messages? (See my previous comment)

test/cli/files.js
pgte: Notice that I'm using the term `file` and not `link`, on purpose ;)

src/core/components/files.js
youngnicks: I think replacing here is still okay as the depth tells us how many nodes `bestMatch` matches and we can use `pathComponents` to build the path. This is essentially what I was doing earlier on in the pull request. I thought of defining if a directory or file was missing, but it wasn't part of the spec so I left it the way it was.

We could have another scenario where a user tries to specify a file under a file thinking it is a directory. What do you think of these errors?

**Missing directory**
`no directory named "wrong-dir" under QmYmW4HiZhotsoSqnv2o1oUusvkRM8b9RweBoH7ao5nki2/init-docs/docs`

**Missing file**
`no file named "does-not-exist" under QmYmW4HiZhotsoSqnv2o1oUusvkRM8b9RweBoH7ao5nki2/init-docs/docs`

**File instead of directory**
`"actually-a-file" is a file not a directory under QmYmW4HiZhotsoSqnv2o1oUusvkRM8b9RweBoH7ao5nki2/init-docs/docs`

pgte: @youngnicks this would be even better! 👍

pgte:
Error messages look great!

To move on, I guess interface-ipfs tests need to be updated, right?

youngnicks:
@pgte Thanks! I already submitted a pull request for `interface-ipfs-core`. I'm just having issues testing it with `js-ipfs-api`. For some reason I'm getting the errors formatted in my prior commit and not the new ones.

https://github.com/ipfs/interface-ipfs-core/pull/241

daviddias:

jacobheun:
Closing as this is stale and there have been significant changes since this was created

PR #1273, opened 2018-03-20
fix: remove unused var

fsdiogo:
Removed an unused var that was producing a linter error.

### DIFF
--- a/src/core/components/swarm.js
+++ b/src/core/components/swarm.js
@@ -3,7 +3,6 @@
 const multiaddr = require('multiaddr')
 const promisify = require('promisify-es6')
 const values = require('lodash.values')
-const PeerId = require('peer-id')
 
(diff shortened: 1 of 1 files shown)


JonKrone:
Thanks @fsdiogo!

@VictorBjelkholm Looking at #1230, and then looking at a couple other PRs, it seems like linting isn't being run during CI anymore. Is this a regression, intentional, or am I missing something?

diasdavid:
It should be run in Circle. Can you add it back it too? Thanks @fsdiogo :)

JonKrone:
update from IRC: @VictorBjelkholm is working on adding it to jenkins 👍

VictorBjelkholm:
I've added linting to the jenkins job now and retried the jenkins jobs for this pr.

PR #1275, opened 2018-03-21
Add CID Instance support to files.cat

### LINKED ISSUE #1229: file.cat API not supporting CIDs
- **Version**: 0.27.7
- **Platform**: OSX
- **Subsystem**: files
#### Description:

https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#cat
Says that in ipfs.files.cat(ipfsPath, cb) 
That ipfsPath can be a CID though it says it should be a “raw Buffer of the cid” or “String base58"

We have been calling files.cat with the CID data structure for a while, 
As reported by Chrome debugging …. 
CID {codec: "dag-pb", version: 0, multihash: Uint8Array(34)}

but my latest update of node packages seems to have broken something… and I’m seeing an error. I could be wrong , but don’t think we were hitting this error before, though its possible we were not using that part of the code. 

Error: path.indexOf is not a function …

Looking at the code, it only handles Buffers or strings, not CIDs.
```
function normalizePath (path) {
  if (Buffer.isBuffer(path)) {  path = toB58String(path) }
  if (path.indexOf('/ipfs/') === 0) { path = path.substring('/ipfs/'.length) }
  if (path.charAt(path.length - 1) === '/') { path = path.substring(0, path.length - 1) }
  return path
}
```

Stack Trace
```
_catPullStream
cat.promisify
Anonymous
Anonymous
 await this.ipfs.files.cat(cid)
```
I wanted to check if this was an IPFS bug, or whether I should convert all my CIDs back to base58 strings if necessary. 

I had thought the direction was towards using CIDs rather than strings or buffers so wanted to check here first before making those edits. If so I’d appreciate what is the “best” way to turn a CID into something to pass to ipfs.files.cat.

jacobheun:
This implements the new spec for ipfs/interface-ipfs-core#243 to resolve #1229 

connects to ipfs/interface-ipfs-core#243
connects to ipfs/js-ipfs-api#721


diasdavid:
Woot! Thanks @jacobheun :)

Issue #874, opened 2017-05-31
"The libp2p node is not started yet" error during discovery


pgte:
Printed message during the tests:

```
assert.js:81
  throw new assert.AssertionError({
  ^
AssertionError: The libp2p node is not started yet
    at Node.dial (/Users/pedroteixeira/projects/ipfs/ipfs-level/node_modules/libp2p/src/index.js:239:5)
    at Node.self._libp2pNode.on (/Users/pedroteixeira/projects/ipfs/ipfs-level/node_modules/ipfs/src/core/components/libp2p.js:29:30)
    at emitOne (events.js:96:13)
    at Node.emit (events.js:188:7)
    at EventEmitter.discovery.on (/Users/pedroteixeira/projects/ipfs/ipfs-level/node_modules/libp2p/src/index.js:71:49)
    at emitOne (events.js:96:13)
    at EventEmitter.emit (events.js:188:7)
    at WebRTCStar._peerDiscovered (/Users/pedroteixeira/projects/ipfs/ipfs-level/node_modules/libp2p-webrtc-star/src/index.js:241:20)
    at Socket.Emitter.emit (/Users/pedroteixeira/projects/ipfs/ipfs-level/node_modules/libp2p-webrtc-star/node_modules/component-emitter/index.js:133:20)
    at Socket.onevent (/Users/pedroteixeira/projects/ipfs/ipfs-level/node_modules/libp2p-webrtc-star/node_modules/socket.io-client/lib/socket.js:267:10)
```

Looks like there may be a race condition here, since `self.isOnline()` returns true [here](https://github.com/ipfs/js-ipfs/blob/860165c496f6a00045f76c14b1cd00658bd65edb/src/core/components/libp2p.js#L27), but the libp2p `isOn()` assertion fails [here](https://github.com/libp2p/js-libp2p/blob/2e326e1619c0337d97612c708e94950dfbd9a6ec/src/index.js#L239).

pgte:
I think I found it: https://github.com/ipfs/js-ipfs/blob/860165c496f6a00045f76c14b1cd00658bd65edb/src/core/components/is-online.js#L5

pgte:
Also, and since this is internal, instead asserting the online status, it should queue the operations for when the node is online (like many database clients do).

In this case, the `.dial()` is being called because we discovered a peer. IMO, we should not ignore this event. Instead, we should queue until every networking component (bitswap and libp2p I guess) is considered to be online.

pgte:
[...]

PR #1278, opened 2018-03-23
chore: move from deprecated multiplex dep to libp2p-mplex

### LINKED ISSUE #1277: libp2p-multiplex is deprecated, use libp2p-mplex instead
Running `npm i ipfs` results in a warning for the deprecated libp2p-multiplex module

```
npm WARN deprecated libp2p-multiplex@0.5.1: use libp2p-mplex instead
```

We should update dependencies to libp2p-mplex

jacobheun:
Resolves #1277


dryajov:
Thanks @jacobheun! Could you please check why CI is not happy?

jacobheun:
Taking a look! It looks like it might be related to https://github.com/ipfs/js-ipfs/issues/874. I've seen this intermittent failure on other builds. I'll see if I can track down the source of libp2p not being ready.

PR #1280, opened 2018-03-24
chore: symlink the jsipfs bin to /usr/local/bin in the Dockerfile

JGAntunes:
Make it possible to run `jsipfs` commands inside the container :+1:


VictorBjelkholm:
Nice, great idea! :+1:

Issue #1178, opened 2018-01-19
Add wrap-with-directory support (HTTP API and Core)


nunofmn:
Related to https://github.com/ipfs-shipyard/ipfs-companion/issues/349

Right now, the `wrap-with-directory` option only works in _CLI_ and support in  _HTTP API_ and _Core_ is missing

I already have a working solution that uses the `wrap` option in [js-ipfs-unixfs-engine Importer API](https://github.com/ipfs/js-ipfs-unixfs-engine#const-import--new-importerdag--options).

_Question:_ it is necessary add support to this option in [interface-ipfs-core](https://github.com/ipfs/interface-ipfs-core) beforehand? (or maybe at least add a tracking issue)

I'm looking forward to open a PR 👍

diasdavid:
Hey @nunofmn o/ Yeah, add it to interface-ipfs-core. You can submit both PRs at the same time :)

diasdavid:
@nunofmn ping :) Still interested in submitting that PR?

musicsmithnz:
I am quite interested in this feature, as it allows to use the human readable filenames and also to have the filename extension which is useful for [browsers?] recognizing the mime type.

wraithgar:
I'm attempting to work on this on the js side of things.  Just getting started ramping up here so your patience is very much appreciated!

musicsmithnz:
I've been looking at a workaround which uses the recursive options and passes in a directory. I have been having trouble with this and I'm not sure exactly why, it might be something to do with asynchronous loading of the files.

wraithgar:
Took a stab at this, see the three PRs linked to above.

diasdavid:
@wraithgar looking good! I'll handle merging and releasing those soon (probably earlier tomorrow) :) Thank you!

musicsmithnz:
Thanks guys I'm looking forward to seeing this.

hacdias:
Closing this since https://github.com/ipfs/js-ipfs/pull/1329 was merged.

PR #1290, opened 2018-04-02
feat: add wrapWithDirectory flag to files.add et al

wraithgar:
For https://github.com/ipfs/js-ipfs/issues/1178

This removes the `-w` flag code from the cli and implements the exact same functionality in the core files component.

There is a `throw` in here that is being eaten (i.e. I did not see it bubble back up when testing this locally on the cli) and I do not know how to make this work, help would be appreciated here.

The reason for the throw is that this feature assumes that you gave a path name with which to wrap in a directory.


diasdavid:
Needs https://github.com/ipfs/js-ipfs-api/pull/730

src/http/api/resources/files.js
```diff
@@ -245,7 +247,7 @@ exports.add = {
       ipfs.files.addPullStream(options),
       pull.map((file) => {
         return {
-          Name: file.path ? file.path : file.hash,
+          Name: file.path, //addPullStream already turned this into a hash if it wanted to
```
wraithgar: This was already being done in addPullStream so I removed it from here, because addPullStream has a new case where it is leaving path as an empty string and wants it that way.

wraithgar:
This has been updated w/ the interface and api updates.

src/core/components/files.js
```diff
@@ -33,16 +35,17 @@ function prepareFile (self, opts, file, callback) {
 
       const b58Hash = cid.toBaseEncodedString()
 
+
       cb(null, {
-        path: file.path || b58Hash,
+        path: opts.wrapWithDirectory ? file.path.substring(WRAPPER.length) : (file.path || b58Hash),
```
wraithgar: I admittedly don't know 100% of what's going on here or why the go implementation behaves this way but having it return an empty path for the wrapper directory feels like an anti-pattern.

wraithgar:
`ipfsd-ctl` will need to be updated to the latest `ipfs-api` for the tests to pass.

src/core/components/files.js
wraithgar: This was explained as "the wrapper is the root"

wraithgar:
needs https://github.com/ipfs/js-ipfsd-ctl/pull/227

VictorBjelkholm:
Any progress on this? Currently needed to have CI passing normally again, as feature was added in interface-ipfs-core and now used for master branch in ipfs/js-ipfs

Issue #1058, opened 2017-11-03
emit event once initialised


pgte:
- **Version**: 0.26.0
- **Platform**: All
- **Subsystem**:

#### Type:

Enhancement

#### Severity:

Very Low

#### Description:

I want to support better offline use cases. I need the peerId to be able to start working locally, without needing for Swarm, Bitswap or Pubsub to consider themselves ready.

AFAIK, right now there is no event that tells me this. `init` is only emitted if the repo needs initialising.
`start` and `ready` are only emitted once the network layer is ready, which may not happen in my case.

Right now I'm polling the IPFS node for the `_peerInfo` property, but I would like to have a better alternative.

I would like to have an event that tells me once the peerId is known.

#### Steps to reproduce the error:

N/A

VictorBjelkholm:
> AFAIK, right now there is no event that tells me this. init is only emitted if the repo needs initialising.
start and ready are only emitted once the network layer is ready, which may not happen in my case.

This is incorrect regarding the `init` event. It's supposed to be emitted when the node is done initializing, not if it needs initializing. Things seem to work currently, check this example:

```
const ipfs = require('./src')

const node = new ipfs({repo: './repo-' + Math.random(), init: {bits: 32}, start: false})
node.once('ready', () => {
  console.log('ready')
  console.log(`current state is "${node.state._state}"`)
})
node.once('start', () => {
  console.log('started but it should have had started')
})
node.once('init', () => {
  console.log('initialized')
  console.log(`current state is "${node.state._state}"`)
})
```

This outputs the following for me:

```
$ node node-init-test.js
initialized
current state is "stopped"
ready
current state is "stopped"
```

However, I would expect `node._peerInfo` to actually have the peerInfo at this point, but it's `undefined`. So calling `node.id()` would not work.

VictorBjelkholm:
Ah, I think I misunderstood a tiny bit. You're saying that if the repo is already initiralized, the `init` event is never emitted, correct? That seems to be true, while the `ready` event is still emitted in that case.

pgte:
@VictorBjelkholm yes, that's the problem. If the node happens to be offline, the `ready` event is not emitted, and I need a peerId in order to be able to work locally...

wraithgar:
Made a PR to address this that goes down the path of correcting the `init` event behavior to emit when the node is done initializing, not only just when it has initialized a new repo.

Mr0grog:
[...]

PR #1293, opened 2018-04-02
fix: init event

wraithgar:
For https://github.com/ipfs/js-ipfs/issues/1058

The init event was only being emitted during initialization of
a new repo.  If one already existed that event was not firing.
This syncs up the behavior between these two situations so that they
log, set state, and emit identically.

### DIFF
--- a/src/core/boot.js
+++ b/src/core/boot.js
@@ -32,6 +32,7 @@ module.exports = (self) => {
       (cb) => {
         self.log('initialized')
         self.state.initialized()
+        self.emit('init');
         cb(null, true)
       }
     ], (err, res) => {
--- a/src/core/boot.js
+++ b/src/core/boot.js
@@ -32,7 +32,7 @@ module.exports = (self) => {
       (cb) => {
         self.log('initialized')
         self.state.initialized()
-        self.emit('init');
+        self.emit('init')
         cb(null, true)
       }
     ], (err, res) => {
--- a/src/core/components/init.js
+++ b/src/core/components/init.js
@@ -22,6 +22,7 @@ module.exports = function init (self) {
         return callback(err)
       }
 
+      self.log('initialized')
       self.state.initialized()
       self.emit('init')
       callback(null, res)


src/core/boot.js
```diff
@@ -32,6 +32,7 @@ module.exports = (self) => {
       (cb) => {
         self.log('initialized')
         self.state.initialized()
+        self.emit('init');
```
diasdavid: This will failing linting (semicolons)

wraithgar: Sorry about that. Linting already reports 87 problems for me in master so that one got lost in the noise.  Pushed a fix.

wraithgar:
@diasdavid I'd be glad to add tests but would need a little help to be pointed in the right direction. As far as I can see in this repo there isn't anything testing the events firing.  Am I just not looking in the right direction or would these be wholly new?

diasdavid:
See https://github.com/ipfs/js-ipfs/blob/master/test/core/create-node.spec.js

Mr0grog:
Isn’t the point of the `init` event to indicate that the *repo* was init’d? This completely changes the semantics of that event and removes a useful signal (you can’t tell from the outside whether `init` will happen; the boot process chooses to do so based on what it finds at the repo location). Also, it will emit *twice* here in the case where the repo does get init’d. That seems bad.

I feel like this event should be named something different that indicates what it is actually about. Maybe `prestart` or `prepared` or `repo-opened` or something?

wraithgar:
It won't fire twice here.

In boot.js, self.init is only called if maybeOpenRepo returns false, and the callback after the emitted init event returns true to the callback.

As far as the semantics of the init event, this is what I thought being discussed in https://github.com/ipfs/js-ipfs/issues/1058. I may have misunderstood.  It would be possible to just emit a 'peer' or `repo`event once the peer is set instead in the pre-start component.

Mr0grog:
> It won't fire twice here.

Ah, you’re right; I misread.

> As far as the semantics of the init event, this is what I thought being discussed in #1058.

I don’t think the semantics of the event really got any discussion there, though. Nobody actually said what the event was doing was wrong *or* unexpected — only that it didn’t fit @pgte’s use case.

I think there are two different things in that issue that make more sense to address:

- Is `ready` always emitting when constructing an IPFS node (assuming it didn’t have any errors)? If not, that’s probably a bug that needs fixing.

- @pgte’s specific ask was “I would like to have an event that tells me once the peerId is known.” Does the above with `ready` solve that? If not, what’s the right event to add? It’s probably something right where you added `init` here (or maybe inside `preStart`), I just think it needs different name because it’s a different thing :)

wraithgar:
If it's a new event it makes sense to do it right in pre-start after self._peerInfo is defined (because from that moment on calls to self.id() will return something).

wraithgar:
The matter of `ready` always firing also does not meet the use case, from what I can tell.  I think he simply wanted to know the moment that a peer was known so that operations that required one could start, without having to wait on the swarm et al operations to consider themselves ready.

Mr0grog:
Fair enough! But maybe useful to wait for feedback from @pgte and some discussion on the original issue to zero in on the right change before continuing with the implementation here?

PR #1294, opened 2018-04-03
Document constructor options more formally

Mr0grog:
This adds more descriptive and organized definitions for all the available options to the `IPFS` constructor as well as a full reference of events and the `start()` and `stop()` methods.

Some open questions here:

- I documented `start()` and `stop()`, but wasn’t sure which of the other methods are considered `public` that we should document.

- I used `node` to refer to the `IPFS` instance throughout example code and in references for methods (`start` and `stop`). That matches the pre-existing usage examples. However, the core API reference uses `ipfs.xyz`. I think we should make both the same for clarity. `node` sounds *slightly* better to me, but I do not have a strong opinion.

- I don’t know enough to say more than what we already had about the `EXPERIMENTAL.dht` option. Is there anything we can say about why someone might want to turn it on or off?

- I did not attempt to document all the `config` options, but I added links directly to the two default config files. Is that OK for now?

- I noticed that the `libp2p` option only supports `modules` and, inside that, only supports `transport` and `discovery` (not `connection` or `DHT`). You also can’t turn *off* any of the default transports or discovery mechanisms.
    - Are there plans to change that?
    - Should we add this to [`src/core/config.js`](https://github.com/ipfs/js-ipfs/blob/master/src/core/config.js) so it can be validated?

### DIFF
--- a/README.md
+++ b/README.md
@@ -602,6 +602,10 @@ HOME=~/.electron-gyp npm install
 
 If you find any other issue, please check the [`Electron Support` issue](https://github.com/ipfs/js-ipfs/issues/843).
 
+#### Have more questions?
+
+Ask for help in our forum at https://discuss.ipfs.io or in IRC (#ipfs on Freenode).
+
 ## Running js-ipfs with Docker
 
 We have automatic Docker builds setup with Docker Hub: https://hub.docker.com/r/ipfs/js-ipfs/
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -183,70 +184,150 @@ node.on('ready', () => {
...
+#### `node.stop([callback])`
 
-node.on('init', () => {})     // Node has successfully finished initing the repo
-node.on('start', () => {})    // Node has started
-node.on('stop', () => {})     // Node has stopped
-```
+Close and stop listening for connections with other IPFS nodes, then release access to the node’s repo.
 
-### [Tutorials and Examples](/examples)
+**Returns a promise** that resolves when the node is completely stopped. You can also pass a callback like `(error) => {...}` or listen for the `stop` event to determine when the node has stopped.
```
diasdavid: Returns a promise if no callback is passed

Mr0grog:
How’s https://github.com/Mr0grog/js-ipfs/blob/43123833af0f1eaebd2b74ba103e12b3d138a185/README.md#nodestartcallback ?

diasdavid:
LGTM, thanks @Mr0grog :)

diasdavid:
@Mr0grog can we have an equivalent PR to js-ipfs-api? It would be good to have both in sync

Mr0grog:
Maybe someone could use this as an example to start from? It took a pretty good chunk of my day to chase through all the code paths and feel confident I was explaining each thing accurately. I need to preserve some time to actually work on my main focus over in `ipfs/docs`!

diasdavid:
Thank you for your contribution, @Mr0grog ❤️

PR #1297, opened 2018-04-04
feat: Provide access to bundled libraries when in browser

vasco-santos:
In this PR, the bundled libraries mentioned in [406](https://github.com/ipfs/js-ipfs-api/issues/406), as well as [ipfs/js-ipfs#525 (comment)](https://github.com/ipfs/js-ipfs/pull/525#issuecomment-260306602) were added the same way as in `js-ipfs-api`.

For this issue, other 2 PRs were created

- [js-ipfs-api](https://github.com/ipfs/js-ipfs-api/pull/732)
- [interface-ipfs-core](https://github.com/ipfs/interface-ipfs-core/pull/252)


alanshaw:
Just wondering why these types are properties of an instance and not of the constructor/separate export?

vasco-santos:
Hi @alanshaw 

At first, I was thinking on a separate export, as you suggest. It is described [here](https://github.com/ipfs/js-ipfs-api/issues/406#issuecomment-378219501) why the solution was the implemented.

alanshaw:
hmm, I'm still not sure I understand the reasoning.

The `util` on `js-ipfs-api` instances contained functions that operate on the instance you've created - that makes sense. Putting non-instance related types/functions in an instance is a little odd. i.e. getting hold of CID: `new IPFS().types.CID` as opposed to `const { CID } = IPFS.types`.

Needless to say, if someone wanted to do something with CID _before_ creating their IPFS node then they can't.

diasdavid:
Let's get a decision on this asap to get both interfaces consistent. I'm fine with having these as static references (meaning, available through class and instance)

diasdavid:
I agree with @alanshaw. What about having both? @alanshaw wanna submit a separate PR to add the types as static exports too?

mitra42:
Did something stall this, I agree with @alanshaw that having this on static is significantly more useful, as we have code that doesn't easily have access to the ipfs node that makes this check by reimporting the "cids"

PR #856, opened 2017-05-19
Awesome Endeavour: DHT Part II

daviddias:
This PR is the second part of the DHT integration in js-ipfs

# Remaining tasks

Main work tracker: https://github.com/libp2p/js-libp2p-dht/pull/1

- [x] Make usage of the DHT efficient (avoid doing multiple query calls for the same file)
  - [x] add .getMany js-ipfs-bitswap - https://github.com/ipfs/js-ipfs-bitswap/issues/163
  - [x] add .getMany js-ipfs-block-service - https://github.com/ipfs/js-ipfs-block-service/issues/79 https://github.com/ipfs/js-ipfs-block-service/pull/81
  - [x] add .get(<cid array>) to js-ipld https://github.com/ipld/js-ipld/issues/132
  - [x] add .get(<cid array>) from the DAG API to js-ipfs-unixfs-engine - https://github.com/ipfs/js-ipfs-unixfs-engine/issues/216
- [x] interface-ipfs-core
  - [x] implement remaining DHT methods in js-ipfs-api
  - [x] Update tests and spec, ensure that both pass in js-ipfs and js-ipfs-api
  - [x] Augment the test suite and spec
    - [x] first pass https://github.com/ipfs/interface-ipfs-core/pull/288
  - [x] Improve the API https://github.com/ipfs/interface-ipfs-core/issues/136 (non critical)
- [x] js-libp2p
  - [x] Make configuring a Node nicer and support multiple Peer Routing + Content Routing systems https://github.com/libp2p/js-libp2p/pull/166
  - [x] libp2p.next() https://github.com/libp2p/js-libp2p/issues/159 (parts needed are shipped)
  - [x] js-libp2p-kad-dht
    - [x] Implement it
    - [x] Integrate in js-libp2p js-libp2p https://github.com/libp2p/js-libp2p/pull/86
    - [x] Integrate in js-libp2p-ipfs-nodejs https://github.com/ipfs/js-libp2p-ipfs-nodejs
    - [x] ~~Integrate libp2p-ipfs-browser~~ will only add DHT there after Relay is in
    - [x] Make random walk be a discovery service https://github.com/libp2p/js-libp2p-kad-dht/issues/7
  - [x] Delegated Routing https://github.com/libp2p/js-libp2p/issues/120
- [x] Implement .dht in js-ipfs
  - [x] core api
  - [x] http-api
  - [x] cli
- [x] interop tests [ipfs/interop#36](https://github.com/ipfs/interop/pull/36)
[...]

PR #1299, opened 2018-04-06
feat: ping

### LINKED ISSUE #928: Implement the `jsipfs ping` command

JGAntunes:
Just opening this for visibility, working on finishing the ping endeavour started at #1202.

- [x]  Dependant on https://github.com/libp2p/js-libp2p-ping/pull/69
- [x]  Missing tests

Also, some relevant notes. Seems like hapijs has compression enabled by default, which means responses  will take into account the `Accept-Encoding` header and will use gzip to compress responses (it seems hapi has no way of disabling compression for a specific route only also). From what I've seen the `go-ipfs` http api doesn't gzip responses by default AFAIK. Anyway, there seems to be an issue when compression is enabled and we're streaming responses - https://github.com/hapijs/hapi/issues/3599 - as it seems the response is actually buffered, compressed and only then it is sent to the client. This results in the actual "streaming" capabilities being lost (the response all at once and only when it is complete). Not sure if this is a relevant issue or not, but though it was worth mentioning since when using the http API from a browser, the `Accept-Encoding` headers will be set and the resulting response will not be streamed.

Closes #928.

### DIFF
--- a/src/cli/commands/ping.js
+++ b/src/cli/commands/ping.js
@@ -1,0 +1,44 @@
+'use strict'
+
+const print = require('../utils').print
+
+module.exports = {
+  command: 'ping <peerId>',
+
+  describe: 'Measure the latency of a connection',
+
+  builder: {
+    count: {
+      alias: 'n',
+      type: 'integer',
+      default: 10
+    }
+  },
+
+  handler (argv) {
+    const peerId = argv.peerId
+    const count = argv.count || 10
+
+    print('PING ' + peerId)
+
+    let noOfTimes = 0
+    let totalTime = 0
+
+    const pingCb = (err, p) => {
+      if (err) {
+        throw err
+      }
+      let time = p.Time
+      totalTime = totalTime + time
+      noOfTimes = noOfTimes + 1
+      print('Pong received: time=' + time + ' ms')
+      if (noOfTimes === count) {
+        print('Average latency: ' + totalTime / count + 'ms')
+      }
+    }
+
+    for (let i = 0; i < count; i++) {
+      argv.ipfs.ping(peerId, pingCb)
+    }
+  }
+}
--- a/src/cli/commands/ping.js
+++ b/src/cli/commands/ping.js
@@ -26,7 +26,7 @@ module.exports = {
     pull(
       argv.ipfs.pingPullStream(peerId, { count }),
       pullCatch(err => {
-        throw err 
+        throw err
       }),
       drain(({ Time, Text }) => {
         // Check if it's a pong
--- a/src/cli/commands/ping.js
+++ b/src/cli/commands/ping.js
@@ -22,7 +22,6 @@ module.exports = {
   handler (argv) {
     const peerId = argv.peerId
     const count = argv.count || 10
-
     pull(
       argv.ipfs.pingPullStream(peerId, { count }),
       pullCatch(err => {
--- a/src/cli/commands/ping.js
+++ b/src/cli/commands/ping.js
@@ -9,7 +9,7 @@ const print = require('../utils').print
 module.exports = {
   command: 'ping <peerId>',
 
-  describe: 'Measure the latency of a connection',
+  description: 'Measure the latency of a connection',
 
   builder: {
     count: {
--- a/src/cli/commands/ping.js
+++ b/src/cli/commands/ping.js
@@ -1,5 +1,9 @@
 'use strict'
 
+const pull = require('pull-stream/pull')
+const drain = require('pull-stream/sinks/drain')
+const pullCatch = require('pull-catch')
+
 const print = require('../utils').print
 
 module.exports = {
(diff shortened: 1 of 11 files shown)


diasdavid:
I believe that meanwhile @vasco-santos started tackling this one as well. @JGAntunes mind syncing up with him to avoid dup work and find a converged solution? Thank you both!

JGAntunes:
Sure, @vasco-santos I'm currently finishing the tests for this and also doing some cleanup. I needed something to test the connections between 2 IPFS nodes hence why I'm working on this. Let me know what's the status on this from your side, I'll gladly drop this if you have something ready already :+1: or we can also work together if you like.

vasco-santos:
Hello @JGAntunes 

I started this on Friday, but I didn't finish it yet. I was implementing using the pull stream (as you did  too), but I didn't notice that you had created https://github.com/libp2p/js-libp2p-ping/pull/69, which would be useful for me 😄After finishing that, I would have to add the tests, as well. Therefore, I think you can keep with this PR.

I cloned your repo, and linked your PR to `js-libp2p-ping` and tested your implementation and it seems good to me, good job!

Just several notes that I was encompassing in my solution:

- Update README.md (ping parameters)
- Update [interface-ipfs-core](https://github.com/ipfs/interface-ipfs-core) spec with ping.

src/cli/commands/ping.js
```diff
@@ -0,0 +1,41 @@
...
+const pull = require('pull-stream/pull')
+const drain = require('pull-stream/sinks/drain')
+const pullCatch = require('pull-catch')
+
+const print = require('../utils').print
+
+module.exports = {
+  command: 'ping <peerId>',
+
+  describe: 'Measure the latency of a connection',
```
vasco-santos: There are several command definitions using `describe` and others using `description`. 

According to [yargs docs](https://github.com/yargs/yargs/blob/master/docs/api.md#completioncmd-description-fn), I think that the correct usage in this case is `description`. In spite of, `describe` working, it is aimed to another use cases, as documented [here](https://github.com/yargs/yargs/blob/master/docs/api.md#describe).

What do you think @JGAntunes and @diasdavid ?

diasdavid: fine either way, pick only one though, consistency is key.

JGAntunes:
Hey @diasdavid, @vasco-santos . I was looking to finish this asap but I have some questions regarding tests. Noticed the `interface-ipfs-core` is responsible for running tests for the core components, seems though that these aren't being run for any of the miscellaneous operations I think - https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/miscellaneous.js - https://github.com/ipfs/js-ipfs/tree/master/test/core/interface, is this by design or is it just something that's lacking?

There seems to be an issue when running a `ping` command with a node with whom we don't have an established connection yet -  https://github.com/libp2p/js-libp2p-ping/issues/70 - hence the need for this on tests - https://github.com/ipfs/js-ipfs/pull/1299/files#diff-9478661c5bacfe4dd5f87a3b88574ca9R69

As for the http-api tests I've noticed you have both using the inject interface from hapi and `js-ipfs` as a client. What's the angle on this? Should both be used ideally? If so where will the major differences rely?

diasdavid:
> However, when I was implementing this, I only achieved a successful ping after connecting to the peer, as well. I don't know if this is the intended behavior or not.

That is because js-ipfs doesn't have the DHT by default, which gives peers Peer Routing. Learn more about it here:

- https://github.com/ipfs/js-ipfs/pull/856
- https://github.com/libp2p/js-libp2p/tree/master/examples/peer-and-content-routing

diasdavid:
@JGAntunes @vasco-santos merged and released https://github.com/libp2p/js-libp2p-ping/pull/69, please update https://github.com/libp2p/js-libp2p and make sure that all tests pass, then we can update all the deps here :)

JGAntunes:
Sorry for delay on this everyone. I think the only thing missing now are the tests on the HTTP API using `js-ipfs-api`. Meanwhile, during ping I know try to use `peerRouting.findPeer` which will fail accordingly if the DHT experimental feature is not enabled:
```
{"Success":true,"Time":0,"Text":"Looking up peer QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z"}
{"Success":false,"Time":0,"Text":"Error: DHT is not available"}
```

Will add the missing tests between today and tomorrow. Let me know if is anything else missing.

JGAntunes:
Added tests for both DHT enabled and disabled. However couldn't find a workaround for this - https://github.com/ipfs/js-ipfs/pull/1299/files#diff-0e68633477fac5c3c897f68c6b28f7a9R187

This is also dependent on https://github.com/ipfs/js-ipfs/pull/1340

Let me know what you think @diasdavid @vasco-santos

diasdavid:
Merging this PR into a branch on the main repo so that I can update and test things.

PR #1274, opened 2018-03-21
cli: add IPFS_PATH info to init command help

bvand:
Connects to https://github.com/ipfs/js-ipfs/issues/1213.

I tried using .usage() from the yargs package, but it just seems to be a different way of doing normal yargs configuration (no special ability to add a prologue).

PR #1302, opened 2018-04-09
fix: Add ipfs path to cli help

### LINKED ISSUE #1213: Add IPFS_PATH to CLI help
It's possible to init a repository at a different location (other than the default `~/.jsipfs`) with setting the `IPFS_PATH` environment variable. This isn't clear from the CLI help. `go-ipfs` has this information:

```
ipfs uses a repository in the local file system. By default, the repo is
located at ~/.ipfs. To change the repo location, set the $IPFS_PATH
environment variable:

    export IPFS_PATH=/path/to/ipfsrepo
```

Add this to the js-ipfs CLI as well.

vasco-santos:
Aiming to have a consistent menu with `go-ipfs`, I added the `IPFS_PATH` as it is presented in Go implementation.

```
ipfs uses a repository in the local file system. By default, the repo is
located at ~/.ipfs. To change the repo location, set the $IPFS_PATH
environment variable:

    export IPFS_PATH=/path/to/ipfsrepo
```

In PR https://github.com/ipfs/js-ipfs/pull/1274, it was added to `jsipfs init --help`, but it was presented before the options description, instead of in the end.

In this solution, I changed the previous implementation. Using `yargs epilog`, I added the `IPFS_PATH` message to the epilogue which is displayed for the following cases:

- `jsipfs`
- `jsipfs --help`
- `jsipfs init --help`
- `jsipfs daemon --help`

Closes #1213

### DIFF
--- a/src/cli/utils.js
+++ b/src/cli/utils.js
@@ -111,3 +111,7 @@ exports.rightpad = (val, n) => {
   }
   return result
 }
+
+exports.ipfsPathHelp = 'ipfs uses a repository in the local file system. By default, the repo is ' +
+  'located at ~/.jsipfs. To change the repo location, set the $IPFS_PATH environment variable:\n\n' +
+  'export IPFS_PATH=/path/to/ipfsrepo\n'
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -29,6 +29,7 @@ const cli = yargs
     default: ''
   })
   .commandDir('commands')
+  .epilog(utils.ipfsPathHelp)
   .demandCommand(1)
   .fail((msg, err, yargs) => {
     if (err) {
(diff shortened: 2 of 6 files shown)


src/cli/utils.js
```diff
@@ -111,3 +111,7 @@ exports.rightpad = (val, n) => {
   }
   return result
 }
+
+exports.ipfsPathHelp = 'ipfs uses a repository in the local file system. By default, the repo is ' +
+  'located at ~/.jsipfs.To change the repo location, set the $IPFS_PATH environment variable:\n\n' +
```
vmx: Missing whitespace between `~/.jsipfs.` and `To change`.

vasco-santos: Good catch! Fixed

Issue #763, opened 2017-02-10
🌟 Parity to IPFS


daviddias:
> Update: I rewrote this first post to make it more useful for tracking the developments of this.

Goal: Add a 'Storage Backend' for js-ipfs that backed by Parity, the Rust implementation of Ethereum. This will enable js-ipfs nodes to fetch content from the Ethereum Network as if it was just a big bucket with all the blocks from the chain.

TASKS

- [x] rust-cid
- [x] Implement an http endpoint in Parity that lets us retrieve the blocks
- [ ] Implement the datastore backend that mounts on top of parity
- [ ] Try it out, a lot :)

kumavis:
from parity `mh-ipfs branch`
```
cargo build --release
./target/release/parity
```
while thats running, you can make requests by cid
```
curl 'http://localhost:5001/api/v0/block/get?arg=z45oqTRzb9a5xyvx5RbfSXH1K5jibyZ4AxnXyYReuLw7KU5veYw'
```

kumavis:
i made this util to help with this https://github.com/kumavis/eth-ipld

diasdavid:
@kumavis let's pick up on this again, now that we have datastore bubbling up all the way :)

kumavis:
whee~

kumavis:
seems simple enough

https://github.com/ipfs/js-datastore-level
https://github.com/ipfs/interface-datastore

how do you mount multiple stores? and can you mount one as read-only?

kumavis:
also - do you have a preferred isomorphic xhr module? otherwise ill use what i usually do

diasdavid:
We shim `stream-http` in js-ipfs-api https://github.com/ipfs/js-ipfs-api/blob/master/package.json#L10

The code for reaching to the block store, given that it exposes the same API as go-ipfs does for blocks, is exactly what js-ipfs-api does https://github.com/ipfs/js-ipfs-api/blob/master/src/api/block.js

On `read-only`, it means that essentially when it calls write, you say 'All good folks', but in reality does nothing.

Right now, ipfs-repo is mounting a single datastore for the blockstore (blocks folder) https://github.com/ipfs/js-ipfs-repo/blob/master/src/index.js#L136-L139, which is them used throughout blockstore code: https://github.com/ipfs/js-ipfs-repo/blob/master/src/blockstore.js#L34-L137

Tasks:

- [ ] Ability to mount more than one datastore for blockstore (make it an array essentially)
[...]

PR #1303, opened 2018-04-09
Create an example for configuring the ipfs repo

jacobheun:
Also demonstrates how to customize the repo lock, which is a new addition

depends on ipfs/js-ipfs-repo#162


examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,78 @@
...
+  },
+
+  // false will disable locking, you can also pass in a custom locker
+  locker: false
+}
+
+// Initialize our repo and IPFS node
+const myRepo = new Repo('/tmp/custom-repo/.ipfs', customRepositoryOptions)
+const node = new IPFS({
+  repo: myRepo
```
diasdavid: @jacobheun I think it makes sense to remove the step of requiring 'ipfs-repo' and set it up manually by just adding a `repoOptions` object to the IPFS constructor options that lets us pass the storagebackends directly.

diasdavid:
@Mr0grog mind reviewing this tutorial?

diasdavid:
@kumavis @hermanjunge witht his addition, it means that a user should be able to plug the parity to IPFS bridge pretty easily. @vmx, you have experience in shipping npm modules that pull rust code, compile it and expose bindings to it, can you help us get the Parity bridge in a nice package?

jacobheun:
@diasdavid I updated the IPFS constructor to accept `repoOptions` and the example is updated to leverage that. Let me know if this is what you were thinking, I think it's cleaner.

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+const series = require('async/series')
+const IPFS = require('ipfs')
+
+// Create our custom options
+const customRepositoryOptions = {
+
+  /**
+   * Storage Backends are fully customizable. Each backend can be stored in seperate services,
+   * or in a single service. Options can be passed into the datastores via the storageBackendOptions
```
Mr0grog: It’s not really the *backend* that’s stored in a service, right? Maybe something like this would be a little clearer?

> IPFS nodes store different information in separate `storageBackends`, or datastores. Each storage backend can use the same type of datastore or a different one — you could store your `keys` in a levelDB database while everything else is in files, for example. (See https://github.com/ipfs/interface-datastore for more about datastores.)

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,77 @@
...
+   * property shown below.
+   */
+  storageBackends: {
+    root: require('datastore-fs'), // version and config data will be saved here
```
Mr0grog: This little description is super helpful. Maybe you could add something similar for the rest of these? Alternatively, maybe you could add some of this info to the docs for `js-ipfs-repo`.

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,77 @@
...
+   * https://github.com/ipfs/js-datastore-s3/tree/master/examples/full-s3-repo
+   */
+  storageBackendOptions: {
+    blocks: {
+      sharding: true,
+      extension: '.data'
+    }
+  },
+
+  // false will disable locking, you can also pass in a custom locker
```
Mr0grog: This comma should be a semicolon.

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,77 @@
...
+    content: Buffer.from('We are using a customized repo!')
+  }, (err, filesAdded) => {
+    if (err) {
+      return cb(err)
+    }
+
+    console.log('\nAdded file:', filesAdded[0].path, filesAdded[0].hash)
+    fileMultihash = filesAdded[0].hash
+    cb()
+  }),
```
Mr0grog: If this were split into steps and you used `async.seq` instead of `async.series`:

```js
(cb) => node.version(cb),
(version, cb) => {
  console.log('Version:', version.version)
  cb()
},
(cb) => node.files.add({
  path: 'test-data.txt',
  content: Buffer.from('We are using a customized repo!')
}, cb),
(filesAdded, cb) => {
  console.log('\nAdded file:', filesAdded[0].path, filesAdded[0].hash)
  cb(null, filesAdded[0].hash)
},
(fileMultiHash, cb) => node.files.cat(fileMultihash, cb),
// etc.
```

You could get rid of the error handling logic so it reads a bit clearer for people trying to follow it. This would remove the need for the `fileMultihash` global, too.

Alternatively, using the promise-based return values would let you do this without a special dependency:

```js
node.on('ready', () => {
  node.version()
    .then(version => console.log('Version:', version.version)
    .then(() => node.files.add({
      path: 'test-data.txt',
      content: Buffer.from('We are using a customized repo!')
    })
    .then(filesAdded => {
      console.log('\nAdded file:', filesAdded[0].path, filesAdded[0].hash)
      return filesAdded[0].hash
    })
    .then(fileMultihash => node.files.cat(fileMultihash))
    // etc.
})
```

Mr0grog: It might also be good to catch any errors at the end whether you use `series`, `seq`, or promises, just to demonstrate good practice.

src/core/index.js
```diff
@@ -43,7 +43,7 @@ class IPFS extends EventEmitter {
 
     if (typeof options.repo === 'string' ||
         options.repo === undefined) {
-      this._repo = defaultRepo(options.repo)
+      this._repo = defaultRepo(options.repo, options.repoPath)
```
Mr0grog: This is a typo, right? In the examples, you set `options.repoOptions`.

If you are going to change the API by adding a new option, would you please document it with [the rest of the constructor options in the README](https://github.com/ipfs/js-ipfs/blob/master/README.md#ipfs-constructor)?

Personally, I’m also a little unsure about the value of this. It’s basically just replacing:

```js
new IPFS({
  repo: new Repo('/some/path', {options}
})
```

with

```js
new IPFS({
  repo: '/some/path',
  repoOptions: {options}
})
```

You don’t have to explicitly import `ipfs-repo` anymore to get a fully customizable repo, but the whole behavior is more complicated as a result (`repoOptions` won’t do anything if you don’t also set `repo` or if you set `repo` to an actual repo instance). There are also now multiple ways to do the same repo customization. I think this makes the API a little less clear.

That said, I’m more here to focus on docs, so I’ll try not to push or say any more on this.

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,77 @@
...
+  (cb) => node.files.cat(fileMultihash, (err, data) => {
+    if (err) {
+      return cb(err)
+    }
+
+    console.log('\nFetched file content:')
+    process.stdout.write(data)
+    cb()
+  })
```
Mr0grog: I wasn’t thinking about this yesterday, but it might also be useful to inspect the filesystem after all these operations and demonstrate that the custom repo options were used. You’d probably want to set options at the top here that aren’t the same as the defaults so their effects are visible, e.g. `extension: '.customblocks'` or something.

src/core/index.js
jacobheun: This is a really good point and I agree. The system has a lot of complexity, so avoiding more options in the API is a good thing in my opinion. I think it also helps delineate using the default repo, versus a custom repo, more clearly.

examples/custom-ipfs-repo/index.js
jacobheun: I think the promise structure is much easier to consume and that's a great point about the error catching. I'll include those changes in the next update I push.

examples/custom-ipfs-repo/custom-locker.js
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Creates the lock. This can be overriden to customize where the lock should be created
+   *
+   * @param {string} dir
+   * @param {function(Error, S3Locker)} callback
+   * @returns {void}
+   */
+  lock (dir, callback) {
+    this.lockPath = this.getLockfilePath(dir)
```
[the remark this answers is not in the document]
jacobheun: This is a good catch. While the system currently only manages a single lock file the existing API in memory and fs don't prohibit multiple locks. I'll get this and the `locked` call added.

examples/custom-ipfs-repo/README.md
```diff
@@ -0,0 +1,24 @@
+# Customizing the IPFS Repo
+
+> This example shows you how to customize your repository, including where your data is stored and how the repo locking is managed.
```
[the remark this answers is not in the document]
Mr0grog: No problem! Glad it was helpful :)

jacobheun:
@Mr0grog I've incorporated your feedback and I've included the latest lock changes from ipfs/js-ipfs-repo#162.

examples/custom-ipfs-repo/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+> npm install
+> npm start
+```
+
+## Other Options
+
+### Custom Repo Lock
+This example sets the repo locker to `false`, preventing any locking from happening. If you would like to control how locking happens, such as with a centralized S3 IPFS Repo, you can pass in your own custom lock. See [custom-lock.js](./custom-lock.js) for an example of a custom lock that can be used for [datastore-s3](https://github.com/ipfs/js-datastore-s3).
```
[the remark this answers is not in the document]
jacobheun: @Mr0grog good catch! I've pushed up this change along with the `node.stop` call.

jacobheun:
@Mr0grog when you have some time can you check the final adjustments I made per your feedback? Thanks!

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,95 @@
...
+    .then((filesAdded) => {
+      console.log('\nAdded file:', filesAdded[0].path, filesAdded[0].hash)
+      return node.files.cat(filesAdded[0].hash)
+    })
+    // Print out the files contents to console
+    .then((data) => {
+      console.log('\nFetched file content:')
+      process.stdout.write(data)
+      console.log('\n\nStopping the node')
+      return node.stop()
```
Mr0grog: Minor technical nit: if you put this *after* the `catch`, then it will still run if there was an error, which is probably desirable:

```js
.then((data) => {
  console.log('\nFetched file content:')
  process.stdout.write(data)
})
// Log out the error, if there is one
.catch((error) => {
  console.log('File Processing Error:', error)
})
.then(() => {
  console.log('\n\nStopping the node')
  return node.stop()
})
```

Otherwise, this all looks lovely! 👍 👍 👍

jacobheun: Good point, I've squashed the change into the previous commit and added a comment about not needing to catch errors on the stop.

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,99 @@
...
+   */
+  storageBackendOptions: {
+    root: {
+      extension: '.ipfsroot', // Used by datastore-fs; Appended to all files
+      errorIfExists: false, // Used by datastore-fs; If the datastore exists, don't throw an error
+      createIfMissing: true // Used by datastore-fs; If the datastore doesn't exist yet, create it
+    },
+    blocks: {
+      sharding: false, // Used by IPFSRepo Blockstore to determine sharding; Ignored by datastore-fs
+      extension: '.ipfsblock',
```
diasdavid: It's currently just `.data`, let's avoid confusing users.

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,99 @@
...
+      errorIfExists: false,
+      createIfMissing: true
+    },
+    keys: {
+      extension: '.ipfskey',
```
diasdavid: Does the keys repo use an extension?

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Storage Backend Options will get passed into the instantiation of their counterpart
+   * in `storageBackends`. If you create a custom datastore, this is where you can pass in
+   * custom constructor arguments. You can see an S3 datastore example at:
+   * https://github.com/ipfs/js-datastore-s3/tree/master/examples/full-s3-repo
+   */
+  storageBackendOptions: {
+    root: {
+      extension: '.ipfsroot', // Used by datastore-fs; Appended to all files
```
diasdavid: No extension here. Check:

```
> ls ~/.jsipfs
blocks    config    datastore keys      repo.lock version
```

examples/custom-ipfs-repo/index.js
```diff
@@ -0,0 +1,99 @@
...
+      errorIfExists: false,
+      createIfMissing: true
+    },
+    keys: {
+      extension: '.ipfskey',
+      errorIfExists: false,
+      createIfMissing: true
+    },
+    datastore: {
+      extension: '.ipfsds',
```
diasdavid: No extension used by default

jacobheun: The reason I added the non-default options was to demonstrate that configuration options to the users. Perhaps I could add in some comments about the defaults here and mention that they are being overridden for demonstration purposes only?

jacobheun:
@diasdavid js-datastore-s3 has a working example of how to use s3 as a full backend, and we reference that in the comments in this example. I think it would be good to avoid duplication of that example to avoid out of date examples as the project evolves, but maybe I can call out that example more prevalently in the README?

diasdavid:
@jacobheun please do, it was not evident to me and it is one of the main reasons on why to upgrade all repos are configured

examples/custom-ipfs-repo/index.js
Mr0grog: Yeah, this would probably be useful to call that out here.

It might also be good to, in the last step of the script, do something like:

```js
.then(() => {
  console.log('Check "/tmp/custom-repo/.ipfs" to see what your customized repository looks like on disk.')
}
```

Mr0grog:
> [the S3 repo] is one of the main reasons on why to upgrade all repos are configured

I don’t think that’s correct — it’s not even used in the example at all. The point of changing all the repo configurations is to make it obvious that what you did affected the output on disk. You can’t see the results if you just set everything to their defaults.

jacobheun:
I've added in some additional comments and updated the readme to try and spell out the differences with the defaults more clearly.

I also updated the full s3 example in the datastore-s3 repo, ipfs/js-datastore-s3#5, to use the S3Lock instead of memory, and fixed a bug with that lock here. That new update has been tested against one of my S3 buckets.

diasdavid:
@Mr0grog what I meant was that this endeavor, "upgrading the ipfs-repo module so that it supports custom locking and update js-ipfs to support custom repos" was started because we wanted to enable users to be able to use an S3 bucket as the storage backend.

Most users won't really need to touch the repo if they are just going to use regular fs. Most users, if not all, will trust the defaults. What is really interesting is to have multiple storage backends, from S3 (requested multiple times by IPFS users) and things like #763.

PR #1305, opened 2018-04-10
feat(bitswap.unwatch) expose bitswap.unwatch to cli and http api

wraithgar:
Does this mean we want to add a bitswap spec to interface-ipfs-core?

The go cli does not print anything on success as far as I can tell: https://github.com/ipfs/go-ipfs/blob/11ce0445cb24b6b99613223fc6d8128c6648a4df/core/commands/bitswap.go#L78

I can remove the print statement in this cli implementation and just test that no error happened instead if we want the output to match too.

### DIFF
--- a/src/cli/commands/bitswap/unwant.js
+++ b/src/cli/commands/bitswap/unwant.js
@@ -1,11 +1,25 @@
 'use strict'
 
+const print = require('../../utils').print
+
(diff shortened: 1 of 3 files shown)


wraithgar:
Just realized I still have this configured to use my fork. Will fix this to go from local branch

PR #1005, opened 2017-09-08
feat: support specify hash algorithm in files.add

alanshaw:
### Specify multihash hashing alogrithm

This allows the optional param `--hash` (CLI) or `?hash=` (HTTP API) to be passed to `files.add` in order to specify the hash algorithm that should be used when generating the multihash(es) for the added content.

Passing this option will set the CID version option to `true`, as per the Go implementation.

### Change `object/get` and `object/data` to accept CIDs

This brings js-ipfs more in line with go-ipfs. Where both `object/get` and `object/data` accept a CID. It's also needed by `js-ipfs-api` when adding files (see further discussion in https://github.com/ipfs/js-ipfs-api/pull/597).

~~**requires https://github.com/ipfs/js-ipfs-unixfs-engine/pull/185**~~
~~**requires https://github.com/ipfs/js-ipfs-api/pull/597**~~

PR #1308, opened 2018-04-12
feat: Add support for specifying hash algorithms in files.add

vasco-santos:
Since [PR 1005](https://github.com/ipfs/js-ipfs/pull/1005/) has been blocked for a while, I went through it. 

As the PR origin is from an external organization, I created a new branch instead of rebasing it in its origin. Shortly, I got the @alanshaw (thanks :D ) implementation with `js-ipfs` master branch and made a few changes, as a consequence of the modifications that occurred meanwhile in `components/objects`, as well as in `resources` and a minor refactor. 

Finally, I added tests for several hash algorithms, since there are more than [300](https://github.com/multiformats/js-multihash/blob/master/src/constants.js). Should I add more to the tests?

### DIFF
--- a/test/cli/files.js
+++ b/test/cli/files.js
@@ -7,7 +7,21 @@ const expect = require('chai').expect
 const path = require('path')
 const compareDir = require('dir-compare').compareSync
 const rimraf = require('rimraf').sync
+const CID = require('cids')
+const mh = require('multihashes')
 const runOnAndOff = require('../utils/on-and-off')
+
+// TODO: Test against all algorithms Object.keys(mh.names)
+// This subset is known to work with both go-ipfs and js-ipfs as of 2017-09-05
+const HASH_ALGS = [
+  'sha1',
+  'sha2-256',
+  'sha2-512',
+  'keccak-224',
+  'keccak-256',
+  'keccak-384',
+  'keccak-512'
+]
 
 describe('files', () => runOnAndOff((thing) => {
   let ipfs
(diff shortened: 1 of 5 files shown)


test/cli/files.js
```diff
@@ -7,8 +7,22 @@ const expect = require('chai').expect
...
+// This subset is known to work with both go-ipfs and js-ipfs as of 2017-09-05
+const HASH_ALGS = [
+  'sha1',
+  'sha2-256',
+  'sha2-512',
+  'keccak-224',
+  'keccak-256',
+  'keccak-384',
+  'keccak-512'
+]
```
diasdavid: @vasco-santos good enough :)

test/cli/files.js
```diff
@@ -7,8 +7,22 @@ const expect = require('chai').expect
 const path = require('path')
 const compareDir = require('dir-compare').compareSync
 const rimraf = require('rimraf').sync
+const CID = require('cids')
+const mh = require('multihashes')
 const runOnAndOff = require('../utils/on-and-off')
 
+// TODO: Test against all algorithms Object.keys(mh.names)
```
diasdavid: No need :)

JonKrone:
Hey @vasco-santos. I've been working with this this weekend and have run into a few pieces that I'd love your thoughts on:

go/js ipfs hash differences:
image
Shouldn't the hashes be the same?

DAG traversal
image

It seems that the DAG commands are not aware that a different hash function was used. They may be casting all hashes to `dag-pb` with something like `new CID(0, 'dag-pb', <hash from specific algorithm>)`. So you can request the specific DAGNode if you know it's hash but it's not possible to traverse the DAGLinks of something added with a non-default hash alg, as you might with large files or folders.

Do you know what's wrong here? Wanted to run this by you before creating an issue.

Issue #1316, opened 2018-04-19
Two peers with same Identity


vmx:
If you create two nodes in parallel, then they end up having the same Identity. When the script below is run you end up with configuration files which contain the same Identity:

```diff
$ diff /tmp/peera/config /tmp/peerb/config
4,5c4,5
<       "/ip4/0.0.0.0/tcp/4012",
<       "/ip4/127.0.0.1/tcp/4013/ws"
---
>       "/ip4/0.0.0.0/tcp/4022",
>       "/ip4/127.0.0.1/tcp/4023/ws"
7,8c7,8
<     "API": "/ip4/127.0.0.1/tcp/5012",
<     "Gateway": "/ip4/127.0.0.1/tcp/9091"
---
>     "API": "/ip4/127.0.0.1/tcp/5022",
>     "Gateway": "/ip4/127.0.0.1/tcp/9092"
```

I tried this on js-ipfs master (64c3bfb45cda27e268655cc9a2bce28eebd09454).

I debugged it a bit, but then I hit a dead end. I don't really understand how the initial configuration is created. It seems that some default configuration is written to disk, which is then overridden.

```JavaScript
'use strict'

const IPFS = require('ipfs')


const configA = {
  "Addresses": {
    "API": "/ip4/127.0.0.1/tcp/5012",
    "Gateway": "/ip4/127.0.0.1/tcp/9091",
    "Swarm": [
      "/ip4/0.0.0.0/tcp/4012",
      "/ip4/127.0.0.1/tcp/4013/ws"
    ]
  },
  "Bootstrap": []
}
const peerA = new IPFS({
  repo: '/tmp/peera',
  config: configA,
  init: {
    emptyRepo: true
  }
})
peerA.on('ready', () => {
  console.log('peerA ready')
})

const configB = {
  "Addresses": {
    "API": "/ip4/127.0.0.1/tcp/5022",
    "Gateway": "/ip4/127.0.0.1/tcp/9092",
    "Swarm": [
      "/ip4/0.0.0.0/tcp/4022",
      "/ip4/127.0.0.1/tcp/4023/ws"
    ]
  },
  "Bootstrap": []
}
const peerB = new IPFS({
  repo: '/tmp/peerb',
  config: configB,
  init: {
    emptyRepo: true
  }
})
peerB.on('ready', () => {
  console.log('peerB ready')
})
```

Mr0grog:
[...]

PR #1323, opened 2018-04-23
Document object form of `init` constructor option

Mr0grog:
I thought I’d done this in the initial round of docs, but apparently I missed it! (Noticed while looking at #1316)


Mr0grog:
I left out `init.log` here because it seemed like more of a private-use feature for the CLI, but I can add it, too, if we want.

PR #1324, opened 2018-04-23
Change default config from JSON file to JS module

### LINKED ISSUE #1316: Two peers with same Identity

Mr0grog:
Because the default node configuration was a JSON file, it was treated as a singleton object. In some parts of the code, data got added to it, leading to race conditions when creating multiple nodes simultaneously.

This is one approach to fixing it: to make creating a shared global object harder, the default config is now a *function* that returns a unique configuration object on each call. (The test is probably valuable whether or not we like this approach.)

Fixes #1316.

License: MIT
Signed-off-by: Rob Brackett <rob@robbrackett.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ Creates and returns an instance of an IPFS node. Use the `options` argument to s
             - `enabled` (boolean): Make this node a relay (other nodes can connect *through* it). (Default: `false`)
             - `active` (boolean): Make this an *active* relay node. Active relay nodes will attempt to dial a destination peer even if that peer is not yet connected to the relay. (Default: `false`)
             
-- `config` (object) Modify the default IPFS node config. Find the Node.js defaults at [`src/core/runtime/config-nodejs.json`](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/config-nodejs.json) and the browser defaults at [`src/core/runtime/config-browser.json`](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/config-browser.json). This object will be *merged* with the default config; it will not replace it.
+- `config` (object) Modify the default IPFS node config. Find the Node.js defaults at [`src/core/runtime/config-nodejs.js`](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/config-nodejs.js) and the browser defaults at [`src/core/runtime/config-browser.js`](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/config-browser.js). This object will be *merged* with the default config; it will not replace it.
   
 - `libp2p` (object) add custom modules to the libp2p stack of your node
     - `modules` (object):
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "main": "src/core/index.js",
   "browser": {
     "./src/core/components/init-assets.js": false,
-    "./src/core/runtime/config-nodejs.json": "./src/core/runtime/config-browser.json",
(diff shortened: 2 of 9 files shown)


Mr0grog:
(CI failures are from in-progress `files` API work, not this.)

diasdavid:
You are correct, the failing tests are not due to this PR, but the linting fail is. Can you update your PR to make sure it passes linting?

Mr0grog:
Ah! I hadn’t realized Jenkins was running different stuff than Travis and didn’t see those failures. Should be fixed now.

I also noticed some of these:

```
310:1   error  Trailing spaces not allowed  no-trailing-spaces
```

Would you mind if I submitted a PR for adding a `.editorconfig` file? Seems like it would be useful (one of the things you can set with it is removal of trailing spaces).

diasdavid:
re: `.editorconfig`, I don't particularly mind, I just worry that then we have to maintain different configs for each editor, otherwise, contributors will then get confused when a rule changes. So far the burden has been on the contributor. I use ALE (https://github.com/w0rp/ale) with StandardJS and it does the job pretty well.

diasdavid:
@Mr0grog also, added you to the JavaScript team on IPFS so that you can _create branches on this repo_, making collaboration easier. Right now, CI is failing because you need to rebase master onto this branch.

PR #1345, opened 2018-05-08
fix: change ^ to ~ on 0.x.x deps

dryajov:


alanshaw:
hey @dryajov what's the reason for this? I was under the impression that ^ has the same semantics for `0.x` versions as ~. Am I missing something?

https://docs.npmjs.com/misc/semver#tilde-ranges-123-12-1

fsdiogo:
Hey @alanshaw, the [js-ipfs guidelines](https://github.com/ipfs/community/blob/master/js-code-guidelines.md#dependency-versions) states that:

> Use ~ for everything below 1.0.0 and ^ for everything above 1.0.0

`~` is for patches, `^` is for minors.

alanshaw:
Ok, my bad! LGTM

alanshaw:
For next time @dryajov please can we have sign off on commits? https://github.com/ipfs/community/blob/master/js-code-guidelines.md#sign-off-on-commits

PR #1350, opened 2018-05-11
perf(cli): load only requested sub-system modules and inline require ipfs and ipfs-api

alanshaw:
This is an alternative to https://github.com/ipfs/js-ipfs/pull/1336 inlining all the requires in every file for the CLI. This is a great idea but I wanted to see if it could be done with minimal disruption as well as [other reasons](https://github.com/ipfs/js-ipfs/pull/1336#issuecomment-387439256).

I had a look at the good work in #1336 and have taken the best bits that provide us with the biggest gains 🚀 

I've changed `src/cli/bin.js` so that `commandDir` now takes an `include` filter. The idea being to only include the command files for the sub-system the user asked for. i.e. if the user types `jsipfs files add` then we only require `src/cli/commands/files.js`.

I've applied the changes @VictorBjelkholm made to `test/cli/daemon.js` verbatim as this was also really beneficial.

Inlining the `require` for js-ipfs and js-ipfs-api is totally worth it as these two have many many dependencies. I've added a comment next to each to make it obvious why it has been done.

As a baseline, `npm run test:node:cli` took **12m57.607s**.

With the work in this PR, it took **6m26.499s**.

I also ran #1336 and it took **5m52.966s**.

So around ~30s slower than in #1336. It was never going to be faster, but it is a significantly smaller changeset, still reduces the test time for the CLI by over 50%, and addresses the concerns I [raised here](https://github.com/ipfs/js-ipfs/pull/1336#issuecomment-387439256).

### DIFF
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -5,6 +5,8 @@
 const yargs = require('yargs')
 const updateNotifier = require('update-notifier')
 const readPkgUp = require('read-pkg-up')
+const fs = require('fs')
+const path = require('path')
 const utils = require('./utils')
 const print = utils.print
 
@@ -15,6 +17,10 @@ updateNotifier({
 }).notify()
 
 const args = process.argv.slice(2)
+
+// Determine if the first argument is a sub-system command
+const commandNames = fs.readdirSync(path.join(__dirname, 'commands'))
+const isCommand = commandNames.includes(`${args[0]}.js`)
 
 const cli = yargs
   .option('silent', {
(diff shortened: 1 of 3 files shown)


diasdavid:
DNS tests are failing pretty reliably but they are not related with this PR 

image

VictorBjelkholm:
Seems to provide the same benefits without all of the hassle of inline requires. Seems it's just the root command which is now the same as master (while my PR made that fast as well) but I think we can live with that taking 0.8, more important to have the commands themselves faster.

Much better solution, thanks for doing this :+1:

Issue #1320, opened 2018-04-23
⚡️ v0.29.0 RELEASE 🚀


diasdavid:
> short tl;dr; of the release

# 🗺 What's left for release

- https://github.com/ipfs/js-ipfs/pull/1045
- https://github.com/ipfs/js-ipfs/pull/1299
- Going through the PRs in flight and picking the ones that are essential or very close to land for this release.

# 🔦 Highlights

> Writing in Progress

# ✅ Release Checklist

- Robustness and quality
  - [ ] Ensure that all tests are passing, this includes:
    - [ ] unit
    - [ ] interop
    - [ ] sharness
  - [ ] Run tests of the following projects with the new release:
    - [ ] [ipfs-pubsub-room](https://github.com/ipfs-shipyard/ipfs-pubsub-room)
    - [ ] [peerpad-core](https://github.com/ipfs-shipyard/peerpad)
    - [ ] [ipfs-log](https://github.com/orbitdb/ipfs-log)
    - [ ] [orbit-db](https://github.com/orbitdb/orbit-db)
- Documentation
  - [ ] Ensure that README.md is up to date
  - [ ] Ensure that all the examples run
- Communication
  - [ ] Create the release issue
  - [ ] Announcements (both pre-release and post-release)
    - [ ] Twitter
    - [ ] IRC
    - [ ] Reddit
  - [ ] Blog post

# 🙌🏽Want to contribute?

Would you like to contribute to the IPFS project and don't know how? Well, there are a few places you can get started:

- Check the issues with the `help wanted` label at the Ready column in our waffle board - https://waffle.io/ipfs/js-ipfs?label=help%20wanted
- Join an IPFS All Hands, introduce yourself and let us know where you would like to contribute - https://github.com/ipfs/pm/#all-hands-call
- Hack with IPFS and show us what you made! The All Hands call is also the perfect venue for demos, join in and show us what you built
- Join the discussion at http://discuss.ipfs.io/ and help users finding their answers.
- Join the [ⒿⓈ⚡️ js-ipfs Dev Team Weekly Sync 🙌🏽](https://github.com/ipfs/js-ipfs/issues/1179) and be part of the Sprint action!

# ⁉️ Do you have questions?

The best place to ask your questions about IPFS, how it works and what you can do with it is at [discuss.ipfs.io](http://discuss.ipfs.io). We are also available at the #ipfs channel on Freenode.

dryajov:
@diasdavid 

[...]

PR #1355, opened 2018-05-15
docs: add api changes to release template

alanshaw:
Adds a section where we can document any API changes that went into a release.

@mitra42 what do you think?

resolves https://github.com/ipfs/interface-ipfs-core/issues/184


diasdavid:
great initiative! @alanshaw mind adding this section for #1320 ?

Issue #1325, opened 2018-04-24
Better error handling in daemon + gateway


victorb:
Today, if there is a error in a stream or any other error within the daemon or the gateway, the entire process crashes. While not a problem for testing, it's a problem when deploying and trying to run a long-term daemon (for example, for the js-ipfs daemons).

I raised this before in https://github.com/ipfs/js-ipfs/pull/1243

Basically, we should be able to gracefully handle errors down the stack without having the daemon and gateway crash. 

Current issue is the following error that happens every ~10 minutes:

```
/usr/src/app/node_modules/pull-pair/index.js:11
      throw new Error('already piped')
      ^

Error: already piped
    at sink (/usr/src/app/node_modules/pull-pair/index.js:11:13)
    at consume (/usr/src/app/node_modules/pull-defer/sink.js:7:17)
    at consume (/usr/src/app/node_modules/pull-defer/sink.js:7:17)
    at Connection.consume (/usr/src/app/node_modules/pull-defer/sink.js:7:17)
    at pull (/usr/src/app/node_modules/pull-stream/pull.js:43:9)
    at Dialer.handle (/usr/src/app/node_modules/multistream-select/src/dialer/index.js:47:5)
    at attemptMuxerUpgrade (/usr/src/app/node_modules/libp2p-switch/src/dial.js:164:10)
    at gotWarmedUpConn (/usr/src/app/node_modules/libp2p-switch/src/dial.js:55:7)
    at Switch.dial (/usr/src/app/node_modules/libp2p-switch/src/dial.js:41:9)
    at _getPeerInfo (/usr/src/app/node_modules/libp2p/src/index.js:238:19)
```

**I'm not saying that we should (or should not) fix this particular error, but rather when this error (and similar) happens, the daemon and gateway should not die but continue running.**

VictorBjelkholm:
@fsdiogo as you're the current OKR owner for this, I've assigned this to you

VictorBjelkholm:
Another error that happens directly after starting daemon right now:

```
Initializing daemon...
Using wrtc for webrtc support
/usr/src/app/src/cli/commands/daemon.js:40
        throw err
        ^

Error: websocket error
    at WS.Transport.onError (/usr/src/app/node_modules/engine.io-client/lib/transport.js:64:13)
    at WebSocket.ws.onerror (/usr/src/app/node_modules/engine.io-client/lib/transports/websocket.js:150:10)
[...]

PR #1359, opened 2018-05-18
[WIP] remove boot from constructor

fsdiogo:
This PR is the start of an effort to standardise the way we propagate and deal with errors.

The objective here is to remove event emitters, as we're only using them to propagate errors during instantiation. Moving `boot` that's inside the IPFS constructor to a factory method that returns a promise is a step towards that.

The discussion started in https://github.com/ipfs/js-ipfs/issues/1325, but check https://github.com/ipfs/js-ipfs/issues/1325#issuecomment-389607930 and https://github.com/ipfs/js-ipfs/issues/1325#issuecomment-389612787 as to why.

This will be a breaking change and apps using ipfs will probably have to be refactored, so that's why we should offer backwards compatibility for some time, to allow developers to update their code.


alanshaw:
https://github.com/ipfs/js-ipfs/pull/2094 is a step towards the idea put forwards in this PR and supercedes much of the work here already.

PR #1360, opened 2018-05-18
feat: mfs implementation

achingbrain:
~~Not ready for merging yet~~ but I'd love some feedback on the approach.  I've placed the mfs commands behind a flag named `--enable-mfs-experiment` - if you enable this it will cause the `ipfs files` subcommands to be the same as the go implementation.


Mr0grog:
Minor note — would you please document the new `EXPERIMENTAL` flag along with [the others in the README](https://github.com/ipfs/js-ipfs/blob/master/README.md#ipfs-constructor)?

achingbrain:
> Was your plan to leave it like this or eventually merge the ipfs-mfs repo with this one

@diasdavid's preference was for a separate module so there are no plans to merge it into this one AFAIK.

test/cli/file.js
```diff
@@ -4,15 +4,15 @@
...
+const dir = 'Qmaj2NmcyAXT8dFmZRRytE12wpcaHADzbChKToMEjBsj5Z'
 
 describe('file ls', () => runOnAndOff((thing) => {
   let ipfs
 
   before(function () {
     this.timeout(50 * 1000)
     ipfs = thing.ipfs
-    return ipfs('files add -r test/fixtures/test-data/recursive-get-dir')
+    return ipfs('add -r test/fixtures/test-data/recursive-get-dir')
```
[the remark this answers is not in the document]
achingbrain: Yes, this is consistent with `go-ipfs`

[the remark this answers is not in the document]
diasdavid: Let's think more about this...

I would be totally down to move forward with https://github.com/ipfs/specs/issues/98 in JS land (both js-ipfs and js-ipfs-api). In fact, I believe we are reaching a time where we need to start pushing forward better and polished APIs, given that we live in a different ecosystem and we need to make things that work for our user needs.

An advantage to ourselves is that we have a SPEC and tests, making this transition smooth and more pleasant.

src/cli/bin.js
```diff
@@ -86,10 +60,37 @@ if (args[0] === 'daemon' || args[0] === 'init') {
...
       cli
+        .commandDir('commands', {
+          visit: (commandObject, pathToFile, filename) => {
+            if (commandObject.command === 'files <command>' && enableMfs) {
+              return null
+            }
+
+            return commandObject
+          }
+        })
```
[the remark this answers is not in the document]
achingbrain: It's excluding any of the existing `ipfs files <command>` commands from loading if `enableMfs` is truthy.

diasdavid:
@achingbrain can we drop the experimental flag? This should work with go-ipfs right off the bat.

achingbrain:
Sure, that’d certainly make things a bit more straightforward.

diasdavid:
@achingbrain is this ready for merge (the title still says WIP)? Are there interop tests?

lidel:
May be too specific, but is there an interop test that checks if  `files.write` produces the same CIDs in js-ipfs and go-ipfs?
We will be changing default CID encoding at some point and such tests would help us coordinate.

achingbrain:
@diasdavid @lidel the interop tests are here: ipfs/interop#25 and there are a whole bunch that assert on CIDs generated by go & js.

src/cli/bin.js
```diff
@@ -86,19 +63,54 @@ if (args[0] === 'daemon' || args[0] === 'init') {
...
+        .catch((arg) => {
+          debug(arg)
+
+          // the argument can have a different shape depending on where the error came from
+          if (arg.message) {
+            print(arg.message)
+          } else if (arg.error && arg.error.message) {
+            print(arg.error.message)
+          } else {
+            print('Unknown error, please re-run the command with DEBUG=ipfs:cli to see debug output')
```
alanshaw: It would be great to have some consistency here - the rest of the cli debug prefixes are `cli:*` and elsewhere in the code we have `jsipfs:*`. One for a separate PR 😉

src/cli/bin.js
```diff
@@ -86,19 +63,54 @@ if (args[0] === 'daemon' || args[0] === 'init') {
...
+          }
+
+          process.exit(1)
```
alanshaw: We always called `cleanup` before even if there was an error. I think this needs to be added here too.

src/cli/commands/dag/get.js
```diff
@@ -41,7 +41,7 @@ module.exports = {
       // * reads as 'agree in'
       if (node._json) {
         delete node._json.multihash
-        node._json.data = '0x' + node._json.data.toString('hex')
+        node._json.data = node._json.data.toString('base64')
```
alanshaw: Is this change meant to be in this PR? Why not also for the if test below?

src/core/index.js
```diff
@@ -133,6 +133,15 @@ class IPFS extends EventEmitter {
       isIPFS: isIPFS
     }
 
+    // ipfs.files
+    const mfs = components.mfs(this)
+
+    Object.keys(mfs).forEach(key => {
+      if (mfs.hasOwnProperty(key)) {
```
alanshaw: Minor: I believe `Object.keys` iterates over own properties so I think this check is redundant, unless I'm missing something?

test/browser.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
+
+const isWebWorker = require('detect-webworker')
+
+if (isWebWorker) {
+  // https://github.com/Joris-van-der-Wel/karma-mocha-webworker/issues/4
+  global.MFS_DISABLE_CONCURRENCY = true
+}
```
alanshaw: Since there is a requirement to either set this variable or do some setup on the main thread (for webworkers) we should mention it and [link to it](https://github.com/ipfs/js-ipfs-mfs#a-note-on-concurrency) from the js-ipfs README so that it's easily discoverable and we don't get issues opened about it here.

Also a separate PR to add a examples for using MFS in webworkers with/without concurrency is needed and would be very 🚀 🎸 ✨ 💖

test/core/interface.spec.js
```diff
@@ -22,59 +22,12 @@ describe('interface-ipfs-core tests', () => {
...
-      }
-    ]
-  })
+  tests.files(defaultCommonFactory)
 
   tests.key(CommonFactory.create({
     spawnOptions: {
       args: ['--pass ipfs-is-awesome-software'],
-      initOptions: { bits: 512 }
+      initOptions: { bits: 1024 }
```
alanshaw: This is for faster boot - `defaultCommonFactory` uses 512 also, is this change needed?

achingbrain: I ran the tests against the go daemon and it refused to start up with 512 and said it should be 1024 as a minimum.  Can change it back if you prefer.

src/core/index.js
achingbrain: From my understanding `Object.keys` returns all properties, including inherited ones whereas `hasOwnProperty` will only return true if it's a non-inherited property.  Happy to change this if you prefer it the other way.

src/cli/commands/dag/get.js
achingbrain: Probably could be separate.  go prints out the data as base64 encoded, we print it out as hex.

src/cli/bin.js
achingbrain: I'd love it if we could standardise on one prefix for all `ipfs` related `DEBUG` values.  Just doing `ipfs:*` instead of `ipfs:*,ipfsd-ctl:*,ipfs-api:*,etc` is my dream.

test/core/interface.spec.js
alanshaw: Yep in js-ipfs-api it's set to 1024 it tests against go-ipfs, but these tests only run against js-ipfs and there's no restriction in place currently.

achingbrain: There probably should be, for the sake of interop though, right?

alanshaw: Yes good point, lets not slow down the tests unnecessarily until that PR comes in though. I'll create an issue for adding it so we can track it.

test/browser.js
achingbrain: I've added a paragraph to the README, will create an example in a separate PR.

diasdavid: What does MFS with concurrency mean?

README.md
```diff
@@ -135,6 +136,20 @@ You can also load it using a `<script>` using the [unpkg](https://unpkg.com) CDN
 ```
 Inserting one of the above lines will make an `Ipfs` object available in the global namespace.
 
+### Use with Web Workers
+
+The `ipfs.files.*` [MFS](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#mutable-file-system) commands use LevelDB to store the CID that corresponds to the current root of the file system.  LevelDB has no concurrency control so a read/write lock at the application level is necessary.  With single-process and [Node.js cluster](https://nodejs.org/api/cluster.html) based apps this is completely transparent but [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) require a little extra setup.  Use the [`observable-webworkers`](https://www.npmjs.com/package/observable-webworkers) module to allow communication with any Web Workers your application creates:
```
diasdavid: I see. This is so that if multiple WebWorkers are spawn, the store is shared?

Note that this is valid for anything that IPFS stores really, in the browser it is IndexedDB for everything.

achingbrain: Yes, if multiple threads mutate the MFS graph at the same time they'll result in different hashes for the root of the graph creating a last-update-wins race condition.  There needs to be some sort of gatekeeper to resolve this.

test/browser.js
diasdavid: nvm. Read the paragraph.

README.md
diasdavid: Multiple threads, each with its IPFS Node or just one IPFS node?

achingbrain: I've flipped this round so that you only need to do the config dance if you've spawned a node on the main process then forked (e.g. node clusters) or spawned nodes on the UI thread and web workers (e.g. browsers) all using the same repo - which hopefully shouldn't be very common.

I'm not sure we should encourage this in the browser at least so I've left it undocumented for the time being but we can discuss this during the locking session at the dev meetup next week.

achingbrain:
no-wai

PR #1358, opened 2018-05-18
fix: Node.js 10 support

jacobheun:
Opening this early to provide more visibility.

**Update**:
The jenkins builds are now running and working for node 10.0. There is an issue in node 10.1 with fs.promises warnings, mentioned below, but there is a fix already merged that should go out with the next release so I am going to leave that for now.

I will be working on getting all the dependency change PRs merged so I can get this PR finalized.

**Things to fix**:
- [x] Get npm install working on node 10
- [x] Resolve [repo not initialized errors](https://github.com/ipfs/js-ipfs/pull/1338)
- [x] Resolve fs.promises ExperimentalWarning in node 10.1 (these go to stderr and cause any exec spawn tests to fail) This is fine in node 10.0 and should be fixed in the next version of node 10 (https://github.com/nodejs/node/pull/20632)
- [x] Get yarn install working for CI (incorrect dependencies being installed)
- [x] Resolve failing tests
- [x] (PR Pending) Fix windows, [node 10 failure in lock-me](https://github.com/dignifiedquire/lock-me/pull/3), due to node 10 fs module bug.

## Dependencies
The following PRs need to be completed prior to this release.
- [x] ipfs/js-datastore-fs#16 merged and released
- [x] ipfs/js-datastore-level#7 merged and released
  - [x] ipfs/js-ipfs-repo#168 merged and released, depends on fs and level PRs    
    - [x] ipld/js-ipld#129 merged and released, causes an issue with npm install of old leveldown

PR #1369, opened 2018-05-29
Node 10

diasdavid:
Let's give Node 10 a spin! :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "ipfs-repo": "~0.22.0",
     "ipfs-unixfs": "~0.1.14",
     "ipfs-unixfs-engine": "~0.29.0",
-    "ipld": "~0.17.0",
+    "ipld": "~0.17.1",
     "ipld-dag-cbor": "~0.12.0",
     "ipld-dag-pb": "~0.14.4",
     "is-ipfs": "~0.3.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -110,7 +110,7 @@
     "ipfs-block": "~0.7.1",
     "ipfs-block-service": "~0.14.0",
     "ipfs-multipart": "~0.1.0",
-    "ipfs-repo": "~0.20.0",
+    "ipfs-repo": "~0.22.0",
```
diasdavid: @jacobheun CI is not happy with the latest ipfs-repo. How did you run the tests before (ref: https://github.com/ipfs/js-ipfs-repo/pull/168#issuecomment-392808793)

jacobheun: This is due to the ipld PR, https://github.com/ipld/js-ipld/pull/129, that needs released. I will update https://github.com/ipfs/js-ipfs/pull/1358 once that is done.

The current versioning scheme results in old leveldown being installed for ipld, which breaks the  install.

jacobheun: @diasdavid I already have a PR open for node 10 that is running tests now, https://github.com/ipfs/js-ipfs/pull/1358. It also includes the fix for the breaking change in ipfs-repo 0.20.1 https://github.com/ipfs/js-ipfs-repo/issues/165, which is the errors your PR is getting.

diasdavid: Missed that one. I had read from your message that you had tested against js-ipfs master. All good then :)

Issue #738, opened 2017-01-29
Enabling experimental features


daviddias:
> Following up https://github.com/ipfs/js-ipfs/pull/735#issuecomment-275927976

We need a way to enable/disable experimental features throughout js-ipfs (that includes all the js-libp2p) in a way that is non-intrusive, easy to add and easy to remove, both in Node.js and the Browser.

A typical way to achieve this, is through env variables, but that might not be the ideal case for Browser bundles.

Thoughts, ideas?

dignifiedquire:
I think these options should be exposed as config options on a library level. We can then pass cli flags in js-ipfs as values for those (node.js) and simply set them directly when starting up the node in the browser. I would really like to stay away from global injections as much as possible for such things.

We could also reuse the existing config file, and add a section `experimental`, in which such things could be configured. This might be even nice for go-ipfs.

diasdavid:
We are now supporting experimental features by passing flags to the `options` object on new IPFS.

Does everyone feel like this is resolved? If so, I'll close the issue :)

haadcode:
I feel the way it's done atm is good, ie. passing `options` to IPFS in the constructor. 

That said, I agree with @dignifiedquire's comment that it'd be nice to have the `experimental` field in the config file as well (this may or may not have been implemented in the past couple of months, ignore if that's already the case).

diasdavid:
Last steps to complete this task:

- [ ] Document how to enable experimental features
- [ ] Support EXPERIMENTAL flags to be passed through config and not options. (so that daemons can also use them easily if needed)

PascalPrecht:
@diasdavid happy to help out with those last tasks (if they still need to be done). 

Any pointers where things should be documented?

daviddias:
This got done meanwhile. Thanks @PascalPrecht :)

PR #1372, opened 2018-05-30
[WIP] feat: read config from repo

achingbrain:
Lets us do things like `jsipfs config --bool EXPERIMENTAL.pubsub true` and have IPFS
respect the flags in daemon and non-daemon mode

Attempts to address ipfs/js-ipfs#738

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -16,6 +16,9 @@ const CID = require('cids')
 const debug = require('debug')
 const extend = require('deep-extend')
 const EventEmitter = require('events')
+const waterfall = require('async/waterfall')
+const series = require('async/series')
+const defaults = require('lodash/defaultsDeep')
 
 const config = require('./config')
 const boot = require('./boot')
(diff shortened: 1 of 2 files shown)


src/core/index.js
```diff
@@ -136,7 +125,49 @@ class IPFS extends EventEmitter {
       isIPFS: isIPFS
     }
 
-    boot(this)
+    series([
+      (cb) => {
+        waterfall([
+          (done) => this._repo.config.get((error, config) => {
+            if (error) {
+              this.log('Could not load config', error)
```
Mr0grog: It’s probably not a huge deal since the logs only show if someone’s explicitly set `DEBUG=jsipfs`, but this will log even when things are just fine — e.g. if running `jsipfs init`. It makes it look like something went wrong even when the error is expected and OK.

Would it be better to check whether the repo exists first or to check `this._options.init`? (Are there other situations when this *should* fail?)

VictorBjelkholm:
@achingbrain resolved the merge conflict here so we could have a CI run, hope you don't mind ([change](https://github.com/ipfs/js-ipfs/pull/1372/commits/faa4c61122c688a1c55b302c605a916e0a973135)).

src/http/api/resources/files.js
```diff
@@ -154,7 +154,19 @@ exports.add = {
...
+        // This is because raw-leaves is not yet implemented in js-ipfs,
+        // and go-ipfs changes the value of raw-leaves to true when
+        // cid-version > 0 unless explicitly set to false.
+        //
+        // This retains feature parity without having to implement raw-leaves.
+        'raw-leaves': Joi.boolean().when('cid-version', {
+          is: 1,
+          then: Joi.boolean().valid(false).required(),
+          otherwise: Joi.boolean().valid(false)
+        }),
```
alanshaw: Eek, bad merge? This shouldn't be here anymore.

VictorBjelkholm: Thanks for catching that, resolved.

PR #1374, opened 2018-05-31
fix: dont fail on uninitialized repo

dryajov:
See this comment for background - https://github.com/ipfs/js-ipfsd-ctl/issues/242#issuecomment-393371387


dryajov:
@alanshaw I agree we should rework that section entirely, but I would rather wait for the rest of the datastores to emit similar error codes instead of introducing ad hoc checks. @hugomrdias is already working on this AFAIK, and this code has it's days counted (hopefully!)

alanshaw:
Right, gotcha, if you wouldn't mind submitting a PR when we can change them all that would be great.

PR #1383, opened 2018-06-04
RFC: allow passing constructor for the libp2p modules

pgte:
WIP

Context: allow overriding the libp2p modules, permitting this type of options:

```js
const ipfs = new IPFS({
  libp2p: { modules }
})

return ipfs

function modules (peerInfo) {
  const appTransport = AppTransport(appName, new WebSocketStar({ id: peerInfo.id }))
  return {
    transport: [ appTransport ],
    discovery: [ appTransport.discovery ]
  }
}
```

Some libp2p modules require the peer id in the constructor. Instead of a fancy DI framework, I propose a simple solution that allows to pass a function that will act as a module constructor.

This could be made to be even more flexible, like, for instance, by allowing some transports to be passed as instances and others as a constructor, like this:

```js
const ipfs = new IPFS({
  libp2p: {
    modules: {
      transport: [
        ws,
        (peerInfo) => myWonderfulTransport(peerInfo.id)
      ]
    }
  }
})
```

(What this doesn't solve is the transport -> discovery coupling, which sometimes may be required. I suggest this by adding `transport.discovery` to the list of discovery modules if it exists).

I prefer this type of simplistic approach over [some others that have been discussed](https://github.com/libp2p/js-libp2p/issues/130), as this does not require libp2p module constructors to implement specific interfaces or require generic DI libs.

### DIFF
--- a/src/core/components/libp2p.js
+++ b/src/core/components/libp2p.js
@@ -19,7 +19,7 @@ module.exports = function libp2p (self) {
           mdns: get(config, 'Discovery.MDNS.Enabled'),
           webRTCStar: get(config, 'Discovery.webRTCStar.Enabled'),
           bootstrap: get(config, 'Bootstrap'),
-          modules: self._libp2pModules,
+          modules: prepareModules(self, self._libp2pModules),
           // EXPERIMENTAL
           pubsub: get(self._options, 'EXPERIMENTAL.pubsub', false),
           dht: get(self._options, 'EXPERIMENTAL.dht', false),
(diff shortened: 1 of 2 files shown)


diasdavid:
@pgte see https://github.com/libp2p/js-libp2p/pull/166. I'm adding support to instances and classes there as well

alanshaw:
I like this approach in general but it sounds like @diasdavid wants to solve the problem in the libp2p layer. I think that's sensible as folks using libp2p directly will run into the same issue.

pgte:
@alanshaw Absolutely, that’s a more permanent solution, I’ll use this one instead. Thanks!

PR #1396, opened 2018-06-12
[WIP] feat: Add rendezvous

mkg20001:
Adds support for [libp2p-rendezvous](https://github.com/libp2p/js-libp2p-rendezvous) protocol

ToDOs:
 - [ ] Make this less hacky
 - [ ] Hide it behind a flag for now?

### DIFF
--- a/src/core/components/libp2p.js
+++ b/src/core/components/libp2p.js
@@ -4,6 +4,7 @@
 const Node = require('../runtime/libp2p-nodejs')
 const promisify = require('promisify-es6')
 const get = require('lodash.get')
+const Rendezvous = require('libp2p-rendezvous')
 
 module.exports = function libp2p (self) {
   return {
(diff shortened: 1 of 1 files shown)


pgte:
@mkg20001 @alanshaw Any news on this? We need this for https://github.com/ipfs-shipyard/peer-star-app/issues/5

alanshaw:
@pgte in lieu of it being a default discovery module can you pass it to `libp2p.peerDiscovery` temporarily in `peer-star-app`?

mkg20001:
@pgte This would have to be properly integrated into libp2p instead of IPFS, I think. Current blocker is that rendezvous requires swarm, and swarm should depend on rendezvous as a discovery module. (chicken&egg problem). There needs to be a better solution to that than just adding more hacks.

alanshaw:
@mkg20001 shall we close this PR for the time being?

mkg20001:
@alanshaw Yes. I think the whole "X requires swarm but is component for swarm" situation requires a refactor anyways

PR #1398, opened 2018-06-14
fix: do not stringify output of object data

achingbrain:
Without this change:

```sh
$ jsipfs object data `dd if=/dev/urandom bs=307200 count=1 | jsipfs files write --create /some-file && jsipfs files stat -h /some-file` | unixfs
{"type":"file","blockSizes":[72481833443311,10909182287855]}
```

e.g.

```
00000000: 0802 18ef bfbd efbf bd12 20ef bfbd efbf  .......... .....
00000010: bd10 20ef bfbd efbf bd02                 .. .......
```

With this change:

```sh
$ jsipfs object data `dd if=/dev/urandom bs=307200 count=1 | jsipfs files write --create /some-file && jsipfs files stat -h /some-file` | unixfs
{"type":"file","blockSizes":[262144,45056]}
```

e.g.

```
00000000: 0802 1880 e012 2080 8010 2080 e002       ...... ... ...
```

Stringifying the data changes it's encoding. Been banging my head against this all morning.

### DIFF
--- a/src/cli/commands/object/data.js
+++ b/src/cli/commands/object/data.js
@@ -17,7 +17,7 @@ module.exports = {
         throw err
       }
 
-      print(data.toString(), false)
+      print(data, false)
     })
   }
 }
--- a/test/cli/object.js
+++ b/test/cli/object.js
@@ -4,6 +4,11 @@
 
 const expect = require('chai').expect
 const runOnAndOff = require('../utils/on-and-off')
+const UnixFs = require('ipfs-unixfs')
+const path = require('path')
+const fs = require('fs')
+const crypto = require('crypto')
+const os = require('os')
 
 describe('object', () => runOnAndOff((thing) => {
   let ipfs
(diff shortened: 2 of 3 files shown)


achingbrain:
Weird that the tests were passing on Circle but exploding in a fairly dramatic fashion on Jenkins.  🤷‍♂️ 

A rebase and it's all looking good.

alanshaw:
Does this need a "BREAKING CHANGE" section in the commit message or should this be considered a bug fix? Is not a chore?!

VictorBjelkholm:
> Weird that the tests were passing on Circle but exploding in a fairly dramatic fashion on Jenkins

Indeed. Could be because of the aggressive caching Circle does of the dependencies.

achingbrain:
> Does this need a "BREAKING CHANGE" section in the commit message

I don't think so, we're not changing an API or interface so it shouldn't break anything, quite the opposite.

> Should this be considered a bug fix? Is not a chore?!

It feels like a chore.  Just kidding, would you like me to change the commit message?

alanshaw:
`chore`s don't make it into the auto generated changelog, if you don't mind changing to fix that would be brill

achingbrain:
> chores don't make it into the auto generated changelog

Fair enough, have updated the commit message.

PR #1400, opened 2018-06-19
feat: ipns locally

vasco-santos:
A working version of **IPNS working locally** is here!  🚀 😎 💪 

Steps:

- [x] Create a new repo (**js-ipns**) like it was made with go in the last week ([go-ipns](https://github.com/ipfs/go-ipns)) and port the related code from this PR to there
- [x] Resolve IPNS names in publish, in order to verify if they exist (as it is being done for regular files) before being published
- [x] Handle remaining parameters in publish and resolve (except ttl). 
- [x] Test interface core spec [interface-ipfs-core#327](https://github.com/ipfs/interface-ipfs-core/pull/327)
- [x] Test interoperability with go. [interop#26](https://github.com/ipfs/interop/pull/26)
- [x] Integrate logging
- [x] Write unit tests
- [x] Add support for the lifetime with nanoseconds precision
- [x] Add Cache
- [x] Add initializeKeyspace
- [x] Republish

Some notes, regarding the previous steps: 

- There is an optional parameter not implemented in this PR, which is `ttl`, since it is still experimental, we can add it in a separate PR.

Finally, thanks @Stebalien for your help and time answering all my questions regarding the IPNS implementation in GO.

Moreover, since there are no specs, and not that much documentation, I have been writing a document with all the IPNS workflow. It is a WIP by now, but it is currently available [here](https://github.com/ipfs/specs/pull/184).

Related PRs:

- [x] [js-ipns#4](https://github.com/ipfs/js-ipns/pull/4)
- [x] [js-ipfs-repo#173](https://github.com/ipfs/js-ipfs-repo/pull/173)
- [x] [js-ipfs#1496](https://github.com/ipfs/js-ipfs/pull/1496)
- [x] [interface-ipfs-core#327](https://github.com/ipfs/interface-ipfs-core/pull/327)
- [x] enable `interface-ipfs-core` tests for IPNS in `js-ipfs`


src/core/ipns/pb/ipnsEntry.js
```diff
@@ -0,0 +1,42 @@
...
+  // Marshal
+  marshal: (ipnsEntry) => {
+    return ipnsEntryProto.encode(ipnsEntry)
+  },
+  // Unmarshal
+  unmarshal: (marsheled) => {
+    return ipnsEntryProto.decode(marsheled)
+  },
+  validityType: ipnsEntryProto.ValidityType
+}
```
hugomrdias:
```
module.exports = {
	create,
	marshal : ipnsEntryProto.encode,
	unmarshal : ipnsEntryProto.decode,
	validityType: ipnsEntryProto.ValidityType
}
```

src/core/components/name.js
```diff
@@ -0,0 +1,133 @@
+'use strict'
+
+const promisify = require('promisify-es6')
+const series = require('async/series')
+const human = require('human-to-milliseconds')
+const path = require('../ipns/path')
+
+const OFFLINE_ERROR = require('../utils').OFFLINE_ERROR
```
hugomrdias: const { OFFLINE_ERROR } = require('../utils')

src/core/ipns/index.js
```diff
@@ -0,0 +1,54 @@
+'use strict'
+
+const peerId = require('peer-id')
```
hugomrdias: const { createFromPrivKey } = require('peer-id')

src/core/ipns/validator.js
```diff
@@ -0,0 +1,50 @@
...
+  const dataForSignature = getDataForSignature(value, validityType, validity)
+
+  // Validate Signature
+  publicKey.verify(dataForSignature, entry.signature, (err, result) => {
+    if (err) {
+      return callback(err)
+    }
+
+    // Validate EOL
+    const validityDate = Date.parse(validity.toString())
```
hugomrdias: consider using https://date-fns.org/

Stebalien: FYI, the only valid time format is RFC3339 with nanosecond precision. That is:

2006-01-02T15:04:05.999999999Z07:00

src/core/components/name.js
```diff
@@ -0,0 +1,133 @@
...
+     * @returns {Promise|void}
+     */
+    publish: promisify((value, resolve = true, lifetime = '24h', ttl, key = 'self', callback) => {
+      if (!self.isOnline()) {
+        return callback(new Error(OFFLINE_ERROR))
+      }
+
+      // Parse ipfs path value
+      try {
+        value = path.parsePath(value)
```
Stebalien: Note: the value can be any arbitrary path.

src/core/ipns/resolver.js
```diff
@@ -0,0 +1,66 @@
...
+
+        return callback(null, res)
+      })
+    } else {
+      return callback(new Error('not implemented yet'))
+    }
+  }
+
+  // https://github.com/ipfs/go-ipfs-routing/blob/master/offline/offline.go
+  resolveLocal (name, publicKey, callback) {
```
Stebalien: Eventually, this should be pushed into a non-IPNS specific routing layer.

src/core/ipns/validator.js
Stebalien: Note: It's extremely important that you only except strings of this exact form. Not validating this has unfortunate security implications.

This also needs to check the validity type.

src/core/components/name.js
vasco-santos: It is really strange to me that the errors are in a `utils.js` file, making those imports really strange. In the future, all errors should be easily accessible from all the project, and we should access them in the same way as @fsdiogo suggested.

src/core/ipns/validator.js
vasco-santos: I will support milliseconds precision at first here and I added a task to improve the precision to nanosecond later (in the list above). 

Basically, JS does not support nanoseconds precision natively and all the libs that I managed to find don't support summing date times with nanosecond precision.

vasco-santos: When you say that I must validate if the string has this exact form, you say that it should be compliant with RFC3339, right?

src/core/ipns/resolver.js
vasco-santos: I had been thinking about the `go-ipfs` implementation for the resolver. 

For one side, I agree with your [approach](https://github.com/ipfs/go-ipfs/blob/master/core/commands/ipns.go#L93). That is, you decide which router the IPNS resolver should use, and this router is completely transparent for the IPNS resolver. However, as we are fetching from an offline and local source (datastore), it also seems strange to me to have the `OfflineRouter`. 

Anyway, I am not sure about where this code should live on. @diasdavid any opinion here?

Stebalien: Basically, it means that we can swap out the router at will. However, given the fact that you're only implementing offline routing, it doesn't really make a difference.

vasco-santos: Being a pluggable router is a considerable advantage. But yeah, that's what I thought while putting the `resolveLocal` in there. 

However, when I integrate the routing I will have to understand if it still makes sense to have the `resolveLocal` inside the IPNS layer.

src/core/ipns/validator.js
```diff
@@ -33,10 +35,12 @@ const verify = (publicKey, entry, callback) => {
     }
 
     // Validate EOL
-    const validityDate = Date.parse(validity.toString())
+    if (validityType === IpnsEntry.validityType.EOL) {
```
Stebalien: We should fail by default if we can't validate the record.

src/core/ipns/validator.js
```diff
@@ -33,10 +35,12 @@ const verify = (publicKey, entry, callback) => {
...
+      const validityDate = Date.parse(validity.toString())
```
Stebalien: So it doesn't get lost in the buried review comments, this needs to explicitly check the date format.

diasdavid:
@Stebalien why go-ipns and not go-libp2p-ipns? Did I miss any important convo?

Stebalien:
> why go-ipns and not go-libp2p-ipns? Did I miss any important convo?

As far as I know, IPNS falls under the umbrella of IPFS, not libp2p.

vasco-santos:
>> why go-ipns and not go-libp2p-ipns? Did I miss any important convo?

> As far as I know, IPNS falls under the umbrella of IPFS, not libp2p.

I see IPNS as a different piece than IPFS and libp2p. But as we don't have an IPNS umbrella, I don't know which would suit best here.

src/core/ipns/validator.js
vasco-santos: already done in [js-ipns](https://github.com/ipfs/js-ipns), more precisely [here](https://github.com/ipfs/js-ipns/pull/1/files#diff-1fdf421c05c1140f6d71444ea2b27638R74)

src/core/ipns/validator.js
vasco-santos: Once `js-ipns` initial implementation PR is merged, I will create an issue in there for tackling this with priority. I have been thinking a lot earlier and reading some stuff about `bigint` for js and I think that I reached a way to add the nanoseconds precision here. I also need to find a way to explicitly check the date format according to the RFC 3339

diasdavid:
I know that there is no spec, but IPNS has been libp2p all along -- https://github.com/libp2p/specs/blob/master/4-architecture.md#46-naming

vasco-santos:
@diasdavid I will change to the `libp2p`organization then. Do you think it should be `js-libp2p-ipns` or `js-libp2p-ipns-record`? I think that the second one suits better because this module does not have the IPNS core code (publish and resolve), but utilities for creating, parsing, and validating IPNS records, which are used for the `publish` and `resolve`. The `go-ipfs` implementation has the IPNS core code, but I think that we could have the `js-libp2p-ipns` with the core implementation of publish and resolve. Unless @Stebalien have other ideas in here.

Yeah @diasdavid ! I have been testing everything to guarantee the interoperability between both. There are two fields which I am not sure right now, which are the record `signature` and the `validity`. 
The signature is really difficult to test now, since it is stored in bytes, and I believe that I don't have a clean way to test this, while not having IPNS routing (I tried to copy the `go-ipfs` repo content and use it in `js` for validating the signature, but it seems that the repos are not compatible). Anyway, I have the interoperability task written in the first comment of this PR and once the [PR](https://github.com/ipfs/js-ipns/pull/1) initial implementation of the `"js-libp2p-ipns-record"` gets merged, I will create an issue to confirm the interoperabillity in that repo.

diasdavid:
@vasco-santos mind doing a drawing of the IPNS architecture first as you see it so that we clearly sync what are the pieces and their respective names?

vasco-santos:
Here I have a small diagram with the interactions that I was describing. Or do you want a more low-level diagram with something like "how ipns works?"

ipns 2x

src/core/ipns/validator.js
Stebalien: I wouldn't feel comfortable merging this until this is fixed. Unfortunately, due to some mistakes in how we make signatures in IPNS, we rely on *precisely* validating the date for security.

Stebalien:
> I know that there is no spec, but IPNS has been libp2p all along

That's a bit odd. I guess one could, e.g., make an IPNS record point to a multiaddr but I still don't feel like it fits in libp2p. It really does feel like a separate thing. I can move it, I just don't want to shuffle things back and forth.

---

As for lang-ipns versus lang-ipns-record: https://github.com/ipfs/go-ipns/issues/1

However, I held off on that as IPNS is record-store independent. That is, I'd almost expect:

* {go,js}-ipns-interface: the publisher/resolver interfaces.
* {go,js}-ipns-value-store: an implementation of the publisher/resolver interfaces that, internally, uses a valuestore (or a record store, depending on what you want to call it).
* {go,js}-ipns-record: the actual IPNS record logic (this package).

vasco-santos:
@Stebalien I would go with that package naming proposal!

src/core/ipns/validator.js
vasco-santos: @Stebalien just added it to [js-ipns PR](https://github.com/ipfs/js-ipns/pull/1#pullrequestreview-131898357)!

diasdavid:
My claim (and others) is that IPNS is just a specialized version of IPRS and that _naming_ is essential piece to Networking and therefore it should be part of libp2p primitives. But you know what, I'll let you paint the bikeshed the color you want as long as we keep making progress here and we having IPNS working locally, over PubSub and DHT as soon as possible :)

vasco-santos:
yeah, let's go forward with this one!

@Stebalien can we change it to the `libp2p`, as well as rename it to `js-libp2p-ipns-record`?

whyrusleeping:
I was asked by the esteemed mr engineer david d# (@diasdavid) to weigh in on the color of this particular bike shed.

In my opinion, IPNS is part of IPFS. IPNS deals explicitly with ipfs hashes/paths. IPRS is part of libp2p, as it deals with more general values. IPRS can be used to implement IPNS, but currently, IPNS is its own thing. If IPNS were less tied to IPFS, then i'd lean towards agreeing with david that its part of libp2p, but thats not the case.

vasco-santos:
All right, thanks for your feedback @whyrusleeping !
By the way, do you have any opinion regarding the `js-ipns` / `go-ipns` naming change to `js-ipns-record` / `go-ipns-record`?

References:

https://github.com/ipfs/js-ipfs/pull/1400#issuecomment-399405650
https://github.com/ipfs/js-ipfs/pull/1400#issuecomment-399558714
https://github.com/ipfs/go-ipns/issues/1

src/core/ipns/validator.js
```diff
@@ -1,54 +0,0 @@
...
-// Sign IPNS Entry for publish
-const sign = (privateKey, value, validityType, validity, callback) => {
-  const dataForSignature = getDataForSignature(value, validityType, validity)
-
-  privateKey.sign(dataForSignature, (err, signature) => {
-    if (err) {
-      return callback(err)
-    }
-    return callback(null, signature)
-  })
```
diasdavid: You know that here you could have just

```
privateKey.sign(dataForSignature, callback)
```

vasco-santos:
At the moment, this PR is almost ready.

I am working on the interop tests (having some problems with it). Moreover, I have not integrated the `ttl` parameter yet, but since it is still experimental in `go-ipfs`, I think I can create a new PR after this one, for adding support for the experimental `ttl`.

src/core/ipns/republisher.js
```diff
@@ -0,0 +1,124 @@
...
+
+      // unmarshal data
+      let record
+      try {
+        record = ipns.unmarshal(dsVal)
+      } catch (err) {
+        const error = `found ipns record that we couldn't convert to a value`
+
+        log.error(error)
+        return callback(error)
```
alanshaw: `return callback(new Error(error))` - and I think that's why it's better called `errMsg` 😉

src/core/ipns/republisher.js
```diff
@@ -0,0 +1,124 @@
...
+    ], (err, value) => {
+      if (err) {
+        return callback(err.code === ERR_NO_ENTRY_FOUND ? null : err)
+      }
+
+      this.publisher.publishWithEOL(privateKey, value, defaultRecordLifetime, callback)
+    })
+  }
+
+  getLastValue (id, callback) {
```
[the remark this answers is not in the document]
alanshaw: rename to `getPreviousValue`? When I first looked at this I assumed it was getting the last item of an array or something.

src/core/ipns/resolver.js
```diff
@@ -0,0 +1,137 @@
...
+const defaultMaximumRecursiveDepth = 32
+
+class IpnsResolver {
+  constructor (routing, repo) {
+    this.routing = routing
+    this.repo = repo
+    this._resolver = undefined // TODO Routing - add Router resolver
+  }
+
+  resolve (name, peerId, options, callback) {
```
alanshaw: ensure name is a string, allow options to be optional

src/core/ipns/resolver.js
```diff
@@ -0,0 +1,137 @@
...
+const ERR_NO_LOCAL_RECORD_FOUND = 'ERR_NO_LOCAL_RECORD_FOUND'
+const ERR_RESOLVE_RECURSION_LIMIT = 'ERR_RESOLVE_RECURSION_LIMIT'
+
+const defaultMaximumRecursiveDepth = 32
+
+class IpnsResolver {
+  constructor (routing, repo) {
+    this.routing = routing
+    this.repo = repo
+    this._resolver = undefined // TODO Routing - add Router resolver
```
alanshaw: Please `_` private properties and methods

src/http/api/routes/name.js
```diff
@@ -0,0 +1,25 @@
...
+      validate: resources.name.resolve.validate
+    }
+  })
+
+  api.route({
+    method: '*',
+    path: '/api/v0/name/publish',
+    config: {
+      handler: resources.name.publish.handler,
+      validate: resources.name.resolve.validate
```
alanshaw: `resources.name.resolve.validate` => `resources.name.publish.validate`

src/cli/commands/name/resolve.js
```diff
@@ -0,0 +1,37 @@
...
+      nocache: argv.nocache,
+      recursive: argv.recursive
+    }
+
+    argv.ipfs.name.resolve(argv['name'], opts, (err, result) => {
+      if (err) {
+        throw err
+      }
+
+      print(`result: ${result}`)
```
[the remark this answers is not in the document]
vasco-santos: It printed before 😛 Fixed!

src/core/components/name.js
```diff
@@ -0,0 +1,160 @@
...
+
+  const pass = ipfsNode._options.pass
+
+  waterfall([
+    (cb) => ipfsNode._keychain.exportKey(kname, pass, cb),
+    (pem, cb) => crypto.keys.import(pem, pass, cb)
+  ], (err, privateKey) => {
+    if (err) {
+      log.error(err)
+      return callback(Object.assign(new Error(err), { code: ERR_CANNOT_GET_KEY }))
```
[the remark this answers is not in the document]
vasco-santos: Bad refactor 👍

src/core/components/name.js
```diff
@@ -0,0 +1,160 @@
...
+        log.error(err)
+        return callback(err)
+      }
+
+      series([
+        (cb) => human(lifetime || '1s', cb),
+        // (cb) => ttl ? human(ttl, cb) : cb(),
+        (cb) => keyLookup(self, key, cb),
+        // verify if the path exists, if not, an error will stop the execution
+        (cb) => resolve.toString() === 'true' ? path.resolvePath(self, value, cb) : cb()
```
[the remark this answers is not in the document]
vasco-santos: Yes! If `resolve` param is true (which is its default value), we should verify if that path already exists (for instance, before publishing an IPNS record, verify if the IPFS path associated was previously added). This way, if it was not found, an error is thrown and the execution of the publish is stopped. Was clear? Or do you have other suggestion for this?

src/core/components/name.js
```diff
@@ -0,0 +1,160 @@
...
+     * @param {function(Error)} [callback]
+     * @returns {Promise|void}
+     */
+    publish: promisify((value, options, callback) => {
+      if (typeof options === 'function') {
+        callback = options
+        options = {}
+      }
+
+      const { resolve = true, lifetime = '24h', key = 'self' } = options
```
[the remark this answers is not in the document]
vasco-santos: You are right! Will change it 👍

src/core/components/name.js
```diff
@@ -0,0 +1,160 @@
...
+      const { nocache = false, recursive = false } = options
+
+      const local = true // TODO ROUTING - use self._options.local
+
+      if (!self.isOnline()) {
+        const error = errors.OFFLINE_ERROR
+
+        log.error(error)
+        return callback(new Error(error))
+      }
```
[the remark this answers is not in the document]
vasco-santos: you are right!

src/core/ipns/publisher.js
```diff
@@ -0,0 +1,191 @@
...
+const ERR_STORING_IN_DATASTORE = 'ERR_STORING_IN_DATASTORE'
+const ERR_UNEXPECTED_DATASTORE_RESPONSE = 'ERR_UNEXPECTED_DATASTORE_RESPONSE'
+
+const defaultRecordTtl = 60 * 60 * 1000
+
+// IpnsPublisher is capable of publishing and resolving names to the IPFS routing system.
+class IpnsPublisher {
+  constructor (routing, repo) {
+    this.routing = routing
+    this.repo = repo
```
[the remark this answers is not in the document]
vasco-santos: all private yeah!

src/core/ipns/publisher.js
```diff
@@ -0,0 +1,191 @@
...
+          log.error(error)
+          return callback(Object.assign(new Error(error), { code: ERR_INVALID_IPNS_RECORD }))
+        }
+      } else if (err.notFound) {
+        if (!checkRouting) {
+          return callback(null, null)
+        }
+        // TODO ROUTING - get
+      } else {
+        const error = `unexpected error getting the ipns record ${peerIdResult.id} from datastore`
```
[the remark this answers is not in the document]
vasco-santos: Totally agree! I would like that we have some guidelines on error handling

src/core/ipns/republisher.js
```diff
@@ -0,0 +1,124 @@
...
+    // We can't currently *do* that because go uses this method for now.
+    this.republishEntry(privateKey, (err) => {
+      if (err) {
+        const error = 'cannot republish entry for the node\'s private key'
+
+        log.error(error)
+        return
+      }
+
+      if (this.ipfs._keychain && Boolean(pass)) {
```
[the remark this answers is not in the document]
vasco-santos: Yes! Maybe I should leave a comment there. When there is no pass, you are not able to access `keychain`, and consequently you cannot decrypt your keys.

src/core/components/name.js
```diff
@@ -0,0 +1,160 @@
...
+        // Calculate lifetime with nanoseconds precision
+        const pubLifetime = results[0].toFixed(6)
+        const privateKey = results[1]
+
+        // TODO IMPROVEMENT - Handle ttl for cache
+        // const ttl = results[1]
+        // const privateKey = results[2]
+
+        // Start publishing process
+        self._ipns.publish(privateKey, value, pubLifetime, callback)
```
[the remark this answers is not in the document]
vasco-santos: Don’t understand the resolved value. The result of: `path.resolvePath(self, value, cb)`? If so, it returns a Dag.Node for `ipfs` or the result of resolving an `ipns` path.

src/core/ipns/republisher.js
```diff
@@ -0,0 +1,124 @@
...
+
+class IpnsRepublisher {
+  constructor (publisher, ipfs) {
+    this.publisher = publisher
+    this.ipfs = ipfs
+    this.repo = ipfs._repo
+  }
+
+  start () {
+    setInterval(() => {
```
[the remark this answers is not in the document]
vasco-santos: true! Fixed now

src/core/ipns/path.js
```diff
@@ -0,0 +1,101 @@
...
+  const cid = new CID(value)
+  CID.validateCID(cid)
+
+  return `/ipfs/${value}`
+}
+
+module.exports = {
+  resolvePath,
+  parsePath
+}
```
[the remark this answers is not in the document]
vasco-santos: I agree that we should centralize all these utilities, as well as others that probably exist across many other repos. We should follow the go team example and create a [go-path repo](https://github.com/ipfs/go-path).

src/core/components/name.js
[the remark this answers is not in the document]
alanshaw: e.g.

1. `/ipns/ipfs.io` resolves to `/ipfs/QmYNQJoKGNHTpPxCBPh9KkDpaExgd2duMa3aF6ytMpHdao`
2. `/ipfs/Qmcxw9wCZTPm4MArAitog7amDqE2qLd6EKz2prpUvykeKG/gif/unicorn-pug.jpg` resolves to
`/ipfs/QmRU85XyXTfFHRetAKSzj6jNGHUNNkrsDAqJqa5k15DvP4`

Expectations:

1. `ipfs name publish --resolve=true /ipns/ipfs.io` publishes `/ipfs/QmYNQJoKGNHTpPxCBPh9KkDpaExgd2duMa3aF6ytMpHdao`
2. `ipfs name publish --resolve=false /ipns/ipfs.io` publishes `/ipns/ipfs.io`
3. `ipfs name publish --resolve=true /ipfs/Qmcxw9wCZTPm4MArAitog7amDqE2qLd6EKz2prpUvykeKG/gif/unicorn-pug.jpg` publishes `/ipfs/QmRU85XyXTfFHRetAKSzj6jNGHUNNkrsDAqJqa5k15DvP4`
4. `ipfs name publish --resolve=false /ipfs/Qmcxw9wCZTPm4MArAitog7amDqE2qLd6EKz2prpUvykeKG/gif/unicorn-pug.jpg` publishes `/ipfs/Qmcxw9wCZTPm4MArAitog7amDqE2qLd6EKz2prpUvykeKG/gif/unicorn-pug.jpg`

Are my expectations incorrect? Also, there might be a bug in go-ipfs... https://github.com/ipfs/go-ipfs/issues/5368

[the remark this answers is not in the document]
vasco-santos: It does the same thing as go-ipfs does: `// verify the path exists`

The docs could be better though 😕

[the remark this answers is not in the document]
alanshaw: Ok I understand now!

src/cli/commands/name/resolve.js
```diff
@@ -0,0 +1,41 @@
...
+    argv.ipfs.name.resolve(argv.name, opts, (err, result) => {
+      if (err) {
+        throw err
+      }
+
+      if (result && result.path) {
+        print(result.path)
+      } else {
+        print(result)
+      }
```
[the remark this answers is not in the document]
vasco-santos: that case happens when the daemon is offline

src/core/ipns/publisher.js
```diff
@@ -0,0 +1,271 @@
...
+        if (!err.notFound) {
+          const errMsg = `unexpected error getting the ipns record ${peerId.id} from datastore`
+
+          log.error(errMsg)
+          return callback(errcode(new Error(errMsg), 'ERR_UNEXPECTED_DATASTORE_RESPONSE'))
+        } else {
+          if (!checkRouting) {
+            return callback(null, null)
+          } else {
+            // TODO ROUTING - get
```
alanshaw: Can we please `return callback(new Error('not implemented'))` for now?

src/core/ipns/republisher.js
```diff
@@ -0,0 +1,161 @@
...
+  stop (cb) {
+    this._canceled = true
+    if (this._timeoutId || !this._onCancel) {
+      // Not running
+      clearTimeout(this._timeoutId)
+      return cb()
+    }
+
+    this._onCancel = cb
+  }
```
alanshaw: I think this needs a bit of a refactor so that start and stop can be called multiple times. Something like this should work:

```js
  constructor (publisher, ipfs) {
    this._publisher = publisher
    this._ipfs = ipfs
    this._repo = ipfs._repo
    this._republishHandle = null
  }

  start () {
    if (this._republishHandle) throw new Error('already running')

    const republishHandle = {
      _onCancel: null,
      _timeoutId: null,
      runPeriodically (fn, period) {
        republishHandle._timeoutId = setTimeout(() => {
          republishHandle._timeoutId = null

          fn((nextPeriod) => {
            // Was republish cancelled while fn was being called?
            if (republishHandle._onCancel) {
              return republishHandle._onCancel()
            }
            // Schedule next
            republishHandle.schedule(fn, nextPeriod || period)
          })
        }, period)
      },
      cancel (cb) {
        // Not currently running a republish, can callback immediately
        if (republishHandle._timeoutId) {
          clearTimeout(republishHandle._timeoutId)
          return cb()
        }
        // Wait for republish to finish then call callback
        republishHandle._onCancel = cb
      }
    }

    const { privKey } = this._ipfs._peerInfo.id
    const { pass } = this._ipfs._options

    republishHandle.runPeriodically((done) => {
      this._republishEntries(privKey, pass, () => done(defaultBroadcastInterval))
    }, minute)

    this._republishHandle = republishHandle
  }

  stop (cb) {
    const republishHandle = this._republishHandle

    if (!republishHandle) {
      return cb(new Error('not running'))
    }

    this._republishHandle = null
    republishHandle.cancel(cb)
  }
```

vasco-santos: you are right Alan! Thanks 🙂

PR #1405, opened 2018-06-20
test: use pin tests

diasdavid:


test/core/node.js
```diff
@@ -0,0 +1,5 @@
+'use strict'
+
+require('./pin')
+require('./pin-set')
+// require('./key-exchange')
```
diasdavid: Does anyone know why these tests were never being run? @richardschneider ??

richardschneider: @diasdavid No idea, the tests should be run.

diasdavid: They weren't being picked up anywhere. I'll add them as a separate PR then.

diasdavid:
Thank you @alanshaw ❤️

PR #1413, opened 2018-06-28
chore(tests): share ipfs instances between tests

wraithgar:
It accounts for the tests being run from either the cli/index.js
file, or the individual files themselves, programatically determining
whether or not it is responsible for shutting down the instance or
cleaning up its repoPath

This speeds up cli tests by about 25% on my local machine, which I know
is not a benchmark in and of itself but it's at least a measurement

### DIFF
--- a/test/cli/index.js
+++ b/test/cli/index.js
@@ -2,8 +2,41 @@
 'use strict'
 
 const fs = require('fs')
+const spawn = require('../utils/on-and-off')
+const parallel = require('async/parallel')
 
 describe('cli', () => {
+  let daemonOff
+  let ipfsd
+  before(function (done) {
+    // CI takes longer to instantiate the daemon,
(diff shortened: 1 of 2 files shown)


alanshaw:
I'm mildly against re-using IPFS nodes between tests. Even re-using nodes within a suite (which we currently do) feels wrong. That said, I could be persuaded.

I've always been of the opinion that each test should start from a clean slate, do any setup it needs and teardown afterwards. It means that any state accumulated during a test does not affect the results of any other tests.

I have an OKR to reduce the total testing time by 50% this quarter so I'll be investigating causes and evaluating solutions (including this one) for the tests taking so long.

wraithgar:
My feeling was that if we are not relying on the state of a node being in any given state for a test to pass, what's the difference between a clean one and a reused one?

PR #1414, opened 2018-06-29
perf: use lodash

alanshaw:
Other modules we depend on also depend on lodash.

screen shot 2018-06-29 at 10 19 53

We actually save around 100KB on bundle size if we use lodash like we use async (require functions we need from the main package) since the functions we require can be deduped because they're already being depended on.


vasco-santos:
All in for this change! 🚀

vasco-santos:
I will change those for:

`js-libp2p-connection-manager`
`js-libp2p-railing`

alanshaw:
Other modules that IPFS are in charge of that are using individual lodash modules that would help our bundle size if they switched:

- `multiaddr`
    - `lodash.filter` -> `lodash/filter`
    - `lodash.map` -> `lodash/map`
- `ipfs-bitswap`
    - `lodash.groupby` -> `lodash/groupBy`
    - `lodash.find` -> `lodash/find`
    - `lodash.sortBy` -> `lodash/sortBy`
    - `lodash.isequalwith` -> `lodash.isEqualWith`
    - `lodash.uniqwith` -> `lodash/uniqWith`
    - `lodash.values` -> `lodash/values`
    - `lodash.debounce` -> `lodash/debounce`
    - `lodash.pullallwith` -> `lodash/pullAllWith`
- `peer-info`
    - `lodash.uniqby` -> `lodash/uniqBy`
- `ipfs-repo`
    - `lodash.has` -> `lodash/has`
    - `lodash.set` -> `lodash/set`
    - `lodash.get` -> `lodash/get`
- `ipld`
    - `lodash.includes` => `lodash/includes`
- `libp2p-switch`
    - `lodash.includes` => `lodash/includes`
- `libp2p-websockets`
    - `lodash.includes` => `lodash/includes`
- `libp2p-tcp`
    - `lodash.includes` => `lodash/includes`
- `libp2p-floodsub`
    - `lodash.values` -> `lodash/values`

hugomrdias:
All maintainers should aim to keep these types of requires consistent. 
There's also other types that should be normalised to keep node and browser happy, like `setimmediate`, `buffer`, `stream` and promisify/nodify are some i can remember from the top of my head.

hugomrdias:
also consider native methods if they can be used in the repo context. `includes` is a good example probably

vasco-santos:
Maybe we should create a endeavour listing everything. What do you think?

Meanwhile, I created a PR also for `js-libp2p-floodsub`

PR #1419, opened 2018-07-04
fix(core/components/dag): fix default hash algorithm for put() api

PascalPrecht:
Turns out we've introduced a little bug in https://github.com/ipfs/js-ipfs/commit/d299ed77c288d82ce807af0ec7a5b2605336626f
in which the default hash algorithm has a typo.

This didn't fail on CI because the `options.hashAlg` passed down to `ipld.put()`
is actually not being used at time of this commit.

For more info see: https://github.com/ipld/js-ipld/blob/d3d78e053ebd3f8c85a8b5579715be8812773d4d/src/index.js#L236-L239

License: MIT
Signed-off-by: Pascal Precht <pascal.precht@gmail.com>

### DIFF
--- a/src/core/components/dag.js
+++ b/src/core/components/dag.js
@@ -19,7 +19,7 @@ module.exports = function dag (self) {
 
       const optionDefaults = {
         format: 'dag-cbor',
-        hashAlg: 'sha2-255'
+        hashAlg: 'sha2-256'
       }
 
       options = options.cid ? options : Object.assign({}, optionDefaults, options)


PascalPrecht:
@alanshaw sorry this one slipped through 🙇🏻‍♂️

alanshaw:
Good catch, thank you 🚀 . Would you be willing to add a test to `interface-ipfs-core` for using a hash algorithm other than the default? We'd need to skip it here until https://github.com/ipld/js-ipld/pull/133 is merged and released.

PascalPrecht:
I can certainly do that. But just to clarify: You want a test that simply checks whether a non-default hashAlg is used?

Cause that's kind of covered in the existing tests: https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/dag/put.js#L55-L81

Although, those don't really check if the resulting `CID` resolves to the configured hash algorithm.

Maybe it makes sense to change those to look more like the one here? https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/dag/put.js#L103-L110 

Otherwise I can also just add a new one that checks on the resolved CID similar to the one above.

PR #1420, opened 2018-07-04
feat: adds data-encoding argument to control data encoding

achingbrain:
This is the js counterpart to ipfs/go-ipfs#5139 and will allow the test in ipfs/interface-ipfs-core#302 to pass.

ipfs/js-ipfs-api#806 needs this to be merged before the interface test linked to above will pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "hoek": "^5.0.3",
     "human-to-milliseconds": "^1.0.0",
     "interface-datastore": "~0.4.2",
-    "ipfs-api": "^22.2.4",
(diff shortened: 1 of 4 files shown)


achingbrain:
This requires go-ipfs 0.4.17 to be released first - not much point merging this until then.

alanshaw:
@achingbrain this will need the fixes in https://github.com/ipfs/js-ipfs/pull/1508 before it'll pass CI

achingbrain:
I was beginning to think CI would never pass. Quick, merge it!

Issue #1422, opened 2018-07-05
Example: Browser video streaming broken


PascalPrecht:
- **Version**: https://github.com/ipfs/js-ipfs/commit/c01a83a67f0bb14b0a6ed22b3a66601ee1a48cde
- **Platform**: Darwin slotmachine-3002.local 17.6.0 Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64 x86_64
- **Subsystem**: js-ipfs

#### Description:

In latest master (https://github.com/ipfs/js-ipfs/commit/c01a83a67f0bb14b0a6ed22b3a66601ee1a48cde at the time of creating this issue), the `browser-video-streaming` example is broken and throws with:

```
index.js:93 Uncaught TypeError: stream.on is not a function
    at index.js:93
    at pull.collect (files.js:290)
    at reduce.js:10
    at drain.js:20
    at filter.js:17
    at map.js:19
    at next (index.js:22)
    at index.js:39
    at Array.<anonymous> (index.js:10)
    at next (index.js:26)
```

First I thought it might've been related to Chrome, as this is an HLS example, so I tried it with Safari. That one doesn't work either.

#### Steps to reproduce the error:
Simply following the example instructions:

- `$ serve .`
- Opening the browser on localhost:<PORT>
- Check console

alanshaw:
@PascalPrecht thanks for reporting - did you find issues with any of the other examples?

alanshaw:
Looks like https://github.com/moshisushi/hlsjs-ipfs-loader/blob/f0df3ff963bfb0ed8b4e6faf3aefa8e69a72d941/src/index.js#L90 needs to use `catReadableStream` instead...

PascalPrecht:
Have been through a couple other examples, not all of them yet as I had to do some other work.

I'll take a look at the hlsjs-ipfs-loader thing.

alanshaw:
Since that's an external dependency I think it would be more valuable to continue through the examples and come back to it later.

PascalPrecht:
Gotcha!

seungwon-kang:
I builded the hlsjs-ipfs-loader (https://github.com/moshisushi/hlsjs-ipfs-loader) and tested it, but it works well. I found that the recent version of it works well.

PascalPrecht:
@seungwon-kang are you saying you couldn't reproduce the bug in the version of the issue?

[...]

PR #1427, opened 2018-07-06
fix: update hlsjs-ipfs-loader version (#1422)

seungwon-kang:
The recent version of hsljs-ipfs-loader works well with no error (#1422)


PascalPrecht:
LGTM

diasdavid:
Thank you @seungwon-kang for the fix and @PascalPrecht for reporting!

PR #1429, opened 2018-07-10
fix: aegir docs fails if outer function is called pin

alanshaw:
It's legal javascript, but the docs command fails if this function is called pin 😕 

You can test this fix out by running `npx aegir docs` in the project directory.


alanshaw:
LGTM 😝

PR #1430, opened 2018-07-10
docs: adds mfs browser example

achingbrain:
Adds an interactive MFS example that runs in the browser!

Also upgrades to the latest js-ipfs-mfs to fix some bugs around handling errors when trying to copy files/directories onto themselves.


achingbrain:
@alanshaw have pushed a fix, please try again.

PR #1433, opened 2018-07-12
docs: fix libp2p peer discovery modules config key

alanshaw:


diasdavid:
@alanshaw seems that there is a pinSet test that might be slow

```
  1) pinSet walkItems visits all non-fanout links of a root node:

     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

```

Issue #1061, opened 2017-11-06
state machine, transitions and events


pgte:
- **Version**: 0.26.0
- **Platform**: All
- **Subsystem**: core, repo, init, swarm, bitswap, libp2p, perhaps others..

#### Type:

Enhancement

#### Severity:

Medium

#### Description:

Initialisation and stopping an IPFS node envolves many states and transitions that are hard to handle, can fail and have many corner cases.

Currently, initialisation states and their transitions are verified and managed explicitly at code level. From my experience, this almost invariantly conducts to bugs and weirdness in app start code.

Instead, a finite state machine that contains all the main, intermediary and error states should be well defined and documented.

The initialisation code should be driven by state transitions, and not the other way around (like it is now). This will avoid double user actions, like starting or stopping the node twice concurrently, or trying to stop after stopped, etc.

----

IMO, special attention should be given so that the implementation:

- is __consistent__: for instance, even if repo initialisation is not required, a `initialised` or equivalent transition event should be fired.
- Some operations should be __idempotent__: starting a node that is started or is starting should not lead to an error, much less triggering the `startup` procedure again.
- Any Irrecoverable error in initialisation should lead to an dead-end "error" state.

diasdavid:
@pgte you have my full blessing to push the changes you have been describing :)

PR #1437, opened 2018-07-12
fix: double pre start

### LINKED ISSUE #1162: Multiple calls to pre-start
It appears (from debugging) that `jsipfs daemon` calls [pre-start]() twice.

| > jsipfs daemon
```
C:\Users\Owner\Documents\GitHub\js-ipfs>jsipfs daemon
Initializing daemon...
  jsipfs booting +0ms
  jsipfs pre-start +175ms
  jsipfs starting +921ms
  jsipfs pre-start +0ms
Swarm listening on /ip4/127.0.0.1/tcp/4003/ws/ipfs/QmWfa2ZtmrbyHxkHUNpMs35rcRSf5bc25D6Fb1NFgA76RA
Swarm listening on /ip4/192.168.178.21/tcp/4002/ipfs/QmWfa2ZtmrbyHxkHUNpMs35rcRSf5bc25D6Fb1NFgA76RA
Swarm listening on /ip4/127.0.0.1/tcp/4002/ipfs/QmWfa2ZtmrbyHxkHUNpMs35rcRSf5bc25D6Fb1NFgA76RA
  jsipfs boot:done null +2s
API is listening on: /ip4/127.0.0.1/tcp/5001
Gateway (readonly) is listening on: /ip4/127.0.0.1/tcp/9090
Daemon is ready
```

alanshaw:
This is a fix for `preStart` being called twice but also a partial refactor of the boot sequence.

The problem is that `preStart` always needs to be called between `init` and `start` and was actually being called in `boot.js` (via the `IPFS` constructor if repo exists or is newly initialized) as well as in `start.js`.

The following permutations of constructing, initializing and starting show how `preStart` was being called twice:

```js
const ipfs = new IPFS({ init: false, start: false })
ipfs.on('ready', () => {
  await ipfs.init()
  await ipfs.start()
})
```

`preStart` is called ✅ 
`preStart` is called once from `ipfs.start()` ✅ 

```js
const ipfs = new IPFS({ init: true, start: false })
ipfs.on('ready', () => {
  await ipfs.start()
})
```

`preStart` is called ✅ 
`preStart` is called once from `boot.js` (via the constructor) and again in `ipfs.start()` ❌ 

```js
const ipfs = new IPFS({ init: true, start: true })
ipfs.on('ready', () => {})
```

`preStart` is called ✅ 
`preStart` is called once from `boot.js` (via the constructor) and again via `boot.js` since it calls `ipfs.start()` ❌ 

---

So having `preStart` called in `boot.js` and `ipfs.start` ensures it is always called, but in most cases it ends up being called twice.

This PR moves the call to `preStart` into `init` and ensures `init` is always called even if the repo already exists.

towards #1061
fixes #1162

### DIFF
--- a/src/cli/commands/daemon.js
+++ b/src/cli/commands/daemon.js
@@ -35,7 +35,7 @@ module.exports = {
     httpAPI = new HttpAPI(process.env.IPFS_PATH, null, argv)
 
     httpAPI.start((err) => {
-      if (err && err.code === 'ENOENT' && err.message.match(/not initialized/i)) {
+      if (err && err.code === 'ENOENT' && err.message.match(/uninitialized/i)) {
         print('Error: no initialized ipfs repo found in ' + repoPath)
         print('please run: jsipfs init')
         process.exit(1)
--- a/src/cli/commands/daemon.js
+++ b/src/cli/commands/daemon.js
@@ -35,7 +35,7 @@ module.exports = {
     httpAPI = new HttpAPI(process.env.IPFS_PATH, null, argv)
 
     httpAPI.start((err) => {
-      if (err && err.code === 'ENOENT' && err.message.match(/Uninitalized repo/i)) {
+      if (err && err.code === 'ENOENT' && err.message.match(/uninitalized/i)) {
         print('Error: no initialized ipfs repo found in ' + repoPath)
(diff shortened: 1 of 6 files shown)


alanshaw:
Correct :D

PR #1439, opened 2018-07-13
chore: update mfs dep to make interop tests pass

achingbrain:
The new version updates the error messages to be the same as go, and fixes the case where `mkdir`ing an existing directory with the `-p` option should not cause an error to occur.


package.json
```diff
@@ -108,7 +108,7 @@
     "ipfs-block": "~0.7.1",
     "ipfs-block-service": "~0.14.0",
     "ipfs-http-response": "~0.1.2",
-    "ipfs-mfs": "~0.0.16",
+    "ipfs-mfs": "~0.0.17",
```
diasdavid: Please release a minor version. Seems that the error message changed and that should be considered a breaking change.

achingbrain: Hmm, I should have done that [ages ago](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase):

> How should I deal with revisions in the 0.y.z initial development phase?
>
>The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.

PR #1443, opened 2018-07-16
fix: fixing circuit-relaying example

dryajov:


dryajov:
thanks for the heads up @PascalPrecht

examples/circuit-relaying/package.json
```diff
@@ -15,11 +15,10 @@
   "license": "MIT",
   "dependencies": {
     "ipfs": "file:../../",
-    "ipfs-pubsub-room": "~0.3.0"
+    "ipfs-pubsub-room": "^1.4.0"
   },
   "devDependencies": {
     "aegir": "^14.0.0",
-    "http-server": "~0.10.0",
```
PascalPrecht: One thing though: Why exactly is this going to be removed? Maybe it shouldn't have been there in the first place, but I assume this has been put there for a reason and it doesn't seem to be necessary for this commit to do its work (fixing that example)

PascalPrecht:
@dryajov I left one more minor comment, sorry didn't catch this earlier.

examples/circuit-relaying/package.json
dryajov: It was missed. I left it there because the example was initially based on browserify and later moved to parcel.

PascalPrecht: 👍 got it!

PascalPrecht:
This LGTM then

PR #1448, opened 2018-07-17
chore: enable dag.put with hashAlg test

alanshaw:
https://github.com/ipld/js-ipld/pull/133 landed and fixed creating dag nodes with a different hashAlg or CID version and we can now enable this test.


PascalPrecht:
🎉

PR #1452, opened 2018-07-18
Proposal: add a promise that resolves on-ready.

mikeal:
This is something I noticed when creating the workshopper. If people want to use async functions they end up having to jump through some very annoying hoops to not use the instance until on-ready.

This removes about 3 lines of text from the workshop boilerplate.


mikeal:
Closing, this branch was built against the wrong reference branch.

PR #1461, opened 2018-07-22
[WIP] fix: add default datastore config

### LINKED ISSUE #1460: Peer default config must have default Datastore config
- **Version**: 0.30.1
- **Platform**: All

#### Severity: High

This is currently a blocker for interop tests for `IPNS`, among other interop tests.

#### Description:

Both `js-ipfs` and `go-ipfs` repos have not been compatible for a long time. 

`go-ipfs` introduced the `datastore_spec` in version `0.4.11` as we can check in the [Changelog](https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#0411-2017-09-14) and more specifically, in the following commits:
- [Write the spec file during initialization only](https://github.com/ipfs/go-ipfs/commit/ce62bed82e66343122b3b67325bf71203a36ba11#diff-8a36823aee47301ed06bf1f9e879b1e9R247)
- [Implement DiskId()](https://github.com/ipfs/go-ipfs/commit/5027b4b14f2fc60e808a39d3b3efc9316d8e511d#diff-8a36823aee47301ed06bf1f9e879b1e9R360)

Taking into consideration the `go-ipfs` repo, it requires that a `datastore_spec` files exists in the repo, the `Datastore` configuration should be in the `config` file and both must match for starting a node with this repo. This way, when we create a repo with a JS daemon (it does not create the `datastore_spec` nor has the `Datastore` configuration in the `config` file) and we try to use it with a GO daemon, it fails.

With a view to make interop tests between both implementations, we need to have a repo that works with both. 

Firstly, it is mandatory that the `js-ipfs-repo` creates the mentioned `datastore_spec` file in the init function as the go implementation does [here](https://github.com/ipfs/go-ipfs/blob/master/repo/fsrepo/fsrepo.go#L292). I created a PR [js-ipfs-repo#173](https://github.com/ipfs/js-ipfs-repo/pull/173) for adding this `datastore_spec` file to the repo.

Moreover, we need to have a default datastore config for the peer. We can provide a config when starting a daemon with the `js-ipfsd-ctl`, but I believe that we should have the same default configuration as `go-ipfs`, and consequently, have the `datastore` default config specified.

There are two PRs referencing this problem in `ipfs/interop`: 
- [ipfs/interop#26](https://github.com/ipfs/interop/pull/26)
- [ipfs/interop#28](https://github.com/ipfs/interop/pull/28)

vasco-santos:
From my analysis, I concluded that our default repo datastore is equal to the Go one, but I would like to have an extra pair of eyes to check it. 

This PR is part of another PR in `js-ipfs-repo` that aims to guarantee the interoperability of repos between `js-ipfs` and `go-ipfs`

It will be necessary to handle other configuration, what is not supported right now. Should I create an issue for tracking it? Or should this be a responsibility of `js-ipfs-repo`? 

Fixes #1460


vasco-santos:
+1 on having an example for this @jacobheun . I went deeper into the `repo` side for creating this set of PRs, but I am still confused about the possible configurations available. I think we have a clear lack of documentation and specs in the repo. 

The problem with this PR is that we can specify the datastore in config, but it will end up with the same configuration anyway, as far as I understood. Correct?

jacobheun:
Yes, we definitely need to get more specs down for the repo, or we're at a higher risk of go and js diverging again in the future.

For configuration we really need to go one way. Currently as listed in that example it's all programatic, but as we implement the spec config we should probably transition to just using that. We'll need to support both to avoid breaking existing projects, but we should deprecate the current format.

Ideally we'd also have configurations for each of the datastores listed clearly in their individual repos so users can easily copy those into the mounts array. I think go has all of those in the go-ipfs repo itself but I think that makes adding new datastore support more cumbersome. 

This PR will require an update to ipfs-repo, which you have started https://github.com/ipfs/js-ipfs-repo/pull/173, to make sure that it fails if there is a mismatch in configuration. We'll want to make sure to add basic tests here as well, with more detailed tests in ipfs/interop.

jacobheun:
> So, js-ipfs should send the provided config to js-ipfs-repo, and we should make sure that js-ipfs-repo fails if the spec provided is not compatible to what it knows. And here in js-ipfs, we should just a test with an incompatible spec, and guarantee that we receive the error. Correct?

Yes, and if the repo already exists and the spec is different it should probably fail as well. Yes, I think tests with a good and bad config would be sufficient here. More unit tests in js-ipfs-repo itself and the interop tests in interop should be sufficient.

vasco-santos:
After a productive call with @jacobheun we created an [endeavour issue](https://github.com/ipfs/js-ipfs/issues/1467) to tackle all the tasks regarding the repo interoperability. This issue aims to tackle the following set of tasks:

#### 2) [ipfs/js-ipfs](https://github.com/ipfs/js-ipfs)

- [ ] **1.**  Send the default/provided config (with datastore.spec)
- [ ] **2.**  Present errors provided by js-ipfs-repo (have tests)
  - **INIT**: if the configuration is not compatible to what repo knows
  - **START**: if the repo is different from the spec

Issue #1459, opened 2018-07-20
Achieving Connectivity Magic


daviddias:
I ran an impromptu and presidential design session with @alanshaw, @olizilla and @achingbrain to walk through some of the design decisions on libp2p and we ended up discussing the Connectivity Magic Objective and what exactly I mean by "creativity".

The tl;dr; there are multiple ways to find file, but all of them have tradeoffs.

### Here is the sketch we ended up doing:

![paper sketches 4](https://user-images.githubusercontent.com/1211152/43016126-bc40e2f6-8c49-11e8-8f6d-c01e67967b58.png)

### Here is an annotated version:

<img width="1297" alt="image" src="https://user-images.githubusercontent.com/1211152/43016400-9ffd1e92-8c4a-11e8-881f-b03d2ad939cb.png">

I can make a quick video explaining this, but perhaps it would be more valuable to have @alanshaw, @olizilla and @achingbrain explain this to @hugomrdias (who owns the KR to get files from the gateways) through a Zoom Recorded Call or Text, so that we check that the memes have indeed been transferred :)

The focus should be around on why the shortcoming exists and the 3 solutions on the solution spectrum (from hacky to ideal but not necessarily memory efficient).

lgierth:
It took me a while to see this: the three steps proposed are: 1) routing via DNS records 2) delegated routing 3) proper DHT routing

diasdavid:
Everyone, thank you for the really awesome sync up. It was lovely to see how everyone came up together so quickly, communicated well their concerns and proposed solutions and eventually converged in a good set of next steps 👏🏽👏🏽👏🏽❤️❤️❤️.

As action items from the call:

- [ ] @lgierth Deploy "Pre Load gateways" preload.gateway.ipfs.io
- [ ] `js-ipfs team` Add new the gateways to the bootstrapper list and hit the preload gateway on every add/get/cat (+ dag.get/put)  (like companion does)
- [ ] `js-ipfs team` release js-ipfs 0.31
- [ ] @hugomrdias integrate it all on js.ipfs.io

Note on hitting the preload gateway: Use the `"x-ipfs-companion-no-redirect"` to avoid getting the preload gateway hit redirected to the local daemon.

lgierth:
Here's the pros/cons of 1a) always connecting to ipfs.io gateways, and 1b) separate preload gateways, that were mentioned in the call:

- ipfs.io gateways
  - Pro: very easy
[...]

PR #1464, opened 2018-07-24
feat: preload content

alanshaw:
refs #1459

This PR adds a new config property `preload`:

```js
new IPFS({
  preload: {
    enabled: false,
    addresses: ['/multiaddr/api/address']
  }
})
```

* `preload.enabled` (default `false`) enables/disabled preloading - **should the default be false?**
* `preload.addresses` array of node API addresses to preload content on. This are the addresses we make a `/api/v0/refs?arg=QmHash` request to, to initiate the preload

**This PR upgrades the following APIs to preload content**. After adding content with `ipfs.files.add` (for example), we make a request to the first preload gateway addresses (providing `preload.enabled` is true), and will fall back to the second etc.

* [x] `dag.put`
* [x] `block.put`
* [x] `object.new`
* [x] `object.put`
* [x] `object.patch.*`
* [x] `mfs.*`

MFS preloading is slightly different - we periodically submit your MFS root to the preload nodes when it changes.

NOTE: this PR adds an option to `dag`, `block` and `object` APIs allowing users to opt out of preloading by specifying `preload: false` in their options object.

TODO

* [x] ~~This PR will make a head request to the root hash of uploaded content, I need to double check that this will preload that hash **as well as** all of it's descendants or if we need to send a preload request for EVERY hash we add~~ Now uses `/api/v0/refs?arg=QmHash` instead
* [x] Add the new bootstrap and gateway addresses for the preload nodes when provided by @lgierth


src/core/preload.js
```diff
@@ -0,0 +1,60 @@
...
+const CID = require('cids')
+const preload = require('./runtime/preload-nodejs')
+
+const log = debug('jsipfs:preload')
+log.error = debug('jsipfs:preload:error')
+
+// Tools like IPFS Companion redirect requests to IPFS gateways to your local
+// gateway. This is a hint to those tools that they shouldn't redirect these
+// requests as they will effectively disable the preloading.
+const redirectOptOutHint = 'x-ipfs-preload'
```
lidel: Something to note is that browser Companion expects [`x-ipfs-companion-no-redirect`](https://github.com/ipfs-shipyard/ipfs-companion/blob/a6583ae8065640b367232d819c509af75e3002b1/add-on/src/lib/ipfs-request.js#L6).
If it does not see it in URL, it will redirect to local gateway, which we don't want (local node on a laptop may go offline).

We can add another URL hint to Companion, but is it necessary?
To limit the number of moving pieces before Dweb Summit I would just use `x-ipfs-companion-no-redirect` for now.

diasdavid:
>  enables/disabled preloading - should the default be false?

I propose that we default to true. It improve user experience a ton.

lidel:
> This PR will make a head request to the root hash of uploaded content, I need to double check that this will preload that hash as well as all of it's descendants or if we need to send a preload request for EVERY hash we add

@alanshaw check Lars' comment at https://github.com/ipfs/js-ipfs/issues/1459#issuecomment-407620240

IIUC with `/api/v0/refs?r=true&arg=<cid>` when adding N files you only need 1 preload request instead of N+1 (one per file + wrapping dir).

alanshaw:
Other APIs that need to be preloaded:

* [ ] `dag.put`
* [ ] `block.put`
* [ ] `object.new`
* [ ] `object.put`
* [ ] `object.patch.*`

alanshaw:
Preload on MFS - do we? It’ll mean that every `cp`, `mv`, `write` etc. will need a preload. It's potentially a lot of churn and a lot of preload requests.

Half of me feels like MFS is your own storage and you don’t really deal with CIDs. Right now if I wanted to share something with someone via IPFS I'd use `ipfs.add` and share the hash, not copy to MFS, stat path, and then share the hash.

If we don't actively preload MFS content, it'll still be accessible (provided the sharer stays online) due to adding the preload nodes to our bootstrap list, but it _won't be as readily available_.

On the other hand, I think it's confusing to have some content preloaded and other content not. Data added to IPFS should always be readily available and preloading everything will ensure that.

I am leaning towards preload MFS also.

@diasdavid @achingbrain @olizilla can have opinions?

diasdavid:
@alanshaw you can set a rule that when MFS gets used, you set a flag that gets checked every 30s (or so) and if it is `true`, it will preload the current root. This way we don't bombard the preload gateways for every node update.

alanshaw:
I think that could work but we'd have to switch to requesting `/api/v0/refs?r=true&arg=QmFoo` (https://github.com/ipfs/js-ipfs/issues/1459#issuecomment-407620240) instead of the HEAD request otherwise we'd have to recurse over everything in MFS a send (potentially) hundreds of requests.

achingbrain:
> it's confusing to have some content preloaded and other content not

This is probably why it's worth doing.  There will be a lot of network activity and extra storage taken up on the preload peers, and the MFS abstraction becomes really leaky when you have to share new CIDs for a given path every time you update something but it's probably better than not being able to get the content someone else has shared with you.

> set a flag that gets checked every 30s (or so) and if it is true, it will preload the current root.

If you wrap `ipfs.dag.put` with something that does the preloading, every MFS operation will get preloaded as that's what it uses internally.  No `setInterval` required.

README.md
```diff
@@ -231,6 +231,9 @@ Creates and returns an instance of an IPFS node. Use the `options` argument to s
         - `enabled` (boolean): Make this node a relay (other nodes can connect *through* it). (Default: `false`)
         - `active` (boolean): Make this an *active* relay node. Active relay nodes will attempt to dial a destination peer even if that peer is not yet connected to the relay. (Default: `false`)
 
+- `preload` (object): Configure external nodes that will preload content added to this node
+    - `enabled` (boolean): Enable content preloading (Default: `true`)
+    - `addresses` (array): Multiaddr API addresses of nodes that should preload content. NOTE: nodes specified here should also be added to your node's bootstrap address list at `config.Boostrap`
```
diasdavid: We will later need to over all the config values to the Config object like we did for js-libp2p.

src/core/components/index.js
```diff
@@ -26,4 +26,4 @@ exports.dht = require('./dht')
 exports.dns = require('./dns')
 exports.key = require('./key')
 exports.stats = require('./stats')
-exports.mfs = require('ipfs-mfs/core')
+exports.mfs = require('./mfs')
```
diasdavid: 👍

src/core/components/mfs.js
```diff
@@ -0,0 +1,24 @@
...
+      }
+
+      opts = Object.assign({}, opts, { preload: false })
+
+      return self.dag.put(node, opts, cb)
+    })
+  })
+
+  return mfs(mfsSelf, mfsSelf._options)
+}
```
diasdavid: 👍

src/core/config.js
```diff
@@ -8,6 +8,10 @@ const schema = Joi.object().keys({
     Joi.string()
   ).allow(null),
   repoOwner: Joi.boolean().default(true),
+  preload: Joi.object().keys({
+    enabled: Joi.boolean().default(true),
+    addresses: Joi.array().items(Joi.multiaddr().options({ convert: false }))
```
achingbrain: Should we require at least one entry here?

src/core/mfs-preload.js
```diff
@@ -0,0 +1,49 @@
...
+  let timeoutId
+
+  const preloadMfs = () => {
+    self.files.stat('/', (err, stats) => {
+      if (err) {
+        timeoutId = setTimeout(preloadMfs, options.interval)
+        return log.error('failed to stat MFS root for preload', err)
+      }
+
+      if (rootCid !== stats.hash) {
```
achingbrain: Unless I'm missing something, if the root cid has not changed, it looks like we stop checking for changes.

alanshaw: Yes good catch! If only there was a test for this...

olizilla:
🚀

PR #1470, opened 2018-07-26
feat: add libp2p generator config option with example

### LINKED ISSUE #1463: Allow for custom libp2p generator in configuration
Currently libp2p is being created by configuration when ipfs boots. This makes it difficult for users to do anything custom with libp2p, even just adding transports that require the nodes peerInfo on startup, https://github.com/libp2p/js-libp2p/issues/222.

By allowing users to specify a generator for libp2p we provide more flexibility for them to customize it to their needs. 

It would be great if users could do something like this:
```js
const ipfs = new IPFS({
    libp2p: (_ipfs) => {
      /* Fancy stuff here with startup info from the ipfs node */

      return new Libp2p({
        peerInfo: _ipfs._peerInfo,
        peerBook: _ipfs._peerInfoBook,
        config: { /* custom config using the above fancy stuff */ }
      })
    }
  })
```

Then when ipfs goes to start libp2p, https://github.com/ipfs/js-ipfs/blob/v0.30.1/src/core/components/libp2p.js#L11, we can run the generator which would have access to boot time information like the nodes peerInfo.

jacobheun:
Resolves #1463 

This adds the ability to pass a libp2p generator function to the ipfs configuration. This makes it easier for users to add custom modules to libp2p that require some startup time properties, like peerInfo (https://github.com/libp2p/js-libp2p/issues/222). 

I have also included an example of how to do this. I think this makes complex libp2p configuration much cleaner and easier to do.

Tests have been added for the old configuration options and the new generator option. I isolated them to avoid spinning up a full ipfs node.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,12 @@ Creates and returns an instance of an IPFS node. Use the `options` argument to s
 
 - `config` (object) Modify the default IPFS node config. Find the Node.js defaults at [`src/core/runtime/config-nodejs.js`](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/config-nodejs.js) and the browser defaults at [`src/core/runtime/config-browser.js`](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/config-browser.js). This object will be *merged* with the default config; it will not replace it.
 
-- `libp2p` (object) add custom modules to the libp2p stack of your node
+- `libp2p` (object or function(ipfs, config)) add custom modules to the libp2p stack of your node
+
+    The libp2p option allows you to build your libp2p node by configuration, or via a generator. If you are looking to just modify the below options, using the object format is the quickest way to get the default features of libp2p. If you need to create a more customized libp2p node, such as with custom transports or peer/content routers that need some of the ipfs data on startup, a generator is a great way to achieve this.
+
+    You can see the generator in action in the [custom libp2p example](examples/custom-libp2p).
+
     - `modules` (object):
         - `transport` (Array<[libp2p.Transport](https://github.com/libp2p/interface-transport)>): An array of Libp2p transport classes/instances to use _instead_ of the defaults. See [libp2p/interface-transport](https://github.com/libp2p/interface-transport) for details.
         - `peerDiscovery` (Array<[libp2p.PeerDiscovery](https://github.com/libp2p/interface-peer-discovery)>): An array of Libp2p peer discovery classes/instances to use _instead_ of the defaults. See [libp2p/peer-discovery](https://github.com/libp2p/interface-peer-discovery) for details. If passing a class, configuration can be passed using the config section below under the key corresponding to you module's unique `tag` (a static property on the class)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -238,7 +238,12 @@ Creates and returns an instance of an IPFS node. Use the `options` argument to s
 
 - `config` (object) Modify the default IPFS node config. Find the Node.js defaults at [`src/core/runtime/config-nodejs.js`](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/config-nodejs.js) and the browser defaults at [`src/core/runtime/config-browser.js`](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/config-browser.js). This object will be *merged* with the default config; it will not replace it.
 
-- `libp2p` (object) add custom modules to the libp2p stack of your node
+- `libp2p` (object or function(ipfs, config)) add custom modules to the libp2p stack of your node
+
+    The libp2p option allows you to build your libp2p node by configuration, or via a generator. If you are looking to just modify the below options, using the object format is the quickest way to get the default features of libp2p. If you need to create a more customized libp2p node, such as with custom transports or peer/content routers that need some of the ipfs data on startup, a generator is a great way to achieve this.
```
alanshaw: I'm wondering if "generator" might be confused with an actual JS [generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*)? IMHO it would be more appropriate to call it a [factory function](https://en.wikipedia.org/wiki/Factory_method_pattern).

README.md
```diff
@@ -238,7 +238,12 @@ Creates and returns an instance of an IPFS node. Use the `options` argument to s
...
+
+    You can see the generator in action in the [custom libp2p example](examples/custom-libp2p).
```
alanshaw: 💯 for adding an example too!

examples/custom-libp2p/index.js
```diff
@@ -0,0 +1,121 @@
...
+ * This is the generator we will use to generate our fully customized libp2p node.
+ *
+ * @param {*} _ipfsNode The ipfs node. This houses the PeerInfo and PeerBook that modules may need
+ * @param {*} _ipfsConfig The config that is fetched from the ipfs-repo
+ * @returns {Libp2p} Our new libp2p node
+ */
+const libp2pGenerator = (_ipfsNode, _ipfsConfig) => {
+  // Set convenience variables to clearly showcase some of the useful things that are available
+  const peerInfo = _ipfsNode._peerInfo
+  const peerBook = _ipfsNode._peerBook
```
alanshaw: hmm, `_peerInfo` and `_peerBook` presumably are pefixed with underscores because they're considered private APIs - I don't think we should encourage people to use them like this!

Would it be simpler and also still solve the issue (https://github.com/libp2p/js-libp2p/issues/222) to just allow passing `peerInfo` and `peerBook` to transports in the same way we do for peer discovery modules? https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L185

I see the value in potentially being able to subclass `Libp2p` and also passing in the IPFS config, but if we don't have a use case for those two features I'd be tempted to go for a simpler approach for now to directly address @pgte's issue and which is also beneficial to people who use libp2p directly too.

alanshaw: Ah, you might have to ignore that last comment (apologies it's late and I should be asleep already), I just looked at the code being used https://github.com/ipfs-shipyard/peer-star-app/blob/master/src/transport/ipfs.js#L26 and the issue is not being able to use the same instance in both `transport` _and_ `peerDiscovery`.

Let me re-review this tomorrow.

README.md
jacobheun: Yes I agree that could get confusing, I will update the references to *factory*.

examples/custom-libp2p/index.js
alanshaw: Unless you think it's valuable to have these accessible outside of libp2p node creation context I'd probably have these passed separately to the factory function rather than exposing a new public API. I'd rather not have to support people randomly mutating the peer info/book.

jacobheun: I'll update it to pass those into the generator.

jacobheun: I've updated the generator to take a single object argument which includes the peerInfo, peerBook, ipfs node options and ipfs node config. The example has also been updated and I fixed the conflicts with master. Should be ready for another look over.

README.md
alanshaw: I think this happened but then got changed back?

jacobheun: wow, I spaced on that one. Came back from vacation saw some inconsistencies between factory/generator and changed everything the wrong way. Everything should be factory now, aside from the branch name.

alanshaw:
@jacobheun I've rebased the branch, just waiting for CI

examples/custom-libp2p/index.js
```diff
@@ -0,0 +1,129 @@
...
+        pubsub: true
+      }
+    }
+  })
+}
+
+// Now that we have our custom factory, let's start up the ipfs node!
+const node = new IPFS({
+  libp2p: libp2pFactory
+})
```
diasdavid: Instead of calling it Factory, let's just reuse the wording around libp2p bundles (used in https://github.com/libp2p/js-libp2p#creating-your-own-libp2p-bundle, the website, slidedecks and other places).

Also, it would be great that instead of passing a factory, we could pass the libp2p instance itself.

jacobheun: I've updated the language to be more bundle centric, and added a link to the example in the main examples readme since it wasn't there yet.

alanshaw:
CI green - @diasdavid is it 👍by you now?

README.md
```diff
@@ -301,8 +301,11 @@ Modify the default IPFS node config. This object will be *merged* with the defau
 | Type | Default |
 |------|---------|
 | object | [`libp2p-nodejs.js`](https://github.com/ipfs/js-ipfs/blob/master/src/core/runtime/libp2p-nodejs.js) in Node.js, [`libp2p-browser.js`](https://github.com/ipfs/js-ipfs/blob/master/src/core/runtime/libp2p-browser.js) in browsers |
+| function | [`libp2p factory`](examples/custom-libp2p) |
```
diasdavid: Was leaving the word `factory` here intentional?

alanshaw:
Ah sorry I only looked at the commit with the change - @jacobheun would you mind changing the rest as well?

jacobheun:
Remainder of the factory language has been updated to bundle.

PR #1475, opened 2018-07-28
feat: preload on content fetch requests

alanshaw:
When JS IPFS requests content not stored locally it needs to ask the peers it knows about to provide it if they have it. The peers are not relays so if they don't have it, they won't just go and find it and provide it to JS IPFS.

However, if we issue a preload request prior to these requests we prompt the preload nodes to fetch the content using their DHT and they can then provide it to JS IPFS.

This fixes the story of adding content to go-ipfs and retrieving it from js-ipfs.


diasdavid:
Put it in a boat 🚣‍♂️

mitra42:
Its breaking - possibly breaking one of my demos - see #1476

alanshaw:
We’re waiting on the CORS settings for the preload nodes to be updated. The failing requests are async and should not effect anything else.

PR #1480, opened 2018-07-30
fix(dag): check dag.put options for plain object

### LINKED ISSUE #1479: dag.put - Cannot read property 'cid' of undefined
Tracking issue for bug reported here: https://github.com/ipfs/js-ipfs/commit/d299ed77c288d82ce807af0ec7a5b2605336626f#r29874058

```js
await ipfs.dag.put(obj) // Error: Cannot read property 'cid' of undefined
```

Assumes options object exists.

Tests added here: https://github.com/ipfs/interface-ipfs-core/pull/338 already

negamaxi:
fixes #1479

### DIFF
--- a/src/core/components/dag.js
+++ b/src/core/components/dag.js
@@ -5,7 +5,6 @@ const CID = require('cids')
 const pull = require('pull-stream')
 const mapAsync = require('async/map')
 const flattenDeep = require('lodash/flattenDeep')
-const isPlainObject = require('lodash/isPlainObject')
 
 module.exports = function dag (self) {
   return {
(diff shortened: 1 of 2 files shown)


src/core/components/dag.js
```diff
@@ -5,16 +5,22 @@ const CID = require('cids')
...
   return {
     put: promisify((dagNode, options, callback) => {
       if (typeof options === 'function') {
         callback = options
-      } else if (options.cid && (options.format || options.hashAlg)) {
-        return callback(new Error('Can\'t put dag node. Please provide either `cid` OR `format` and `hashAlg` options.'))
-      } else if ((options.format && !options.hashAlg) || (!options.format && options.hashAlg)) {
-        return callback(new Error('Can\'t put dag node. Please provide `format` AND `hashAlg` options.'))
+        options = {}
+      } else if (isPlainObject(options)) {
```
PascalPrecht: How about instead of introduce `isPlainObject()` we'd just introduce a check on whether `options` is defined:

```
if (options && options.cid && (options.format || options.hasAlg)) {
  ...
} else if (/* same thing here */) 
```

And then we'd remove the `else` block and always do sth. like `options = options || {}`

PascalPrecht:
This LGTM but there seems to be a hiccup with CI.. maybe someone with privilege can restart the job to see if that fixes it?

alanshaw:
Thanks @negamaxi ✨ . I have rebased this. Fixes to failing and temperamental tests have been merged to master so lets see what CI says now :D

PR #1482, opened 2018-08-01
fix: fix content-type detection by doing a fall-back based on the ext…

satazor:
The JS IPFS gateway was only responding with the correct content-type for some mime-types, see https://github.com/sindresorhus/file-type#supported-file-types.
This commit now fall-backs to detecting based on the extension as well.

Note that SVGs aren't supported by the `file-type` module.


src/http/gateway/resources/gateway.js
```diff
@@ -12,6 +12,21 @@ const Stream = require('readable-stream')
 const { resolver } = require('ipfs-http-response')
 const PathUtils = require('../utils/path')
 
+function detectContenType (ref, chunk) {
```
alanshaw: Minor typo here `detectContenType` - missing "t" on "conten"

src/http/gateway/resources/gateway.js
```diff
@@ -12,6 +12,21 @@ const Stream = require('readable-stream')
...
+  let fileSignature
+
+  // try to guess the filetype based on the first bytes
+  // note that `file-type` doesn't support svgs, thereore we assume it's a svg if ref looks like it
+  if (!ref.endsWith('.svg')) {
```
alanshaw: Is this check necessary?

satazor: Yes, file-type doesn’t support svgs and might actually return invalid mine types. They state that in their readme.

src/http/gateway/resources/gateway.js
satazor: Will fix once the workshop I’m attending on dweb ends. :p

src/http/gateway/resources/gateway.js
alanshaw: Ah ok, would you mind adding a comment to say that please?

alanshaw: Just saw there’s already a comment there sorry!

satazor:
The typo should be fixed!

PR #1485, opened 2018-08-04
feat: Allow pregenerated ids

mkg20001:
Needed for https://github.com/ipfs/js-ipfsd-ctl/pull/284

### DIFF
--- a/src/core/components/init.js
+++ b/src/core/components/init.js
@@ -61,14 +61,13 @@ module.exports = function init (self) {
           return cb(new Error('repo already exists'))
         }
 
-        if (opts.pregenId) {
-          // Use pregenerated Id
-          self.log('using pregenerated id')
-          cb(null, opts.pregenId)
-        } else if (process.env.IPFS_PREGENERATED_PRIVATE_KEY) {
-          // Use pregenerated Id from env
-          self.log('using pregenerated id from env')
-          peerId.createFromPrivKey(Buffer.from(process.env.IPFS_PREGENERATED_PRIVATE_KEY, 'base64'), cb)
+        if (opts.privateKey) {
+          self.log('using user-supplied private-key')
+          if (typeof opts.privateKey === 'object') {
+            cb(null, opts.privateKey)
+          } else {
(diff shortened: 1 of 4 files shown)


src/core/components/init.js
```diff
@@ -61,10 +61,20 @@ module.exports = function init (self) {
...
+        } else if (process.env.IPFS_PREGENERATED_PRIVATE_KEY) {
+          // Use pregenerated Id from env
+          self.log('using pregenerated id from env')
+          peerId.createFromPrivKey(Buffer.from(process.env.IPFS_PREGENERATED_PRIVATE_KEY, 'base64'), cb)
+        } else {
+          // Generate peer identity keypair + transform to desired format + add to config.
+          opts.log(`generating ${opts.bits}-bit RSA keypair...`, false)
+          self.log('generating peer id: %s bits', opts.bits)
+          peerId.create({ bits: opts.bits }, cb)
+        }
```
diasdavid: Instead of calling it "pregen ID", just say that it takes a Private Key.

diasdavid: And make it an option and not a env variable.

alanshaw: +1 for an option. i.e. a top level `privateKey` option to the constructor that you can also pass as a CLI option `--private-key=...`.

**Also needs to be documented in the README.**

mkg20001:
@alanshaw Added test

mkg20001:
Also the `--private-key` option would be needed in go-ipfs, too.

Mr0grog:
I know I’m way late to this, but reading it, I worry that it won’t be clear that the private key is being used to generate a *peer ID* here (as opposed to, say, encrypting the repo contents on disk or something).

At the very least, I think it would be really helpful to state that this is the key for generating a peer ID in the README & CLI help. I’m kind of wondering if a different name for the option might be better, though, like `peerIdKey` or something  :\

alanshaw:
@mkg20001 I have rebased and tweaked the README a little - hope that's ok :D

Lets get CI green and then get this merged!

alanshaw:
@Mr0grog IMHO just `private-key` is ok here since it is the private key for your IPFS node.

PR #1487, opened 2018-08-06
[WIP] perf: faster cli tests

VictorBjelkholm:
Some hacking on getting the CLI tests to be faster. (please disregard the messy git history, will be cleaned up)

Still bunch of tests that are commented out to not run yet.

Preliminary results look nice so far, this branch can run the entire CLI suite in 30 seconds compared to almost six minutes which it takes for me on the master branch. Granted, some tests are still not running, but just the fact that each test is now takes ms to run instead of one second, makes me pretty confident it will run a lot faster, even with all tests.

Every command has been been changed to signal when it's done.

Most of the "real" changes are in `ipfs-exec`, where instead of shelling out, we now `require` and manually execute the command. This leads to the good performance gains, as nodejs doesn't need to parse entire js-ipfs every time we run a command.


VictorBjelkholm:
Current situation
```
yarn test:node:core     102.67s user  5.20s system  42%   cpu 4:14.16 total
yarn test:node:http     295.62s user  21.05s system 69%   cpu 7:38.15 total
yarn test:node:gateway  4.44s user    0.19s system  107%  cpu 4.290 total
yarn test:node:cli      28.14s user   1.67s system  91%   cpu 32.450 total

```

VictorBjelkholm:
@alanshaw this PR updates the way CLI tests are being run. Many changes as I've added a callback so we know when commands finished. 

My testing shows that the `test:node:cli` tests with this change takes around 40 seconds to run through rather than 6 minutes.

Let me know if the approach is sound and I'll rebase, and clean it up so we can get it merged.

VictorBjelkholm:
Closing as https://github.com/ipfs/js-ipfs/pull/1553 would be a better solution.

Issue #1406, opened 2018-06-20
Taking control of Error Handling, reducing js-ipfs Daemon crashes


fsdiogo:
Hey everyone, this issue has the purpose of gathering everything related to the `js-ipfs` daemon crashes and error handling.

Right now the daemon is unstable, there are a lot of uncaught exceptions and they crash the process when an error occurs down the stack.

There is a lot of work to be done around the robustness of the daemon and gateway, so this will serve as a roadmap to the state of each problem.

## Roadmap

The related issues will be tracked using the table below.

Each one of them will have a priority, the same as our labels, from `P0 - Critical` to `P4 - Very Low`.

If possible, instructions of how to manually reproduce the error/crash and then tests to cause those errors, so we can track down the problem and start working on it.

When that problem is solved we'll link the solution or PR here, and mark the `Status` column as solved.

| Issue | Description | Priority | Manual Reproduction | Test Reproduction | Fix | Status |
| :--: | -- | :--: | --| -- | -- | :--: |
| [#1326](https://github.com/ipfs/js-ipfs/issues/1326) | WebSocket connection error | `P1` |  |  |  | :x: |
| [#1325](https://github.com/ipfs/js-ipfs/issues/1325#issue-317208909) [#1156](https://github.com/ipfs/js-ipfs/issues/1156) | Already piped | `P1` |  |  | [#1458](https://github.com/ipfs/js-ipfs/issues/1458) | ✅ |
| [#1331](https://github.com/ipfs/js-ipfs/issues/1331) | Invalid block | `P2` |  |  | https://github.com/ipfs/js-ipfs/pull/1378 | ✅ |

## Improving debuggability

We should strive to make debugging as smooth as possible.

Using [debug](https://github.com/visionmedia/debug) to log meaningful events, namespacing the module you are on, is super helpful. It's easier to track logs and check the specific module that is writing them.

<img width="898" alt="debug" src="https://user-images.githubusercontent.com/33324750/41724338-3b3c6efa-7565-11e8-80cd-40acd5a19161.png">

Usage is pretty straightforward too:
```js
const debug = require('debug')
const log = debug('libp2p:switch')

log('adding WebsocketStar')
```

For more info, check the [usage](https://github.com/visionmedia/debug#usage) section.

## Improving error handling

We should use [error codes](https://nodejs.org/api/errors.html#errors_node_js_error_codes) like Node does.

[...]

PR #1506, opened 2018-08-12
fix: improper input validation

alanshaw:
This PR adds a variety of fixes and tests for invalid inputs passed to the IPFS node that would crash the daemon.

The module `err-code` is used to add code's to error messages that the client can rely on and we can use to assert on in our tests.

refs https://github.com/ipfs/js-ipfs/issues/1406
also sent as part of this work https://github.com/libp2p/js-libp2p/pull/232
related to https://github.com/ipfs/interface-ipfs-core/issues/300


src/core/components/object.js
```diff
@@ -210,14 +217,14 @@ module.exports = function object (self) {
...
 
-      let mh
+      let mh, cid
 
       try {
         mh = normalizeMultihash(multihash, options.enc)
+        cid = new CID(mh)
       } catch (err) {
-        return callback(err)
+        return setImmediate(() => callback(errCode(err, 'ERR_INVALID_CID')))
```
jacobheun: If `normalizeMultihash` throws the error users are going to get confused since this will always say the cid is the problem. We should check if it's an unsupported hash error first.

alanshaw:
@jacobheun can I get a 👍 now? (will also wait for a jenkins 💚 too)

alanshaw:
> The module err-code is used to add code's to error messages that the client can rely on and we can use to assert on in our tests.

@diasdavid wanted to double check this was ok by you?

diasdavid:
@alanshaw what's the pros/cons of err-code that you see?

alanshaw:
Pros:

* Error messages aren't part of the public API
    * Allows us to fix typos, reword or just change error messages without breaking user code
    * Gives users something more stable to allow them to detect and deal with specific errors
* Augmented Error object with a `code` property instead of Error sub-classes:
    * Doesn't require user code to import Error classes (just easier)
    * Can be easily serialized/deserialized for transport over HTTP for example
* Enhances test quality
    * People know they shouldn't rely on error messages in tests but they don't know what they should do instead. We end up with tests of negative utility that just assert an error exists and no more. This can easily lead to false positives
* Reduces the boilerplate to assigning a `code` property to an error instance
* It's fairly standard practice to assign a code to an error and is even seen in Node.js also so it should be familiar to users and IPFS/libp2p developers

Note that when I mention "user" above I'm referring to developers using js-ipfs.

Cons:

* We have to remember to add an error code to error messages we create
* It adds 22 lines of code to our bundle size :P

---

Additionally, I'm aware of the following difficulties in introducing this:

* Not all errors will have an error code

    ...and they never will. We can't make dependencies add error codes and we can't add error codes to developer errors. With buy in from the JS teams we can perhaps add them to many of the projects we have control of. We'll not be able to add them in one shot, but if we can slowly add them then we'll gradually improve error handling in the code base and user land

* We have no agreement with the go-ipfs team on what the codes should be

    A conversation to be had. I'm not sure it'll be possible/appropriate to talk about this on a per-error basis - we might have to find consensus on the code when we find errors created for similar reasons.

    It's ok to change the code...in either code base! It's the same as if we changed the message right now, but doing so _should_ automatically cause the developer to consider it a change that might break user code and so we'll be able to make better choices about the version numbers we pick for our releases.

    Right now I bet that if we changed an error message we wouldn't think twice about it breaking user code regardless of the reality!

* The HTTP API and the CLI does not communicate them to users

    I would like to send a PR at some point to make this happen.

PR #1512, opened 2018-08-16
fix: stub out call to fetch for ipfs.dns test in browser

alanshaw:
Stubs `self.fetch` to return a static CID for calls to `https://ipfs.io/api/v0/dns?arg=ipfs.io`.

Removes dependency on external service.

### DIFF
--- a/test/utils/dns-fetch-stub.js
+++ b/test/utils/dns-fetch-stub.js
@@ -1,9 +1,8 @@
-/* eslint-env browser */
 'use strict'
 
-const _fetch = typeof self === 'undefined' ? null : self.fetch
-
-module.exports = () => {
+// Create a fetch stub with a fall through to the provided fetch implementation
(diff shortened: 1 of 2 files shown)


test/utils/dns-fetch-stub.js
```diff
@@ -0,0 +1,17 @@
...
+module.exports = () => {
+  return function () {
+    if (arguments[0].startsWith('https://ipfs.io/api/v0/dns?arg=ipfs.io')) {
+      return Promise.resolve({
+        json: () => Promise.resolve({
+          Path: '/ipfs/QmYNQJoKGNHTpPxCBPh9KkDpaExgd2duMa3aF6ytMpHdao'
+        })
+      })
+    }
+    return _fetch.apply(this, arguments)
```
achingbrain: Can `_fetch` be null here?

alanshaw: yes, better now?

Issue #1510, opened 2018-08-15
jsDelivr complains about package size exceeded the configured limit of 50 MB


kevinsimper:
`Package size exceeded the configured limit of 50 MB`

I think it is important as links in the readme are to jsDelivr

https://www.jsdelivr.com/package/npm/ipfs

![image](https://user-images.githubusercontent.com/1126497/44174389-4d516380-a0e3-11e8-8691-112ec268adc6.png)

mikeal:
Is there a way to configure this to only care about/load the dist directory?

MartinKolarik:
Hi, Martin from jsDelivr here. Technically we are able to serve packages even if they exceed this limit and I could add an exception for `ipfs`.

The reason for having a limit by default is that npm only provides tarballs with all files so we always have to fetch everything, even if nothing else than `dist/` is needed. The bigger the package, the higher the chance this takes too long and some requests occasionally fail.

We also find that very few packages actually need to exceed this limit. In your case, 63 MB out of 80 MB are just the examples. In my opinion, those do not belong to the npm package at all and should be added to `.npmignore`.

alanshaw:
fixed in https://github.com/ipfs/js-ipfs/pull/1513 and released in 0.31.7

kevinsimper:
@alanshaw Awesome!

PR #1513, opened 2018-08-17
fix: npm publishes examples folder

alanshaw:
refs https://github.com/ipfs/js-ipfs/issues/1510


kevinsimper:
Why not all examples? :)

PR #1521, opened 2018-08-21
docs: better start and stop examples

alanshaw:
This PR adds separate examples for each of the different methods of starting/stopping a node (promises, callbacks, events) instead of a single confusing example with all the methods in one code snippet.

This makes for easier copy/paste.

Also switches promise method to use async/await syntax.


alanshaw:
Awesome, thanks @Mr0grog :D

alanshaw:
Merging as this PR is approved and only changes README.md - test failures are for issues fixed in master already.

PR #1524, opened 2018-08-22
docs: add pin API doc links

alanshaw:
Adds the docs that are in the js-ipfs-api README for pin.

Also removes code blocks from the non-code headings...

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,7 +428,7 @@ The IPFS core API provides all functionality that is not specific to setting up 
 
 The core API is grouped into several areas:
 
-#### `Files`
+#### Files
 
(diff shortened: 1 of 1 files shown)


Mr0grog:
Are test failures some other dependency mismatch that’s currently happening?

alanshaw:
> Are test failures some other dependency mismatch that’s currently happening?

Yes, bug fix in a dependency has caused some tests to fail here - hold tight, will be resolved asap.

Thanks for taking a look ❤️

alanshaw:
Merging as this PR is approved and only changes README.md - test failures are for issues fixed in master already.

PR #1523, opened 2018-08-21
chore: update ipfs-unixfs-engine dependency

alanshaw:
fixes #1518

@lidel do you have time to verify this fix?

Issue #1518, opened 2018-08-19
ipfs.files.add with {cidVersion: 1} returns CID  of an empty file


lidel:
> This was originally found as a part of  #1440 and  https://github.com/ipfs/js-ipfs-http-response/pull/9

- **Version**: js-ipfs 0.31.6 
- **Platform**: Node
- **Subsystem**: 

#### Description: 

For some inputs `ipfs.files.add` with `{cidVersion: 1}` returns a [CIDv1 of an empty file](http://cid-utils.ipfs.team/#zdj7WiLc855B1KPRgV7Fh8ivjuAhePE1tuJafmxH5HmmSjqaD), instead of CID for the actual input buffer.

#### Steps to reproduce the error:

I attach test case that illustrates the issue: [js-ipfs-add-cidv1-bug-repro.zip](https://github.com/ipfs/js-ipfs/files/2300603/js-ipfs-add-cidv1-bug-repro.zip)

To run it: `yarn && yarn test`  (or `npm install &&  npm test`)

PR #1525, opened 2018-08-22
feat: use package-table vs custom script

diasdavid:


diasdavid:
CI is failing for unrelated changes to this PR. @alanshaw was there a PR merged that wasn't green?

alanshaw:
> was there a PR merged that wasn't green?

Not on my watch!

https://github.com/ipfs/js-ipfs/issues/1518 (TLDR; `ipfs add --cid-version=1` causes js-ipfs to give back a CID that addresses an _empty_ DAG node) was fixed in `ipfs-unixfs-engine`, released, and caused some tests to break here.

Turns out the tests were expecting the wrong CID so I created this PR to fix them https://github.com/ipfs/js-ipfs/pull/1523

In doing so it highlighted a different problem with interop - a small file created with `--cid-version=1 --raw-leaves=true` is getting a different CID in js-ipfs to go-ipfs.

I am in the process of creating a new PR to interop to ensure it doesn't happen again, and also another PR to `ipfs-unixfs-engine` to address the issue.

I appreciate this is an undesirable state to have master in and I'm working as fast as I can to restore it to green.

diasdavid:
Understood and thank you for being on top of it, @alanshaw! ❤️

PR #1537, opened 2018-09-02
feat(gateway): X-Ipfs-Path, Etag, Cache-Control, Suborigin

lidel:
This PR makes js-ipfs Gateway return the same headers as HTTP Gateway exposed by go-ipfs, namely:

- `X-Ipfs-Path`: requested IPFS Path of returned resource
- `Etag`: multihash of returned payload
- `Cache-Control`: disable cache for directory listings and errors,
  enable heavy caching for immutable assets from /ipfs/ namespace
- `Suborigin`: use root CID in base32 and literal prefix to conform
  to current suborigin spec (https://github.com/ipfs/in-web-browsers/issues/66)

cc https://github.com/ipfs/js-ipfs-http-response/issues/10

### DIFF
--- a/src/http/gateway/resources/gateway.js
+++ b/src/http/gateway/resources/gateway.js
@@ -8,6 +8,7 @@ const toPull = require('stream-to-pull-stream')
 const fileType = require('file-type')
 const mime = require('mime-types')
 const Stream = require('readable-stream')
+const CID = require('cids')
 
 const { resolver } = require('ipfs-http-response')
 const PathUtils = require('../utils/path')
@@ -41,6 +42,7 @@ module.exports = {
       ref: `/ipfs/${request.params.cid}`
     })
   },
+
   handler: (request, reply) => {
     const ref = request.pre.args.ref
     const ipfs = request.server.app.ipfs
@@ -120,6 +122,15 @@ module.exports = {
 
         let response = reply(stream2).hold()
 
+        // Etag maps directly to an identifier for a specific version of a resource
+        // TODO: change to .cid.toBaseEncodedString() after switch to new js-ipfs-http-response
+        response.header('Etag', `"${data.multihash}"`)
+
+        // Set headers specific to the immutable namespace
+        if (ref.startsWith('/ipfs/')) {
+          response.header('Cache-Control', 'public, max-age=29030400, immutable')
+        }
+
         pull(
           toPull.source(stream),
           pull.through((chunk) => {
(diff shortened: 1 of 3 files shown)


src/http/gateway/resources/gateway.js
```diff
@@ -120,6 +121,15 @@ module.exports = {
 
         let response = reply(stream2).hold()
 
+        // Etag maps directly to an identifier for a specific version of a resource
+        // TODO: change to .cid.toBaseEncodedString() after switch to new js-ipfs-http-response
+        response.header('Etag', `"${data.multihash}"`)
```
alanshaw: is `data.multihash` already a string?

src/http/gateway/resources/gateway.js
```diff
@@ -148,5 +158,22 @@ module.exports = {
...
+  },
+
+  afterHandler: (request, reply) => {
+    const response = request.response
+    if (response.statusCode === 200) {
+      const ref = request.pre.args.ref
+      response.header('X-Ipfs-Path', ref)
+      if (ref.startsWith('/ipfs/')) {
+        const rootCid = ref.split('/')[2]
+        const CID = request.server.app.ipfs.types.CID
```
alanshaw: Can we just `require` this instead?

lidel: Done.

src/http/gateway/resources/gateway.js
lidel: Yes, it is a string already. `Etag` spec [requires wrapping it with `""`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#Directives)

Note: the source of this value may change after https://github.com/ipfs/js-ipfs-http-response/pull/9 lands  – it will add CID object as `data.cid`. It just needs to be a unique string, so we can use cid or a raw multihash – both are fine for this.

lidel:
@alanshaw if it is okay to add http gateway tests to the interop repo, I will PR tests there. 
(I was under impression it is ipfs-specific on-the-wire test suite, does it include the http-wire?)

Should I PR before or after this one lands?

lidel:
Sound good. I've added interop for this to my TODO list. 

(We will need HTTP API interop tests for https://github.com/ipfs/js-ipfs/pull/1540, so  this one will not be the only case of looking at HTTP headers)

lidel:
@alanshaw  I noticed there is a lot of code duplication between companion, service worker gateway and gateway exposed by js-ipfs.  

Perhaps instead of this PR, we should go in the direction described in https://github.com/ipfs/js-ipfs-http-response/issues/6? 

Would  be great to read your thoughts  there :raised_hands:

PR #1540, opened 2018-09-03
[WIP] feat: support chunked add requests

hugomrdias:
This PR supports this https://github.com/ipfs/js-ipfs-api/pull/851 in the ipfs-api

todo:

- [x] use file per chunk, resume should simpler with this
- [ ] ~~try not using files and add directly~~
- [ ] maxBytes value ???

next(in a new PR): 
- resumable add
- concurrent chunk request
- delete zombie tmp files

These need more work so we don't add corrupted data.
- check duplicates
- check missing chunks
- check chunk size matches content-range
- in the last request right before adding check that full size matches the number of chunk in the folder


src/http/api/routes/files.js
```diff
@@ -2,7 +2,66 @@
...
+const fs = require('fs')
+const path = require('path')
+const tempy = require('tempy')
+const multipart = require('ipfs-multipart')
+const pull = require('pull-stream')
+const toPull = require('stream-to-pull-stream')
+const pushable = require('pull-pushable')
 
+const streams = []
+const filesDir = tempy.directory()
```
lidel: Do we have a plan for what happens with temp files produced by failed/aborted  uploads?
Leaving them forever may be a surface for self-administered DoS attacks :)

lidel: We should remove temporary files after successful upload, but also have a plan for what happens with temp files produced by failed/aborted  ones. I don't think we can assume temp dir will be cleared by the OS. 
Leaving them forever may be a self-administered DoS attack (eventually running out of disk space)

hugomrdias: yes definitely should go like this
- success adding ? delete file
- errors during session ? define a way to remove only the last chunk from the file ( or have one file per chunk ) this would allow for resumable add's
- on new session start async check the temp folder and delete all files older than 1 hour
- also we need to handle open streams, too many can become a problem

src/http/api/routes/files.js
```diff
@@ -37,13 +136,106 @@ module.exports = (server) => {
...
   })
 
+  api.route({
+    method: 'POST',
+    path: '/api/v0/add-chunked',
+    config: {
+      payload: {
+        parse: false,
+        output: 'stream',
+        maxBytes: 1000 * 1024 * 1024
```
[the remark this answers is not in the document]
hugomrdias: >How do we pick what is the default max chunk size? Eg. why 1GB and not 256MB ?

we need this value to be high for the non chunked path

>Is this going to be a hardcoded limit, or a configuration option that can be passed to js-ipfs constructor?

don't know. what you think ?

from the hapi documentation when output=stream, maxBytes doesn't matter **but** this doesnt seem to be the case. Needs further investigation, maybe it's a matter of upgrading hapi dunno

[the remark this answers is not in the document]
lidel: I can't think of anything right now.
Unless you see good use case for customizing it, lets go with hardcoded value for now.
Exposing it as a config property can be contributed later in separate PR.

hugomrdias:
Yes I'm gonna do that, just getting this ready for you and others. I have a couple more commits to push with tests and a big fix for a bug I found today.

src/http/api/utils/add-experimental.js
```diff
@@ -0,0 +1,146 @@
...
+  const options = {
+    cidVersion: query['cid-version'],
+    rawLeaves: query['raw-leaves'],
+    progress: query.progress ? progressHandler : null,
+    onlyHash: query['only-hash'],
+    hashAlg: query.hash,
+    wrapWithDirectory: query['wrap-with-directory'],
+    pin: query.pin,
+    chunker: query.chunker
+  }
```
alanshaw: We should extract this logic into a function e.g. `queryToAddOptions` in `http/api/resouces/files.js` and use it in `add` and `addExperimental` and just pass the result of calling it to this function. It'll save us having to keep track of another place where these options are constructed.

src/http/api/utils/add-experimental.js
```diff
@@ -0,0 +1,146 @@
...
+
+  if (!input) {
+    return null
+  }
+  const match = input.match(regex)
+
+  return [match[1], Number(match[2])]
+}
+
+const createMultipartReply = (readStream, request, reply, cb) => {
```
alanshaw: IMHO this should just construct and return a stream that can be passed to reply in the handler.

alanshaw:
@hugomrdias any plans to continue this PR or shall we close?

hugomrdias:
let it be open i would like to come back to this

jacobheun:
Closing as this is quite stale.

PR #1541, opened 2018-09-05
chore: more test stabilization

alanshaw:
* Increases the preload test timeouts
* Reduces the number of bits for the Peer ID generated in bitswap CLI tests
* Adds `waitFor` utility to wait for an async function to return true before continuing
* Ensures temp repos are cleaned up after tests
* Increases default test timeout from 5s to 10s


package.json
```diff
@@ -25,15 +25,15 @@
...
-    "test:bootstrapers": "IPFS_TEST=bootstrapers aegir test -t browser -f test/bootstrapers.js",
+    "test": "aegir test -t node -t browser -t webworker --no-cors --timeout=10000",
+    "test:node": "aegir test -t node --timeout=10000",
+    "test:browser": "aegir test -t browser --no-cors --timeout=10000",
+    "test:webworker": "aegir test -t webworker --no-cors --timeout=10000",
+    "test:node:core": "aegir test -t node -f test/core/**/*.js --timeout=10000",
+    "test:node:http": "aegir test -t node -f test/http-api/index.js --timeout=10000",
+    "test:node:gateway": "aegir test -t node -f test/gateway/index.js --timeout=10000",
+    "test:node:cli": "aegir test -t node -f test/cli/index.js --timeout=10000",
+    "test:bootstrapers": "IPFS_TEST=bootstrapers aegir test -t browser -f test/bootstrapers.js --timeout=10000",
```
diasdavid: I would really like to avoid global timeouts and instead focus on understanding which tests are actually timing out by seeing the explicit `this.timeout`

PR #1543, opened 2018-09-05
fix: libp2p records for ipns should be signed

vasco-santos:
There are two types of IPNS records. IPNS records for being stored locally and IPNS records for being used in the network.

IPNS records for being stored locally are used for understanding which is the most recent record since it is stored by the peer who published it.

IPNS records for being used in the network, are used as a common [libp2p-record](https://github.com/libp2p/js-libp2p-record) and are intended to be used with the DHT and Pubsub.

This way, the `libp2p-record` should have a signature and that signature should be validated on the resolve, in order to guarantee that the record was not changed in the network. 

NOTE: The current implementation guarantees that the `IPNS Record` is not changed, since it has its own signature, but the `libp2p-record` used must be used as a generic `libp2p-record`, with its own signature as well.

### DIFF
--- a/src/core/ipns/publisher.js
+++ b/src/core/ipns/publisher.js
@@ -40,7 +40,7 @@ class IpnsPublisher {
           return callback(err)
         }
 
-        this._putRecordToRouting(record, peerId, callback)
+        this._putRecordToRouting(record, peerId, privKey, callback)
       })
     })
   }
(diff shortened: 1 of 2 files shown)


alanshaw:
Can you please explain why this is necessary? Seems as though `serializeSigned` does the same as serialize, but has the side effect of setting a `signature` property on the record. The record is then discarded, so all that appears to be happening is just verifying that is _can_ be signed without error.

vasco-santos:
@alanshaw edited the initial description of the PR. Does it makes sense to you now?

PR #1548, opened 2018-09-07
fix: allow keychain interactions without a password

achingbrain:
go-ipfs allows you to list keys, create keys, etc without using a keychain password.  This PR brings js-ipfs in line with that though will need a few other PRs merging first, if we think this is a good idea.


test/cli/key.js
```diff
@@ -19,6 +19,7 @@ describe('key', () => runOnAndOff.off((thing) => {
     this.timeout(40 * 1000)
 
     return ipfs(`${pass} key gen ${name} --type rsa --size 2048`)
+      .then(() => ipfs(`${pass} key list -l`))
```
VictorBjelkholm: This change changes the assertion as now it's testing on the output of `key list` rather that the name is included in the output when running `key gen`

achingbrain: `js-ipfs` used to do: 

```console
$ jsipfs key gen foobar
generated Qmb1Wj3YR4PGgscvXJ3JgaiX6GCQFXLwP6Qo7c5kNR6PQK foobar
```

but to bring it inline with `go-ipfs` it now outputs:

```console
$ jsipfs key gen foobar
Qmb1Wj3YR4PGgscvXJ3JgaiX6GCQFXLwP6Qo7c5kNR6PQK
```

so the name is no longer in the output and the test would fail.  In order to verify that we have generated a key with the passed name, I changed the test to examine the output of `jsipfs key list` after generating the key which includes the name.

test/cli/key.js
```diff
@@ -5,7 +5,7 @@ const expect = require('chai').expect
 const runOnAndOff = require('../utils/on-and-off')
 const hat = require('hat')
 
-describe('key', () => runOnAndOff.off((thing) => {
+describe.only('key', () => runOnAndOff.off((thing) => {
```
alanshaw: Reminder to remove this before merge ;)

Stebalien:
So, go-ipfs has never supported encrypted keys but I believe it has been on the "todo" list for a while (I can see some remnants in the code). However, IMO, it makes sense to support keys without a password (e.g., for system daemons).

Note: the password should not be passed as a command-line flag. Ideally, on repo init, IPFS would ask for a password (or allow a `--no-password` flag, or something like that). Then, on start, if the keys are encrypted, IPFS would ask for the decryption password (and fail if STDIN is closed).

Passing a password on the command-line pretty much defeats the point:

1. It'll likely get logged to disk somewhere.
2. It'll show up in `ps`.

daviddias:

achingbrain:
This was fixed!

PR #1551, opened 2018-09-10
fix: mfs preload test

alanshaw:
JS makes no guarantees a function you schedule with `setTimeout` will execute before another function you schedule _even_ if the second function is scheduled for execution _after_ the first.

This PR fixes the MFS preload test to wait for the expected CIDs to have been preloaded instead of assuming they would be preloaded after a certain time.

### DIFF
--- a/test/core/mfs-preload.spec.js
+++ b/test/core/mfs-preload.spec.js
@@ -7,6 +7,7 @@ const dirtyChai = require('dirty-chai')
 const expect = chai.expect
 chai.use(dirtyChai)
 
+const waitFor = require('../utils/wait-for')
 const mfsPreload = require('../../src/core/mfs-preload')
 
 const createMockFilesStat = (cids = []) => {
@@ -15,11 +16,12 @@ const createMockFilesStat = (cids = []) => {
 }
 
 const createMockPreload = () => {
-  return function preload (cid, cb) {
-    preload.cids = preload.cids || []
+  const preload = (cid, cb) => {
     preload.cids.push(cid)
     cb()
   }
+  preload.cids = []
+  return preload
 }
 
 describe('MFS preload', () => {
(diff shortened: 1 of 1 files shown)


test/core/mfs-preload.spec.js
```diff
@@ -41,16 +43,17 @@ describe('MFS preload', () => {
...
-      }, statCids.length * (interval * 2))
+      const test = (cb) => {
+        // Slice off any extra CIDs it processed
+        const cids = mockPreload.cids.slice(0, expectedPreloadCids.length)
+        if (cids.length !== expectedPreloadCids.length) return cb(null, false)
+        cb(null, cids.every((cid, i) => cid === expectedPreloadCids[i]))
+      }
+
+      waitFor(test, { name: 'CIDs to be preloaded' }, (err) => {
+        expect(err).to.not.exist()
```
VictorBjelkholm: rather than `expect(err).to.not.exist()` we should be using `if (err) return done(err)` so we properly finish off the test when it errors

VictorBjelkholm: rather than `expect(err).to.not.exist()` we should be using `if (err) return done(err)` so we properly finish off the test when it errors

VictorBjelkholm: rather than `expect(err).to.not.exist()` we should be using `if (err) return done(err)` so we properly finish off the test when it errors

alanshaw: Do you mean so that we always call `preloader.stop`?

I'm not sure I understand why using `done(err)` is necessary in this case and not elsewhere in the tests?

alanshaw: ping @VictorBjelkholm

alanshaw:
Ok I'm going to merge this as for now so that we can get other PRs passing. We can fix/update later if needs be. All green except one CI failure, which seems to be a CI issue not an issue with this code.

PR #1556, opened 2018-09-11
fix: ipns publish resolve option overwritten

vasco-santos:
`yargs-promise` [adds](https://github.com/eddywashere/yargs-promise/blob/master/index.js#L14) resolve/reject properties to `argv`. 

This way, `resolve` option was being overwritten to a function and, as it was not `false`, it was defined always as true for the `publish` operation.

Publish should use the `r` alias instead, as resolve will always be overwritten to a function.

### DIFF
--- a/src/cli/commands/name/publish.js
+++ b/src/cli/commands/name/publish.js
@@ -9,6 +9,7 @@ module.exports = {
 
   builder: {
     resolve: {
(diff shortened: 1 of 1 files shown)


alanshaw:
Merging as errors are problems with jenkins and not related to this PR

Issue #1557, opened 2018-09-11
custom-ipfs-repo example broken


alanshaw:
With dependencies upgraded to:

"datastore-fs": "~0.5.0",
"ipfs-repo": "~0.23.1"

```console
Error: unexpected error getting the ipns record  Q,ALPl|���;GVM�t�Н����ʨ from datastore
    at ReadFileContext._repo.datastore.get (/Users/alan/Code/protocol-labs/js-ipfs/src/core/ipns/publisher.js:198:35)
    at ReadFileContext.callback (/Users/alan/Code/protocol-labs/js-ipfs/examples/custom-ipfs-repo/node_modules/graceful-fs/graceful-fs.js:78:16)
    at FSReqWrap.readFileAfterOpen [as oncomplete] (fs.js:235:13)
Emitted 'error' event at:
    at done (/Users/alan/Code/protocol-labs/js-ipfs/src/core/components/init.js:21:14)
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/async/internal/once.js:12:16
    at next (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/async/waterfall.js:21:29)
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/async/internal/onlyOnce.js:12:16
    at parallel (/Users/alan/Code/protocol-labs/js-ipfs/src/core/components/init.js:130:13)
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/async/internal/parallel.js:39:9
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/async/internal/once.js:12:16
    at iteratorCallback (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/async/eachOf.js:58:13)
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/async/internal/onlyOnce.js:12:16
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/async/internal/parallel.js:36:13
```

jacobheun:
So the end error that's showing up here is a bit misleading as it's a could of different errors all kind of wrapping each other. Fun stuff!

[...]

PR #1558, opened 2018-09-12
fix: ipns datastore get not found

vasco-santos:
In the context of the bug described on #1557 

Different datastores are providing different errors for `NOT FOUND`.

Needs:

- [x] [interface-datastore#22](https://github.com/ipfs/interface-datastore/pull/22)
- [x] [js-ipfs-repo#176](https://github.com/ipfs/js-ipfs-repo/pull/176)


jacobheun:
@vasco-santos interface-datastore@0.5.0 has been released. Most of the datastores have also been released, this should be good to finish up. js-ipfs-mfs should hopefully be released tomorrow.

vasco-santos:
I will update it now! Thanks @jacobheun

PR #1559, opened 2018-09-13
feat: ipns over pubsub

vasco-santos:
This PR addresses the `IPNS over Pubsub` feature.

It needs the following PRs:

- [x] [js-ipfs-api#846](https://github.com/ipfs/js-ipfs-api/pull/846)
- [x] [interface-ipfs-core#361](https://github.com/ipfs/interface-ipfs-core/pull/361)
- [x] [interface-ipfs-core#395](https://github.com/ipfs/interface-ipfs-core/pull/395)
- [x] [js-datastore-pubsub#6](https://github.com/ipfs/js-datastore-pubsub/pull/6)
- [x] enable `interface-ipfs-core` tests for `ipns-pubsub`(after `interface-ipfs-core` new release)
- [x] [ipfs/js-datastore-pubsub#9](https://github.com/ipfs/js-datastore-pubsub/pull/9)

Unblocks:

- [ ] interop tests [ipfs/interop#39](https://github.com/ipfs/interop/pull/39)


vasco-santos:
@diasdavid regarding the weekly sync question:

[js-datastore-pubsub](https://github.com/ipfs/js-datastore-pubsub) is used as middleware for pubsub to be used as a datastore with TieredDatastore. This way, we can combine multiple `put` and `get` in the `IPNS` land transparently with a "single" operation.

We provide to the `IPNS` the [routing](https://github.com/ipfs/js-ipfs/pull/1559/files#diff-8d1a7002931f907b006489a19a204fd8R60), as a `TieredDatastore` composed by the `OfflineDatastore` and the `PubsubDatastore`.  (NOTE: It now receives the `OfflineDatastore`, but this will be replaced with the `DHT`, once it gets ready). After that, when an `IPNS` operation happens, such as a publish, the `IPNS` just [puts](https://github.com/ipfs/js-ipfs/pull/1559/files#diff-79e126ac6c795e54feb81c978d905333R118) it to the routing.

All in all, `pubsub-datastore` [receives](https://github.com/ipfs/js-datastore-pubsub#setup) a datastore (repo's datastore in this case), as well as a pubsub implementation. It provides the same API as a general store (`put`, `get`, ...) and translates those to what they really represent in the `pubsub` world (`publish` and `subscribe`). After that, it also takes care of the `subscriptions handling` and their consequent store in the node's repo datastore.

Moreover, `go-team` uses a similar approach with [go-libp2p-pubsub-router](https://github.com/libp2p/go-libp2p-pubsub-router), but their module is specific for `IPNS`, while `pubsub-datastore` is agnostic to where it is used. In addition, they also use the same logic for the `TieredDatastore` [here](https://github.com/ipfs/go-ipfs/blob/master/core/core.go#L517).

I hope to have been clear with this explanation. Let me know if you agree, or if you have any other direction that this sould go.

src/core/ipns/publisher.js
```diff
@@ -108,13 +108,13 @@ class IpnsPublisher {
...
-    this._repo.datastore.put(key, rec.serialize(), (err, res) => {
+    // Add record to routing (buffer key)
+    this._routing.put(key.toBuffer(), rec.serialize(), (err, res) => {
       if (err) {
         const errMsg = `ipns record for ${key.toString()} could not be stored in the routing`
 
-        log.error(errMsg)
-        return callback(errcode(new Error(errMsg), 'ERR_STORING_IN_DATASTORE'))
+      log.error(errMsg)
+      return callback(errcode(new Error(errMsg), 'ERR_STORING_IN_DATASTORE'))
```
dirkmc: Nit: should this error message have a different label? eg ERR_PUTTING_TO_ROUTING?

dirkmc:
This is a really nice approach 👍

src/core/ipns/routing/pubsub.js
```diff
@@ -0,0 +1,206 @@
...
+
+    callback(null, keys.routingKey.toBuffer())
+  }
+
+  /**
+   * Get pubsub subscriptions related to ipns.
+   * @param {function(Error, Object)} callback
+   * @returns {void}
+   */
+  getSubscriptions (callback) {
```
alanshaw: Why can't we just use `Object.keys(this._subscriptions)` here?

src/core/ipns/routing/pubsub.js
```diff
@@ -0,0 +1,206 @@
...
+          }
+        }
+      })
+
+      callback(null, strings)
+    })
+  }
+
+  /**
+   * Get pubsub subscriptions related to ipns.
```
alanshaw: Get -> cancel

src/core/ipns/routing/pubsub.js
```diff
@@ -0,0 +1,206 @@
...
+
+    // Verify peerId validity
+    try {
+      PeerId.createFromBytes(Buffer.from(name))
+    } catch (err) {
+      const errMsg = `ipns key is not a valid peer ID`
+
+      log.error(errMsg)
+      return callback(errcode(new Error(errMsg), 'ERR_INVALID_IPNS_KEY'))
+    }
```
alanshaw: Why do we need to do this?

src/core/ipns/routing/pubsub.js
```diff
@@ -0,0 +1,206 @@
...
+      // Iterate over subscriptions
+      const strings = []
+      res.forEach((subs) => {
+        const ns = subs.substring(0, ipnsNSLength)
+        const subscription = this._subscriptions[subs]
+
+        if (ns === ipnsNS && subscription) {
+          // Verify valid PeerID
+          try {
+            PeerId.createFromBytes(Buffer.from(subscription))
```
alanshaw: Why are we doing this here and not when we try to `get`?

src/core/ipns/routing/pubsub.js
```diff
@@ -0,0 +1,206 @@
...
+
+const ipnsNS = '/ipns/'
+const ipnsNSLength = ipnsNS.length
+
+// Pubsub aims to manage the pubsub subscriptions for IPNS
+class Pubsub {
+  constructor (node) {
+    const pubsub = node._libp2pNode.pubsub
+    const localDatastore = node._repo.datastore
+    const peerId = node._peerInfo.id
```
alanshaw: I'd pass these directly to the constructor so that it is more explicit what dependencies this class has.

test/cli/name-pubsub.js
```diff
@@ -0,0 +1,253 @@
...
+        done()
+      })
+    })
+
+    it('should get error getting the available subscriptions', function (done) {
+      ipfsA('name pubsub subs').catch((err) => {
+        expect(err).to.exist() // error as it is disabled
+
+        done()
+      })
```
alanshaw: Assert on the error message here - it's the best we can do for now

test/cli/name-pubsub.js
```diff
@@ -0,0 +1,253 @@
...
+    })
+
+    it('should get error canceling a subscription', function (done) {
+      ipfsA('name pubsub cancel /ipns/QmSWxaPcGgf4TDnFEBDWz2JnbHywF14phmY9hNcAeBEK5v').catch((err) => {
+        expect(err).to.exist() // error as it is disabled
+
+        done()
+      })
```
alanshaw: Assert on the error message here - it's the best we can do for now

test/cli/name-pubsub.js
```diff
@@ -0,0 +1,253 @@
...
+                ipfsA(`name resolve ${nodeBId.id}`).then((res) => {
+                  expect(res).to.exist()
+                  expect(res).to.satisfy(checkAll([cidAdded])) // value propagated to node B
+
+                  done()
+                })
+              })
+            })
+          })
+        })
```
alanshaw: Please chain these properly (applies to all tests)

test/cli/name-pubsub.js
```diff
@@ -0,0 +1,253 @@
...
+    after((done) => parallel(nodes.map((node) => (cb) => node.stop(cb)), done))
+
+    it('should get disabled state of pubsub', function (done) {
+      ipfsA('name pubsub state').then((res) => {
+        expect(res).to.exist()
+        expect(res).to.have.string('disabled')
+
+        done()
+      })
+    })
```
alanshaw: Mocha supports promises. You should remove `done` and return the promise or use an async function. Mixing promises and callbacks is confusing.

src/core/components/start.js
```diff
@@ -34,6 +40,25 @@ module.exports = (self) => {
...
+          ipnsStores.push(pubsub)
+        }
+
+        // NOTE: Until the IPNS over DHT is not ready, it is being replaced by the local repo datastore
+        // When DHT is added, If local option enabled, should receive offlineDatastore as well
+        const offlineDatastore = new OfflineDatastore(self._repo)
+        ipnsStores.push(offlineDatastore)
+
+        const routing = new TieredDatastore(ipnsStores)
+        self._ipns = new IPNS(routing, self, pubsub)
```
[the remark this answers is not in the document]
alanshaw: You could perhaps use this function in `isNamePubsubEnabled` instead of having to duplicate options checking...

src/core/config.js
```diff
@@ -28,6 +28,7 @@ const schema = Joi.object().keys({
   }).allow(null),
   EXPERIMENTAL: Joi.object().keys({
     pubsub: Joi.boolean(),
+    namesysPubsub: Joi.boolean(),
```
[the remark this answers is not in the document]
vasco-santos: This change of naming is done in [js-ipfs/http/index.js#L78](https://github.com/ipfs/js-ipfs/blob/0627b06ebd038fa20b83f14e20e666603eb2ed72/src/http/index.js#L78). `go-ipfs` uses the `namesysPubsub` name, but I felt that `ipnsPubsub`was a better naming for us. 

This mapping of names also happens for other experimental flags, as you can see. Do you prefer to have the same `namesysPubsub` in everywhere? I just maintained it in the daemon perspective to have it equal to `go-ipfs`.

src/core/ipns/routing/pubsub.js
```diff
@@ -0,0 +1,206 @@
...
+        const id = toB58String(key.slice(ipnsNSLength))
+
+        this._subscriptions[stringifiedTopic] = id
+
+        log(`subscribed pubsub ${stringifiedTopic}: ${id}`)
+      }
+
+      if (err) {
+        return callback(err)
+      }
```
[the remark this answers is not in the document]
vasco-santos: When a pubsub "get" happens, we have the `subscribe` happening, no matter if we could find the record locally or not. This way, the error is handled after storing the subscription.

I will add a comment to make it more understandable.

src/core/ipns/routing/pubsub.js
vasco-santos: We have different encodings between regular `pubsub` and `ipns pubsub` topics.

Test scenario:
```sh
> ipfs name resolve /ipns/QmWc7wfpSQf1rGLijVYAWxmVh7mSXzHsF8daqShaMCHeR9
> ipfs name pubsub subs
/ipns/QmWc7wfpSQf1rGLijVYAWxmVh7mSXzHsF8daqShaMCHeR9
> ipfs pubsub ls
/ipns/ z�?��)�%�o���c���M���!��-V͜
```

That is the reason why I added the `_subscriptions`, which maps the bytes from the real pubsub subscription with the friendly identifier for ipns pubsub users.

src/core/components/name.js
```diff
@@ -157,11 +158,11 @@ module.exports = function name (self) {
```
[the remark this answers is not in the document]
vasco-santos: I tried and the `preStart` runs if a command is executed with no daemon running.

src/core/ipns/path.js
```diff
@@ -13,13 +13,13 @@ const resolvePath = (ipfsNode, name, callback) => {
   if (isIPFS.ipnsPath(name)) {
     log(`resolve ipns path ${name}`)
 
-    const local = true // TODO ROUTING - use self._options.local
+    const local = ipfsNode._options.local
```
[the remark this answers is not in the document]
vasco-santos: Oh, this is not needed anymore. The routing is handled outside 🙂

src/core/components/name.js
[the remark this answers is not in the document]
vasco-santos: I tested it and running a command with no daemon running, the `preStart` runs as well.

alanshaw:
@vasco-santos would you mind updating the PR comment and labels?

vasco-santos:
done @alanshaw

src/core/components/start.js
```diff
@@ -2,11 +2,13 @@
 
 const series = require('async/series')
 const Bitswap = require('ipfs-bitswap')
+const get = require('lodash/get')
```
hugomrdias: use smaller package 
https://bundlephobia.com/result?p=lodash.get@4.4.2
vs
https://bundlephobia.com/result?p=dlv@1.1.2
```suggestion
const get = require('dlv')
```

src/core/config.js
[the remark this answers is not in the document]
hugomrdias: just pick one and use it everywhere

src/core/ipns/routing/offline-datastore.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 const { Key } = require('interface-datastore')
+const Record = require('libp2p-record').Record
```
hugomrdias:
```suggestion
const { Record } = require('libp2p-record')
```

src/core/ipns/routing/pubsub-datastore.js
```diff
@@ -0,0 +1,171 @@
...
+
+const errcode = require('err-code')
+const debug = require('debug')
+const log = debug('jsipfs:ipns:pubsub')
+log.error = debug('jsipfs:ipns:pubsub:error')
+
+const ipnsNS = '/ipns/'
+const ipnsNSLength = ipnsNS.length
+
+// Pubsub aims to manage the pubsub subscriptions for IPNS
```
hugomrdias:
```suggestion
// Pubsub datastore aims to manage the pubsub subscriptions for IPNS
```

src/core/ipns/routing/pubsub-datastore.js
```diff
@@ -0,0 +1,171 @@
...
+
+const withIs = require('class-is')
+
+const errcode = require('err-code')
+const debug = require('debug')
+const log = debug('jsipfs:ipns:pubsub')
+log.error = debug('jsipfs:ipns:pubsub:error')
+
+const ipnsNS = '/ipns/'
+const ipnsNSLength = ipnsNS.length
```
hugomrdias: i predict these 2 lines will be used in more places maybe they can be exported from the ipns package ?

src/core/ipns/routing/pubsub-datastore.js
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * Get pubsub subscriptions related to ipns.
+   * @param {function(Error, Object)} callback
+   * @returns {void}
+   */
+  getSubscriptions (callback) {
+    const subscriptions = Object.values(this._subscriptions)
+
+    return callback(null, subscriptions.map((sub) => `/ipns/${sub}`))
```
hugomrdias: use ipnsNS const here
```suggestion
    return callback(null, subscriptions.map((sub) => `${ipnsNs}${sub}`))
```

src/core/ipns/routing/pubsub-datastore.js
```diff
@@ -0,0 +1,171 @@
...
+    // Unsubscribe topic
+    try {
+      const bufTopic = Buffer.from(stringifiedTopic)
+
+      this._pubsubDs.unsubscribe(bufTopic)
+    } catch (err) {
+      return callback(err)
+    }
+
+    delete this._subscriptions[stringifiedTopic]
```
hugomrdias: this really needs to be deleted can't be just null ? `delete` can be bad for perf

src/core/ipns/routing/pubsub-datastore.js
```diff
@@ -0,0 +1,171 @@
...
+    log(`unsubscribed pubsub ${stringifiedTopic}: ${name}`)
+
+    callback(null, {
+      canceled: true
+    })
+  }
+}
+
+// exports = module.exports = IpnsPubsubDatastore
```
hugomrdias:
```suggestion
```

src/core/components/start.js
vasco-santos: Cool, thanks! Will add `dlv` to `package.json` as well.

src/core/ipns/routing/pubsub-datastore.js
```diff
@@ -0,0 +1,167 @@
...
+      log.error(errMsg)
+      return callback(errcode(new Error(errMsg), 'ERR_INVALID_KEY'))
+    }
+
+    if (!Buffer.isBuffer(value)) {
+      const errMsg = `received value ${value} is not a buffer`
+
+      log.error(errMsg)
+      return callback(errcode(new Error(errMsg), 'ERR_INVALID_VALUE_RECEIVED'))
+    }
```
alanshaw: You can remove these validations - `datastore-pubsub` does them and this method does not use `key` or `value` directly.

src/http/index.js
```diff
@@ -75,6 +75,7 @@ function HttpApi (repo, config, cliArgs) {
             pass: cliArgs && cliArgs.pass,
             EXPERIMENTAL: {
               pubsub: cliArgs && cliArgs.enablePubsubExperiment,
+              ipnsPubsub: cliArgs && cliArgs.enableNamesysPubsub,
```
alanshaw: 🙏 would you mind defaulting `cliArgs` to `{}` so we don't have to `&&` so much?

```js
cliArgs = cliArgs || {}
```

src/core/components/start.js
alanshaw: I'm fine with using `dlv` over `loadash.get` but all other occurances of `lodash.get` need to be replaced as well otherwise we're just adding more to our bundle.

src/core/components/start.js
```diff
@@ -2,11 +2,13 @@
 
 const series = require('async/series')
 const Bitswap = require('ipfs-bitswap')
+const get = require('dlv')
```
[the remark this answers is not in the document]
alanshaw: Also see https://github.com/ipfs/js-ipfs/pull/1559#discussion_r237823774

src/core/components/start.js
vasco-santos: If you agree, I can create other PR with all those occurrences changed to `dlv`.

src/core/index.js
```diff
@@ -166,6 +166,11 @@ class IPFS extends EventEmitter {
       this.log('EXPERIMENTAL pubsub is enabled')
     }
     if (this._options.EXPERIMENTAL.ipnsPubsub) {
+      if (!this._options.EXPERIMENTAL.pubsub) {
+        this.log('EXPERIMENTAL pubsub is enabled to use IPNS pubsub')
+        this._options.EXPERIMENTAL.pubsub = true
+      }
```
daviddias: Does go-ipfs do IPNS over PubSub as an experimental flag? Why not just one flag that enables all of PubSub

vasco-santos: Yes, `go-ipfs` uses an experimental flag for IPNS over Pubsub: `--enable-namesys-pubsub`

daviddias: "Experimental features on Alpha software"

alanshaw:
@vasco-santos released `interface-ipfs-core@0.89` with https://github.com/ipfs/interface-ipfs-core/pull/395

vasco-santos:
Thanks! Updated the `package.json` with it

PR #1583, opened 2018-09-19
Clarify preload behavior

parkan:
Preload is confusing because it happens on both add and get, and because it isn't documented anywhere in depth. This clarifies what happens a bit. We should also write out the behavior in more detail like with circuit relay above (possibly in the form of a tutorial for setting up your own preload node)

see also https://github.com/ipfs/js-ipfs/issues/1459#issuecomment-422079307


alanshaw:
Thanks @parkan 😎

PR #1591, opened 2018-09-25
fix: report correct size for raw dag nodes

### LINKED ISSUE #1585: ipfs.files.add with {cidVersion: 1}  returns valid hash with undefined size
> This was originally found as a part of  #1440 and  https://github.com/ipfs/js-ipfs-http-response/pull/9

- **Version**: js-ipfs 0.32.2
- **Platform**: Node
- **Subsystem**: 

#### Description: 

For some inputs `ipfs.files.add` with `{cidVersion: 1}` returns valid CID however `size` field is `undefined`. The problem does not occur if `{cidVersion: 0}` is used. 

##### Result of ipfs.files.add (CIDv0)
```js
{ path: 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP',
    hash: 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP',
    size: 20 }
```

#####  Result of ipfs.files.add (CIDv1)
```js
{ path: 'zb2rhdTDKmCQD2a9x2TfLR61M3s7RmESzwV5mqgnakXQbm5gp',
    hash: 'zb2rhdTDKmCQD2a9x2TfLR61M3s7RmESzwV5mqgnakXQbm5gp',
    size: undefined }
```

Did not look into details, but I attach test case that illustrates the issue:  [js-ipfs-add-cidv1-missing-size-bug-repro.zip](https://github.com/ipfs/js-ipfs/files/2407377/js-ipfs-add-cidv1-missing-size-bug-repro.zip).

To run it: `yarn && yarn test`  (or `npm install &&  npm test`).
Hope this helps.

achingbrain:
If you specify `cidVersion: 0`, for go compatibility you will have raw leaf nodes. If your data fits in one dag node, the root node will be a raw node for the same reason, and `ipfs.object.get` will return a buffer.

This PR inspects the returned node to see if it's a buffer, if it is, use the `length` property, otherwise use the DAGNode `size` property.

Fixes #1585

### DIFF
--- a/test/core/files.spec.js
+++ b/test/core/files.spec.js
@@ -83,5 +83,30 @@ describe('files', () => {
         done()
       })
     })
+
+    it('should add a file with a v1 CID', (done) => {
+      ipfs.files.add(Buffer.from([0, 1, 2]), {
+        cidVersion: 1
+      }, (err, files) => {
+        expect(err).to.not.exist()
+        expect(files.length).to.equal(1)
+        expect(files[0].hash).to.equal('zb2rhiNedvrkpYhcrgtpmhKk5UPzcgizgSXaQLYXNY745BmYP')
+        expect(files[0].size).to.equal(3)
+        done()
+      })
+    })
+
+    it('should add a file with a v1 CID and not raw leaves', (done) => {
+      ipfs.files.add(Buffer.from([0, 1, 2]), {
+        cidVersion: 1,
+        rawLeaves: false
+      }, (err, files) => {
+        expect(err).to.not.exist()
+        expect(files.length).to.equal(1)
+        expect(files[0].hash).to.equal('zdj7WcDSFNSsZkdkbpSDGeLsBtHbYKyvPQsaw6PpeeYdGqoAx')
+        expect(files[0].size).to.equal(11)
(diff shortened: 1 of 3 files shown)


test/core/files.spec.js
```diff
@@ -83,5 +83,30 @@ describe('files', () => {
...
+
+    it('should add a file with a v1 CID and not raw leaves', (done) => {
+      ipfs.files.add(Buffer.from([0, 1, 2]), {
+        cidVersion: 1,
+        rawLeaves: false
+      }, (err, files) => {
+        expect(err).to.not.exist()
+        expect(files.length).to.equal(1)
+        expect(files[0].hash).to.equal('zdj7WcDSFNSsZkdkbpSDGeLsBtHbYKyvPQsaw6PpeeYdGqoAx')
+        expect(files[0].size).to.equal(11)
```
achingbrain: N.b. this is bigger than the input buffer due to the protobuf overhead.  The other test uses raw nodes, so no the data hasn't been wrapped in a protobuf.

PR #1595, opened 2018-09-27
feat: show Web UI url in daemon output

olizilla:
- make it easier for people to discover the Web UI
- tidy up inconsistencies across the log lines
- add IPFS to the first line. confirm to new users they started what they expected to.

fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/815

**Before**
```console
$ jsipfs daemon
Initializing daemon...
Swarm listening on /ip4/127.0.0.1/tcp/4003/ws/ipfs/QmPmrqh4ybdUJwGchEecuj2GBXCxKL5iapW9qmk8pwgCGs
Swarm listening on /ip4/127.0.0.1/tcp/4002/ipfs/QmPmrqh4ybdUJwGchEecuj2GBXCxKL5iapW9qmk8pwgCGs
Swarm listening on /ip4/192.168.1.244/tcp/4002/ipfs/QmPmrqh4ybdUJwGchEecuj2GBXCxKL5iapW9qmk8pwgCGs
API is listening on: /ip4/127.0.0.1/tcp/5002
Gateway (readonly) is listening on: /ip4/127.0.0.1/tcp/9090
Daemon is ready
```

**After**
```console
$ node src/cli/bin.js daemon
Initializing IPFS daemon...
Swarm listening on /ip4/127.0.0.1/tcp/4003/ws/ipfs/QmPmrqh4ybdUJwGchEecuj2GBXCxKL5iapW9qmk8pwgCGs
Swarm listening on /ip4/127.0.0.1/tcp/4002/ipfs/QmPmrqh4ybdUJwGchEecuj2GBXCxKL5iapW9qmk8pwgCGs
Swarm listening on /ip4/192.168.1.244/tcp/4002/ipfs/QmPmrqh4ybdUJwGchEecuj2GBXCxKL5iapW9qmk8pwgCGs
API listening on /ip4/127.0.0.1/tcp/5002
Gateway (read only) listening on /ip4/127.0.0.1/tcp/9090
Web UI available at http://127.0.0.1:5002/webui
Daemon is ready
```
- Announces the Web UI 🎉
- No colons, simply siding with what most of the log lines do currently
- Use "listening on" consistently rather than "**is** listening on"
- Use "read only" not "readonly"
- Initializing **IPFS** daemon... to confirm to the user they started the thing they expected.

I'd keep the "Daemon is ready" copy the same though as it's the kind of output that people might be scraping from scripts to check that it started without error, and changing it would frustrate.

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "form-data": "^2.3.2",
     "hat": "0.0.3",
     "interface-ipfs-core": "~0.78.0",
-    "ipfsd-ctl": "~0.39.1",
+    "ipfsd-ctl": "ipfs/js-ipfsd-ctl#fix/cli-scraping",
     "mocha": "^5.2.0",
     "ncp": "^2.0.0",
     "nexpect": "~0.5.0",
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "form-data": "^2.3.2",
     "hat": "0.0.3",
     "interface-ipfs-core": "~0.78.0",
-    "ipfsd-ctl": "ipfs/js-ipfsd-ctl#fix/cli-scraping",
+    "ipfsd-ctl": "~0.39.3",
     "mocha": "^5.2.0",
     "ncp": "^2.0.0",
     "nexpect": "~0.5.0",
--- a/src/cli/commands/daemon.js
+++ b/src/cli/commands/daemon.js
@@ -28,7 +28,7 @@ module.exports = {
   },
 
   handler (argv) {
-    print('Initializing daemon...')
+    print('Initializing IPFS daemon...')
 
     const repoPath = utils.getRepoPath()
 
--- a/src/http/index.js
+++ b/src/http/index.js
@@ -155,8 +155,9 @@ function HttpApi (repo, config, cliArgs) {
         api.info.ma = uriToMultiaddr(api.info.uri)
         gateway.info.ma = uriToMultiaddr(gateway.info.uri)
 
-        console.log('API is listening on: %s', api.info.ma)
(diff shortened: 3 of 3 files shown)


olizilla:
@alanshaw the alignment isn't like when people create work for everyone by aligning their variable declarations; in this case it's more like we're showing tabular data (or at least a definition list) so lets give it a table-like layout for readability.

You will get a PR with a CID for the new Web UI when we fix all the issues you raised 😜

achingbrain:
I'd love a `--quiet` option that doesn't show anything so as to not see anything in test output, etc.

alanshaw:
@achingbrain the [silent option](https://github.com/ipfs/js-ipfs/blob/549f2f63e19bc8ae9dfcd7a785f083f29d3c069f/src/cli/bin.js#L23-L28) should probably do this!

Stebalien:
> @Stebalien would you accept a PR to make the same changes in go-ipfs?

Absolutely! (once go-ipfs has had it's webui version updated to the revamped version, that is...)

alanshaw:
@olizilla I think this is going to need changes to ipfsd-ctl before we can merge it.

olizilla:
Sure I'll take a look.

package.json
```diff
@@ -72,7 +72,7 @@
     "form-data": "^2.3.2",
     "hat": "0.0.3",
     "interface-ipfs-core": "~0.78.0",
-    "ipfsd-ctl": "~0.39.1",
+    "ipfsd-ctl": "ipfs/js-ipfsd-ctl#fix/cli-scraping",
```
diasdavid: Can we get this released before releases js-ipfs

PR #1600, opened 2018-09-28
chore: update dependencies

alanshaw:


alanshaw:
Merging as the only failure is jenkins related!

PR #1612, opened 2018-10-02
chore: correct build badge

VictorBjelkholm:
Was pointing to the wrong build


VictorBjelkholm:
Cancelled the CI run for this PR as there is no code changes. (which would probably be a nice feature to have by default in JS pipelines now when I think about it)

PR #1617, opened 2018-10-04
fix: use class is function on ipns

### LINKED ISSUE #1615: Fails to start for the first time due to 'datastore key does not have a valid format'
- **Version**: 0.32.3
- **Platform**: ipfs-companion in Firefox and Chrome
- **Subsystem**: looks like related to newly added IPNS

#### Description:

After switching from js-ipfs `0.31.7` to `0.32.3` switching to embedded node in companion  produces an error:

> 

What is interesting, is that if I switch to External node and then back again to Embedded one, second time js-ipfs tries to initialize is  successful and everything works as expected. 

I did not look into it deeper, but my guess is that there is  some sort first-run init that happens after the error occurs that fixes datastore(?) and makes following runs in the same context to work fine. Maybe IPNS tries to publish key before it is stored or something?

cc @vasco-santos  in case its something obvious 

To build Companion with js-ipfs v0.32.3 and run in Firefox (but the same issue is in Chromium):

1. `git clone https://github.com/ipfs-shipyard/ipfs-companion.git`
1. `cd ipfs-companion`
1. `git reset --hard  21eb92bee2b8ea99daa3fb1836431bd63f047c24`
1. `npm run dev-build`
1. `npm run firefox`

Then, go to _Preferences_ and switch to _Embedded_ node. Initial try will fail, following ones will succeed.

vasco-santos:
As `instance of` fails using different versions of a module, it was changed to use `class-is`.

Needs:

- [x] [js-peer-id#84](https://github.com/libp2p/js-peer-id/pull/84)
- [x] [interface-datastore#24](https://github.com/ipfs/interface-datastore/pull/24)

Fixes #1615

### DIFF
--- a/package.json
+++ b/package.json
@@ -153,7 +153,7 @@
     "once": "^1.4.0",
     "path-exists": "^3.0.0",
     "peer-book": "~0.8.0",
-    "peer-id": "~0.11.0",
+    "peer-id": "~0.12.0",
     "peer-info": "~0.14.1",
     "progress": "^2.0.0",
     "promisify-es6": "^1.0.3",
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "ipld": "~0.17.3",
     "ipld-dag-cbor": "~0.12.1",
     "ipld-dag-pb": "~0.14.6",
-    "ipns": "~0.2.0",
+    "ipns": "~0.3.0",
     "is-ipfs": "~0.4.2",
     "is-pull-stream": "~0.0.0",
     "is-stream": "^1.1.0",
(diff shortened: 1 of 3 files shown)


vasco-santos:
@alanshaw the `peerId` instance is obtained through [js-ipfs/core/ipns/publisher.js#L33](https://codecov.io/gh/ipfs/js-ipfs/src/master/src/core/ipns/publisher.js#L33). Considering that `_putRecordToRouting` is a private function [js-ipfs/core/ipns/publisher.js#L53](https://codecov.io/gh/ipfs/js-ipfs/src/master/src/core/ipns/publisher.js#L53) maybe we should remove this validation. 

I am not able to test this, unless I test the private function directly or mock the `PeerId.createFromPrivKey` function.

PR #1618, opened 2018-10-04
chore: remove ipld-dag-cbor dependency

vmx:
The `ipld-dag-cbor` module isn't used anywhere, hence remove it
as a direct dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,6 @@
     "ipfs-unixfs": "~0.1.15",
     "ipfs-unixfs-engine": "~0.32.3",
     "ipld": "~0.17.3",
-    "ipld-dag-cbor": "~0.12.1",
     "ipld-dag-pb": "~0.14.6",
     "ipns": "~0.2.0",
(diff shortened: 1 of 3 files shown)


vmx:
@vasco-santos I can see the point for other modules. But I don't think you should ever need direct access to ipld-dag-cbor. And if you do, just bundle it yourself. The problem I see here is that if we give access here, it might even be a different version from what js-ipld is using. Hence I think it would cause more pain than benefits.

vasco-santos:
I see. And what about `dagPB`? Should we keep it?

vmx:
I would say IPLD formats are implementation detail, so I'd also be for removing `ipld-dag-pb` if everyone agrees.

vmx:
I'm closing this one in favour of removing all IPLD Formats:

 - https://github.com/ipfs/js-ipfs/pull/1626
 - https://github.com/ipfs/js-ipfs-api/pull/872

PR #1623, opened 2018-10-08
fix: datastore not found error on ipns publish

### LINKED ISSUE #1620: Uncaught Error: unexpected error getting the ipns record
- **Version**: 0.32.3
- **Platform**: Browser
- **Subsystem**: ?

#### Description:

Error starting IPFS node:

```
Uncaught Error: unexpected error getting the ipns record  �F�W��if4w~}k��*X�%v�5Q3W�AC0t� from datastore
      at _repo.datastore.get (node_modules/ipfs/src/core/ipns/publisher.js:184:35)
      at node_modules/levelup/lib/levelup.js:168:14
      at node_modules/encoding-down/index.js:51:21
```

This doesn't happen in v0.32.2.

I start IPFS with the following config options: https://github.com/ipfs-shipyard/peer-star-app/blob/master/src/transport/ipfs.js#L15

vasco-santos:
Check for both error types when hitting a `datastore` not found error.

Closes #1620

### DIFF
--- a/src/core/ipns/publisher.js
+++ b/src/core/ipns/publisher.js
@@ -177,7 +177,7 @@ class IpnsPublisher {
       let result
 
       if (err) {
-        if (err.code !== 'ERR_NOT_FOUND') {
+        if (err.code !== 'ERR_NOT_FOUND' && !err.notFound) {
           const errMsg = `unexpected error getting the ipns record ${peerId.id} from datastore`
 
           log.error(errMsg)


vasco-santos:
With [peer-star-app#72](https://github.com/ipfs-shipyard/peer-star-app/pull/72), we do not need to make this modification, as all the datastores should have the error code uniformization

PR #1646, opened 2018-10-19
chore: upgrade libp2p record version

vasco-santos:
The `libp2p-record` module was updated to interop with `go-libp2p-record`.  As a result, the author and signature were removed, as well as all the logic inherent to the signature.

More info [js-libp2p-record#8](https://github.com/libp2p/js-libp2p-record/pull/8)


alanshaw:
@vasco-santos is this best to get in for the 0.33 release to minimise the time people are creating incompatible IPNS records for?

vasco-santos:
yeah! I think it is best to get it in `0.33`

PR #1647, opened 2018-10-19
feat: update to Web UI 2.0

alanshaw:
🎉 

This isn't the official 2.0 release, but it _is_ the current master. Updating it here so we can release an RC today and we'll upgrade when the official release is available.


alanshaw:
LGTM ;)

PR #1656, opened 2018-10-23
[WIP] fix: better input validation for add

alanshaw:
requires https://github.com/ipfs/interface-ipfs-core/pull/376


alanshaw:
Update: just waiting on next release of IPFS API and then this can be mreged

PR #1661, opened 2018-10-23
fix: correct output in webpack example

vmx:
The webpack example printed out the contents of the file in hex
(`1041011081081113211911111410810032102114111109321191019811297991071011003273807083`)
instead of a properly formated string (`hello world from webpacked IPFS`).


alanshaw:
Checked locally and works for me.

PR #1662, opened 2018-10-24
feat: add support to pass config in the init cmd

hugomrdias:
This PR adds support to pass node config in the init cmd. The main purpose for this is to improve the speed of ipfsd-ctl and all the tests that use ctl.

example: `ipfs init '{"Addresses":{"Swarm":["/ip4/0.0.0.0/tcp/4003"]}}'`
This merges the input config with the default config, just like the node api already does.

Check ipfsd-ctl PR related to this one for speed improvements https://github.com/ipfs/js-ipfsd-ctl/pull/303

Right now we need to spawn 4 child_processes to start a node in ctl
- ipfs init
- ipfs config show
- ipfs config replace
- ipfs daemon

With this PR we would only need 2
- ipfs init <condig>
- ipfs daemon

And with further changes we could do `ipfs daemon --init --config "{}"`

Right now i think we have a couple of problems with this
- go ipfs init needs the config to be in a file and needs it to be a full config, no merging 
- other config changing cmds only do replacing no merging

Questions: 
- How can we get the Go default config to manually do the merging? 
- Can we get Go to do merging? 
- Can we do this change in js without Go supporting it?
- Shouldn't we have a `ipfs config merge` ?

Go version works like this https://docs.ipfs.io/reference/api/cli/#ipfs-init

### DIFF
--- a/src/cli/commands/init.js
+++ b/src/cli/commands/init.js
@@ -4,13 +4,15 @@ const utils = require('../utils')
 const print = utils.print
 
 module.exports = {
-  command: 'init',
-
+  command: 'init [config] [options]',
   describe: 'Initialize a local IPFS node',
-
   builder (yargs) {
     return yargs
       .epilog(utils.ipfsPathHelp)
+      .positional('config', {
+        describe: 'Node config, this should JSON and will be merged with the default config. Check https://github.com/ipfs/js-ipfs#optionsconfig',
+        type: 'string'
+      })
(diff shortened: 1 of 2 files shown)


hugomrdias:
@alanshaw swap deep-extend with defaults-deep with last commit

alanshaw:
Just waiting for CI

diasdavid:
image

PR #1663, opened 2018-10-24
fix: add missing dependencies

hugomrdias:
This PR also add a npm script to check missing dependencies this should in the future go into aegir.


alanshaw:
@hugomrdias please use `~` for versions < 0

hugomrdias:
lol i did just didn't save the file :/ 

fixed with the last commit

package.json
```diff
@@ -95,6 +96,8 @@
     "cid-tool": "~0.1.0",
     "cids": "~0.5.3",
     "debug": "^3.1.0",
+    "deep-extend": "~0.6.0",
+    "electron-webrtc": "~0.3.0",
```
alanshaw: "electron-webrtc" and "wrtc" seem to be excluded on purpose. Could you check the reason behind this?

package.json
```diff
@@ -167,14 +172,18 @@
     "pull-stream": "^3.6.8",
     "pull-stream-to-stream": "^1.3.4",
     "pull-zip": "^2.0.1",
+    "pump": "^3.0.0",
     "read-pkg-up": "^4.0.0",
     "readable-stream": "2.3.6",
     "receptacle": "^1.3.2",
+    "safe-buffer": "^5.1.2",
```
alanshaw: Is this needed? Only one use in the code, and we don't use it anywhere else that we're using buffers.

package.json
hugomrdias: Can you give me more context on 'excluded on purpose' ?

hugomrdias: `wrtc` and `electron-webrtc` break libp2p 

```js
/Users/hugomrdias/code/js-ipfs-ci/node_modules/libp2p/src/index.js:145
      if (t.filter(multiaddrs).length > 0) {
            ^

TypeError: t.filter is not a function
    at _modules.transport.forEach (/Users/hugomrdias/code/js-ipfs-ci/node_modules/libp2p/src/index.js:145:13)
    at Array.forEach (<anonymous>)
    at Node.start (/Users/hugomrdias/code/js-ipfs-ci/node_modules/libp2p/src/index.js:136:29)
    at gotConfig (/Users/hugomrdias/code/js-ipfs-ci/src/core/components/libp2p.js:95:26)
    at store.get (/Users/hugomrdias/code/js-ipfs-ci/node_modules/ipfs-repo/src/config.js:45:9)
    at fs.readFile (/Users/hugomrdias/code/js-ipfs-ci/node_modules/datastore-fs/src/index.js:218:7)
    at /Users/hugomrdias/code/js-ipfs-ci/node_modules/graceful-fs/graceful-fs.js:78:16
    at FSReqWrap.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:53:3)

```

@jacobheun can you help with this ? i have no idea what is going on :/

alanshaw: some context and maybe the fix: https://github.com/libp2p/interface-transport/issues/38#issuecomment-432063553

jacobheun: @hugomrdias I'll take a look.

jacobheun: I found the issue, it's with `deep-extend` https://github.com/ipfs/js-ipfs/blob/master/src/core/index.js#L48. When the instance of `libp2p-webrtc-star` is passed in via options, the methods on the class end up getting stripped off. I'll push up a commit to remove `deep-extend` in favor of `@nodeutils/defaults-deep` as it doesn't mutate.

hugomrdias: thank you @jacobheun 
deep-extend can do immutable if you add an empty object as the first argument. Still https://www.npmjs.com/package/assign-deep is the best package for this type of stuff its one of the smallest, handles non-plain-object like promise correctly, works exactly like object.assign but deep, etc. `@nodeutils/defaults-deep` depends on full lodash so its the biggest of them all.
Please if you agree with the above pls consider using `assign-deep` everything one of my OKR is to reduce the bundle size so this would help :)

Check https://codesandbox.io/s/742p0rjvy0 for comparison

jacobheun: @hugomrdias I think we can look at doing that in a subsequent PR. `deep-extend` doesn't work because it's doing cloning and turning instances into plain objects, and the filter methods get missed in the process.  We can look at `assign-deep` or one of it's sister modules, we just need to make sure it's merging objects similarly to `@nodeutils/defaults-deep`.

hugomrdias: agreed !

alanshaw:
@hugomrdias everything is failing 😞 - would you mind taking a look?

hugomrdias:
did someone merged or rebased this branch ?

hugomrdias:
Updated aegir to make tests run and rebased with master.

Please don't rebase branches you don't own.

diasdavid:
@hugomrdias I don't think anyone rebased, otherwise you would see double authors.

package.json
```diff
@@ -87,6 +88,7 @@
     "cid-tool": "~0.1.0",
     "cids": "~0.5.3",
     "debug": "^3.1.0",
+    "electron-webrtc": "~0.3.0",
```
diasdavid: This is deliberately not installed (as it is not needed)

package.json
```diff
@@ -154,13 +158,16 @@
...
+    "pump": "^3.0.0",
     "read-pkg-up": "^4.0.0",
     "readable-stream": "2.3.6",
     "receptacle": "^1.3.2",
     "stream-to-pull-stream": "^1.7.2",
     "tar-stream": "^1.6.1",
     "temp": "~0.8.3",
     "update-notifier": "^2.5.0",
+    "varint": "^5.0.0",
+    "wrtc": "~0.2.1",
```
diasdavid: This is deliberately not installed (as it is not needed)

alanshaw:
ping @hugomrdias can we get this PR in order? People are reporting [duplicates of issues](https://github.com/ipfs/js-ipfs/issues/1708) that this PR will fix.

package.json
achingbrain: Perhaps it should be moved to peerDependencies?

package.json
hugomrdias: any background on this? to better understand ?

diasdavid: https://github.com/ipfs/js-ipfs#is-there-webrtc-support-for-js-ipfs-with-nodejs

hugomrdias:
tests passing ship it

package.json
```diff
@@ -168,12 +172,17 @@
...
     "yargs-promise": "^1.1.0"
   },
   "optionalDependencies": {
     "prom-client": "^11.1.3",
     "prometheus-gc-stats": "~0.6.0"
   },
+  "peerDependencies": {
+    "electron-webrtc": "~0.3.0",
+    "wrtc": "~0.2.1"
+  },
```
alanshaw: We get the following warning now when installing dependencies for js-ipfs:

```console
npm WARN ipfs@0.33.1 requires a peer of electron-webrtc@~0.3.0 but none is installed. You must install peer dependencies yourself.
npm WARN ipfs@0.33.1 requires a peer of wrtc@~0.2.1 but none is installed. You must install peer dependencies yourself.
```

This makes people think that those dependencies are required when they are not. @achingbrain @hugomrdias do you think these should be moved to `optionalDependencies` instead?

alanshaw: 🙄 it shouldn't break...reading the docs, that seems to be the whole point of `optionalDependencies`...

alanshaw: Ok, remove them for now as they were before...?

PR #1665, opened 2018-10-24
fix: remove unused deps

hugomrdias:
This PR removes unused deps.

When this https://github.com/ipfs/js-ipfs/pull/1663 is merged, we will be able to use `npm run dep-check -- --extra` to check for unused packages


alanshaw:
Has exposed this issue: https://github.com/ipfs/aegir/pull/293

hugomrdias:
I'm on it

Issue #1666, opened 2018-10-24
Consolidate the http-api tests


jacobheun:
Currently all of the http-api tests in the inject folder, https://github.com/ipfs/js-ipfs/tree/master/test/http-api/inject, create the http-api from source and test it directly. This is the behavior I'd expect from the tests.

We also have the tests in the root http-api folder, https://github.com/ipfs/js-ipfs/tree/master/test/http-api, that test the api by proxy by spinning up an ipfs daemon. 

The cli tests end up testing similar code paths as the http-api folder as far as I can tell. This redundancy seems unnecessary and adds a substantial amount of time to the tests. 

Locally, running `time npm run test:node:http` for the current suite takes 8.5 minutes:
```sh
real    8m32.132s
user    6m50.012s
sys     1m2.530s
```
The same run testing only the inject files takes only 6.3 seconds
```sh
real    0m6.368s
user    0m4.764s
sys     0m0.743s
```

Can we get rid of the http-api tests using daemons?

diasdavid:
💯 𝐘Ⓔ𝒮 :D

jacobheun:
In this realm, although it should probably become it's own ticket, I was testing some changes to cli tests. I think we can realistically remove the need for any daemons in the js-ipfs test suite, and limit process forking to just verifying the cli entry.

Currently the CLI tests each spawn a daemon controlled by `ipfsd-ctl`. Ultimately, this ends up either testing the core code (no daemon commands) or http-api (daemon on commands). It would be much more efficient to test `src/cli/bin.js` to just ensure it's invoking the proper handlers with appropriate args, leveraging the right ipfs type (http-api vs core vs `ipfs-api`), and then test the handlers directly.  This also has the nice side effect of making things much easier to debug, because we're not spawning another process.

The testing on the handlers can also be fairly minimal, we just need to ensure that they are interfacing with their respective portion of the ipfs code properly, and transforming the data properly for the CI. This should result in a pretty significant reduction in cli test time. I have a proof of concept working for this, but am short on time at the moment to be able to redo the whole cli test suite.

hugomrdias:
@jacobheun that poc is anywhere on the interwebs ?  would love to have a look

jacobheun:
@hugomrdias I need to tidy it up a bit and then will push it up. Just finished some of it before todays meetings.

jacobheun:
[...]

PR #1672, opened 2018-10-26
test: remove unnecessary api tests

jacobheun:
The http-api test suite is doing redundant work. The removed tests were spinning up a daemon to run the http-api tests by proxy. This duplicates the code paths the cli tests are already using. Also, the inject http-api tests test the api directly, without spinning up a daemon. This is a better, more performant approach and makes these deleted tests redundant.

The risk of exposing future code stability issues due to removing these tests is extremely low since, as mentioned above, the api is being tests twice: in isolation and via the cli.

## Test Time Savings

The following table shows a cost savings in testing time as shown on jenkins between one of the latest master builds and this branch. While this comparison is isolated to a single build of each branch, the build times for other builds of the respective branches are not significantly different and the magnitude of the change warrants this PR.

|  | Mac 8.x | Mac 10.x | Linux 8.x | Linux 10.x | Win 8.x | Win 10.x | Coverage |
|--- |--- |--- |--- |--- |--- |--- |--- |
|[master 22](https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs/detail/master/22/pipeline)| 29m 3s | 27m 38s | 22m 54s | 21m 48s | 28m 4s | 28m 8s | 27m 30s |
|[api-cleanup 2](https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs/detail/test%2Fapi-cleanup/2/pipeline/)| 15m 37s | 15m 15s | 14m 0s | 13m 0s | 16m 16s | 16m 46s | 16m 5s |
|Time reduction* | -46% | -45% | -39% | -40% | -42% | -40% | -41% |

\* Time reduction percentages will fluctuate based on jenkins performance.

Resolves https://github.com/ipfs/js-ipfs/issues/1666


diasdavid:
@jacobheun can you get CI to become green?

alanshaw:
@jacobheun did you check that there is an equivalent CLI / http-inject test for each http-daemon test that has been removed?

jacobheun:
@alanshaw no, I haven't done that full comparison yet, I'll do a run through.

There are inject tests that need to get added as there are endpoints not being tested in either. There's no reason we shouldn't have 100% coverage on the api endpoints.  I think it would be good to couple that work with testing updates mentioned at https://github.com/ipfs/js-ipfs/issues/1666#issuecomment-433993559

jacobheun:
I am going to close this for now, I think more discussion needs to occur on #1666 and that should become a larger testing discussion. We can open this later if it makes sense to proceed.

Looking through the tests I realized I removed the `interface` (interface-ipfs-core), which is removing a large suite of daemon tests. While this is highly redundant a significant audit needs to happen to make sure we're optimizing the tests appropriately. More thorough, mocked, http inject tests should be added before we remove the interface tests.

PR #1673, opened 2018-10-27
docs: add IPFS Architecture Diagram

daviddias:
I took some time to design the current IPFS Architecture so that we could have more informed discussions.

architecture

You can see it as png, txt or a monodraw file.

This diagram presents what the architecture is today and not what is next (the dashed boxes for the internal pieces also don't exist but that is where they fit the puzzle).


diasdavid:
image

vmx:
You mentio

> They represent what is there today and not what is next.

Graph Service/GraphSync isn't a thing yet, so I think it shouldn't be in the diagram.

diasdavid:
@vmx the internal pieces in boxes with dashed pattern mean that they are in progress/missing but that is where they fit on the picture today.

vmx:
@diasdavid K, cool. It would be nice if the diagram would have a legend what dashed boxes mean.

alanshaw:
@diasdavid can we replace the [diagram here](https://github.com/ipfs/js-ipfs#ipfs-core-architecture) with it please?

diasdavid:
@alanshaw absolutely. I was expecting a bit more question/discussion, but if everyone simply agrees with this and likes it, then all good to me! :)

alanshaw:
> @alanshaw absolutely. I was expecting a bit more question/discussion, but if everyone simply agrees with this and likes it, then all good to me! :)

Ok cool! I think it's great and there's not a lot to say about it since that IS the architecture. It's much better than the existing diagram already. Some feedback tho:

* Could possibly be a bit wider? Feels a little squashed.
* Graph service - maybe put "(IPLD)" in the diagram
* `pubsub`, `IPNS`, `DHT`, `Swarm` do they belong in here?
* I think some explanatory text underneath (when added to the README) would be beneficial (basically your annotations from https://github.com/ipfs/js-ipfs/pull/1673#issuecomment-433633559)

alanshaw:
ping @daviddias

daviddias:
Added to my TODO list for this week

daviddias:
Applied CR :)

PR #1678, opened 2018-10-29
fix: remove example eslint config

hugomrdias:
This PR removes eslint config from webpack example to solve the missing esling-config-react when running `aegir lint`. This is one way to go.

Another option is to stop linting examples with aegir, and have them setup specific eslint configs.

Please give feedback about these two options and which one we should adopt.

Related:
https://github.com/ipfs/aegir/pull/293/files
https://github.com/ipfs/js-ipfs/pull/1665#issuecomment-433896935


vasco-santos:
I agree that we should go with this approach, instead of adding the `eslint-plugin-react` with all the aegir projects. 

Moreover, Aegir should support custom configs according to the projects needs.

hugomrdias:
There's already a this PR https://github.com/ipfs/aegir/pull/250 about custom eslint configs support

Issue #1684, opened 2018-10-31
ping tests are failing


diasdavid:
```
  1) HTTP API
       ## http-api spec tests
         /ping
           returns 200 and the response stream with the ping result:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/imp/code/js-ipfs/test/node.js)


  2) HTTP API
       ## http-api spec tests
         /pubsub
           /peers
             returns 200 if no topic is provided:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/imp/code/js-ipfs/test/node.js)


  3) interface-ipfs-core over ipfs-api tests
       .ping
         should fail when pinging an unknown peer:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/imp/code/js-ipfs/test/node.js)


  4) interface-ipfs-core over ipfs-api tests
       .ping
         should fail when pinging an invalid peer:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/imp/code/js-ipfs/test/node.js)


  5) interface-ipfs-core over ipfs-api tests
       .pingPullStream
         should fail when pinging an unknown peer over pull stream:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/imp/code/js-ipfs/test/node.js)


  6) interface-ipfs-core over ipfs-api tests
       .pingPullStream
         should fail when pinging an invalid peer over pull stream:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/imp/code/js-ipfs/test/node.js)


  7) interface-ipfs-core over ipfs-api tests
       .pingReadableStream
         should fail when pinging an unknown peer over readable stream:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/imp/code/js-ipfs/test/node.js)


  8) interface-ipfs-core over ipfs-api tests
       .pingReadableStream
         should fail when pinging an invalid peer over readable stream:
[...]

PR #1685, opened 2018-10-31
fix ping tests

diasdavid:
For https://github.com/ipfs/js-ipfs/issues/1684


test/http-api/inject/pubsub.js
```diff
@@ -91,17 +91,6 @@ module.exports = (http) => {
...
-      it('returns 200 if no topic is provided', (done) => {
-        api.inject({
-          method: 'GET',
-          url: `/api/v0/pubsub/peers`
-        }, (res) => {
-          expect(res.statusCode).to.equal(200)
-          expect(res.result.Strings).to.be.eql([])
-          done()
-        })
-      })
```
diasdavid: This test was no longer valid. I've checked go-ipfs and now it is indeed supposed for IPFS to return the list of peers we are pubsub'ing with when no topic is provided.

src/core/components/ping-pull-stream.js
```diff
@@ -27,7 +27,7 @@ module.exports = function pingPullStream (self) {
       if (err) {
         log.error(err)
         source.push(getPacket({ success: false, text: err.toString() }))
-        source.end(err)
+        source.end()
```
diasdavid: The errors are part of the API. passing the error was making it so that the response would never get to the client!

diasdavid:
Seems there are still some more issues on ping :(

src/core/components/ping-pull-stream.js
```diff
@@ -27,7 +27,8 @@ module.exports = function pingPullStream (self) {
       if (err) {
         log.error(err)
         source.push(getPacket({ success: false, text: err.toString() }))
-        source.end(err)
+        source.push(err)
+        source.end()
```
diasdavid: This made the tests pass but I'm not 100% convinced this is exactly what we want.

diasdavid:
After debugging a bit more, I really think the problem is with the tests themselves. https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/ping/ping.js#L58-L64 seems to be an antipattern to me. We can't expect to have error and results all at the same time.

diasdavid:
Now I also understand why people were reporting that Ping was not working because DHT is not available:

```
AssertionError: expected [Error: Error: DHT is not available] to not exist
```

It should instead say, that it couldn't find the Peer.

src/core/components/ping-pull-stream.js
alanshaw: We need to differentiate between a real failure (invalid peer ID) and unavailable.

diasdavid:
Almost there. last mile is figuring out while the error doesn't propagate to the tests when using core directly (ipfs-api over http-api is fine)

```
Test Node.js


  cli
    ping
      ✓ ping host (955ms)
      ✓ ping host with --n option (791ms)
      ✓ ping host with --count option (908ms)

  HTTP API
Swarm listening on /ip4/127.0.0.1/tcp/0/ws/ipfs/QmQ2zigjQikYnyYUSXZydNXrDRhBut2mubwJBaLXobMt3A
Swarm listening on /ip4/127.0.0.1/tcp/62841/ipfs/QmQ2zigjQikYnyYUSXZydNXrDRhBut2mubwJBaLXobMt3A
API listening on /ip4/127.0.0.1/tcp/62842
Gateway (read only) listening on /ip4/127.0.0.1/tcp/62843
Web UI available at http://127.0.0.1:62842/webui
    ## http-api spec tests
      /ping
        ✓ returns 400 if both n and count are provided
        ✓ returns 400 if arg is not provided
        ✓ returns 200 and the response stream with the ping result

  interface-ipfs-core over ipfs-api tests
    .ping
      ✓ should send the specified number of packets
      ✓ should fail when pinging a peer that is not available
      ✓ should fail when pinging an invalid peer Id
    .pingPullStream
      ✓ should send the specified number of packets over pull stream
      ✓ should fail when pinging an unknown peer over pull stream
      ✓ should fail when pinging an invalid peer id over pull stream
    .pingReadableStream
      ✓ should send the specified number of packets over readable stream
      ✓ should fail when pinging peer that is not available over readable stream
      ✓ should fail when pinging an invalid peer id over readable stream

  interface-ipfs-core tests
    .ping
Swarm listening on /ip4/127.0.0.1/tcp/62915/ipfs/QmeSfaoj3aAEJUH5T7hJ7ULu1evo8j1TewyQhzvPESfaR3
Swarm listening on /ip4/127.0.0.1/tcp/62917/ipfs/QmX98kZpNKipM1EifTj6wqsYS6mmAQX37oYNmVpWPz8dv3
      ✓ should send the specified number of packets
      1) should fail when pinging a peer that is not available
      2) should fail when pinging an invalid peer Id
    .pingPullStream
Swarm listening on /ip4/127.0.0.1/tcp/62924/ipfs/QmY31MzrqTECyDz1MBHWtdMc4DyDPAMvoTXfZSDtYvhQ7K
Swarm listening on /ip4/127.0.0.1/tcp/62928/ipfs/QmTttTG2nVztU7uw4Xq76qwZEQrQxvG5tnmR8Cw5bBQGXG
      ✓ should send the specified number of packets over pull stream
      3) should fail when pinging an unknown peer over pull stream
      4) should fail when pinging an invalid peer id over pull stream
    .pingReadableStream
Swarm listening on /ip4/127.0.0.1/tcp/62935/ipfs/QmVZaeQC65aPEpkrwEnp5SUsvdxXEph4168tfij6i3aFG2
Swarm listening on /ip4/127.0.0.1/tcp/62939/ipfs/QmPLxuVK8hqzdTV3RzZP8hdrChuykNXGuNqdoSFChg4iqs
      ✓ should send the specified number of packets over readable stream
      5) should fail when pinging peer that is not available over readable stream
      6) should fail when pinging an invalid peer id over readable stream
```

src/core/components/ping-pull-stream.js
diasdavid: I've separated the getPeer from the runPing steps

src/http/api/resources/ping.js
```diff
@@ -43,6 +43,9 @@ exports.get = {
     const responseStream = toStream.source(source)
     const stream2 = new PassThrough()
     pump(responseStream, stream2)
-    return reply(stream2).type('application/json').header('X-Chunked-Output', '1')
+    // TODO: here lies the issue
+    // The fact that ping can fail after it starts writting some success ndjson
+    // responses is not very friendly to how Hapi expects this stream to work
+    reply(stream2).type('application/json').header('X-Chunked-Output', '1')
```
diasdavid: Here, I believe this to be the problem

src/http/api/resources/ping.js
```diff
@@ -23,9 +24,19 @@ exports.get = {
   handler: (request, reply) => {
```
diasdavid: @alanshaw @hugomrdias @achingbrain et al, can I get your eyes to go through this handler. This stream tagliatelle is a nightmare came true.

achingbrain: Stream tagliatelle - I’m going to use that one

PR #1695, opened 2018-11-05
test: dont bootstrap unless it's needed

jacobheun:
This removes bootstrap nodes from all tests that don't need them. This should also help improve test determinism by preventing outside node influence. This also has the pleasant side effect of not having test nodes connect to our production bootstrap nodes.

A nice **future** update would be to set up specific bootstrap nodes when we do need to test bootstrapping so we're not hitting production at all.

### DIFF
--- a/test/core/interface.spec.js
+++ b/test/core/interface.spec.js
@@ -39,6 +39,9 @@ describe('interface-ipfs-core tests', () => {
       initOptions: { bits: 512 },
       EXPERIMENTAL: {
         dht: true
+      },
+      config: {
+        Bootstrap: []
       }
     }
   })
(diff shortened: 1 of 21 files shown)


test/core/interface.spec.js
```diff
@@ -154,7 +157,10 @@ describe('interface-ipfs-core tests', () => {
 
             if (typeof config === 'function') {
               cb = config
-              config = undefined
```
alanshaw: Still need to set config to `null` here otherwise it'll still point to the callback function and you won't get your empty bootstraps ;)

jacobheun: Added it back. I thought I already did, but my brain was wrong. Thanks for catching it!

jacobheun:
Ci is all green now 💚

PR #1696, opened 2018-11-05
refactor: use types test from interface-ipfs-core

vmx:
The test for types also exists in interface-ipfs-core, which is the
central test repository for both js-ipfs and js-http-api tests.
Pulling in the test once from there is enough.


vmx:
CI is green, only circleci failed.

PR #1701, opened 2018-11-07
refactor: ipns routing logic moved to instantiation

vasco-santos:
Currently, the `ipns` routing logic is inside the `publish`/ `resolve` logic.

Aiming to allow easily pluggable routings to `ipns`, I created this PR where the routing logic was moved from the `ipns.publish` / `ipns.resolve` to the instantiation of `ipns`. 

With this, we can add a `TieredDatastore` to `ipns` with an array of datastores. This array can have the node repo's datastore, the DHT, the [ipfs/js-datastore-pubsub](https://github.com/ipfs/js-datastore-pubsub), among others in the future. As a result, inside `ipns` logic, we will have `routing.put` and `routing.get` abstracted to all the enabled datastores.

The routing received by `ipns` will differ according to the node state. In the `init` and `pre-start` step, it will only receive the `OfflineDatastore` aka Repo's datastore. This is necessary to the offline features of `IPNS`, such as `initializeKeyspace`. Finally, in the end of the `start` step, that is, when all the services are running, `ipns` routing will have the mentioned `TieredDatastore`, containing all the intended datastores.

Moreover, the `routing` keys used to `put` records and public keys, where modified according to the new ones being used on `go-ipfs`

**Note**: this logic was implemented in [ipfs/js-ipfs/pull/1559](https://github.com/ipfs/js-ipfs/pull/1559), which is currently blocked due to an interop problem with the topic ([ipfs/interop#39](https://github.com/ipfs/interop/pull/39)). This way, I extracted this from that PR, in order to allow the implementation of `IPNS over DHT`. Once this is merged, I will also rebase the `Pubsub` PR with this, maintaing only the pubsub specific code in that PR.

### DIFF
--- a/src/core/components/start.js
+++ b/src/core/components/start.js
@@ -4,6 +4,10 @@ const series = require('async/series')
 const Bitswap = require('ipfs-bitswap')
 const setImmediate = require('async/setImmediate')
 const promisify = require('promisify-es6')
+const { TieredDatastore } = require('datastore-core')
+
+const IPNS = require('../ipns')
+const OfflineDatastore = require('../ipns/routing/offline-datastore')
 
 module.exports = (self) => {
   return promisify((callback) => {
@@ -34,6 +38,20 @@ module.exports = (self) => {
       },
       (cb) => self.libp2p.start(cb),
       (cb) => {
+        // Setup online routing for IPNS with a tiered routing composed by a DHT and a Pubsub router (if properly enabled)
+        const ipnsStores = []
+
+        // TODO Add IPNS pubsub if enabled
+
+        // NOTE: Until the IPNS over DHT is not ready, it is being replaced by the local repo datastore
+        // When DHT is added, If local option enabled, should receive offlineDatastore as well
+        const offlineDatastore = new OfflineDatastore(self._repo)
+        ipnsStores.push(offlineDatastore)
+
+        // Create ipns routing with a set of datastores
+        const routing = new TieredDatastore(ipnsStores)
+        self._ipns = new IPNS(routing, self)
+
         self._bitswap = new Bitswap(
           self._libp2pNode,
           self._repo.blocks,
--- a/src/core/components/start.js
+++ b/src/core/components/start.js
@@ -43,14 +43,14 @@ module.exports = (self) => {
 
         // TODO Add IPNS pubsub if enabled
 
-        // NOTE: Until the IPNS over DHT is not ready, it is being replaced by the local repo datastore
-        // When DHT is added, If local option enabled, should receive offlineDatastore as well
+        // NOTE: IPNS routing is being replaced by the local repo datastore while the IPNS over DHT is not ready
+        // When DHT is added, if local option enabled, should receive offlineDatastore as well
         const offlineDatastore = new OfflineDatastore(self._repo)
         ipnsStores.push(offlineDatastore)
 
         // Create ipns routing with a set of datastores
         const routing = new TieredDatastore(ipnsStores)
-        self._ipns = new IPNS(routing, self)
+        self._ipns = new IPNS(routing, self._repo, self._peerInfo, self._keychain, self._options)
 
         self._bitswap = new Bitswap(
           self._libp2pNode,
--- a/src/core/ipns/republisher.js
+++ b/src/core/ipns/republisher.js
@@ -18,10 +18,12 @@ const defaultBroadcastInterval = 4 * hour
 const defaultRecordLifetime = 24 * hour
 
 class IpnsRepublisher {
-  constructor (publisher, ipfs) {
+  constructor (publisher, repo, peerInfo, keychain, options) {
     this._publisher = publisher
-    this._ipfs = ipfs
-    this._repo = ipfs._repo
+    this._repo = repo
+    this._peerInfo = peerInfo
+    this._keychain = keychain
+    this._options = options
     this._republishHandle = null
   }
 
@@ -62,8 +64,8 @@ class IpnsRepublisher {
       }
     }
 
-    const { privKey } = this._ipfs._peerInfo.id
-    const { pass } = this._ipfs._options
+    const { privKey } = this._peerInfo.id
+    const { pass } = this._options
 
     republishHandle.runPeriodically((done) => {
       this._republishEntries(privKey, pass, () => done(defaultBroadcastInterval))
@@ -98,8 +100,8 @@ class IpnsRepublisher {
       }
 
       // keychain needs pass to get the cryptographic keys
-      if (this._ipfs._keychain && Boolean(pass)) {
-        this._ipfs._keychain.listKeys((err, list) => {
+      if (this._keychain && Boolean(pass)) {
+        this._keychain.listKeys((err, list) => {
           if (err) {
             log.error(err)
             return
@@ -107,7 +109,7 @@ class IpnsRepublisher {
 
           each(list, (key, cb) => {
             waterfall([
-              (cb) => this._ipfs._keychain.exportKey(key.name, pass, cb),
+              (cb) => this._keychain.exportKey(key.name, pass, cb),
               (pem, cb) => crypto.keys.import(pem, pass, cb)
             ], (err, privKey) => {
               if (err) {
--- a/src/core/ipns/republisher.js
+++ b/src/core/ipns/republisher.js
@@ -100,7 +100,7 @@ class IpnsRepublisher {
       }
 
       // keychain needs pass to get the cryptographic keys
-      if (this._keychain && Boolean(pass)) {
+      if (pass) {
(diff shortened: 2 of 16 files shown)


vasco-santos:
Tested `ipfs/interop` => 🍏

src/core/ipns/resolver.js
```diff
@@ -102,30 +84,35 @@ class IpnsResolver {
...
     })
   }
 
-  // resolve ipns entries locally using the datastore
-  _resolveLocal (name, peerId, callback) {
-    const { ipnsKey } = ipns.getIdKeys(fromB58String(name))
+  // resolve ipns entries from the provided routing
+  _resolveName (name, callback) {
+    const peerId = PeerId.createFromB58String(name)
+    const { routingKey } = ipns.getIdKeys(fromB58String(name))
```
fsdiogo: Can't you use the peerId here? This way you're duplicating the calculation. 

```js
const { routingKey } = ipns.getIdKeys(peerId.toBytes())
```

vasco-santos: Thanks, updated!

src/core/components/start.js
```diff
@@ -34,6 +38,20 @@ module.exports = (self) => {
...
+        // TODO Add IPNS pubsub if enabled
+
+        // NOTE: Until the IPNS over DHT is not ready, it is being replaced by the local repo datastore
+        // When DHT is added, If local option enabled, should receive offlineDatastore as well
+        const offlineDatastore = new OfflineDatastore(self._repo)
+        ipnsStores.push(offlineDatastore)
+
+        // Create ipns routing with a set of datastores
+        const routing = new TieredDatastore(ipnsStores)
+        self._ipns = new IPNS(routing, self)
```
alanshaw: It would be easier to test and easier to understand what dependencies this module has if we just passed the things we need e.g. repo, keychain, peerId etc. instead of `self`

src/core/ipns/resolver.js
```diff
@@ -102,30 +83,35 @@ class IpnsResolver {
...
 
-    this._repo.datastore.get(ipnsKey, (err, dsVal) => {
-      if (err) {
-        const errMsg = `local record requested was not found for ${name} (${ipnsKey})`
+    // TODO DHT - get public key from routing?
+    // https://github.com/ipfs/go-ipfs/blob/master/namesys/routing.go#L70
+    // https://github.com/libp2p/go-libp2p-routing/blob/master/routing.go#L99
+
+    this._routing.get(routingKey.toBuffer(), (err, res) => {
+      if (err || !res) {
```
alanshaw: Please split into two if statements so we don't mask errors as not found.

src/core/ipns/resolver.js
```diff
@@ -102,30 +83,35 @@ class IpnsResolver {
...
+      if (!Buffer.isBuffer(res)) {
         const errMsg = `found ipns record that we couldn't convert to a value`
 
         log.error(errMsg)
         return callback(errcode(new Error(errMsg), 'ERR_INVALID_RECORD_RECEIVED'))
       }
 
-      const record = Record.deserialize(dsVal)
+      const record = Record.deserialize(res)
       const ipnsEntry = ipns.unmarshal(record.value)
```
alanshaw: `try/catch` around these incase of bad values from the network?

src/core/ipns/routing/offline-datastore.js
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Put a value to the local datastore indexed by the received key properly encoded.
+   * @param {Buffer} key identifier of the value.
+   * @param {Buffer} value value to be stored.
+   * @param {function(Error)} callback
+   * @returns {void}
+   */
+  put (key, value, callback) {
+    // encode key properly - base32(/ipns/{cid})
+    const routingKey = new Key('/' + encodeBase32(key), false)
```
alanshaw: Needs a `try/catch` if `encodeBase32` can throw...

src/core/ipns/routing/offline-datastore.js
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Get a value from the local datastore indexed by the received key properly encoded.
+   * @param {Buffer} key identifier of the value to be obtained.
+   * @param {function(Error, Buffer)} callback
+   * @returns {void}
+   */
+  get (key, callback) {
+    // encode key properly - base32(/ipns/{cid})
+    const routingKey = new Key('/' + encodeBase32(key), false)
```
alanshaw: Needs a `try/catch` if `encodeBase32` can throw...

src/core/ipns/routing/offline-datastore.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * Put a value to the local datastore indexed by the received key properly encoded.
+   * @param {Buffer} key identifier of the value.
+   * @param {Buffer} value value to be stored.
+   * @param {function(Error)} callback
+   * @returns {void}
+   */
+  put (key, value, callback) {
```
alanshaw: Do we need to validate `key` in any way?

alanshaw:
P.S. commit message prefixes should be "refactor: " for this ;)

vasco-santos:
@alanshaw addressed your review, changed the commit message prefix and added a bunch of tests to the `/core/ipns`.

src/core/ipns/republisher.js
```diff
@@ -98,16 +100,16 @@ class IpnsRepublisher {
       }
 
       // keychain needs pass to get the cryptographic keys
-      if (this._ipfs._keychain && Boolean(pass)) {
-        this._ipfs._keychain.listKeys((err, list) => {
+      if (this._keychain && Boolean(pass)) {
```
alanshaw: Can `this._keychain` be `null`?

src/core/ipns/resolver.js
```diff
@@ -89,19 +89,33 @@ class IpnsResolver {
...
 
     this._routing.get(routingKey.toBuffer(), (err, res) => {
-      if (err || !res) {
+      if (err) {
         const errMsg = `record requested was not found for ${name} (${routingKey}) in the network`
 
         log.error(errMsg)
-        return callback(errcode(new Error(errMsg), 'ERR_NO_NETWORK_RECORD_FOUND'))
+        return callback(errcode(new Error(errMsg), 'ERR_NO_RECORD_FOUND'))
+      }
```
alanshaw: Is this the only error that can occur - not found? We need to destinguish between not found and other errors.

src/core/ipns/resolver.js
```diff
@@ -89,19 +89,33 @@ class IpnsResolver {
...
+
+      if (!res) {
+        const errMsg = `record requested was empty for ${name} (${routingKey}) in the network`
+
+        log.error(errMsg)
+        return callback(errcode(new Error(errMsg), 'ERR_EMPTY_RECORD_FOUND'))
       }
```
alanshaw: I was thinking a `null` value for `res` would signify not found...but it looks like that's not the case?

If an error is raised for not found then I don't think we need to deal with this - I'd remove this and let it continue through to `Record.deserialize`/`ipns.unmarshal` where it will fail...

src/core/ipns/resolver.js
```diff
@@ -111,8 +125,15 @@ class IpnsResolver {
         return callback(errcode(new Error(errMsg), 'ERR_INVALID_RECORD_RECEIVED'))
       }
```
alanshaw: Do we need this check here? Does `Record.deserialize`/`ipns.unmarshal` deal with this? - it should do!

src/core/ipns/routing/utils.js
```diff
@@ -1,8 +1,7 @@
 'use strict'
 
-const base32 = require('base32.js')
+const multibase = require('multibase')
 
 module.exports.encodeBase32 = (buf) => {
-  const enc = new base32.Encoder()
-  return enc.write(buf).finalize()
+  return multibase.encode('base32', buf)
```
alanshaw: If you only want the base32 encoding of `buf` you'll need to slice off the multibase codec from the front, or require and use the base32 module directly https://github.com/multiformats/js-multibase/blob/master/src/base32.js.

src/core/ipns/resolver.js
vasco-santos: Yes, I should do the same as in `/ipns/publisher`, where I differentiate the not found error, from the unexpected other errors.

src/core/ipns/resolver.js
vasco-santos: agree that differentiating the errors above is enough

alanshaw:
@vasco-santos would you mind rebasing this now the [pin fixes](https://github.com/ipfs/js-ipfs/pull/1719) are in?

vasco-santos:
It is already running the CI rebased 😀

PR #1671, opened 2018-10-26
[WIP] feat: upgrade the Files API

daviddias:
For https://github.com/ipfs/interface-ipfs-core/pull/378

PR #1703, opened 2018-11-09
docs: remove unexistent alias

fsdiogo:
I stumbled upon this when working on https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/18.

The documentation stated that `ipfs.files.get` was an alias to `ipfs.get`. In the CLI and HTTP API it's the other way around: `ipfs.get` is an alias to `ipfs.files.get`, as you can see [here](https://github.com/ipfs/js-ipfs-api/blob/348a1443bd1069c86a4199670a454d2bfc1e8965/src/get.js#L8).

But on the Core API those alias don't exist, so we may as well remove them from the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,13 +495,13 @@ The core API is grouped into several areas:
 #### Files
 
 - [files](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md)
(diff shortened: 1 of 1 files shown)


diasdavid:
From the perspective of the user coming out of go-ipfs, files.get is an alias to .get.

https://github.com/ipfs/js-ipfs/pull/1671 will shuffle these around

PR #1704, opened 2018-11-09
perf: lazy load IPLD formats

alanshaw:
This PR uses the new `loadFormat` option for IPLD to lazily require IPLD formats in order to reduce the startup time for the node.

If you're feeling like you've seen this before then, for reference:

The PR https://github.com/ipld/js-ipld/pull/164 undid the work done in https://github.com/ipld/js-ipld/pull/145 and https://github.com/ipld/js-ipld/pull/178 re-enabled the ability to do so. This PR makes use of this new ability to lazy load the formats.


mikeal:
Love this, +1, but we should maybe start another thread about trying to make this work with code splitting in webpack as well. Getting these out of bundles would be a big win :)

alanshaw:
> Can we go even one more lever deeper in the lazyness and make the formats loaded through IPFS if the node gets a CID for which it doesn't have the IPLD Format code yet?

❤️ YES 🤯

IPFS module loader? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Dynamic_Imports

//cc @achingbrain

PR #1716, opened 2018-11-16
Upgrade to latest libp2p, 0.24

jacobheun:
This bumps all libp2p modules to their latest versions.
And also updates the libp2p-custom example with the latest versions and fixes the usage of `libp2p-railing` which is now `libp2p-bootstrap`


alanshaw:
Merging as CI failure is because macos is out of disk space

PR #1718, opened 2018-11-19
fix: remove electron-webrtc and wrtc for now

hugomrdias:


alanshaw:
@hugomrdias pin tests were fixed in https://github.com/ipfs/js-ipfs/pull/1719 - would you mind rebasing this to ensure CI passing?

PR #1722, opened 2018-11-19
chore: add renovate

hugomrdias:
This PR adds renovate service config.

Renovate works like greenkeeper but has filters. We have been talking about having a dependencies checker service to automatically do PRs with new versions but only for our packages. 

The config in the PR only enables packages matching `^ipfs` for now so we can test if this works well for us. 

Check this https://github.com/hugomrdias/js-ipfs/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc to see what the PRs look like.

for this to work we need this service https://github.com/marketplace/renovate active for this repo. /cc @daviddias

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,26 @@
     "coverage-publish": "aegir-coverage publish",
     "dep-check": "npx dependency-check package.json './test/**/*.js' './src/**/*.js'"
   },
+  "renovate": {
+    "extends": [
+      "config:base",
+      ":preserveSemverRanges"
+    ],
+    "packageRules": [
+      {
+        "packagePatterns": [
+          ".*"
+        ],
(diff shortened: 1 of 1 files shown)


alanshaw:
Currently tests are always failing due to half complete circle config - we need to sort that out before we enable this.

hugomrdias:
Ok I'll try to fix the config, do you feel it's useful to also activate gitlab CI but with the pre release and release in manual mode?

alanshaw:
@hugomrdias yes lets do it ;)

alanshaw:
I'm going to close this - we've enabled greenkeeper and it seems to be working quite well (now that we have stable CI). Will revisit if it becomes problematic.

PR #1725, opened 2018-11-20
feat: ipns over dht

vasco-santos:
Here is an initial implementation of `IPNS over DHT`. Until the dht is enabled by default in `js-ipfs` we need to use the `--enable-dht-experiment` experimental flag, or it will only use the local repo as datastore.

### Needs:

- [x] [libp2p/js-libp2p#288](https://github.com/libp2p/js-libp2p/pull/288)
- [x] [libp2p/js-libp2p-kad-dht#57](https://github.com/libp2p/js-libp2p-kad-dht/pull/57)
- [x] `js-libp2p` release with [libp2p/js-libp2p#288](https://github.com/libp2p/js-libp2p/pull/288)

### Unblocks:

- [ ] [ipfs/interop#47](https://github.com/ipfs/interop/pull/47)


dirkmc:
@vasco-santos it's great to see you moving ahead with this work

It looks like the code only gets a single IPNS value from the DHT, whereas in the go version it retrieves multiple versions of the IPNS value (I believe the default is 16).

The go version provides [a streaming API](https://github.com/ipfs/go-ipfs/blob/master/namesys/interface.go#L91). Do you plan to support streaming in JS?

vasco-santos:
hey @dirkmc 

So the stream was added recently to `go-ipfs`, in the latest `ipfs@.4.18`iirc!

We should add support for it in `js-ipfs`, as well, but I believe the main focus now should be to have this initial implementation as go had before and have interop between `go` and `js`. After that, I think it will be reasonable to add a stream API for this. We have been discussing adding stream's in the API for other things, and probably we should look into this too.

Just other detail, we currently do the `dht.getMany` with 16 occurences too, in the `js-libp2p-kad-dht` [libp2p/js-libp2p-kad-dhtsrc/private.js#L271](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/private.js#L271). However, we select the best record afterwards (using a select function provided), and return the "best" value.

dirkmc:
@vasco-santos I agree about the streaming interface, it can wait until a later version. The reason I bring it up is because in the go version of IPNS, retrieving those 16 records takes so long that it almost always times out, so IPNS is painfully slow. I added an option to retrieve less than 16 records to mitigate this problem, but the real solution is streaming.

I think a lot of people were turned away from IPNS and even IPFS because of IPNS slowness so it would be great to get ahead of the issue as soon as possible

src/core/ipns/publisher.js
```diff
@@ -174,45 +161,55 @@ class IpnsPublisher {
...
+    try {
+      result = ipns.unmarshal(data)
+    } catch (err) {
+      const errMsg = `found ipns record that we couldn't convert to a value`
 
-        log.error(errMsg)
-        return callback(null, null)
-      }
+      log.error(errMsg)
+      return callback(null, null)
```
alanshaw: Do you need to callback with the error here?

src/core/ipns/publisher.js
```diff
@@ -174,45 +161,55 @@ class IpnsPublisher {
...
+    let result
+    try {
+      result = ipns.unmarshal(data)
+    } catch (err) {
+      const errMsg = `found ipns record that we couldn't convert to a value`
 
-        log.error(errMsg)
-        return callback(null, null)
-      }
+      log.error(errMsg)
```
alanshaw: Please log the error!

src/core/ipns/publisher.js
```diff
@@ -174,45 +161,55 @@ class IpnsPublisher {
...
+        try {
+          keys = ipns.getIdKeys(peerId.toBytes())
+        } catch (err) {
+          log.error(err)
+          return callback(err)
+        }
+
+        this._routing.get(keys.routingKey, (err, res) => {
+          if (err) {
+            log(`error when determining the last published IPNS record for ${peerId.id}`)
```
alanshaw: Please log the error!

src/core/ipns/publisher.js
```diff
@@ -174,45 +161,55 @@ class IpnsPublisher {
...
             return callback(null, null)
```
alanshaw: Callback with the error?

alanshaw:
@vasco-santos would you mind rebasing this now that the IPNS over pubsub work is merged? 🙏

src/core/ipns/resolver.js
```diff
@@ -96,14 +98,15 @@ class IpnsResolver {
...
-    // https://github.com/ipfs/go-ipfs/blob/master/namesys/routing.go#L70
-    // https://github.com/libp2p/go-libp2p-routing/blob/master/routing.go#L99
-
-    this._routing.get(routingKey.toBuffer(), (err, res) => {
-      if (err) {
+    auto({
+      // Name should be the hash of a public key retrievable from ipfs.
+      // We retrieve public key to add it to the PeerId, as the IPNS record may not have it.
+      pubKey: (cb) => this._routing.get(routingPubKey.toBuffer(), cb),
+      record: (cb) => this._routing.get(routingKey.toBuffer(), cb)
```
alanshaw: These will be done in parallel - is that the desired behaviour? If so you can probably just use `async/parallel` instead.

src/core/ipns/resolver.js
```diff
@@ -116,9 +119,24 @@ class IpnsResolver {
         return callback(errcode(new Error(errMsg), 'ERR_NO_RECORD_FOUND'))
       }
 
+      // If public key was found in the routing, add it to the peer id
+      // otherwise, wait to check if it is embedded in the record.
+      if (res.pubKey) {
```
alanshaw: Can this be `null` and not 404 error?

vasco-santos: If we fail to get the public key, but we can get the record, we should keep going, in order to try to resolve the record correctly. Two main reasons for this:

- the IPNS record will almost for sure have the public key in it, and consequently, if we end up failing to get the public key for any reason, we should try to keep going until checking if it has the public key
- https://github.com/ipfs/js-ipfs/blob/master/src/core/ipns/publisher.js#L77 in JS land, we always put the public key in the ipns record. So, we are only putting it to the routing, in order to be consistent with the old go nodes that do not embed it in the record. All the public key logic will be removed in a near future, and when `go-ipfs` does that, we will not break as we keep goind in this case. (the code should be removed in that time too).

dirkmc: Once the call to get the record completes, can we check at that point if the public key is in the record, and if it is, don't bother waiting until the call to get the public key completes?

vasco-santos: I can change to your suggestion @dirkmc . I did not do that because doing two `dht.get` in parallel would be faster than doing both in series. But, as we will almost everytime have it in the record, I think it may be better to go with your suggestion. Let's wait for @alanshaw feedback

dirkmc: I don't think it's necessary to do them in series. It would be something like:

```Javascript
let complete = 0
const onComplete = once(function () {
  // ...
})
getRecord((err, rec) => {
  complete++
  if (complete === 2 || rec.hasPubKey()) onComplete()
})
getPubkeyFromNetwork((err, pubk) => {
  complete++
  if (complete === 2) onComplete()
})
```

dirkmc:
W00t 🎉

PR #1728, opened 2018-11-23
feat: move to travis-ci.com

hugomrdias:
Needs:
https://github.com/ipfs/aegir/pull/324


.gitlab-ci.yml
```diff
@@ -0,0 +1,99 @@
+image: hugomrdias/node-alpine:test
+stages:
+  - check
+  - test
+  - cov
+  - pre-release
+  - release
```
alanshaw: I think we should comment these last two out until we have a step for them

.gitlab-ci.yml
```diff
@@ -0,0 +1,99 @@
...
+    - win
+  artifacts:
+    paths:
+      - .nyc_output/
+
+test-browser-chrome:
+  image: hugomrdias/node-chrome:test
+  stage: test
+  script:
+    - AEGIR_BROWSERS=ChromeDocker yarn test:browser
```
alanshaw: yarn here also?

.gitlab-ci.yml
```diff
@@ -0,0 +1,99 @@
...
+  - yarn
+
+check:
+  stage: check
+  script:
+    - yarn lint
+    - yarn dep-check
+    # - npm install --no-lockfile @commitlint/config-conventional @commitlint/cli --save-dev
+    # - npx commitlint --extends=@commitlint/config-conventional --from=$CI_COMMIT_BEFORE_SHA --to=$CI_COMMIT_SHA
+    - yarn build
```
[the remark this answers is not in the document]
hugomrdias: i can change to npm run no problem i just what to use yarn to install packages because its faster without lock file.

vasco-santos:
Cooool! When can we get this merged?

achingbrain:
The GitLab build for this PR took 86 minutes.  Was that expected?  https://gitlab.com/ipfs-ci/js-ipfs/pipelines/38534308

hugomrdias:
No!! node 11 takes a long time dunno why yet probably our fault. 
look at the node 10 job time 35 mins, so full pipeline 40+- mins

And remember this is running in gitlab linux shared containers , win/mac in a single mac book pro (win inside virtualbox) and no cache tweaks like travis has (ill add that later).

TLDR ignore how long it takes and look at how it works, features, interface. 

Pipeline duration can only improve with dedicated machines and cache.

jacobheun:
I have a couple questions, but I think since the plan is to test this for a while in parallel, these aren't blockers for this PR.

* Where can you view the code coverage report? In Jenkins it links off to codecov.io so you can view the breakdown. I'm not seeing where the artifacts are uploaded to. It would be great to log that out in the step if we can, so it's easy to find.
* Do we have plans to add an aegir script to update the gitlab yml file if we pick gitlab? In Jenkins we have global configuration to handle updating node versions. This should probably be in an aegir check step so we don't end up with old pipelines.


Overall the individual job reporting at the pipeline level is much nicer than we have now, and individual retries is fantastic.

hugomrdias:
pipeline-cov-compressor

the second one, yes probably

achingbrain:
It does seem like a vast improvement 😄

.gitlab-ci.yml
```diff
@@ -0,0 +1,97 @@
...
+stages:
+  - check
+  - test
+  - cov
+
+variables:
+  DOCKER_DRIVER: overlay2
+
+before_script:
+  - yarn
```
alanshaw: Still yarn? Just to clarify, I'd rather not use yarn in CI as I think most devs here use npm (including myself) and I want to keep the dev friction to a minimum when eventually we have a problem with npm installing a different directory structure to yarn and it causing problems.

.gitlab-ci.yml
```diff
@@ -0,0 +1,89 @@
...
+  stage: test
+  script:
+    - AEGIR_BROWSERS=ChromeDocker npm run test:browser
+    - AEGIR_BROWSERS=ChromeDocker npm run test:webworker
+  artifacts:
+    paths:
+      - coverage/
+
+test-browser-firefox:
+  allow_failure: true
```
alanshaw: Does this meant that the overall build will pass even if this step fails? What's the reason for adding this currently?

.gitlab-ci.yml
```diff
@@ -0,0 +1,89 @@
+image: hugomrdias/node-alpine:test
```
alanshaw: Can we publish this under "ipfs" or "protocol labs" or something? What's the difference between this image and regular `node-alpine`?

.aegir.js
```diff
@@ -21,6 +22,39 @@ module.exports = {
...
+        // We must disable the Chrome sandbox when running Chrome inside Docker (Chrome's sandbox needs
+        // more permissions than Docker allows by default)
+        flags: ['--no-sandbox']
+      }
+    },
+    client: {
+      mocha: {
+        bail: true,
+      }
+    },
```
alanshaw: Will some of this eventually move into Aegir? Seems weird that we have so much Aegir config overrides when it is basically built to service PL JS projects.

hugomrdias: we are adding features here that aegir doesn't support yet, but it will and this will all be removed.

alanshaw:
@hugomrdias are you happy to merge this or is there more you want to add?

hugomrdias:
i would like to have at least chrome green, it seems the error is network related i did a restart lets see

daviddias:
@hugomrdias can't wait for you to put this one on a boat 🛳 \o/ :D

daviddias:
image

Nice! Sad that Windows still takes so long though.

PR #1729, opened 2018-11-26
fix: better error message when pubsub is not enabled

alanshaw:
If pubsub is disabled there's no `pubsub` property on the libp2p node and we get the error "Cannot read property 'subscribe' of undefined". This is really confusing. This PR detects disabled pubsub and throws an error with a better message.

### DIFF
--- a/src/core/components/pubsub.js
+++ b/src/core/components/pubsub.js
@@ -2,6 +2,11 @@
 
 const promisify = require('promisify-es6')
 const setImmediate = require('async/setImmediate')
+const errCode = require('err-code')
+
+const errPubsubDisabled = () => {
+  return errCode(new Error('pubsub experiment is not enabled'), 'ERR_PUBSUB_DISABLED')
+}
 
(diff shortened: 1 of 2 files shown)


alanshaw:
I think what @vasco-santos suggested is an interesting idea, but it would need to wait for the `async/await` refactor, otherwise we'd still need code to deal with the throw in js-ipfs and return it in the callback.

`libp2p.pubsub` can be a getter that throws if pubsub is not enabled. When we only have an async/await API, an app using libp2p will (likely) have a `try/catch` around the call anyway:

```js
class Lib2p {
  /* ... */
  get pubsub () {
    if (!this.options.pubsub.enabled) throw new Error('pubsub not enabled')
    return this._pubsub
  }
}
```

```js
const libp2p = new Libp2p

try {
  libp2p.pubsub.subscribe(/*...*/)
} catch (err) {
  console.log(err.message) // pubsub is not enabled
}
```

PR #1730, opened 2018-11-27
[WIP] refactor: object APIs write methods now return CIDs

alanshaw:
For the back story on this change, please see: https://github.com/ipfs/interface-ipfs-core/pull/388#pullrequestreview-173866270

BREAKING CHANGE: Object API refactor.

Object API methods that write DAG nodes now return a [CID](https://www.npmjs.com/package/cids) instead of a DAG node. Affected methods:

* `ipfs.object.new`
* `ipfs.object.patch.addLink`
* `ipfs.object.patch.appendData`
* `ipfs.object.patch.rmLink`
* `ipfs.object.patch.setData`
* `ipfs.object.put`

Example:

```js
// Before
const dagNode = await ipfs.object.new()
```

```js
// After
const cid = await ipfs.object.new() // now returns a CID
const dagNode = await ipfs.object.get(cid) // fetch the DAG node that was created
```

IMPORTANT: `DAGNode` instances, which are part of the IPLD dag-pb format have been refactored.

These instances no longer have `multihash`, `cid` or `serialized` properties.

This effects the following API methods that return these types of objects:

* `ipfs.object.get`
* `ipfs.dag.get`

See https://github.com/ipld/js-ipld-dag-pb/pull/99 for more information.

### DIFF
--- a/src/core/components/object.js
+++ b/src/core/components/object.js
@@ -98,7 +98,7 @@ module.exports = function object (self) {
                 self._preload(cid)
               }
 
-              cb(null, node)
+              cb(null, cid)
             })
           })
         }
@@ -150,7 +150,7 @@ module.exports = function object (self) {
             self._preload(cid)
           }
 
-          callback(null, node)
+          callback(null, cid)
         })
       })
     }),
@@ -209,9 +209,11 @@ module.exports = function object (self) {
             return callback(err)
           }
 
-          self.object.get(cid, {
-            preload: options.preload
-          }, callback)
+          if (options.preload !== false) {
+            self._preload(cid)
+          }
+
+          callback(null, cid)
         })
       }
     }),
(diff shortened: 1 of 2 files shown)


alanshaw:
This PR is just waiting for an update to `ipfsd-ctl` and then it should be good to merge. It is ready for a review in the mean time.

img/overview.txt
```diff
@@ -1,17 +1,17 @@
...
+
+                     offline mode - uses IPFS core directly
+                  ┌────────────────────────────────────────────┐
+                  │                                            │
+                  │                                            │
+                  │ online mode - uses IPFS through http-api   │
+ ┌────────────┐   │                          ┌─────────────┐   │    ┌─────────┐
+ │            │   │ ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─    │             │   │    │         │
+ │    CLI     │───┴──  ipfs-http-client  ├──▶│  HTTP-API   │───┴───▶│IPFS Core│
+ │            │     └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─    │             │        │         │
```
daviddias: Makes so much more sense now! :D

PR #1733, opened 2018-11-28
fix: allow disabling mfs preload from config

achingbrain:
Use the `preload.enabled` config key to turn on/off periodic mfs preloading.

### DIFF
--- a/test/core/mfs-preload.spec.js
+++ b/test/core/mfs-preload.spec.js
@@ -74,8 +74,6 @@ describe('MFS preload', () => {
   })
 
   it('should disable preloading MFS', function (done) {
-    this.timeout(80 * 1000)
-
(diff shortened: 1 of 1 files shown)


test/core/mfs-preload.spec.js
```diff
@@ -56,4 +72,22 @@ describe('MFS preload', () => {
       })
     })
   })
+
+  it('should disable preloading MFS', function (done) {
+    this.timeout(80 * 1000)
```
alanshaw: Surely this test doesn't need a timeout adjustment?

achingbrain: Have removed it.

alanshaw:
Would you mind rebasing for green CI please?

achingbrain:
Done

PR #1737, opened 2018-11-28
chore: update ipfs-repo

hugomrdias:
update ipfs-repo but still pinned to be safe

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "ipfs-http-response": "~0.2.1",
     "ipfs-mfs": "~0.5.2",
     "ipfs-multipart": "~0.1.0",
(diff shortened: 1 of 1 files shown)


alanshaw:
> still pinned to be safe

Not sure I understand why? It was pinned because 0.25.1 was breaking builds if this is resolved we can unpin right?

hugomrdias:
Mainly to force a PR if we need to make more tweaks, as this issue doesn't happen always. In the first change we didn't do a PR and we only found the problem already in master.

alanshaw:
Would you mind rebasing for green CI please?

hugomrdias:
ci green

PR #1741, opened 2018-11-29
fix: ipns datastore key

vasco-santos:
When the `base32` lib was changed, the new one uses `lowercase`, which brokes the interop with `go-ipfs` for the `IPNS working locally`


alanshaw:
LGTM - I've restarted the build to see if goes green...

PR #1746, opened 2018-12-03
[WIP] feat: add errors with codes and utils

hugomrdias:
This PR adds a proposal to improve the error handling beyond using `err-code`, with these changes we create static errors backed by error classes and printf style formatted messages.
This approach is a simplified version of node's errors https://github.com/nodejs/node/blob/master/lib/internal/errors.js, i guess if it's good for node's code base it should also be good for us.

The benefits : 
- "standard" being based on node's version we can assume it works well in a large code base like ours
- errors are static we require them instead of just adding a string as a code property, this is less error prone
- messages are more stable because they are defined with the error and error code
- messages output is more flexible and advanced, it uses node util.format so things like objects and json use util.inspect for a nicer output.
- errors are sub-classes of native primitive error type this allows for more complex error matching
- we can match using error codes e.g. `err.code === ERR_IPFS_MISSING_IPLD_FORMAT.code`
- errors are directly associated with error messages, this gives the dev more context about the error use case which in turn will lead to a better error selection.

Please give feedback about this PR if we all agreed with this i'll change the rest of code to this pattern. //cc @ipfs/javascript-team

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "err-code": "^1.1.2",
     "file-type": "^10.2.0",
     "fnv1a": "^1.0.1",
+    "format-util": "^1.0.3",
     "fsm-event": "^2.1.0",
     "get-folder-size": "^2.0.0",
     "glob": "^7.1.3",
--- a/src/core/errors.js
+++ b/src/core/errors.js
@@ -78,7 +78,9 @@ function makeError (Base, key) {
 
 function E (sym, val, def) {
   messages.set(sym, val)
-  return makeError(def, sym)
+  const c = makeError(def, sym)
+  c.code = sym
+  return c
 }
 
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Error_types
--- a/src/core/errors.js
+++ b/src/core/errors.js
@@ -1,0 +1,89 @@
+'use strict'
+
+const format = require('format-util')
+const assert = require('assert')
+
+const messages = new Map()
+const { defineProperty } = Object
+
+/**
+ * Get formatted message for a error code
+ *
+ * @see https://nodejs.org/api/util.html#util_util_format_format_args
+ * @param {string} key - Error code
+ * @param {Array} args - Placeholder values for the message
+ * @returns {string} - Formatted message
+ */
+function getMessage (key, args) {
+  const msg = messages.get(key)
+
+  if (typeof msg === 'function') {
+    assert(
+      msg.length <= args.length, // Default options do not count.
+      `Code: ${key}; The provided arguments length (${args.length}) does not ` +
+        `match the required ones (${msg.length}).`
+    )
+    return msg.apply(null, args)
+  }
+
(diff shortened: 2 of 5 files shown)


vasco-santos:
I also agree with moving on with this approach for the errors. 

My biggest concern is that with so many modules, I don't think that the best approach is to have errors defined in all those modules and export them so that they can be used by the others. At least, we should have the more generic errors in their own module.

alanshaw:
I like this. We do need to deal with the case where we get an error from a dependency and want to "wrap it" to preserve the original (as @vmx noted). This proposal doesn't allow us to do that right now but `err-code` does.

@vmx - my concern with adding a `cause` property is that it gets hidden in browser console output:

## Chrome (cannot see `cause` in throw or log):
screenshot 2018-12-06 at 09 55 25

## Firefox (cannot see `cause` in throw or log):
screenshot 2018-12-06 at 09 55 39

## Node.js (it's ok, but we don't see the stack from wrapper...not entirely sure that matters though!):
screenshot 2018-12-06 at 09 55 58

@hugomrdias as a suggesstion, [`explain-error`](https://github.com/dominictarr/explain-error) actually preserves the error and might give you the basis for a proposal to allow us to support this...

Another idea, if @vmx, @vasco-santos and @jacobheun are on board with this we should pull the error creation code out into a module so that IPLD and libp2p can also use it.

vasco-santos:
> Another idea, if @vmx, @vasco-santos and @jacobheun are on board with this we should pull the error creation code out into a module so that IPLD and libp2p can also use it.

I am totally on this

vmx:
Another things I came across while working on some code. Let's assume you delete several items by CID, but that's not always successful, so you might get an error message. You would want to know on which CID it failed.

You can add this information to the error message text, but you won't have programmatic access to it.

So what about changing the `fprint` like formatting strings (where arguments are an array) to more Template Literal-like ones (where arguments are named).

Example:

```JS
// Instead of
new ERR_IPFS_MISSING_IPLD_FORMAT(codec)
// You would do 
new ERR_IPFS_MISSING_IPLD_FORMAT({ codec: codec })
// Or if you're fancy 
new ERR_IPFS_MISSING_IPLD_FORMAT({ codec })
```

The format string would be:

```JS
// Instead of
'Missing IPLD format %s'
// You'd be using
`Missing IPLD format ${args.codec}`
```

The advantage would be that the error object itself could have a field called e.g. `args`, `context` or `values` where you have access to the properties you passed in. This way you would have programmatic access to the custom things that caused the error.

AuHau:
Is there a plan to continue with this? I think that there is a big improvement space for error handling in `js` packages and this would be a great step forward.

I would also suggest having a one master table where all error codes would be collected from all `js` packages. Maybe together which packages can throw it? (but that could be hard to maintain)

alanshaw:
@hugomrdias do you plan to move forwards with this or shall we close?

hugomrdias:
i plan to include this in this quarter work

jacobheun:
Closing as stale

PR #1765, opened 2018-12-09
docs: add https://discuss.ipfs.io/c/announcements to release checklist

daviddias:
I believe this would be a valuable add, watcha think?


daviddias:
@jacobheun @vasco-santos @vmx wanna do the same for js-libp2p and js-ipld?

jacobheun:
👍 PR added for libp2p

PR #1770, opened 2018-12-10
docs: add code example for creating your own peer ID

alanshaw:


alanshaw:
Merging - this is just a docs change! I believe it to be correct (I copied the example on the Peer ID repo). Shout if you think this is not useful or incorrect!

PR #1772, opened 2018-12-10
refactor: modularise files

alanshaw:
This is basically a mechanical change to separate `files-regular.js` into multiple modules - easier to read, debug and maintain.


alanshaw:
Merging as test failure seems unrelated.

PR #1773, opened 2018-12-10
feat: add from url/stream

alanshaw:
~~depends on https://github.com/ipfs/js-ipfs/pull/1772~~

### DIFF
--- a/src/core/components/files-regular/add-from-url.js
+++ b/src/core/components/files-regular/add-from-url.js
@@ -1,0 +1,41 @@
+'use-strict'
+
+const { URL } = require('url')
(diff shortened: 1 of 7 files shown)


src/core/components/files-regular/add-from-url.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
+
+const { URL } = require('url')
+const fetch = require('../../runtime/fetch-nodejs')
+
+module.exports = (self) => {
+  return async (url, options, callback) => {
```
achingbrain: async methods that take callbacks. The async/await refactor cannot come soon enough.

PR #1777, opened 2018-12-12
feat: add from fs

### LINKED ISSUE #1625: Cannot add multiple files from command line
screenshot 2018-10-09 at 10 57 17

alanshaw:
This PR adds a new method [`addFromFs`](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#addfromfs) allowing users to more easily add files from their file system without having to specify every single file to add.

In the browser the user will receive a "not available" error.

I've pulled out a module `glob-source.js` - call it with some file paths and it returns a pull stream source that can be piped to `ipfs.addPullStream`.

This PR comes with the following added benefits:

* `ipfs add` on the CLI uses `glob-source.js` - **nice and DRY**
* `glob-source.js` uses the events that the `glob` module provides allowing the globbing to be a `pull-pushable`, which means that matched paths can begin to be added before all the globbing is done - **faster**
* `ipfs add` now supports adding multiple paths, fixes #1625 - **better**
* `ipfs add --progress=false` doesn't calculate the total size of the files to be added anymore! It didn't need to do that as the total was completely discarded when progress was disabled. It means we can add BIGGER directories without running into memory issues - **stronger**

---

* [x] depends on #1773


src/utils/files/glob-source.js
```diff
@@ -0,0 +1,111 @@
...
+  )
+}
+
+function normalizePathWithType (path, cb) {
+  parallel({
+    stat: cb => fs.stat(path, cb),
+    realpath: cb => fs.realpath(path, cb)
+  }, (err, res) => {
+    if (err) return cb(err)
+    cb(null, { path: res.realpath, type: res.stat.isDirectory() ? 'dir' : 'file' })
```
achingbrain: What happens if `path` is a socket, symbolic link, block device, etc?

alanshaw: Undefined. Needs tests! :D

src/utils/files/glob-source.js
```diff
@@ -0,0 +1,112 @@
+'use strict'
+
+const fs = require('fs')
+const Path = require('path')
+const isString = require('lodash/isString')
+const pull = require('pull-stream')
+const glob = require('glob')
```
hugomrdias: can't we just use https://www.npmjs.com/package/fast-glob ? it's faster and supports streams so we can go directly to pull-stream

src/core/runtime/add-from-fs-nodejs.js
```diff
@@ -0,0 +1,20 @@
...
+
+const promisify = require('promisify-es6')
+const pull = require('pull-stream')
+const globSource = require('../../utils/files/glob-source')
+const isString = require('lodash/isString')
+
+module.exports = self => {
+  return promisify((...args) => {
+    const callback = args.pop()
+    const options = isString(args[args.length - 1]) ? {} : args.pop()
```
hugomrdias: This should be enough we don't box strings 
```suggestion
    const options = typeof args[args.length - 1] == 'string' ? {} : args.pop()
```

Trying to avoid anything from lodash lots of unnecessary use of it that increase bundle size.

src/utils/files/glob-source.js
```diff
@@ -0,0 +1,112 @@
...
+* @param {String} ...paths File system path(s) to glob from
+* @param {Object} [options] Optional options
+* @param {Boolean} [options.recursive] Recursively glob all paths in directories
+* @param {Boolean} [options.hidden] Include .dot files in matched paths
+* @param {Array<String>} [options.ignore] Glob paths to ignore
+* @param {Boolean} [options.followSymlinks] follow symlinks
+* @returns {Function} pull stream source
+*/
+module.exports = (...args) => {
+  const options = isString(args[args.length - 1]) ? {} : args.pop()
```
hugomrdias: This should be enough we don't box strings 
```suggestion
  const options = typeof args[args.length - 1] === 'string' ? {} : args.pop()
```

Trying to avoid anything from lodash lots of unnecessary use of it that increase bundle size.

src/utils/files/glob-source.js
alanshaw: I tried it, and it's not significantly faster than `glob`, in fact in most tests it's slower (left is `glob`, right is `fast-glob`).

screenshot 2018-12-13 at 15 49 43

hugomrdias: even using streams instead of glob events ? seems weird but if its faster like this im good with it

alanshaw: I know weird, maybe it is the stream to pull stream conversion? Anyway even weirder is that the benchmarks for `fast-glob` do not run against `glob` 😱 They benchmark `globby`, `bash-glob`, `tiny-glob` and `glob-stream`...but not regular `glob` 🙄.

src/utils/files/glob-source.js
alanshaw: `fs.realpath` has been removed - it wasn't necessary.

Symlinks work fine for both files and directories. I've added some tests here: https://github.com/ipfs/interface-ipfs-core/pull/421 but I'm not sure if these will work on windows.

For a socket or block device, it essentially boils down to if `fs.createReadStream` can be used to read it then it can be added https://github.com/ipfs/js-ipfs/blob/4258046dadc357e87dd3ba5db445c269dde09c33/src/utils/files/glob-source.js#L55

I'm unsure how I'd test these edge cases right now...but it's worth noting the behaviour would be the [same as before](https://github.com/ipfs/js-ipfs/blob/6d46e2e9bf5d36dc7f8dcbc92b97347e193719d3/src/cli/commands/add.js#L79) (in the case of `ipfs add`) and there's currently no special provisions for these things in [`addFromFs` in `js-ipfs-http-client`](https://github.com/ipfs/js-ipfs-http-client/blob/master/src/utils/prepare-file.js).

Given that all the current tests are passing and CI is green, I'm feeling as though we should get this merged so that we can release 0.34 and deal with bug reports for sockets or block devices when they come in. Does that sound sensible?

alanshaw:
@hugomrdias is this gtg now?

Issue #1721, opened 2018-11-19
⚡️ v0.34.0 RELEASE 🚀


alanshaw:
> Ahh yeah, it got fast

# 🗺 What's left for release?

* [x] [Refactor files API](https://github.com/ipfs/js-ipfs/pull/1671)
* [x] [Upgrade to latest ipld-dag-pb](https://github.com/ipfs/js-ipfs/pull/1702)
* [x] [Refactor Object API](https://github.com/ipfs/js-ipfs/pull/1730)
* [x] [`--cid-base` option](https://github.com/ipfs/js-ipfs/pull/1552)
* [x] [CID version agnostic get](https://github.com/ipfs/js-ipfs-repo/pull/184) and [#1757](https://github.com/ipfs/js-ipfs/pull/1757)
* ~~[DHT](https://github.com/ipfs/js-ipfs/pull/856)~~ bumped to next release
* [x] [IPNS over pubsub](https://github.com/ipfs/js-ipfs/pull/1559)
* [x] [IPNS over DHT](https://github.com/ipfs/js-ipfs/pull/1725)
* [x] [`addFromURL`](https://github.com/ipfs/js-ipfs/pull/1773), [`addFromStream`](https://github.com/ipfs/js-ipfs/pull/1773)
* [x] [`addFromFs`](https://github.com/ipfs/js-ipfs/pull/1777)
* [x] [Support for HAMT directories in MFS](https://github.com/ipfs/js-ipfs/pull/1734)

# 🔦 Highlights

## 🏎 Reading Protobuf DAG nodes is super fast

Reading Protobuf DAG nodes just got a serious speed boost as well as a memory reduction. Previously Protobuf DAG nodes (dag-pb nodes for short) carried a multihash property (a CIDv0) with them. This used up loads of CPU time calculating the hash for a given node as it was retrieved from storage and with the advent of CIDv1 there was no guarantee the hashing algorithm and length were correct.

So, we removed it! 🦖

Ok so that had the unfortunate consequence of making the object API a whole lot less useful when writing data - the DAG nodes you got back were basically the data you put in. Hence the object API was refactored to return CIDs instead of DAG nodes for write operations...and we all lived happily ever after.

## 🚤 Adding content is blazingly fast

We found a big bottleneck (or should I say small bottleneck?) that meant the bigger the file you added to IPFS, the longer it would take - oh no 😭! Good news folks, we've seen the speed of adding large files increase by 50% or more in some cases! 💥

[...]

PR #1791, opened 2018-12-21
node.add -> node.files.add

schollz:
I found this helpful as there seems to have been an API change.


schollz:
@vasco-santos Okay, I'm going to try - let me know if I am getting it wrong

alanshaw:
@schollz the examples are correct for the code in master that will be released in 0.34.

More info on the breaking change here https://github.com/ipfs/js-ipfs/issues/1721

PR #1793, opened 2018-12-22
feat: use ws-star-multi instead of ws-star

### LINKED ISSUE #1619: inconsistent node startup behavior depending on attempted ws-star swarm binding
- **Version**: 0.32.3
- **Platform**: node
- **Subsystem**:  js-libp2p-websockets

#### Description: 

In the browser, we generally have to use websocket-star swarm addresses (since binding TCP is not available), however these addresses are temperamental and depend on the signaling server being available. Node startup will abort if the specified swarm address cannot be bound:

```
>  const options = {
...    "config": {
.....      "Addresses": {
.......        "Swarm": [
.......                 "/dns4/example.com/tcp/443/wss/p2p-websocket-star/" // bad
.......        ]
.......      }
.....    },
...    "EXPERIMENTAL": {
.....      "pubsub": true
.....    }
...  }
undefined
> const node = new IPFS(options)
undefined
> Error: websocket error
    at WS.Transport.onError (/home/arkadiy/tmp/dweb-transport/node_modules/engine.io-client...
> node.isOnline()
false
```

or even if *ANY* ws failures occur before an address could be bound:

```
>  const options = {
...    "config": {
.....      "Addresses": {
.......        "Swarm": [
.......                 "/dns4/example.com/tcp/443/wss/p2p-websocket-star/", // bad
.......                 "/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star/" // good
.......
.......        ]
.......      }
.....    },
...    "EXPERIMENTAL": {
.....      "pubsub": true
.....    }
...  }
undefined
>
> const node = new IPFS(options)
undefined
> Error: websocket error
    at WS.Transport.onError (/home/arkadiy/tmp/dweb-transport/node_modules/engine.io-client/lib/transport.js:64:13)
    at WebSocket.ws.onerror (/home/arkadiy/tmp/dweb-transport/node_modules/engine.io-client/lib/transports/websocket.js:150:10)
    at WebSocket.onError (/home/arkadiy/tmp/dweb-transport/node_modules/engine.io-client/node_modules/ws/lib/EventTarget.js:109:16)
    at WebSocket.emit (events.js:182:13)
    at WebSocket.EventEmitter.emit (domain.js:442:20)
    at WebSocket.finalize (/home/arkadiy/tmp/dweb-transport/node_modules/engine.io-client/node_modules/ws/lib/WebSocket.js:182:41)
    at ClientRequest._req.on (/home/arkadiy/tmp/dweb-transport/node_modules/engine.io-client/node_modules/ws/lib/WebSocket.js:653:12)
    at ClientRequest.emit (events.js:182:13)
    at ClientRequest.EventEmitter.emit (domain.js:442:20)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:546:21)
> node.isOnline()
false
```

however, if we succeed before we fail, the node starts up fine(-ish) and even lies about what addresses it's listening on:

```
>  const options = {
...    "config": {
.....      "Addresses": {
.......        "Swarm": [
.......                 "/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star/", // good
.......                 "/dns4/example.com/tcp/443/wss/p2p-websocket-star/", // bad
.......
.......        ]
.......      }
.....    },
...    "EXPERIMENTAL": {
.....      "pubsub": true
.....    }
...  }
undefined
> const node = new IPFS(options)
undefined
> Swarm listening on /dns4/example.com/tcp/443/wss/p2p-websocket-star/ipfs/QmSmwDi3AmMm3pFbyvzmRZ3FfLtNAtYv5ie7ispER1kGUB // are you really?
Swarm listening on /dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star/ipfs/QmSmwDi3AmMm3pFbyvzmRZ3FfLtNAtYv5ie7ispER1kGUB

> node.isOnline()
true
> node.swarm.addrs(a => console.log(a))
null
```

finally, if we specify an empty swarm address set the node happily goes online:
```
>  const options = {
...    "config": {
.....      "Addresses": {
.......        "Swarm": [
.......        ]
.......      }
.....    },
...    "EXPERIMENTAL": {
.....      "pubsub": true
.....    }
...  }
undefined
> const node = new IPFS(options)
undefined
> node.isOnline()
true
```

(the latter is useful if you want to dial another node directly, for example)

there seem to be at least two problems here:
- assumptions about node startup success based on TCP (where failure to bind a port is, reasonably, a fatal error) do not apply in more fragile environments like ws relay
- *no* swarm addresses and *no successfully bound* swarm addresses are not equivalent
- logic with multiple ws-star addresses is broken

some related discussion in https://github.com/ipfs-shipyard/peer-pad-core/issues/13, though the fix is probably at js-ipfs level

mkg20001:
This replaces ws-star with ws-star-multi transparently

resolves #1619

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,6 +140,7 @@
     "libp2p-tcp": "~0.13.0",
     "libp2p-webrtc-star": "~0.15.5",
     "libp2p-websocket-star": "~0.10.0",
+    "libp2p-websocket-star-multi": "^0.4.0",
     "libp2p-websockets": "~0.12.0",
     "lodash": "^4.17.11",
     "mafmt": "^6.0.2",
--- a/package.json
+++ b/package.json
@@ -139,7 +139,6 @@
     "libp2p-secio": "~0.11.0",
     "libp2p-tcp": "~0.13.0",
     "libp2p-webrtc-star": "~0.15.5",
-    "libp2p-websocket-star": "~0.10.0",
     "libp2p-websocket-star-multi": "~0.4.0",
     "libp2p-websockets": "~0.12.0",
     "lodash": "^4.17.11",
--- a/package.json
+++ b/package.json
@@ -140,7 +140,7 @@
     "libp2p-tcp": "~0.13.0",
     "libp2p-webrtc-star": "~0.15.5",
     "libp2p-websocket-star": "~0.10.0",
-    "libp2p-websocket-star-multi": "^0.4.0",
+    "libp2p-websocket-star-multi": "~0.4.0",
     "libp2p-websockets": "~0.12.0",
     "lodash": "^4.17.11",
     "mafmt": "^6.0.2",
--- a/src/core/runtime/libp2p-browser.js
+++ b/src/core/runtime/libp2p-browser.js
@@ -2,17 +2,21 @@
 
 const WS = require('libp2p-websockets')
 const WebRTCStar = require('libp2p-webrtc-star')
-const WebSocketStar = require('libp2p-websocket-star')
+const WebSocketStarMulti = require('libp2p-websocket-star-multi')
 const Multiplex = require('libp2p-mplex')
 const SECIO = require('libp2p-secio')
 const Bootstrap = require('libp2p-bootstrap')
 const libp2p = require('libp2p')
 const defaultsDeep = require('@nodeutils/defaults-deep')
+const multiaddr = require('multiaddr')
 
 class Node extends libp2p {
   constructor (_options) {
     const wrtcstar = new WebRTCStar({ id: _options.peerInfo.id })
-    const wsstar = new WebSocketStar({ id: _options.peerInfo.id })
+
+    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => ~addr.indexOf('p2p-websocket-star'))
+    _options.peerInfo.multiaddrs.replace(wsstarServers.map(multiaddr), '/p2p-websocket-star') // the ws-star-multi module will replace this with the chosen ws-star servers
+    const wsstar = new WebSocketStarMulti({ servers: wsstarServers, id: _options.peerInfo.id, ignore_no_online: !wsstarServers.length || _options.wsStarIgnoreErrors })
 
     const defaults = {
       modules: {
--- a/src/core/runtime/libp2p-browser.js
+++ b/src/core/runtime/libp2p-browser.js
@@ -14,7 +14,7 @@ class Node extends libp2p {
   constructor (_options) {
     const wrtcstar = new WebRTCStar({ id: _options.peerInfo.id })
 
-    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => ~addr.indexOf('p2p-websocket-star'))
+    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => addr.includes('p2p-websocket-star'))
     _options.peerInfo.multiaddrs.replace(wsstarServers.map(multiaddr), '/p2p-websocket-star') // the ws-star-multi module will replace this with the chosen ws-star servers
     const wsstar = new WebSocketStarMulti({ servers: wsstarServers, id: _options.peerInfo.id, ignore_no_online: !wsstarServers.length || _options.wsStarIgnoreErrors })
 
--- a/src/core/runtime/libp2p-browser.js
+++ b/src/core/runtime/libp2p-browser.js
@@ -14,6 +14,8 @@ class Node extends libp2p {
   constructor (_options) {
     const wrtcstar = new WebRTCStar({ id: _options.peerInfo.id })
 
+    // this can be replaced once optional listening is supported with the below code. ref: https://github.com/libp2p/interface-transport/issues/41
+    // const wsstar = new WebSocketStar({ id: _options.peerInfo.id })
     const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => addr.includes('p2p-websocket-star'))
     _options.peerInfo.multiaddrs.replace(wsstarServers.map(multiaddr), '/p2p-websocket-star') // the ws-star-multi module will replace this with the chosen ws-star servers
     const wsstar = new WebSocketStarMulti({ servers: wsstarServers, id: _options.peerInfo.id, ignore_no_online: !wsstarServers.length || _options.wsStarIgnoreErrors })
--- a/src/core/runtime/libp2p-nodejs.js
+++ b/src/core/runtime/libp2p-nodejs.js
@@ -14,7 +14,7 @@ const multiaddr = require('multiaddr')
 
 class Node extends libp2p {
   constructor (_options) {
-    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => ~addr.indexOf('p2p-websocket-star'))
+    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => addr.includes('p2p-websocket-star'))
     _options.peerInfo.multiaddrs.replace(wsstarServers.map(multiaddr), '/p2p-websocket-star') // the ws-star-multi module will replace this with the chosen ws-star servers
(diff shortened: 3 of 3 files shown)


alanshaw:
@mkg20001 please can I get some context? We can already pass multiple discovery/transports to libp2p, what are the benefits of using this module?

mkg20001:
@alanshaw It allows passing multiple ws-star servers in a way that makes the node only fail to start if **all** servers are down (or optionally never at all), as compared to previous behavior where the app would fail if **any** server was down

lidel:
@alanshaw IIUC this is a fix for https://github.com/ipfs-shipyard/ipfs-share-files/issues/63#issuecomment-444565592 (cc @olizilla @fsdiogo) and https://github.com/ipfs/js-ipfs/issues/1619#issuecomment-449373215  (cc @satazor)

parkan:
is this superseded by https://github.com/libp2p/js-libp2p-websocket-star/issues/70 ?

mkg20001:
@parkan It depends. This is a fix that could be released _now_. Stardust still needs to be polished which could take a week or two.

alanshaw:
> @alanshaw It allows passing multiple ws-star servers in a way that makes the node only fail to start if **all** servers are down (or optionally **never** at all), as compared to previous behavior where the app would fail if **any** server was down

Yes, that is preferable! Ok so, the other thing I don't understand is why this is implemented as a separate module and not as a fix to `libp2p` or `libp2p-websocket-star`?

mkg20001:
Because that would only allow for ignoring _any_ down server or _none_
This also allows to ignore _some_ and throw if _all_ are down

This module "listens" on `/p2p-websocket-star` which internally makes it start listen on all the different server, ignoring the offline ones.

Later it swaps `/p2p-websocket-star` for the actual addresses (which is actually done by libp2p in the same way `/tcp/0` gets replaced by the actual port)

mkg20001:
> I can see how that's true if we were to "fix" libp2p-websocket-star, but isn't this behaviour desirable for all libp2p transports? Why not fix it in libp2p?

Libp2p has no concept of an "optional listening address",so that would need to get invented first. Additionally there is this problem that not listening on TCP for ex. disables the ability to dial `/tcp` addrs, even though TCP still could work without listening on any address. So if tcp silently fails the whole app suddenly is unable to dial anything.

jacobheun:
We've had some discussion that would allow us to support optional listening addresses in libp2p, and I think that it's reasonable for us to get this implemented relatively soon. The idea relates to https://github.com/libp2p/interface-transport/issues/41, which would support an array of listening addresses. 

The idea being, that we could pass additional options to the switch, and subsequently to the transports to determine our level of tolerance on listening. So if configured to be very tolerant, we could end up failing to listening to any addresses but still continue to run.

Some additional context (edited for ease of reading) from the comments around this discussion:

> Maybe the transport should have an option for retrying with a basic backoff and jitter. If retrying is not specified and we have 0 servers then an error is passed back. Otherwise, if we fail to listen on any addresses we will do a graceful backoff to retry listening. This could be repeated in the event the server goes down while we're listening.
We could make the retry logic on by default instead of having to specify it.

>I think this might be a good option to pass to the switch and let it handle errors passed back by the listeners. We could also set the level of fatal, (no fatal errors, fatal if all transports failed to listen, fatal if one transport failed to listen), so users could get further control. We could do "fatal if all transports failed to listen" by default.
If the transport has retry turned on, it would never return an error, so the switch options wouldn't matter in this case. But with other transports in use that becomes a lot more useful.

parkan:
@alanshaw @jacobheun the idea of supporting this across a variety of transports is sound, though the semantics of failing to bind a port are pretty different for ws-star vs TCP so the behavior would need to be configurable

however, the issue described in #1619 (and fixed in this PR) is a **showstopper bug** that prevents multiple prospective users from shipping otherwise working js-ipfs code, and I am having a very hard time explaining to them why we haven't put a fix in place (especially given that we've had this PR up for almost a month) -- I would say this is among the **top 3 blockers to js-ipfs adoption**, and by far the easiest of those to fix

I could advise them to inject ws-star-multi themselves, but I am concerned that it will again drift out of sync with js-ipfs core and break in a subsequent update

would it be possible to ship this as is, with a note to re-evaluate once we have a comprehensive, configurable solution across libp2p?

package.json
```diff
@@ -140,6 +140,7 @@
     "libp2p-tcp": "~0.13.0",
     "libp2p-webrtc-star": "~0.15.5",
     "libp2p-websocket-star": "~0.10.0",
```
alanshaw: Can `libp2p-websocket-star` be removed?

package.json
```diff
@@ -140,6 +140,7 @@
...
+    "libp2p-websocket-star-multi": "^0.4.0",
```
alanshaw: Please use ~ for pre 1.0 dependencies.

src/core/runtime/libp2p-browser.js
```diff
@@ -2,17 +2,21 @@
...
 const libp2p = require('libp2p')
 const defaultsDeep = require('@nodeutils/defaults-deep')
+const multiaddr = require('multiaddr')
 
 class Node extends libp2p {
   constructor (_options) {
     const wrtcstar = new WebRTCStar({ id: _options.peerInfo.id })
-    const wsstar = new WebSocketStar({ id: _options.peerInfo.id })
+
+    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => ~addr.indexOf('p2p-websocket-star'))
```
alanshaw:
```suggestion
    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => addr.includes('p2p-websocket-star'))
```

src/core/runtime/libp2p-nodejs.js
```diff
@@ -3,17 +3,20 @@
...
 const Multiplex = require('libp2p-mplex')
 const SECIO = require('libp2p-secio')
 const libp2p = require('libp2p')
 const defaultsDeep = require('@nodeutils/defaults-deep')
+const multiaddr = require('multiaddr')
 
 class Node extends libp2p {
   constructor (_options) {
-    const wsstar = new WebSocketStar({ id: _options.peerInfo.id })
+    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => ~addr.indexOf('p2p-websocket-star'))
```
alanshaw:
```suggestion
    const wsstarServers = _options.peerInfo.multiaddrs.toArray().map(String).filter(addr => addr.includes('p2p-websocket-star'))
```

package.json
mkg20001: Depends. This forces ws-star-multi to use specifically that version because ws-star-multi has a `^` requirement instead of a `~`. So I'd leave it here just to make the version more specific

mkg20001:
Btw, should I move ws-star-multi to the libp2p org?

parkan:
thank you @alanshaw + @mkg20001 🙏

package.json
alanshaw: I think ~ and ^ have the same semantics for pre 1.0 dependencies.

alanshaw: > Um, but why [#1793 (comment)](https://github.com/ipfs/js-ipfs/pull/1793#discussion_r249005091) then, though?

It's in our commit linter for CI and is in our [JS code guidelines](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#dependency-versions).

> AFAIK `~` matches same-minor and same-major, while `~` matches same-major only.

`^0.2.3 := >=0.2.3 <0.3.0` from https://www.npmjs.com/package/semver#caret-ranges-123-025-004 - `^` has a different meaning for 0.x versions (which just so happens to be the same as `~`).

i.e. unless there's another reason to keep it, it can be removed - it'll have no bearing on what version `libp2p-websocket-star-multi` uses as they have equivalent version ranges.

mkg20001: > AFAIK ~ matches same-minor and same-major, while ~ matches same-major only.

:man_facepalming: one of them should be `^`, but seems wrong anyways

> i.e. unless there's another reason to keep it, it can be removed - it'll have no bearing on what version libp2p-websocket-star-multi uses as they have equivalent version ranges.

Agree

alanshaw:
Merging as windows CI failure is unrelated and fixed by https://github.com/ipfs/js-ipfs/pull/1834

PR #1803, opened 2019-01-03
chore: update deps

alanshaw:


package.json
```diff
@@ -62,7 +62,7 @@
   },
   "homepage": "https://github.com/ipfs/js-ipfs#readme",
   "devDependencies": {
-    "aegir": "^17.0.1",
+    "aegir": "^17.1.1",
```
vasco-santos: you can update `aegir` now to `18.0.3`

jacobheun:
@alanshaw the test failures with handling control signals is due to Event Emitter Warnings being registered as test errors. Without resolving the underlying emitter warnings, we could either pass `NODE_NO_WARNINGS=1` to the execa call for the test suite, or do something along the lines of https://github.com/ipfs/js-ipfs/commit/f894a3ff89d4c4d75be4b28ce186a5984e49d452, to log out the warnings and continue with the test so we at least are getting visibility to those issues. 

Since code didn't change before this started occurring it's likely due to a patch dependency update adding another listener somewhere.

alanshaw:
@jacobheun it would be awesome if you have a moment to eyeball https://github.com/ipfs/js-ipfs/pull/1803/commits/60dc0db47ff151436345785d39d7d6ccb8236d18 - it's fallout from https://github.com/libp2p/js-libp2p-bootstrap/pull/86

jacobheun:
@alanshaw the fix looks reasonable, nice to see we're discovering peers so quickly :).  Moving the event listener outside `peer:discovery` is definitely an improvement.

package.json
alanshaw: I'm going to hold off on that one for this release and do that upgrade in 0.35

alanshaw:
I'm going to get this merged as it'll fix master. Windows CI failure is known occasional issue "Database is not open" - unrelated to the changes here.

PR #1805, opened 2019-01-05
Tutorial README namespace fix.

ug02fast:
It's out of date.


alanshaw:
@ug02fast thanks for the contribution but that is actually the old API, it is correct for js-ipfs 0.34.

Issue #1767, opened 2018-12-09
Sunset WebSockets Star


daviddias:
With DHT and Circuit Relay coming to their full potential in v0.34.0 https://github.com/ipfs/js-ipfs/issues/1721, I want to propose the sunset of the WebSockets Star transport in favor of the power combo DHT Peer Routing + Circuit Relay.

This issue serves the purpose to gather thoughts and reach Consensus, plus outline the migration path necessary (tests, what's missing to do, etc).

pgte:
To me websocket-star provides not only peer routing and relay, but it also peer discovery.

For instance, In peer-star, we are using the transport discovery events to gather knowledge about the existence of peers. After we discover a new peer, we ask it if they're interested in participating in the app, etc...

I think we need something like [the rendezvous protocol](https://github.com/libp2p/specs/pull/56) implemented and some nodes that implement it running so that we're able to do service discovery, which is something I think that DHT Peer Routing + Circuit relay does not provide us.

@daviddias does this make sense to you?

daviddias:
@pgte it does, thank you for bringing up the Rendezvous Protocol. Where can I track it's progress, blockers and ETA?

pgte:
@daviddias besides that spec, I'm not sure. I wish they had an "Awesome Endeavour" so we could track this on one thread...

daviddias:
@mkg20001 the focus will be to continue investing on libp2p solutions that don't rely on external and inflexible central points. That is, Rendezvous, Circuit Relay and full Peer Routing adopted by default.

mkg20001:
@daviddias And when will they be ready to get shipped? This really just needs a few more days to be ready. Rendezvous and other solutions don't seem to be really progressing in terms of finalization.

mkg20001:
Aside from that there are things I want to mention: 

[...]

PR #1812, opened 2019-01-12
[WIP] feat: add stardust behind flag "EXPERIMENTAL.stardust"

mkg20001:
This adds support for the stardust protocol behind the flag 
EXPERIMENTAL.stardust

This pr is part of the endeavour to replace ws-star with stardust for 
the time being until a better solution is found.

See the discussion at 
https://github.com/libp2p/js-libp2p-websocket-star/issues/70#issuecomment-453601740 
for more information

Ref #1767


mkg20001:
Yes. https://github.com/libp2p/js-libp2p-crypto/pull/125 (a pr where I need some help with adding the `.encrypt()` method with webcrypto) https://github.com/multiformats/js-multiaddr/pull/78 (needs merge&release) https://github.com/multiformats/js-mafmt/pull/35 (needs merge&release, depends on previous pr)

alanshaw:
Unless I'm missing something, we don't need an experimental flag for this since you can enable it through constructor config already. I think we should instead add a stardust [example](https://github.com/ipfs/js-ipfs/tree/master/examples) and mention it in the README as an alternative transport that people can use.

PR #1819, opened 2019-01-15
[WIP] feat: default preload false in ci

### LINKED ISSUE #1815: Disable preloading in test environments
For all tests other than the ones testing preloading we should disable preloading. We can detect a CI environment variable and default the preload option to false if not explicitly set.

alanshaw:
BREAKING CHANGE: Content preloading on preload nodes is now _disabled by default_ in test and CI environments. This can be overridden by explicitly enabling preloading in test/CI by setting  `preload.enabled` to `true` in the IPFS constructor args.

resolves #1815

cc @eefahy


achingbrain:
Couldn't we configure this as part of the build environment or during the test suite setup instead of changing the application code to have test-specific code paths?

alanshaw:
We could, but this also mitigates against preload happening when run in CI by people using js-ipfs in their applications.

alanshaw:
> The users of this module will have to do zero changes to their codebase

In production, no, but if users have integration tests that rely on two nodes exchanging content via the preload nodes then they'll need to explicitly set `preload` to `true`.

What is "breaking" anyway? 😆

alanshaw:
...need to fix the browser issues before this can be merged

PR #1830, opened 2019-01-18
[WIP] feat: load IPLD formats lazily from IPFS

alanshaw:


src/core/components/files-regular/cat-pull-stream.js
```diff
@@ -16,7 +16,8 @@ module.exports = function (self) {
     ipfsPath = normalizePath(ipfsPath)
     const pathComponents = ipfsPath.split('/')
     const restPath = normalizePath(pathComponents.slice(1).join('/'))
-    const filterFile = (file) => (restPath && file.path === restPath) || (file.path === ipfsPath)
+    const fileName = restPath.includes('/') ? restPath.split('/').pop() : restPath
+    const filterFile = (file) => (restPath && file.path === restPath) || (file.path === ipfsPath) || file.path === fileName
```
alanshaw: ooh, nte to self I think this a bug in JS IPFS that needs ot be fixed in a separate PR

lidel:
I feel this is the future :artificial_satellite:, but like others mentioned, really rough around the edges today.

Some thoughts about raised issues: 

- **can we make sure this ipfs.cat will always work?**: I assume pinning assets will be part of release dance. Still, if cat fails for some reason, we could fallback to HTTP by loading library from our public gateway or URL specified by developer.
- **devs no longer own their production env:** valid concern, but if someone cares about full control it may be enough to have a mechanism to opt-out from lazy-loading by providing a hardcoded list of ipld-formats. 
- **external sourcemap:** would `sourceMappingURL` pointing at `.map` at our public gateway do the trick?
- **ensuring versions from [`ipld-formats-browser.js`](https://github.com/ipfs/js-ipfs/blob/974e60f9ff8c154781830e25de269c1219320836/src/core/runtime/ipld-formats-browser.js)  are in sync with `package.json`:** this sounds like something a build script could validate: look at `browser` section of `package.json`  and throw an error if the right side does not match versions currently resolving for the left side. 
  - To make this less cumbersome: perhaps code could be left untouched, but the logic responsible for lazy-loading from IPFS moved to aegir or a webpack plugin? It could detect browser build and transparently replace first calls to specific imports with lazy-load logic, and adding them to IPFS behind the scenes. I feel if we don't reach that level of automation, maintenance cost may be too high.

src/core/components/files-regular/cat-pull-stream.js
alanshaw: https://github.com/ipfs/js-ipfs/pull/1920

PR #1831, opened 2019-01-18
docs: fix browser video streaming example

alanshaw:
https://github.com/moshisushi/hlsjs-ipfs-loader/pull/7 was merged and released so this can now be fixed.


moshisushi:
I think the demo linked to in the "Putting it all together" section should be removed as well, since it uses the old version of the library.

Issue #1664, opened 2018-10-24
ipfs.libp2p should be the ipfs._libp2pNode


daviddias:
It is confusing for users that libp2p is not available through the .libp2p key and instead it is on ipfs._libp2pNode. Anyone opposed to this change?

anorth:
I looked at this as an easy first change, but was thrown by the existing assignment to `.libp2p`: https://github.com/ipfs/js-ipfs/blob/d215502b915c43c3ffe4f6475b38aaa63ec3d7f8/src/core/index.js#L150

Is this explicitly requesting to move the component reference out of the way to a different key and punch through to this internal node reference?
https://github.com/ipfs/js-ipfs/blob/d215502b915c43c3ffe4f6475b38aaa63ec3d7f8/src/core/components/libp2p.js#L72

alanshaw:
@anorth what's the status on this? I see you've made some progress. Do you have the time to bring it over the line? 😜 

If not then it would be ideal if you could submit a PR that someone else can finish off.

anorth:
Sorry I didn't bring it home–I switched away after struggling with tests. I look forward to finding some other low-hanging changes as I get to know the project.

PR #1832, opened 2019-01-18
feat: make ipfs.libp2p the libp2p node rather than component

anorth:
This is an attempt on #1664–see discussion there. It may not be complete as I had trouble running unit tests (at master). I am posting this by way of abandoning it and welcome someone taking it over to finish it off.

I'm not sure this is 100% the right approach; in particular the `this.libp2p = undefined` in `core/index.js` is different to surrounding code. I don't know the project well enough to suggest an improvement here.

resolves https://github.com/ipfs/js-ipfs/issues/1664

### DIFF
--- a/src/core/components/dht.js
+++ b/src/core/components/dht.js
@@ -30,7 +30,7 @@ module.exports = (self) => {
 
       options = options || {}
 
-      self._libp2pNode.dht.get(key, options.timeout, callback)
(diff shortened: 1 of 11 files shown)


alanshaw:
This looks good! The windows test failure is unrelated. I'll review this asap. Thanks again.

Issue #1670, opened 2018-10-25
Awesome Endeavour: Async Iterators


alanshaw:
JS IPFS supports two types of stream at the API level, but uses pull streams for internals. If I was working on js-ipfs at the time I'd have made the same decision. Since then, async/await became part of the JS language and the majority of JavaScript runtimes now support async/await, async iterators and for/await/of (i.e. no need to transpile). These tools give us the power to stream data without needing to rely on a library.

Just because there are new language features available doesn't mean we should switch to using them. It's a significant upheaval to change the core interface spec and its implementations (js-ipfs, js-ipfs-api etc.) without good reason.

That said, it has become apparent that there are a growing number of good reasons to do this:

* Reduction in bundle size - no need to bundle two different stream implementations, and their eco-system helper modules, no need for the `async` module.
* Reduce `npm install` time - fewer dependencies to install.
* Allows us to remove a bunch of plumbing code that converts Node.js streams to pull streams and vice versa.
* Reduces API surface area, no `addPullStream`, `addReadableStream`.
* Building an `interface-ipfs-core` compatible interface becomes a whole lot easier, no dual promise/callback API and no multiple stream implementation variations of the same function. It would also reduce the number of tests in the `interface-ipfs-core` test suite for the same reasons.
* [Node.js readable streams are now async iterators](http://2ality.com/2018/04/async-iter-nodejs.html) thanks to [#17755](https://github.com/nodejs/node/pull/17755)!
* Of note, it is trivial to convert from [pull stream to (async) iterator](https://github.com/alanshaw/pull-stream-to-async-iterator) and [vice versa](https://github.com/alanshaw/async-iterator-to-pull-stream).
* Unhandled throws that cannot be caught will no longer be a problem
* Better stack traces, stacks no longer clipped at async boundaries, [`await` stack traces better than promise stack traces](https://mathiasbynens.be/notes/async-stack-traces)
* A modern, up to date and cutting edge API will aid community contributions and adoption.

The rough plan is:

1. Drop support for dual callback/promise based APIs
1. Expose only APIs that return promises or iterators for async actions
1. Use async/await over then/catch when dealing with promises

This will require significant discussion and coordination from the JS teams. We'll need to reach agreement on the best API to expose for each module and manage releases carefully.

[...]

PR #1834, opened 2019-01-21
fix: race condition causing Database is not open error

alanshaw:
[`cleanup`](https://github.com/ipfs/js-ipfs/blob/7d9b006d0b0542651cbaa540d5f22a0112ae09bd/src/cli/bin.js#L109) closes the DB but `yargs-promise` does not wait for async stuff in the command handler to finish, and executes that promise chain immediately after the handler is executed. So it's a race condition. In windows, _sometimes_, the database is closed before the async stuff from the handler completes.

This PR changes the CLI command handlers to always pass a promise to `resolve` function that `yargs-promise` adds to the context (`argv`). This makes `yargs-promise` wait for it to be resolved before continuing the promise chain and closing the DB.

Since I had to edit all of the commands to make them use the `resolve` function and introduce promises, I decided to take the opportunity to refactor the CLI commands to use async/await. It should help towards https://github.com/ipfs/js-ipfs/issues/1670.

Also, w00t! - removed almost 200 LOC

### DIFF
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -59,7 +59,11 @@ if (args[0] === 'daemon' || args[0] === 'init') {
     .completion()
     .command(require('./commands/daemon'))
     .command(require('./commands/init'))
-    .parse(args)
+
+  new YargsPromise(cli).parse(args)
+    .then(({ data }) => {
+      if (data) print(data)
+    })
 } else {
   // here we have to make a separate yargs instance with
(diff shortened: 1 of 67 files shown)


vasco-santos:
I will update the DHT CLI for [ipfs/js-ipfs#856](https://github.com/ipfs/js-ipfs/pull/856) according to the changes here!

PR #1840, opened 2019-01-23
feat: exit command line tool on unhandled Promise rejections

vmx:
Currently the code base is slowly switching from callback style to
async/await. This means that there are more cases where there are
unhandled Promise rejections, sometime they even can't easilt be
prevented.

The command line interface now exits just like there was a normal
exception.

### DIFF
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -17,6 +17,12 @@ updateNotifier({
   updateCheckInterval: 1000 * 60 * 60 * 24 * 7 // 1 week
 }).notify()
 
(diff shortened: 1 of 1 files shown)


vmx:
I did, but this change https://github.com/ipfs/js-ipfs/commit/6066c97d08cdacb9f7f7034f8565c687f2610810#diff-d231953efc1e3b19e5f18bf32ebe8f92 fixed it.

My problem was that the `ipfs.object.get()` call used a callback, but internally I made a Promise-based call. The subsequent `throw err` then lead to an unhandled Promise rejection.

PR #1844, opened 2019-01-24
refactor: migrate http api to use hapi 18

alanshaw:
**DEAR REVIEWERS, SORRY THAT THIS PR IS HUGE 😞**

---

This PR refactors the HTTP API for use with Hapi 18.

The current code works with Hapi 16. This is now two major versions behind the current version and we need to update our code to keep up to date with the latest bug/security fixes and performance improvements.

Hapi 17 introduced significant breaking changes, along with a switch from a callback API to a promises API hence the code changes to support this are also significant.

This PR includes:
* A switch to using `async`/`await` with `try`/`catch` over `.then`/`.catch` for easier to understand and more succinct code - Hapi 17+ uses this syntax and it's supported in Node.js since version 8
* Numerous fixes where API was returning a string error message - now returns a proper JSON object
* Numerous fixes where API was returning 500 for a client error - now returns 400
* Makes full use of the error handler, instead of repeating code to return an error response with a particular code, it instead throws errors and uses `boom` to throw errors with specific HTTP status codes e.g.
    ```js
    // Replaces this:
    return reply({
      Message: "Argument 'arg' is required",
      Code: 0
    }).code(400)

    // ...with this:
    throw Boom.badRequest("Argument 'arg' is required")
    ```
* Perf improvement to `/api/v0/add` - response is no longer buffered in memory, instead it is streamed to the client

refs https://github.com/ipfs/js-ipfs/issues/1670

* [x] depends on https://github.com/ipfs/js-ipfs-mfs/pull/36


hugomrdias:
// Replaces this:
return reply({
  Message: "Argument 'arg' is required",
  Code: 0
}).code(400)

// ...with this:
throw Boom.badRequest("Argument 'arg' is required")

i think the reply was like that for Go interop, am i wrong ?

alanshaw:
> i think the reply was like that for Go interop, am i wrong ?

The error handler catches the error and transforms it into a `{ Message, Code, Type }` JSON response so we're just removing the boilerplate.

alanshaw:
Slightly more urgent now - Hapi 16 depends on `cryptiles@3.1.4` which has a security vulnerability. refs https://github.com/ipfs/js-ipfs/issues/1862

alanshaw:
Okay, I'm going to merge this. It makes me nervous having a PR with so much change hanging around! It LGTM 🤣 and all the tests are passing. Lets PR for any breaks / changes after merge.

PR #1850, opened 2019-01-29
chore: rename local option to offline

### LINKED ISSUE #1778: Rename --local to --offline
Related to: https://github.com/ipfs/go-ipfs/issues/4182
Go-ipfs proposal: https://github.com/ipfs/go-ipfs/pull/5825#discussion_r240721033

The local flag was *supposed* to mean "do this on the client, not the daemon" (I think?). However, that hasn't been true for a while. Now it just means "don't use the network for this operation".

Proposal:

1. Rename `--local` to `--offline` (replacing the `daemon --offline` flag with the global `--offline` flag.
2. Make it apply *everywhere*.

vasco-santos:
Closes #1778


alanshaw:
This PR addresses point 1 in #1778 but does not resolve point 2. I've updated the PR description so that the issue is not closed when this is merged.

Issue #1762, opened 2018-12-07
Remove ready event from basic usage


mikeal:
Now that we've started going down the "promises everywhere" rabbit hole it should be possible to remove waiting for the ready event from the default usage.

There's a few different ways we could go about this. Off the top of my head I think it would be nice to have a property on the object that is a promise that is resolved when ready.

```javascript
let ipfs = new IPFS()
await ipfs.ready
```

And we could create a convenience method called `create()` that did this automatically.

```javascript
let ipfs = await IPFS.create()
```

Thoughts?

haadcode:
I'd personally love to see this happen! 👍I think adding/changing what is suggested here would improve the usability and readability a ton (=2 loc). 

We had a discussion about it in the past in https://github.com/ipfs/js-ipfs/issues/1083 (see the different proposal for the API), but as you said @mikeal, things have changed now with using promises everywhere. Thanks for bringing it up again! ❤️

pgte:
@mikeal As a frequent user of js-ipfs I also think something like this is really necessary — thanks for bringing this up!

I would like, though, to make a distinction between creating and starting, and would like to make `.start()` a method.
(The creation can only be called once, but starting would be idempotent.)

Because in some user-land packages we can be using IPFS node instances that we haven't created ourselves, we don't really know the start state, and simply doing something like this to ensure the node is started would help a lot on reducing the complexity of some processes:

```js
// a) starts or,
// b) if starting, returns when started or
// c) if started, returns immediately
await ipfs.start() 
```

Also, I think we should decide between calling the state `started` versus `ready`. I'm fine with either, but `await ipfs.start()` is more intuituitive to me than `await ipfs.ready()`.

Also, I think `const ipfs = await IPFS.create(options)` should, by default, start the node (unless `options.start` is false).

Nonetheless, the user could still ensure it's started by doing `await start()`.

mikeal:
@pgte I ran into this just the other day calling `await ipfs.start()` when it was already started seeing the exception :)

I agree that it's confusing to have these two states, `ready` and `started`. My main question is, why is the `ready` state necessary? I understand that there are async things we want to do upon creation but why don't we just begin all of the API methods with `await ipfs.ready` and hide this state change from the user and effectively just queue up any method calls that are waiting for the ready state?

pgte:
@mikeal that would be very helpful, (much alike a lot of node.js database clients where it queues operations or queries while not started).

[...]

PR #1878, opened 2019-02-13
[WIP] feat: ready promise

pkafei:
Related to https://github.com/ipfs/js-ipfs/issues/1762

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -187,6 +187,14 @@ class IPFS extends EventEmitter {
     }
 
     boot(this)
+    this.once('ready', () => this.__ready = true)
+  }
+  get ready () {
+    return new Promise((resolve, reject) => {
+      if (this.__ready) return resolve(this)
+      this.on('ready', () => resolve(this))
+      this.on('error', reject)
+    })
   }
 }
 
@@ -194,4 +202,9 @@ exports = module.exports = IPFS
 
 exports.createNode = (options) => {
   return new IPFS(options)
+}
+
+exports.create = (options) => {
+  let node = new IPFS(options)
+  return node.ready
 }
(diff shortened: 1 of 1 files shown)


mikeal:
RE: documentation. Do we want to update the README to move all the examples to use this?

alanshaw:
> RE: documentation. Do we want to update the README to move all the examples to use this?

Yes, this should be the recommended way of obtaining a ready node.

Separately, I think we should stop advertising the callback API (both in the README and the examples) so predominantly. Instead just document that all the API methods support callbacks. I want new users to be using promises so that when https://github.com/ipfs/js-ipfs/issues/1670 lands it's easier for them to upgrade.

re: the [examples here](https://github.com/ipfs/js-ipfs/tree/master/examples) I think we should submit a separate PR to switch them _all_ to using async await style, I'd rather not mix `await ipfs.ready` with callbacks.

alanshaw:
@pkafei ping

pkafei:
> @pkafei ping

@alanshaw Saw comment and will give detailed response shortly.

pkafei:
> Yes, this should be the recommended way of obtaining a ready node.
> 
> Separately, I think we should stop advertising the callback API (both in the README and the examples) so predominantly. Instead just document that all the API methods support callbacks. I want new users to be using promises so that when #1670 lands it's easier for them to upgrade.
> 
> re: the examples here I think we should submit a separate PR to switch them all to using async await style, I'd rather not mix await ipfs.ready with callbacks.

Just want to reiterate the scope of this ticket. Here I'm implementing ready promise, and I'll update the README (docs) and add a test. 

@mikeal and I plan on updating all the callbacks in the `js-ipfs` codebase to use async await in future tickets. I think it would be better for us to update all of the examples _after_ we have refactored all of the callbacks to async await. Alternatively we can update the examples as we go.

alanshaw:
@pkafei do you need some help getting this finished?

pkafei:
@alanshaw Yes, I need help writing the test.

PR #1879, opened 2019-02-14
fix: dht browser disabled

vasco-santos:
The `libp2p` defaults were not allowing the browser config to disable the `dht` for browser nodes. This PR fixes that issue

### DIFF
--- a/test/core/dht.spec.js
+++ b/test/core/dht.spec.js
@@ -7,45 +7,96 @@ const dirtyChai = require('dirty-chai')
 const expect = chai.expect
 chai.use(dirtyChai)
 
+const isNode = require('detect-node')
+
 const IPFSFactory = require('ipfsd-ctl')
 const IPFS = require('../../src/core')
 
 describe('dht', () => {
-  let ipfsd, ipfs
+  describe('enabled', () => {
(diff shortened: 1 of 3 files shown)


vasco-santos:
Thanks @daviddias 

I added a test running in the browser and receiving the error

test/core/dht.spec.js
```diff
@@ -7,45 +7,95 @@ const dirtyChai = require('dirty-chai')
...
+      }
+    })
+
+    describe('put', () => {
+      it('should callback with error for DHT not available', async () => {
+        let res
+        try {
+          res = await ipfs.dht.put(Buffer.from('a'), Buffer.from('b'))
+        } catch (err) {
+          expect(err).to.exist()
```
alanshaw: Is there an error code or message we can assert on?

vasco-santos: No! But, I can add it to `js-libp2p`. However, we will have to bump the `rc`. What do you think?

alanshaw: We need to ensure the error is because the DHT is disabled and not because of anything else. Whatever you can do to make that happen is good 👍

PR #1902, opened 2019-02-27
refactor: use unixfs exporter with CID instances

alanshaw:
Allows IPFS to use the unixfs exporter that exports CID instances _without_ API changes.

Next step is to get the unixfs importer to do the same and then make a breaking interface JS IPFS core API change to ensure we're always returning CID instances not strings for add, ls etc etc.

refs https://github.com/ipfs/interface-js-ipfs-core/issues/394

Depends on:

* [x] https://github.com/ipfs/js-ipfs-mfs/pull/44
* [x] https://github.com/ipfs/js-ipfs-unixfs-exporter/pull/19


package.json
```diff
@@ -108,11 +108,12 @@
     "ipfs-block-service": "~0.15.1",
     "ipfs-http-client": "^29.0.0",
     "ipfs-http-response": "~0.2.1",
-    "ipfs-mfs": "~0.9.1",
+    "ipfs-mfs": "github:ipfs/js-ipfs-mfs#refactor/use-exporter-w-cid-instances",
```
achingbrain:
```javascript
    "ipfs-mfs": "~0.10.1",
```

PR #1907, opened 2019-03-04
chore: disable dht discovery

vasco-santos:
Currently `js-ipfs` was having a bad time with CPU usage. 

In this context, we decided to go with a temporary change on `dht.put`, as well as disabling the `random-walk` discovery. We will benefit from discovery through the queries that we will be running for now. 

With this, we will get `js-ipfs` stable again, as well as have the DHT available and have time to implement a proper solution to solve this issue.

More details about the problem and discussion of the solution:

- [libp2p/js-libp2p-kad-dht#86](https://github.com/libp2p/js-libp2p-kad-dht/issues/86)

### DIFF
--- a/src/core/components/libp2p.js
+++ b/src/core/components/libp2p.js
@@ -79,7 +79,7 @@ function defaultBundle ({ datastore, peerInfo, peerBook, options, config }) {
         kBucketSize: get(options, 'dht.kBucketSize', 20),
         enabled: get(options, 'offline', false) ? false : undefined, // disable if offline
         randomWalk: {
-          enabled: get(options, 'dht.randomWalk.enabled', true)
+          enabled: false // disabled waiting for https://github.com/libp2p/js-libp2p-kad-dht/issues/86
         },
         validators: {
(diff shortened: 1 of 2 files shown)


vasco-santos:
I did not add tests for the `randomWalk` disabled, as `js-libp2p` / `js-libp2p-kad-dht` already do that.

src/core/components/libp2p.js
```diff
@@ -79,7 +79,7 @@ function defaultBundle ({ datastore, peerInfo, peerBook, options, config }) {
         kBucketSize: get(options, 'dht.kBucketSize', 20),
         enabled: get(options, 'offline', false) ? false : undefined, // disable if offline
         randomWalk: {
-          enabled: get(options, 'dht.randomWalk.enabled', true)
+          enabled: false
```
alanshaw: Please could you add a comment explaining the reason for this being disabled?

vasco-santos:
Thanks for pointing out Alan. Added the issue in a comment

PR #1911, opened 2019-03-06
refactor: normalise debug logging prefix from jsipfs to just ipfs

alanshaw:
Shorter prefix ftw, MFS already uses "ipfs". Hooray \o/


achingbrain:
Ah, so much better.

PR #1914, opened 2019-03-11
fix: only dial to unconnected peers

alanshaw:
Depends on:

* [x] https://github.com/libp2p/js-libp2p-switch/pull/309
* [x] https://github.com/moxystudio/node-proper-lockfile/pull/74

### DIFF
--- a/src/core/components/libp2p.js
+++ b/src/core/components/libp2p.js
@@ -19,10 +19,11 @@ module.exports = function libp2p (self, config) {
   const libp2p = createBundle({ options, config, datastore, peerInfo, peerBook })
   let discoveredPeers = []
 
+  const noop = () => {}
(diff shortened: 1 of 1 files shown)


src/core/components/libp2p.js
```diff
@@ -21,7 +21,9 @@ module.exports = function libp2p (self, config) {
 
   const putAndDial = peerInfo => {
     peerBook.put(peerInfo)
```
jacobheun:
```suggestion
    peerInfo = peerBook.put(peerInfo)
```

When we discover a peer, we may not have the latest information from our PeerBook, such as the connected multiaddress. Doing `peerBook.put` handles merging this information and the new PeerInfo instance is returned. Updating the peerInfo instance, before performing a check, should make sure we have the correct information.

PR #1920, opened 2019-03-12
fix: cat deeply nested file

alanshaw:
The exporter was exporting the file, but the filter was filtering it out because the exported path includes only `file.txt` but the filter was expecting `path/file.txt`.

So when you cat `/ipfs/QmHash/file.txt` it worked, but when you cat `/ipfs/QmHash/path/file.txt` it did not because the exported file had path `file.txt` but the filter was looking for `path/file.txt`.

Depends on:

* [x] https://github.com/ipfs/interface-js-ipfs-core/pull/448

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "execa": "^1.0.0",
     "form-data": "^2.3.3",
     "hat": "0.0.3",
-    "interface-ipfs-core": "~0.97.1",
+    "interface-ipfs-core": "ipfs/interface-js-ipfs-core#test/add-test-for-deeply-nested-file",
     "ipfsd-ctl": "~0.41.0",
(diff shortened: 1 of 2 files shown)


alanshaw:
Merging because the CI error is unrelated:

```console
  1) cli
       dht
         query:
     Error: Command failed: /Users/travis/build/ipfs/js-ipfs/src/cli/bin.js dht query QmRhsb2HQUGSrZstoMqiYunqgDzw566igxtgdTwxbgrGuQ
connect ECONNREFUSED 127.0.0.1:52581
      at makeError (node_modules/execa/index.js:174:9)
      at Promise.all.then.arr (node_modules/execa/index.js:278:16)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

PR #1921, opened 2019-03-12
feat: Provide access to multicodec along with other libraries (#1913)

### LINKED ISSUE #1913: Please consider exposing multicodec along with other dependencies
I'm working on https://github.com/gozala/ipdf/ where I need to encode codec format in the buffer before encrypting it so during encryption I can figure out what format to use in decoding.

Provided an ipfs node instance implementation has access to everything it needs to do it's job except of [`multicodec`](https://github.com/multiformats/multicodec) that IPFS implementation internally depends upon. Please consider exposing it along with other dependencies in `ipfs.types.` as that would avoid pulling in separately library & any versioning mismatches that could arise from that.

niinpatel:
Resolves #1913


alanshaw:
This looks great @niinpatel! Thank you. Would you be willing to open a PR to do the same thing for [`ipfs-http-client`](https://github.com/ipfs/js-ipfs-http-client)?

niinpatel:
@alanshaw Sure, I'll do it today. ☺️

PR #1926, opened 2019-03-13
perf: lower connection manager limits

alanshaw:
I've observed that over ~600 peers the node becomes less stable, this lowers the default limits until we can address performance issues in libp2p.

### DIFF
--- a/src/core/runtime/config-browser.js
+++ b/src/core/runtime/config-browser.js
@@ -28,8 +28,8 @@ module.exports = () => ({
   ],
   Swarm: {
     ConnMgr: {
-      LowWater: 600,
-      HighWater: 900
+      LowWater: 200,
+      HighWater: 500
     }
   }
 })
--- a/src/core/runtime/config-nodejs.js
+++ b/src/core/runtime/config-nodejs.js
@@ -41,8 +41,8 @@ module.exports = () => ({
   ],
   Swarm: {
     ConnMgr: {
-      LowWater: 600,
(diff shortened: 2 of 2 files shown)


src/core/runtime/config-browser.js
```diff
@@ -28,8 +28,8 @@ module.exports = () => ({
   ],
   Swarm: {
     ConnMgr: {
-      LowWater: 600,
-      HighWater: 900
+      LowWater: 200,
+      HighWater: 500
```
jacobheun: We can try this, but I think 500 is going to end up being too high for the browser. I think it's more reasonably going to be in the range of 50-200.

alanshaw: I'm going to get some testing done for this in the browser and adjust it accordingly, for now just setting it to be the same as Node.js.

alanshaw:
Merging because the CI failure is unrelated and will be resolved in a subsequent PR:

```console
  1) cli
       dht
         query:
     Error: Command failed: /Users/travis/build/ipfs/js-ipfs/src/cli/bin.js dht query Qmbkf4u9FFDM5i5PywnpgLuPQg8tP8qBBWXQTWEswYdxwp
connect ECONNREFUSED 127.0.0.1:52855
      at makeError (node_modules/execa/index.js:174:9)
      at Promise.all.then.arr (node_modules/execa/index.js:278:16)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

PR #1930, opened 2019-03-13
feat: adds http DAG api

achingbrain:


test/http-api/dag.js
```diff
@@ -0,0 +1,111 @@
...
+      ipfs.dag.get(`${cid.toBaseEncodedString()}/foo`, (err, result) => {
+        expect(err).to.not.exist()
+        expect(result.value).to.equal('bar')
+
+        done()
+      })
+    })
+  })
+
+  it('returns value with a path as a separate argument', (done) => {
```
alanshaw: Is this covered by Is this covered by https://github.com/ipfs/interface-js-ipfs-core/blob/8d7467b207425536b16e196ba7b12e8ce306d7fa/src/dag/get.js#L171-L177?

test/http-api/dag.js
```diff
@@ -0,0 +1,111 @@
...
+  })
+
+  it('returns error for request with invalid argument', (done) => {
+    ipfs.dag.get('invalid', { enc: 'base58' }, (err, result) => {
+      expect(err.message).to.include("invalid 'ipfs ref' path")
+      done()
+    })
+  })
+
+  it('returns value', (done) => {
```
alanshaw: Is this covered by https://github.com/ipfs/interface-js-ipfs-core/blob/8d7467b207425536b16e196ba7b12e8ce306d7fa/src/dag/get.js#L101-L114?

test/http-api/dag.js
```diff
@@ -0,0 +1,111 @@
...
+
+      ipfs.dag.get(cid, 'foo', (err, result) => {
+        expect(err).to.not.exist()
+        expect(result.value).to.equal('bar')
+
+        done()
+      })
+    })
+  })
+})
```
alanshaw: Shall we move the rest of these to `interface-ipfs-core`? They're not testing anything that only JS IPFS supports and there's some overlap with the tests you added here https://github.com/ipfs/js-ipfs-http-client/pull/957

src/http/api/resources/dag.js
```diff
@@ -0,0 +1,237 @@
...
+          throw Boom.boomify(err, { message: 'Failed to put node' })
+        }
+
+        if (request.query.pin) {
+          await ipfs.pin.add(cid)
+        }
+
+        resolve(h.response({
+          Cid: {
+            '/': cid.toBaseEncodedString(request.query.cidbase)
```
alanshaw: Please use the [`cidToString`](https://github.com/ipfs/js-ipfs/blob/master/src/utils/cid.js) utility so this doesn't throw and auto upgrades the CID to v1 when appropriate.

e.g. `cidToString(cid, { base: request.query['cid-base'] })`

src/http/api/resources/dag.js
```diff
@@ -0,0 +1,237 @@
...
+            format: request.query.format,
+            hashAlg: request.query.hash
+          })
+        } catch (err) {
+          throw Boom.boomify(err, { message: 'Failed to put node' })
+        }
+
+        if (request.query.pin) {
+          await ipfs.pin.add(cid)
+        }
```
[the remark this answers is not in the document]
grantlouisherman: hey @alanshaw I will wait on this to get merged before opening a PR.

src/http/api/resources/dag.js
```diff
@@ -0,0 +1,237 @@
...
+      throw Boom.boomify(err, { message: 'Failed to get dag node' })
+    }
+
+    if (key.codec === 'dag-pb' && result.value) {
+      if (typeof result.value.toJSON === 'function') {
+        result.value = result.value.toJSON()
+      }
+
+      if (Buffer.isBuffer(result.value.data)) {
+        result.value.data = result.value.data.toString(request.query.dataencoding)
```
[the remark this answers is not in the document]
achingbrain: No, you're right - I thought I'd seen that convention in the codebase but appear to have been seeing things.

src/http/api/resources/dag.js
```diff
@@ -0,0 +1,237 @@
...
+
+    try {
+      result = await ipfs.dag.get(key, path)
+    } catch (err) {
+      throw Boom.boomify(err, { message: 'Failed to get dag node' })
+    }
+
+    if (key.codec === 'dag-pb' && result.value) {
+      if (typeof result.value.toJSON === 'function') {
+        result.value = result.value.toJSON()
```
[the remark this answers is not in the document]
achingbrain: According to the [HTTP API docs](https://docs.ipfs.io/reference/api/http/#response-23) `/dag/get` should return a `text/plain` response body, which seems like an odd decision.

test/http-api/dag.js
achingbrain: I've refactored the tests to use `.inject` so we're testing the HTTP API contract and not the `ipfs-http-client` module.  You are quite right, most of the tests are covered by the interface tests.

PR #1933, opened 2019-03-14
fix: handle subdomains for ipfs.dns

### LINKED ISSUE #1932: ipfs.dns does not handle subdomains correctly
When I do ipfs.dns('docs.ipfs.io'). It throws an error 'ENODATA'. I did some digging and found out that this error is thrown by nodejs' `dns.resolveTxt` method when it does not find any txt records. 

Now, to handle subdomain correctly, we need to append `_dnslink.` string to it, which we do here
https://github.com/ipfs/js-ipfs/blob/3ef1ed42581152701d002efaaff4215c96f60930/src/core/runtime/dns-nodejs.js#L20-L22

But, the problem is, the error is way before the line gets to be executed, here: 

https://github.com/ipfs/js-ipfs/blob/3ef1ed42581152701d002efaaff4215c96f60930/src/core/runtime/dns-nodejs.js#L11

The easiest solution to this problem would be to check for err.code 'ENODATA' also along with 'ENOTFOUND', etc. in the if condition. So, the next lines get a chance to be executed. 

Basically something like this:

```
      if (err.code !== 'ENOTFOUND' && err.code !== 'ERR_DNSLINK_NOT_FOUND' && err.code !== 'ENODATA') throw err
```
Would fix the problem. 

What do you think?

niinpatel:
resolves #1932 

`dns.requireTxt` throws 'ENODATA' if no txt records are found, so I included a condition to check for that error and check the `_dnslink.<domain>` so that subdomain is resolved correctly. 

Also added a few tests for checking if subdomain resolves correctly.

### DIFF
--- a/src/core/runtime/dns-nodejs.js
+++ b/src/core/runtime/dns-nodejs.js
@@ -7,8 +7,8 @@ const errcode = require('err-code')
 module.exports = (domain, opts, callback) => {
   resolveDnslink(domain)
     .catch(err => {
-      // If the code is not ENOTFOUND or ERR_DNSLINK_NOT_FOUND then throw the error
-      if (err.code !== 'ENOTFOUND' && err.code !== 'ERR_DNSLINK_NOT_FOUND') throw err
+      // If the code is not ENOTFOUND or ERR_DNSLINK_NOT_FOUND or ENODATA then throw the error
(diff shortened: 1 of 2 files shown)


niinpatel:
dns-subdomain

PR #1935, opened 2019-03-14
feat: recursive dnslink lookups

### LINKED ISSUE #1931: `ipfs.dns` does not have "recursive" option
Version:
js-ipfs version: 0.35.0-rc.0
Platform:
Windows 10 64-bit
Google Chrome Version 73.0.3683.75
Subsystem: `ipfs.dns`

#### Description:
go-ipfs implementation has `--recursive` option to resolve dns links. 
js-ipfs throws this error: "Not enough non-option arguments: got 0, need at least 1"

**go-ipfs**: 
`ipfs dns --recursive ipfs.io`
resolves to: /ipfs/QmTxvVbLQTm9qvh4wsdhehDGPy92LJSYiyAg5sh3X2ZaUE

**js-ipfs**:
`jsipfs dns --recursive ipfs.io`: 
throws: "Not enough non-option arguments: got 0, need at least 1"

niinpatel:
fixes #1931 . 

Also, incorporated the suggestions from https://github.com/ipfs/interface-js-ipfs-core/issues/377  that `ipfs dns` should be recursive by _default_.


niinpatel:
Here's the quick demo of this feature working on the cli. I've included tests to make sure it works in other places like http-api, browsers, etc. 

ipfs dns demo

niinpatel:
@alanshaw can you review this one when you can?

alanshaw:
Thanks @niinpatel - I will look at this as soon as I get a moment!

niinpatel:
Hey @alanshaw , did you get a chance to look at it?

niinpatel:
@alanshaw did you get a chance to check it out?

src/cli/commands/dns.js
```diff
@@ -7,15 +7,21 @@ module.exports = {
...
+      alias: 'r',
+      desc: 'Resolve until the result is not a DNS link'
+    },
     format: {
       type: 'string'
     }
   },
 
-  handler ({ getIpfs, domain, resolve }) {
+  handler ({ getIpfs, domain, resolve, ...opts }) {
```
alanshaw: Please could you use long hand for now - I don't know how good support for this feature is in browsers/Node.js and we don't transpile.

```suggestion
  handler ({ getIpfs, domain, resolve, recursive, format }) {
```

src/core/runtime/dns-nodejs.js
```diff
@@ -2,10 +2,29 @@
...
 const _ = require('lodash')
+const isIPFS = require('is-ipfs')
 const errcode = require('err-code')
 
+const maxRecursiveDepth = 32
+
 module.exports = (domain, opts, callback) => {
-  resolveDnslink(domain)
+  // recursive is true by default, it's set to false only if explicitly passed as argument in opts
+  const recursive = opts.recursive === undefined || opts.recursive.toString() === 'true'
```
alanshaw: Default true for `null`/`undefined` and boolean value of argument otherwise:

```suggestion
  const recursive = opts.recursive == null ? true : Boolean(opts.recursive)
```

src/core/runtime/dns-nodejs.js
```diff
@@ -2,10 +2,29 @@
 
 const dns = require('dns')
 const _ = require('lodash')
+const isIPFS = require('is-ipfs')
 const errcode = require('err-code')
 
+const maxRecursiveDepth = 32
```
alanshaw:
```suggestion
const MAX_RECURSIVE_DEPTH = 32
```

test/core/dns.spec.js
```diff
@@ -0,0 +1,62 @@
...
+      expect(res).to.match(/\/ipfs\/.+$/)
+    })
+  })
+
+  it('should resolve subdomain docs.ipfs.io', () => {
+    return ipfs.dns('docs.ipfs.io').then(res => {
+      // matches pattern /ipfs/<hash>
+      expect(res).to.match(/\/ipfs\/.+$/)
+    })
+  })
```
alanshaw: Could you please move these tests to https://github.com/ipfs/interface-js-ipfs-core/blob/master/src/miscellaneous/dns.js?

src/core/runtime/dns-nodejs.js
niinpatel: `Boolean('false')` actually returns `true` not `false`. Arguments passed via query parameters are of `string` type and we need to parse those too. Once easy solution I just now came up with is this: 

`const recursive = opts.recursive.toString() !== 'false'`

If recursive is `false` or `"false"` then only we parse it as false. `true` when passed anything else, including `null` and `undefined`

test/core/dns.spec.js
niinpatel: Done https://github.com/ipfs/interface-js-ipfs-core/pull/455

src/core/runtime/dns-nodejs.js
niinpatel: Edit: That wouldn't work either because error will be thrown if opts.recursive is undefined. 

The original one is probably the best solution after all, with slight modifications: 

`const recursive = opts.recursive == undefined || opts.recursive.toString() !== 'false'`

opts.recursive === undefined \|\| opts.recursive.toString() === 'true'
 





opts.recursive === undefined \|\| opts.recursive.toString() === 'true'

alanshaw: > Arguments passed via query parameters are of string type and we need to parse those too

It should be passed as a boolean - this is the documented type for the value. If a string is being passed then we need to [fix our HTTP endpoint](https://github.com/ipfs/js-ipfs/blob/master/src/http/api/resources/dns.js) to properly parse query strings to the expected types for the calls to core.

We can't guarantee the passed value is a boolean but we _can_ ensure that the value is a boolean for our code by casting it.

niinpatel: That makes sense. Right now I don't think query parameters are being parsed and "type-casted" before passing to the core API.  For this PR, I've included it here: 

https://github.com/ipfs/js-ipfs/blob/fb114f1ece896130ea5eddc59248c304b0ae6036/src/http/api/resources/dns.js#L14-L18

niinpatel:
I've applied all the suggested changes.

PR #1949, opened 2019-03-19
Show room feature in Exchange Files Example

### LINKED ISSUE #1632: Exchange Files Example: Show room feature
image

Currently, it is impossible to guess that there is the room feature on the exchange files example. It would be great to add a label at the top of the website of "create room"

niinpatel:
As mentioned in issue #1632, currently, it is impossible to guess that there is the room feature on the exchange files example.

So, I added a textbox and a button which can be used for subscribing to a different room. Here's a small preview of the feature: 

ipfs rooms demo (2)

closes #1632


alanshaw:
That is super cool @niinpatel 🚀 - I'll review as soon as I have a sec!

Issue #1826, opened 2019-01-17
⚡️ v0.35.0 RELEASE 🚀


alanshaw:
> Small, acyclic and flexible, just the way you like it

# 🗺 What's left for release?

* [x] ~~[DHT](https://github.com/ipfs/js-ipfs/pull/856)~~ delayed until a future release
* [x] [Connections limiting](https://github.com/ipfs/js-ipfs/pull/1872)
* [x] [Upgrade API for Hapi 18](https://github.com/ipfs/js-ipfs/pull/1844)
* [x] [Rename `--local` to `--offline`](https://github.com/ipfs/js-ipfs/pull/1850)
* [x] [Add `--enable-preload` to enable/disable preloading for daemons](https://github.com/ipfs/js-ipfs/pull/1909)
* [x] [Export types and utilities statically](https://github.com/ipfs/js-ipfs/pull/1908)
* [x] [API/Gateway listen on multiple addresses](https://github.com/ipfs/js-ipfs/pull/1903)
* [x] [Only dial to unconnected peers](https://github.com/ipfs/js-ipfs/pull/1914)
* [x] [HTTP DAG API](https://github.com/ipfs/js-ipfs/pull/1930)
* [x] [ipfs.ls file sizes interop](https://github.com/ipfs/js-ipfs-unixfs-exporter/pull/20)
* [x] [Dependency updates](https://github.com/ipfs/js-ipfs/pull/1929)

# 🔦 Highlights

## ✂️ Huge bundle size cuts!

One of the biggest goals for JS IPFS is to enable the distributed web in the browser. Part of that is being a good web citizen and minimising the amount of code we ship, to ensure IPFS downloads quicker, `npm install`'s quicker, loads quicker and utilizes fewer resources. It's especially important on resource constrained mobiles and IoT devices.

The good news you want to hear is that we've now chopped around **26%** off the bundle size of JS IPFS since 0.33, that's a reduction of _over_ a quarter of the code we were previously shipping! This is an amazing improvement and a great reason to upgrade.

## 🌳 DAG HTTP API

The HTTP API now has endpoints for DAG operations like `get`, `put` and `resolve`. Hooray! It was actually super easy so we just decided to do it spontaneously. I know, we're so zany.

On a more serious note, this brings JS IPFS another step closer to feature parity with Go IPFS. We're pushing forwards with this in _every_ release 🚀 - stay tuned!

## 🙉 Multiple address listening for API and Gateway servers

[...]

PR #1951, opened 2019-03-20
docs: add DHT config documentation

alanshaw:
Following up on https://github.com/ipfs/js-ipfs/issues/1826#issuecomment-471920430


alanshaw:
Browser build failures are due to an [upstream dependency change](https://github.com/babel/babel/pull/9493) that has broken the browser build.

Issue #1947, opened 2019-03-19
CLI getIpfs uses wrong args object


hugomrdias:
- **Version**:
- **Platform**:
- **Subsystem**:

#### Description: 
this line 
https://github.com/ipfs/js-ipfs/blob/master/src/cli/bin.js#L48
passes an argv obj that wasn't created with our yargs config

This makes cmds like `ipfs add --silent ./path/file` to instantiate IPFS with `{ silent :  './path/file' }`

Right now master doesn't fail because `silent` wasn't added to the validation yet, still this behaviour should be fixed.

grassias:
If we changed the line https://github.com/ipfs/js-ipfs/blob/master/src/cli/bin.js#L48
into: `const getIpfs = utils.singleton(cb => utils.getIPFS(cli.argv, cb))`?
That'll do it?

alanshaw:
Sadly no, `argv` is an accessor that causes a handler for the command to be invoked. At the point where `getIpfs` is called we're already in a handler so we need a new yargs which doesn't have the commands handlers. All we're missing is the ability to parse the global options (`--silent` and `--pass`).

I think we just need a new function to create a new yargs with the global options that we can use and assign to `cli` as well as use in `getIpfs`.

hugomrdias:
made some suggestions in the #1955 PR

alanshaw:
Thanks again @grassias ❤️

PR #1955, opened 2019-03-21
Create helper function to parse global options

grassias:


grassias:
This PR aims to address issue #1947 after @alanshaw suggestions.

hugomrdias:
i propose to do something like this 

```js

  let getIPFS = null
  cli
    .commandDir('commands')
    .middleware(argv => {
      getIPFS = argv.getIpfs = utils.singleton(cb => utils.getIPFS(argv, cb))
      return argv
    })
    .help()
    .strict()
    .completion()
  let exitCode = 0

  try {
    const { data } = await new YargsPromise(cli).parse(args)
    if (data) print(data)
  } catch (err) {
    debug(err)

    // the argument can have a different shape depending on where the error came from
    if (err.message || (err.error && err.error.message)) {
      print(err.message || err.error.message)
    } else {
      print('Unknown error, please re-run the command with DEBUG=ipfs:cli to see debug output')
    }

    exitCode = 1
  } finally {
    // If an IPFS instance was used in the handler then clean it up here
    if (getIPFS.instance) {
      try {
        const cleanup = getIPFS.rest[0]
        await cleanup()
      } catch (err) {
        debug(err)
        exitCode = 1
      }
    }
  }

  if (exitCode) {
    process.exit(exitCode)
  }
}
```

this way we use only one yargs instance

grassias:
I'd like so but I'm not sure yet about how to handle it. Is there any test utility/method that exposes the instantiated IPFS object produced by cmds?

hugomrdias:
to do this we need to change the way we test the cli check the example below

https://codesandbox.io/s/mqq66j9v6x

PR #1965, opened 2019-03-28
feat: better error message on ipfs cat

niinpatel:
resolves: #1145 

The command `ipfs cat <hash>/wrong-dir/wrong-file` 
previously returned error message: `No such file`

Following the suggestions given in [this thread](https://github.com/ipfs/js-ipfs/pull/1270#pullrequestreview-105635422), I changed the error message to:
`no file named "wrong-dir/wrong-file" under <hash>`
 
Also, the CI check will fail because test case for this is in [interface-ipfs-core](https://github.com/ipfs/interface-js-ipfs-core/blob/0f256bf399a8e3318dab5927f2670882d5551640/src/files-regular/cat.js#L192-L194). So we'll need to change the tests in interface-ipfs-core repo.

### DIFF
--- a/src/core/components/files-regular/cat-pull-stream.js
+++ b/src/core/components/files-regular/cat-pull-stream.js
@@ -14,8 +14,9 @@ module.exports = function (self) {
     options = options || {}
 
     ipfsPath = normalizePath(ipfsPath)
+
     const pathComponents = ipfsPath.split('/')
-    const fileNameOrHash = pathComponents[pathComponents.length - 1]
+    const cid = pathComponents[0]
 
     if (options.preload !== false) {
       self._preload(pathComponents[0])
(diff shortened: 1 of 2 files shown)


src/core/components/files-regular/cat-pull-stream.js
```diff
@@ -14,26 +14,40 @@ module.exports = function (self) {
...
     }
 
     const d = deferred.source()
 
+    let closestMatchedLink
+
     pull(
-      exporter(ipfsPath, self._ipld, options),
-      pull.filter(file => file.path === fileNameOrHash),
+      exporter(cid, self._ipld, options),
```
alanshaw: If you export the CID not CID+path you'll have to traverse through the entire graph until you get to the path you were looking for. This could make `cat` VERY slow for big directories if the content is not in your repo already (and be slow anyway, even if it is).

cc @achingbrain

achingbrain: +1 that - if you pass the path as well as the CID (e.g. `/ipfs/Qmfoo/path/to/file.txt`) the the unixfs exporter is smart enough to not load whole DAGs for sharded directories, etc which saves a load of time & bandwidth.  Maybe this change is better made there?

alanshaw:
Closing as this needs a better approach.

PR #1968, opened 2019-03-29
Fix ws port example, static refs in exchange files in browser

dirkmc:
- Port is currently shown as 9999, should be 4003
- Update screenshot with new port number
- Use static reference to Ipfs.Buffer


dirkmc:
Good catch - updated

PR #1976, opened 2019-04-03
fix: stop IPNS republisher ASAP

alanshaw:
This PR refactors the stop logic to perform stop actions in parallel for various components (where it is safe to do so). This means that libp2p gets stopped at the same time as the republisher. When libp2p stops it closes all connections meaning that the republsher is able to complete a running republish sooner.

Before, we had to wait for republish to complete before stop happens (since there's no way to cancel a DHT put right now).


alanshaw:
wow, did not expect this to cause this:

```console
  1) interface-ipfs-core over ipfs-http-client tests
       .key.list
         "after all" hook:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\Users\travis\build\ipfs\js-ipfs\test\node.js)
```

PR #1977, opened 2019-04-03
fix: avoid logging http errors when its logger is not on

jacobheun:
If any logger is turned on via `DEBUG`, such as `DEBUG=libp2p:switch` the api errors get logged to console, which makes viewing the actual logs I want rather annoying. 

Rather than checking the `DEBUG` value for http logging, this change has it check the status of the debug logger itself. Rather than defaulting to `error` level, it now defaults to `fatal`.

If you use `DEBUG=ipfs:http-api` or `DEBUG=ipfs:http-api:error`, you will still get the appropriate log level.


alanshaw:
@jacobheun I've rebased for green CI

PR #1981, opened 2019-04-08
fix: move mfs cmds and safer exit

hugomrdias:
This PR moves mfs cli commands to parser.js to make them available for tests and also tweak a bit error handling.


src/cli/bin.js
```diff
@@ -41,18 +38,15 @@ async function main (args) {
...
-        }
-      }
-    })
+
+  onExit(() => {
+    // If an IPFS instance was used in the handler then clean it up here
+    if (getIpfs && getIpfs.instance) {
+      const cleanup = getIpfs.rest[0]
+
+      cleanup.catch(err => debug(err))
```
alanshaw:
```suggestion
      cleanup().catch(err => debug(err))
```

src/cli/bin.js
```diff
@@ -41,18 +38,15 @@ async function main (args) {
...
+    }
+  })
```
alanshaw: I don't think we're able to do async operations on process exit? Do we need a synchronous cleanup instead?

alanshaw:
ping @hugomrdias - any chance you can get this one ready to merge?

src/cli/bin.js
```diff
@@ -3,79 +3,65 @@
...
+  print(err.message)
+  debug(err)
+})
 
-  const cli = new YargsPromise(parser)
+// Check for node version
+if (!semver.satisfies(process.versions.node, pkg.engines.node)) {
+  console.error(`Please update your Node.js version to ${pkg.engines.node}`)
+  process.exit(1)
+}
```
alanshaw: The `uncaughtException`/`unhandledRejection` handlers and node version check need to be above the `require` statements as they were because any one of them could cause an error or use code that is not supported in the version of Node.js that the user is running.

src/cli/bin.js
```diff
@@ -3,79 +3,65 @@
...
+process.once('uncaughtException', (err, origin) => {
+  if (origin === 'uncaughtException') {
+    print(err.message)
+    debug(err)
+  }
+})
+process.once('unhandledRejection', (err) => {
+  print(err.message)
+  debug(err)
+})
```
alanshaw: Can we log these out as they were before? ...(maybe `console.error` not `console.info`?)

These unexpected errors aren't always easy to reproduce so getting hold of a stack trace after the fact could be difficult and time consuming. You'd have to run the command again in debug mode and somehow replicate the circumstances that trigger the error.

Users also rarely run in debug mode so won't see the stack either. That translates to it being more difficult for us to begin to debug if users open an issue with just an error message and not a stack trace.

It also makes it harder for users to self debug - users unfamiliar with the `debug` module need to figure out how to turn it on in order to get the stack.

test/cli/config.js
```diff
@@ -58,7 +58,7 @@ describe('config', () => runOnAndOff((thing) => {
     })
 
     it('set a config key with invalid json', () => {
-      return ipfs.fail('config foo {"bar:0} --json')
+      return ipfs.fail('config foo {"bar:0"} --json')
```
alanshaw: I think this test is meant to fail!

test/cli/id.js
```diff
@@ -1,25 +1,43 @@
...
-    return ipfs('id').then((res) => {
-      const id = JSON.parse(res)
-      expect(id).to.have.property('id')
-      expect(id).to.have.property('publicKey')
-      expect(id).to.have.property('addresses')
-    })
+    expect(data).to.eq('{\n  "id": "id",\n  "publicKey": "publicKey"\n}')
   })
-}))
+})
```
alanshaw: This is quite complicated, would it be better to just require the command and call the handler, passing in the mocks and asserting on the result?

src/cli/bin.js
hugomrdias: to do that it's better to just let them throw and remove the handlers all together

alanshaw: Screenshot 2019-07-23 at 14 00 14

First is with:

```js
process.on('uncaughtException', err => {
  console.error(err)
  process.exit(1)
})
```

...second is without.

Not much in it TBH. It's kinda nice to have the Node.js internal stack greyed out.

My main concern is having the stack logged for these unexpected errors and not behind a debug flag 😉

test/cli/config.js
hugomrdias: and still does!
if we don't close the quotes the stdin keeps open for manual input

test/cli/id.js
hugomrdias: yes i know thats one of the things im gonna do next, do injection with https://github.com/yargs/yargs/blob/master/docs/api.md#parseargs-context-parsecallback

this will become just

```js
it('should output formatted json string', async () => {
    const { data } = await cli.parse('id', {
      ipfs: {
        id: Promise.resolve({ id: 'id', publicKey: 'publicKey' })
      }
    })

    expect(data).to.eq('{\n  "id": "id",\n  "publicKey": "publicKey"\n}')
})


```

no spies, no mocks, no stubs just plain simple js

PR #1983, opened 2019-04-10
fix: use libp2p auto dial

jacobheun:
Moves all peer discovery dialing logic to libp2p, which prevents dialing discovered nodes when we have enough peers.

* [x] depends on https://github.com/libp2p/js-libp2p/pull/349


jacobheun:
Added it to the libp2p config section of the readme.

jacobheun:
I've also updated the custom libp2p example and preemptively added the dependency as `0.25.0` for libp2p. I've gone through all the libp2p examples to update those, https://github.com/libp2p/js-libp2p/pull/353. 

I'll do some final testing this afternoon and do a release of 0.25 tonight or first thing tomorrow morning.

cc @vasco-santos

PR #1988, opened 2019-04-11
fix: ipns using offline datastore if no dht

vasco-santos:
When the DHT is not enabled, we must add the `OfflineDatastore` to for IPNS instead of the DHT

### DIFF
--- a/src/core/components/start.js
+++ b/src/core/components/start.js
@@ -2,14 +2,11 @@
 
 const series = require('async/series')
 const Bitswap = require('ipfs-bitswap')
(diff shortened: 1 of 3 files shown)


alanshaw:
Can we get the build green and add a test?

alanshaw:
@vasco-santos I've merged this into #1991 as I think they now depend on each other!

vasco-santos:
@alanshaw just added a new commit to skip the ipns over dht tests. You should grab it as well, or skip in your PR.

PR #1990, opened 2019-04-12
chore: update libp2p version

jacobheun:
fix: update configurations for libp2p-switch

Libp2p 0.25 is out! This updates the dependency and adds configuration for libp2p switch to better help with connection creation and management.


alanshaw:
LGTM, windows failure seems unrelated

Issue #2012, opened 2019-05-01
RFC: js-ipfs Garbage Collection


dirkmc:
This issue is to discuss how best to implement Garbage Collection in js-ipfs

### The go-ipfs Garbage Collector
We would like to learn from the experience of go-ipfs when building a Garbage Collector for js-ipfs.

#### Triggers

- Manual: `ipfs repo gc`
- Daemon flag: `--enable-gc` causes GC to run
  - only if storage exceeds `StorageGCWatermark`% of `StorageMax` (90% of 10G by default)
  - periodically every `GCPeriod` (1 hour by default)
  - ~~when a file is added to unixfs~~ ([currently disabled](https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/unixfs.go#L44))

#### Algorithm

[Source code](https://github.com/ipfs/go-ipfs/blob/master/pin/gc/gc.go#L32)
```Go
// GC performs a mark and sweep garbage collection of the blocks in the blockstore
// first, it creates a 'marked' set and adds to it the following:
// - all recursively pinned blocks, plus all of their descendants (recursively)
// - bestEffortRoots, plus all of its descendants (recursively)
// - all directly pinned blocks
// - all blocks utilized internally by the pinner
//
// The routine then iterates over every block in the blockstore and
// deletes any block that is not found in the marked set.
func GC(ctx context.Context, bs bstore.GCBlockstore, dstor dstore.Datastore, pn pin.Pinner, bestEffortRoots []cid.Cid) <-chan Result {
```
Note that `bestEffortRoutes` currently only contains the [MFS root](https://github.com/ipfs/go-ipfs/blob/master/core/corerepo/gc.go#L77)

#### Requirements
- Manually Garbage Collect all blocks that are not pinned, not in MFS and not used internally
  `ipfs repo gc`: mark and sweep - remove all unreachable blocks
- Passively Garbage Collect old blocks when new blocks are added to the blockstore
  `ipfs daemon --enable-gc` causes GC to run
  - whenever a block is added to blockstore
  - only if storage exceeds `StorageGCWatermark`% of `StorageMax` (90% of 10G by default)
  - remove only excess blocks, least recently accessed blocks first

#### Algorithm for passive GC
- Track reference counts using persistent reference counting algorithm
  - reference count increment:
    - each time a block is pinned directly or indirectly
      - pinner must ensure that multiple pins of same hash are only registered once
    - on mfs add
  - reference count decrement:
[...]

PR #2022, opened 2019-05-08
Implement Garbage Collection

dirkmc:
resolves https://github.com/ipfs/js-ipfs/issues/2012

Depends on
- [x] https://github.com/ipfs/js-ipfs/pull/2004
- [x] https://github.com/ipfs/js-ipfs-http-client/pull/992
- [x] https://github.com/ipfs/interface-js-ipfs-core/pull/462
- [x] https://github.com/achingbrain/mortice/pull/1

TODO:
- [x] Core (mark and sweep)
- [x] CLI
- [x] http interface
- [x] interface-js-ipfs-core tests https://github.com/ipfs/interface-js-ipfs-core/pull/462
- [x] nodejs-specific tests
- [x] Locking
- [x] Tests for locking


dirkmc:
Note: I needed to refactor the pinning code in order to share code with some GC functionality. The functionality hasn't changed, things have just moved around somewhat.

wehriam:
@dirkmc thanks for working on this, and looking forward to the finish line.

We're using JS-IPFS to run 24/7 video streams in the browser; it is fairly quickly filling up all available disk. 

Please let me know if there's anything I can do to assist.

dirkmc:
@wehriam that's great to hear, thanks! It would be great to get some feedback from some different use cases. Would you like to try out this branch to see if it helps with your use case? I can provide instructions if you're not sure how

wehriam:
@dirkmc thanks! We did a build off of [feat/gc](https://github.com/ipfs/js-ipfs/tree/feat/gc) and are testing it now. It seems to be working well so far.

Are you planning support for [Datastore options similar to those in go-ipfs](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#datastore)? They aren't in the [config parser](https://github.com/ipfs/js-ipfs/blob/master/src/core/config.js) so I wasn't sure.

We're currently using the [navigator.storage.estimate()](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/estimate) method (when available) to trigger GC.

Are all unpinned blocks removed on garbage collection? It would be great to have a way to remove them on an LRU basis so browser-based swarms could serve as an edge cache for large files.

dirkmc:
@wehriam the implementation of GC on this branch must be manually invoked, and removes all blocks that are unpinned and not in MFS. We are [currently discussing](https://github.com/ipfs/js-ipfs/issues/2012) how best to implement automatic GC, including using an LRU heuristic.

You can see usage information for the IPFS repo with [`ipfs.repo.stat()`](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/REPO.md#repostat)

wehriam:
@dirkmc I have a fiddle running here - https://jsfiddle.net/t9a5ogc1/ - you can see usage stats in the console.

wehriam:
This works well in our testing. Are there any blockers for a final release? It appears https://github.com/ipfs/interface-js-ipfs-core/pull/460 has been successfully merged.

dirkmc:
@wehriam thanks for helping out with testing this in a real world scenario.

We are currently all hands on deck getting ready for [IPFS camp](https://camp.ipfs.io), so it's likely this will be released soon after, once people have had time to code review it.

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,135 @@
...
+    return this.lock('readLock', lockedFn, cb)
+  }
+
+  writeLock (lockedFn, cb) {
+    return this.lock('writeLock', lockedFn, cb)
+  }
+
+  lock (type, lockedFn, cb) {
+    if (typeof lockedFn !== 'function') {
+      throw new Error(`first argument to ${type} must be a function`)
```
alanshaw:
```suggestion
      return cb(new Error(`first argument to ${type} must be a function`))
```

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,135 @@
...
+      log(`[${lockId}] ${type} started`)
+      lockedFn((err, res) => {
+        this.emit(`${type} release`, lockId)
+        log(`[${lockId}] ${type} released`)
+        err ? reject(err) : resolve(res)
+      })
+    })
+
+    const lock = this.mutex[type](locked)
+    return lock.then(res => cb(null, res)).catch(cb)
```
alanshaw:
```suggestion
    return lock.then(res => cb(null, res), cb)
```

Is it better to do this instead? My understanding is that in this case if `cb(null, res)` throws then we get an uncaught promise rejection, whereas with `.then(res => cb(null, res)).catch(cb)` if `cb(null, res)` throws then `cb` is called _again_ with the error.

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,135 @@
+'use strict'
+
+const mortice = require('mortice')
+const pull = require('pull-stream')
```
alanshaw: Please import just the things you need from `pull-stream` like we do with `async`.

package.json
```diff
@@ -159,13 +160,15 @@
     "merge-options": "^1.0.1",
     "mime-types": "^2.1.21",
     "mkdirp": "~0.5.1",
+    "mortice": "dirkmc/mortice#fix/read-then-write",
```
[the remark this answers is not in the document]
dirkmc: @achingbrain could we please have a v1.2.2 mortice release

src/core/components/pin/gc-lock.js
dirkmc: The reason I throw here instead of returning an error in the callback is that not providing the right arguments to the function is a programmer error. Do you think that's the right approach?

src/core/components/pin/gc-lock.js
dirkmc: You're right, good catch 👍

package.json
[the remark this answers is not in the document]
achingbrain: `v1.2.2` was released a month ago?

```
$ npm view mortice versions
npm info it worked if it ends with ok
npm info using npm@6.4.1
npm info using node@v10.15.3
npm info attempt registry request try #1 at 3:38:13 PM
npm http request GET https://registry.npmjs.org/mortice
npm http 200 https://registry.npmjs.org/mortice
 
[ '0.0.1',
  '0.0.2',
  '1.0.0',
  '1.0.1',
  '1.1.0',
  '1.2.0',
  '1.2.1',
  '1.2.2' ]

npm timing npm Completed in 343ms
npm info ok 
```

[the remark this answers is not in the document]
achingbrain: https://www.npmjs.com/package/mortice/v/1.2.2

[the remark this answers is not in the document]
achingbrain: Weirdly libraries.io only shows it at v1.2.1 🤷‍♂  https://libraries.io/npm/mortice

src/core/components/pin/gc-lock.js
achingbrain: Elsewhere we use the [`assert`](https://nodejs.org/api/assert.html) module to guard against programmer errors.

package.json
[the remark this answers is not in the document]
dirkmc: Ahh thanks! I didn't see a release commit for 1.2.2 on the repo so I got confused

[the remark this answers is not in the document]
achingbrain: Ah, that was my fault - forgot to push it & the tags. All done now.

src/core/components/pin/gc-lock.js
dirkmc: Thanks, that's a better idea 👍

alanshaw:
Needs `interface-ipfs-core@0.106`

src/core/components/pin/gc-lock.js
alanshaw: 👍 yep good point

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,138 @@
+'use strict'
+
+const assert = require('assert')
+const mortice = require('mortice')
+const pull = require('pull-stream')
```
achingbrain:
```suggestion
const pull = require('pull-stream/pull')
```

https://www.npmjs.com/package/pull-stream#minimal-bundle

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+
+    self._gcLock.writeLock((lockCb) => {
+      parallel([
+        // Get all blocks from the blockstore
+        (cb) => self._repo.blocks.query({ keysOnly: true }, cb),
+        // Mark all blocks that are being used
+        (cb) => createMarkedSet(self, cb)
+      ], (err, [blocks, markedSet]) => {
+        if (err) {
+          log(`Error - ${err.message}`)
```
alanshaw: Please log the full error object and add an appropriate message - this is so we get the stack trace - it's no use if someone turns on `DEBUG` and is not able to debug due to no stack.

```suggestion
          log(`Failed to fetch all block keys and create marked set`, err)
```

Please also fix other instances!

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+          return lockCb(err)
+        }
+
+        // Delete blocks that are not being used
+        deleteUnmarkedBlocks(self, markedSet, blocks, start, (err, res) => {
```
alanshaw: Can we keep `start` local to this function and log out "complete" in the callback here instead?

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+const CID = require('cids')
+const base32 = require('base32.js')
+const parallel = require('async/parallel')
+const mapLimit = require('async/mapLimit')
+const { Key } = require('interface-datastore')
+const log = require('debug')('ipfs:gc')
+
+// Limit on the number of parallel block remove operations
+const BLOCK_RM_CONCURRENCY = 256
+const MFS_ROOT_DS_KEY = new Key('/local/filesroot')
```
alanshaw: Can this be imported from `ipfs-mfs`? Would be nice if we didn't have to reach all the way in to https://github.com/ipfs/js-ipfs-mfs/blob/eb04d6d4cb9e3cfdb617ad29a271dc59af9dfd87/src/core/utils/constants.js#L13 cc @achingbrain

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+    }),
+
+    // The MFS root and all its descendants
+    (cb) => ipfs._repo.root.get(MFS_ROOT_DS_KEY, (err, mh) => {
+      if (err) {
+        if (err.code === 'ERR_NOT_FOUND') {
+          log(`No blocks in MFS`)
+          return cb(null, [])
+        }
+        return cb(new Error(`Could not get MFS root from datastore: ${err.message}`))
```
alanshaw: It would be nice not to lose the stack from this error! There's quite a lot of instances in this PR where this is happening and it'll make debugging really difficult.

If you want to add context to the error then I'd recommend using [`explain-error`](https://www.npmjs.com/package/explain-error).

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+    log(`Found ${cids.length} MFS blocks`)
+    // log('  ' + cids.join('\n  '))
+    callback(null, cids)
+  })
+}
+
+// Delete all blocks that are not marked as in use
+function deleteUnmarkedBlocks (ipfs, markedSet, blocks, start, callback) {
+  // Iterate through all blocks and find those that are not in the marked set
+  // The blocks variable has the form { { key: Key() }, { key: Key() }, ... }
```
alanshaw: Do you mean?:

```suggestion
  // The blocks variable has the form [ { key: Key() }, { key: Key() }, ... ]
```

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+  const msg = `Marked set has ${markedSet.size} unique blocks. Blockstore has ${blocks.length} blocks. ` +
+    `Deleting ${unreferenced.length} blocks.` + (errCount ? ` (${errCount} errors)` : '')
+  log(msg)
+  // log('  ' + unreferenced.join('\n  '))
+
+  mapLimit(unreferenced, BLOCK_RM_CONCURRENCY, (cid, cb) => {
+    // Delete blocks from blockstore
+    ipfs._repo.blocks.delete(cid, (err) => {
+      const res = {
+        cid: cid.toString(),
```
alanshaw: Leave it as a CID?

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+    callback(null, res.concat(delRes))
+  })
+}
+
+function dsKeyToCid (key) {
+  // Block key is of the form /<base32 encoded string>
+  const decoder = new base32.Decoder()
+  const buff = decoder.write(key.toString().slice(1)).finalize()
+  return new CID(Buffer.from(buff))
+}
```
alanshaw: Would be nice if this util was exported from repo.

src/http/api/resources/repo.js
```diff
@@ -1,9 +1,25 @@
...
+  },
+
+  async handler (request, h) {
+    const streamErrors = request.query['stream-errors']
+    const { ipfs } = request.server.app
+    const res = await ipfs.repo.gc()
+
+    const filtered = res.filter(r => !r.err || streamErrors)
+    const response = filtered.map(r => {
+      return { Err: r.err, Key: !r.err && { '/': r.cid } }
```
alanshaw: `r.err` is an `Error`? Does it need to be changed to `r.err.message`?

`r.cid` this should be a CID instance right? Does it need a `toString()`?

test/cli/repo.js
```diff
@@ -18,4 +19,30 @@ describe('repo', () => runOnAndOff((thing) => {
       expect(out).to.eql(`${repoVersion}\n`)
     })
   })
+
+  // Note: There are more comprehensive GC tests in interface-js-ipfs-core
```
alanshaw: Can we also have a test (or two) that mocks IPFS and asserts that `stream-errors` is passed from argv to the mock correctly and that the output is formatted appropriately given canned results from the mock?

Ideally I'd like to move towards having just a few e2e CLI tests (since we have comprehensive core tests) and many unit tests for the handlers that simply ensure args are converted appropriately for core and that output from core is formatted appropriately in the CLI.

I think it would also help to explicitly group types of tests in a directory "unit", "e2e" etc.

refs https://github.com/ipfs/js-ipfs/issues/2131
cc @hugomrdias

src/core/components/block.js
```diff
@@ -81,17 +81,19 @@ module.exports = function block (self) {
             cb(null, new Block(block, cid))
           })
         },
-        (block, cb) => self._blockService.put(block, (err) => {
-          if (err) {
-            return cb(err)
-          }
+        (block, cb) => self._gcLock.readLock((_cb) => {
```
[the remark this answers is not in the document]
dirkmc: I think we don't want to write to the blockstore while GC is running, because depending on where GC is up to while iterating through the blocks, the block may or may not be GCed, so it results in non-deterministic behaviour.

src/core/components/block.js
```diff
@@ -100,7 +102,8 @@ module.exports = function block (self) {
       } catch (err) {
         return setImmediate(() => callback(errCode(err, 'ERR_INVALID_CID')))
       }
-      self._blockService.delete(cid, callback)
+
+      self._gcLock.writeLock((cb) => self._blockService.delete(cid, cb), callback)
```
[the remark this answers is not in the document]
dirkmc: We want to make sure that adding a block and removing a block are exclusive operations.

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,138 @@
...
+const pullThrough = require('pull-stream/throughs/through')
+const pullAsyncMap = require('pull-stream/throughs/async-map')
+const EventEmitter = require('events')
+const log = require('debug')('ipfs:gc:lock')
+
+class GCLock extends EventEmitter {
```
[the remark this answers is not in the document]
dirkmc: It simplifies testing, eg [whereever p-event is used](https://github.com/ipfs/js-ipfs/pull/2022/files#diff-c3c5bb981f284f576bb39b60a89014d4R76) in gc.spec.js

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,138 @@
...
+          // Request the lock
+          this.lock = this.mutex[this.type](() => this.locked())
+          // If there is an error, it gets passed through to the caller using
+          // pull streams, so here we just catch the error and ignore it so
+          // that there isn't an UnhandledPromiseRejectionWarning
+          this.lock.catch(() => {})
+        }
+
+        // Wait for the mutex to give us permission
+        this.lockReady.then(() => cb(null, i))
```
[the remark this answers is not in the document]
dirkmc: Calling `cb` will actually execute the locked piece of code - we don't want to do that until the mutex has given us permission

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,138 @@
...
+    )
+  }
+
+  // Releases the lock
+  release () {
+    return pullThrough(null, (err) => {
+      log(`[${this.lockId}] ${this.type} (pull) released`)
+      this.emitter.emit(`${this.type} release`, this.lockId)
+      this.releaseLock(err)
+    })
```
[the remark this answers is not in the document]
dirkmc: That's a great point, locking with pull streams is very tricky. Maybe it's better to take out the pull stream locking code, and instead have an explicit `lock` and `release` at the start and end of the locked portion of code?

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+
+// Perform mark and sweep garbage collection
+module.exports = function gc (self) {
+  return promisify((callback) => {
+    const start = Date.now()
+    log(`Creating set of marked blocks`)
+
+    self._gcLock.writeLock((lockCb) => {
+      parallel([
+        // Get all blocks from the blockstore
```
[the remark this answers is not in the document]
dirkmc: Oh really? Does `repo.blocks.query({ keysOnly: true }, cb)` get filtered somehow?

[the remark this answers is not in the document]
alanshaw: I just meant the line of code afterwards is getting the keys not the _blocks_.

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+        (cb) => self._repo.blocks.query({ keysOnly: true }, cb),
```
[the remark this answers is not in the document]
dirkmc: I think go-ipfs uses streaming, but we don't have a streaming API until we switch to async/await

src/core/components/pin/gc.js
```diff
@@ -0,0 +1,153 @@
...
+        // Mark all blocks that are being used
+        (cb) => createMarkedSet(self, cb)
+      ], (err, [blocks, markedSet]) => {
```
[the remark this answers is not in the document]
dirkmc: These aren't really keys or blocks, each one is `{ key: Key() }`. So I will change the name to blockKeys

src/core/components/pin/gc.js
dirkmc: Agreed :+1:

src/core/components/pin/gc.js
dirkmc: That's true, @vasco-santos did this in https://github.com/ipfs/js-ipfs/pull/2184/files#diff-47300e7ecd8989b6246221de88fc9a3cR142

vasco-santos: As we eventually want to move the repo codebase to this repo, I think we should have it here. What do you think?

src/core/components/pin/gc.js
dirkmc: I wanted to add context for a user who gets something like 'ENOENT file descriptor invalid' and has no idea what it means. explain-err looks like a great way to help with this 👍

src/core/components/pin.js
```diff
@@ -2,171 +2,25 @@
...
-
-      // save root to datastore under a consistent key
-      cb => repo.datastore.put(pinDataStoreKey, root.multihash, cb)
-    ], (err, res) => {
-      if (err) { return callback(err) }
-      self.log(`Flushed pins with root: ${root}`)
-      return callback(null, root)
-    })
-  }
+  const pinManager = new PinManager(self._repo, dag, self.log)
```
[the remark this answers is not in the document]
dirkmc: It's mostly a refactor - moving things into separate files and everything related to pinning into its own directory. But there are also some changes, like adding locking in some places

src/cli/commands/repo/gc.js
```diff
@@ -1,16 +1,37 @@
...
-    argv.resolve((async () => {
-      const ipfs = await argv.getIpfs()
-      await ipfs.repo.gc()
+  handler ({ getIpfs, quiet, streamErrors, resolve }) {
+    resolve((async () => {
+      const ipfs = await getIpfs()
+      const res = await ipfs.repo.gc()
+      for (const r of res) {
+        if (r.err) {
+          streamErrors && print(r.err, true, true)
```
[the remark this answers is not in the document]
dirkmc: You're right this is not actually streaming - unfortunately this parameter name is inherited from go-ipfs: `--stream-errrors`

At present the repo blocks API isn't streaming so it doesn't really make a difference if we present a streaming API. Maybe better to wait until we convert js-ipfs to async/await?

src/core/components/pin/gc.js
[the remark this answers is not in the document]
dirkmc: Thanks I'll fix the comment

daviddias:
Very quick question while I'm on the Core Implementations call: Does js-ipfs GC take into account the ephemerallity of IndexedDB? Check the LRU policy to see how this will affect pinning, gc and even fetching of large datasets https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria#LRU_policy

src/core/components/pin/gc.js
dirkmc: Yes I think that makes sense, I'll add it

dirkmc: Actually sorry I'm just noticing that @alanshaw is suggesting this be in repo, which I think makes more sense

src/cli/commands/repo/gc.js
```diff
@@ -1,16 +1,37 @@
...
+        } else {
+          print((quiet ? '' : 'Removed ') + r.cid)
```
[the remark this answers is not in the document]
alanshaw: Is it the same behaviour as go-ipfs?

src/cli/commands/repo/gc.js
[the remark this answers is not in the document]
alanshaw: Can you open an issue to track this task?

alanshaw:
Thanks for raising that @daviddias!

If I understand this article correctly, it seems browser storage space (IndexedDB+Web Storage+more) is limited on disk space at a number of levels and when it becomes full, LRU eviction comes into play.

LRU deletes storage for an _entire origin_ at a time, so we won't randomly lose blocks of data from our datastore which, in short means that this policy doesn't really effect our implementation of GC.

That said, actually having a GC implementation in JS IPFS will reduce the chance of LRU eviction happening in the first instance.

CID in subdomain will help with not losing ALL data for ALL sites you access via the gateway. Currently `ipfs.io/ipfs/QmHash0` and `ipfs.io/ipfs/QmHash1` are at the same origin so if LRU eviction picks the `ipfs.io` origin you'll lose ALL data the sites at `ipfs.io/ipfs/QmHash0` and `ipfs.io/ipfs/QmHash1` were storing.

With CID in subdomain `QmHash0.ipfs.dweb.link` and `QmHash1.ipfs.dweb.link` are separate origins (although in the same "group") so LRU eviction will be more granular.

It's worth mentioning that even with CID in subdomain all the sites accessed will still be classed as being in the same "group" and will have the same group limit (less than the global limit).

cc @lidel

src/cli/commands/repo/gc.js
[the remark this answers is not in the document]
dirkmc: https://github.com/ipfs/js-ipfs/issues/2266

src/cli/commands/repo/gc.js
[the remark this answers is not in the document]
dirkmc: The way [the code is written](https://github.com/ipfs/go-ipfs/blob/master/core/commands/repo.go#L79) is a little confusing, but yes, the overall effect is that setting --stream-errors to false results in errors not being emitted to the response stream

obo20:
@dirkmc Something that isn't immediately clear to me upon reading this PR: 

Does this implementation of GC "stop the world" for the IPFS node? By this I mean, can operations such as adding / pinning / unpinning take place while garbage collecting?

While some garbage collection is better than no garbage collection at all, I want to bring up these issues on the go-ipfs side of things:
* https://github.com/ipfs/go-ipfs/issues/4382
* https://github.com/ipfs/go-ipfs/issues/6114

I know for us, this inability to GC while maintaining normal operations is fairly problematic and is something we're constantly having to create work arounds for.

dirkmc:
@obo20 [we discussed](https://github.com/ipfs/js-ipfs/issues/2012) whether to implement something more complex (eg reference counting) as part of this PR but landed on doing the simplest thing first, which is stop-the-world mark and sweep

obo20:
@dirkmc That makes sense. From your experiences in implementing the current version are there any major concerns you have with implementing a passive version GC? As far as I'm aware, I don't think anybody has touched GC for awhile so it would be nice to hear your perspective on things.

dirkmc:
Something I didn't consider initially is that the mark-and-sweep model can defer a lot of work until GC time, which is useful for some use cases in which the performance of adding files is very important, and performance of GC is less important (because it can be performed at specific moments when the system is not resource constrained). For example if a user wants to regularly add a large directory of files into IPFS, and perform GC in between. Reference counting would require creating a reference for every block.

obo20:
@dirkmc Just to clarify on my end, is the mark-and-sweep model you're referring to essentially this: 

1) We start garbage collection and start marking everything we need to collect. 
2) If we get a new file add operation, we pause GC and take note of every block that was just added
3) Compare the new blocks from the add operation against the current list and see if any blocks marked for "gc" were pinned during the new add operation. If so, remove them from the list to be GC'd
4) Resume the GC process, pausing again each time a locking operation (such as pinning) takes place.

dirkmc:
It's simpler than that - we just stop the world while performing mark-and-sweep. ie no blocks are allowed to be changed until GC completes. So if a file is added during GC it will pause until GC has completed before adding the file.

obo20:
@dirkmc From what you've seen is there any hidden dangers I'm not considering in the approach I described in my previous comment? 

Note, I'm not asking you to change this PR. I'm more curious about what the next steps could be to make this whole GC process truly passive.

dirkmc:
No worries these are good questions 👍 

I can see a couple of potential issues
- for a system with high churn, GC may never actually be allowed to complete
- pausing and resuming GC when a new file is added is more complex than it may appear. We'd need to
  - keep track of what has already been GCed and what is yet to be GCed
  - keep track of what new blocks are created
  - compare those lists
  - manage locking and unlocking between GC, the pinner and the blockstore
- all that adds to code complexity and requires non-deterministic memory usage

obo20:
Thanks for answering the questions. This is something pretty important to us, so any work in the GC area is of high interest to me. From what you're saying it sounds like ref counting is probably still the best approach, even though it takes up extra resources.

dirkmc:
I think it really depends on the use case. Persistent reference counting is fairly complex, because in order to be performant it will probably require using a write-ahead log. On the other hand it should simplify the locking that we are doing currently.

obo20:
Could it be something that users choose in a config setting? Or would that be too complex to account for both GC strategies in one codebase? 

You're right in that many casual users will be fine locking their nodes during a GC, as often these nodes are on laptops that only occasionally would add something. However anybody running IPFS in production would likely take the ref counting strategy to avoid locking their infrastructure up.

Having both strategies available would be nice, as long as it didn't drastically complicate the code-base

dirkmc:
I think we need to consider that choice carefully. This PR is to implement the simplest GC mechanism, once it is merged and working then we will look at more sophisticated strategies.

Stebalien:
I'd avoid having both. The downside of refcounting is complexity. Supporting both mark-and-sweep _and_ refcounting just adds more complexity.

But dirk is right, this is all beyond the scope of this PR.

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,138 @@
...
+class GCLock extends EventEmitter {
+  constructor (repoOwner) {
+    super()
+
+    // Ensure that we get a different mutex for each instance of GCLock
+    // (There should only be one GCLock instance per IPFS instance, but
+    // there may be multiple IPFS instances, eg in unit tests)
+    const randId = (~~(Math.random() * 1e9)).toString(36) + Date.now()
+    this.mutex = mortice(randId, {
+      singleProcess: repoOwner
```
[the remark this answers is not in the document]
hugomrdias: for any "global this" resolution please use this approach https://mathiasbynens.be/notes/globalthis 
never trust the user to pass the correct global scope!

@achingbrain you could add this directly in mortice

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
hugomrdias: please please don't use events, use signals or hooks instead 🙏 

http://blog.millermedeiros.com/callbacks-promises-signals-and-events/

this can be just a couple of hooks functions passed in the options, it will be much easier to debug, discover, mock, etc.. no need to clean listeners

test/cli/repo.js
hugomrdias: yes please only test cli logic in cli tests, you can check the ipns cli tests for examples 

there are still some weird in there because of how the cli works, but we will make that better this quarter

test/core/gc.spec.js
```diff
@@ -0,0 +1,245 @@
...
+  before(function (done) {
+    this.timeout(20 * 1000)
+    repo = createTempRepo()
+    let config = { Bootstrap: [] }
+    if (isNode) {
+      config.Addresses = {
+        Swarm: ['/ip4/127.0.0.1/tcp/0']
+      }
+    }
+    ipfs = new IPFS({ repo, config })
```
hugomrdias: why dont you use ipfsd-ctl ?

test/core/gc.spec.js
```diff
@@ -0,0 +1,245 @@
+/* eslint max-nested-callbacks: ["error", 8] */
+/* eslint-env mocha */
+'use strict'
+
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const expect = chai.expect
+chai.use(dirtyChai)
+
+const isNode = require('detect-node')
```
hugomrdias: i would suggest you use our ipfs-utils for this https://github.com/ipfs/js-ipfs-utils/blob/master/src/env.js#L15 it's safer

test/core/gc.spec.js
```diff
@@ -0,0 +1,245 @@
...
+    add2: () => ipfs.object.put({ Data: 'obj put 2', Links: [] }),
+    resToCid: (res) => res.toString()
+  }, {
+    name: 'block put',
+    add1: () => ipfs.block.put(Buffer.from('block put 1'), null),
+    add2: () => ipfs.block.put(Buffer.from('block put 2'), null),
+    resToCid: (res) => res.cid.toString()
+  }]
+
+  describe('locks', function () {
```
hugomrdias: all theses tests are specific to js-ipfs ? or can be interface-core tests ?

src/cli/commands/repo/gc.js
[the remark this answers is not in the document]
alanshaw: True, the don't get emitted, but they are returned...I wonder if the CLI prints them at the end?

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
alanshaw: It would be nice to not have code in src that is exclusively for testing. Is there another way you can perform these tests?

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
alanshaw: When you get the _first_ chunk, `this.lock` will be `undefined`, it gets created and you wait for the lock to be ready. When ready you call the callback `cb`.

For the second chunk, `this.lock` will **not** be `undefined` and you will have acquired the lock so you can call `cb` straight away.

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
alanshaw: That's kinda what you're doing isn't it?...but in a helper function that adds both `take` and `release` to the pipeline.

How would you see this working for `addPullStream`?

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
dirkmc: Fair enough - I will try writing it using sinon. Would that make sense to you Hugo?

test/core/gc.spec.js
dirkmc: These tests are quite specific and need to be able to control timing using events / hooks, so I think it makes most sense for them to be here

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
hugomrdias: note: if you mix callbacks with promises **always** call the callback inside a setImmediate

src/core/components/pin/gc.js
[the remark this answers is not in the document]
alanshaw: To be clear, I meant the entire GC process.

We can always "fake" streaming for APIs we cannot yet stream from e.g. `pull.values()` and that would save us a bunch of refactoring when we're able to stream from them.

As it stands, we're in the middle of a big refactor to switch to async await and async iterators so we're going to have to refactor this in some way anyway.

test/core/gc.spec.js
alanshaw: It would be great to remove the `detect-node` dependency if this is available in `ipfs-utils` already.

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
hugomrdias: sinon is gonna be tricky for some of them but yeah it's better.

im trying to find a better solution but without a declarative way to handle the lifecycle of this code it's hard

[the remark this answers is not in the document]
dirkmc: hmm yeah I'm not really seeing an easy way to do it, because the PullLocker object gets created when a pull lock is needed, but the lock is only taken when the pull stream starts

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
dirkmc: @alanshaw you're right, that makes sense 👍

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
dirkmc: You're right there isn't really another way to do it. I think your suggestion makes the most sense. We'd need to think about how to implement that in a way that doesn't make this even more complex.

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
hugomrdias: https://v8.dev/features/globalthis

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
hugomrdias: Fresh out of Twitter https://twitter.com/davidmarkclem/status/1151540675093127168?s=20 check up and down the thread

src/core/components/pin/gc-lock.js
```diff
@@ -0,0 +1,135 @@
...
+const EventEmitter = require('events')
+const log = require('debug')('ipfs:gc:lock')
+
+class GCLock extends EventEmitter {
+  constructor (repoOwner) {
+    super()
+
+    // Ensure that we get a different mutex for each instance of GCLock
+    // (There should only be one GCLock instance per IPFS instance, but
+    // there may be multiple IPFS instances, eg in unit tests)
```
daviddias: IPFS locks the repo when using it. There should never be two IPFS nodes running over the same repo. If you happen to caught it not behaving as expected, then it is a bug on the ipfs-repo module.

package.json
```diff
@@ -139,13 +141,16 @@
...
     "mime-types": "^2.1.21",
     "mkdirp": "~0.5.1",
+    "mortice": "^1.2.2",
     "multiaddr": "^6.0.5",
     "multiaddr-to-uri": "^4.0.1",
     "multibase": "~0.6.0",
     "multicodec": "~0.5.1",
     "multihashes": "~0.4.14",
     "multihashing-async": "~0.6.0",
+    "nanoid": "^2.0.3",
```
daviddias: We already use `hat`, does `nanoid` add anything that `hat` doesn't provide?

hugomrdias: Doesn't rely on `Math`

src/core/components/pin/gc-lock.js
dirkmc: @daviddias the issue here is that mortice caches the lock by name, so in order to avoid different unit tests interfering with each other we need to either make sure the name is unique, or clear the require cache for mortice before each test (debatably also a good option)

dirkmc:
There are several concerns raised in the comments. I'd like to summarize the main ones here so they don't get lost:

1. gc-lock.js is complex and mixes concepts (EventEmitter, callbacks, Promises)
  - The Promises come from using the mortice locking library. We are using it for a couple of reasons:
    - it is already used for locking by MFS
    - we are moving js-ipfs toward an async/await based API, so this should save us some future work
  - EventEmitter is used because it makes it easier to test things that depend on specific timing of locks
    - I tried using sinon instead, but it requires internal knowledge of how the class works, so it makes things even more complex
    - I agree with Alan that it's better not to add functionality to src that is only used in test
    - GCLock is a singleton constructed when the API is initiated, so I'm not sure if there's a better approach? I'm open to suggestions
  - The mixing of concepts is partly a reflection of js-ipfs, as a large project that is moving from one paradigm to another
    - My thinking was that it's best to try to move all the complexity into one place and test it thoroughly
    - There are comprehensive tests in [gc-lock.spec.js](https://github.com/ipfs/js-ipfs/pull/2022/files#diff-d95cb8e4b7ed2d460b544f2e77da0de2)

2. Slow reading of the source or draining of the sink may cause ipfs.add() to delay GC from running

3. There should be a streaming API for GC

4. We should add tests to make sure CLI parameters are passed from argv to the core correctly

5. GC should be smarter (reference counting, origin storage limits etc)

My feeling is that given how complex this PR already is, and the number of ideas raised, it's probably worth getting a first implementation of GC out there and adding the nice-to-haves in further PRs. I think for this PR we should try to:
- add the CLI tests
- (maybe) replace EventEmitter with a different mechanism

Thoughts?

daviddias:
Thank you for summarizing the feedback, @dirkmc. My suggestion is to label these with their level of importance/"criticality" and then contemplate if it will be problematic to merge the most critical ones or if it will be easy to refactor/revert in the future.

I would say that specially number 1 should not be overlooked as it signals significant code complexity that reduced the ability of some folks to review it fully. My advice is for you to refactor the code while your mind still has all of the ideas and how it works vs. waiting for someone else to have to untangle it in the future

src/core/components/pin/gc-lock.js
hugomrdias: we can also improve `mortice` to be more atomic right ?

dirkmc: @achingbrain I'm not sure on the answer to this one, could you comment?

dirkmc:
@daviddias makes sense, thanks 👍 

I will add an abstraction over mortice to present a callback-based API, I think that should help to simplify things. I'll also see if there's some way to simplify the pull-stream locking.

alanshaw:
> There are several concerns raised in the comments. I'd like to summarize the main ones here so they don't get lost:
>
>     1. gc-lock.js is complex and mixes concepts (EventEmitter, callbacks, Promises)
>
>     * The Promises come from using the mortice locking library. We are using it for a couple of reasons:
>
>       * it is already used for locking by MFS
>       * we are moving js-ipfs toward an async/await based API, so this should save us some future work

Can we create a small shim for `mortice` to pull the callback/Promise mixing out of the core code and make it more readable/maintainable and in keeping with the rest of the core code?

>     * EventEmitter is used because it makes it easier to test things that depend on specific timing of locks
>
>       * I tried using sinon instead, but it requires internal knowledge of how the class works, so it makes things even more complex
>       * I agree with Alan that it's better not to add functionality to src that is only used in test
>       * GCLock is a singleton constructed when the API is initiated, so I'm not sure if there's a better approach? I'm open to suggestions

Just a thought, but you could create a pretty minimal helper class that adds in the events you're after for testing. It would remove EventEmitter from `src` at least:

```js
// Something like this...?
class GCLockTester extends GCLock {
  constructor (emitter) {
    super()
    this._emitter = emitter
  }

  lock (type, lockedFn, cb) {
    this._emitter.emit(`${type} request`, lockId)
    return super.lock(type, cb => {
      this._emitter.emit(`${type} start`, lockId)
      lockedFn((err, res) => {
        this._emitter.emit(`${type} release`, lockId)
        cb(err, res)
      })
    }, cb)
  }
}
```

In tests:

```js
describe('GCLock', () => {
  let ipfs, lockEmitter

  before(() => {
    // ...spawn ipfs node
    lockEmitter = new EventEmitter()
    ipfs._gcLock = new GCLockTester(lockEmitter)
  })

  it('test', () => {
    const addLockRequested = pEvent(lockEmitter, 'readLock request') // ...for example
  })
})
```

When we switch to async/iterators the pull streams code can probably be removed and locking will be easy like:

```js
ipfs.add = source => importer((async function * () {
  await ipfs._gcLock.take()
  for await (const chunk of source) yield chunk
  await ipfs._gcLock.release()
})())
```

>     1. Slow reading of the source or draining of the sink may cause ipfs.add() to delay GC from running

Low priority, nothing will be broken by this not being addressed. I think we can easily address this on the sink side.

>     2. There should be a streaming API for GC

Medium-High priority. In this PR we don't read all the data into memory but we do read all the keys. Nodes hosting HUGE datasets may experience out of memory from day one.

>
>     3. We should add tests to make sure CLI parameters are passed from argv to the core correctly

Low priority - I figure this is just two short tests with mocks.

>
>     4. GC should be smarter (reference counting, origin storage limits etc)

Medium priority - definitely a separate PR(s) though

dirkmc:
@alanshaw I like these ideas, thanks 👍 

> We don't read all the data into memory but we do read all the keys. Nodes hosting HUGE datasets may experience out of memory from day one

In version 0.26.6 of js-ipfs-repo, [repo.blockstore.query()](https://github.com/ipfs/js-ipfs-repo/blob/v0.26.6/src/blockstore.js#L61) returns all the blocks from the query in an array.

The next version of js-ipfs-repo is 0.27.0 which replaces async/await with callbacks.

So we would probably need to either
- fork js-ipfs-repo at version 0.26.6 to change the API to return a pull stream
- switch to using the async/await version of js-ipfs-repo
- do some kind of hack that reaches into the internals of js-ipfs-repo

alanshaw:
🙄 only forwards! Time to bring in some more async/await deps.

src/utils/mutex.js
```diff
@@ -0,0 +1,72 @@
...
+  * @param {function(releaseLock)} lockedFn A function that runs the locked piece of code and calls releaseLock when it completes
+  * @param {function(err, res)} [cb] A function that is called when the locked function completes
+  * @returns {void}
+  */
+  _lock (type, lockedFn, cb) {
+    assert(typeof lockedFn === 'function', `first argument to CBLock.${type}() must be a function`)
+
+    if (typeof cb === 'undefined') {
+      cb = noop
+    }
```
hugomrdias: can we do 
```
_lock (type, lockedFn, cb = noop) {
```

and remove

```
if (typeof cb === 'undefined') {
      cb = noop
}
```

?

src/core/components/pin/gc-lock.js
```diff
@@ -54,26 +48,30 @@ class PullLocker {
...
       })
-    )
+    })
   }
 
   // Releases the lock
   release () {
     return pullThrough(null, (err) => {
+      // When the stream completes, release the lock
       this.releaseLock(err)
```
hugomrdias: this doesn't need to be protected ? like

```
if(this.releaseLock) {
	this.releaseLock(err)
	this.releaseLock = null
}else{
	throw new Error('No lock to release')
}
```

dirkmc: Because `release()` is called after the stream completes, it shouldn't be possible to get here without a lock having been acquired (and `this.releaseLock` being set)

hugomrdias: what happen if `lockedPullFn()` doesnt work properly, throws or break flow ?

src/core/components/pin/gc-lock.js
achingbrain: > mortice caches the lock by name

The locks need to be accessible to the mortice module outside of the creating context in order to have the default lock work transparently across multiple cluster workers or web workers as implemented.

So if we're creating multiple in-process IPFS nodes, they'll fight over the lock name even if they use different repos.  Presumably we could make the lock name a config option or use something unique to the repo like the peer ID?

src/core/components/pin/gc-lock.js
dirkmc: `lockedPullFn()` returns a pull stream, so it depends on what that pull-stream does

hugomrdias: what im asking is if `lockedPullFn()` does something wrong is this code still safe ?

src/utils/mutex.js
```diff
@@ -0,0 +1,68 @@
...
+const noop = () => {}
+
+// Wrap mortice to present a callback interface
+class Mutex {
+  constructor (repoOwner, options = {}) {
+    // Ensure that we get a different mutex for each instance of the lock
+    const randId = nanoid()
+    this.mutex = mortice(randId, {
+      singleProcess: repoOwner
+    })
```
hugomrdias: we need to handle the `global` option
either we do this here or inside mortice, imo the second option is better.

i have added globalThis support to ipfs-utils here https://github.com/ipfs/js-ipfs-utils/blob/master/src/globalthis.js

`const globalThis = require('ipfs-utils/src/globalthis')`
or even just 
`require('ipfs-utils/src/globalthis')`

@achingbrain thoughts ? should we PR mortice ?

src/core/components/pin/gc-lock.js
dirkmc: I don't believe that will affect anything in the `PullLocker.release()` function - the second callback to pullThrough() should only be invoked once the stream ends.

If an exception is thrown from within the pull-stream returned by lockPullFn() I believe it will be thrown outside the control of the locking class.

src/utils/mutex.js
achingbrain: > we need to handle the `global` option

I'm not sure I follow - what's the problem you are trying to solve?

hugomrdias: mortice defaults the global option to `global` which isnt safe for all js envs

hugomrdias: with my suggestion you can remove the option and it will always have the right globalThis

achingbrain: Gotcha. Has this caused problems?

I've put a PR together that removes the option [here](https://github.com/achingbrain/mortice/pull/2).

achingbrain: I ask because mortice has tests that runs across browsers, webworkers, node and node cluster and they all pass, so if it's detecting the wrong global as it stands, I'd like to know about it!

dirkmc:
Rebased onto master - brings in a [pinning fix](https://github.com/ipfs/js-ipfs/pull/2228) and [better print handling](https://github.com/ipfs/js-ipfs/pull/2232)

src/utils/mutex.js
hugomrdias: you assume browserify is used and global is defined not always true

src/core/components/pin/gc-lock.js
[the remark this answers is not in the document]
alanshaw: Sounds like this is being resolved upstream 🐟 🚀

alanshaw:
@hugomrdias: want to give this a final review?

The current plan is to get this merged as it stands and then follow up with a PR to switch to streaming in order to avoid out of memory for big data sets.

src/core/components/pin/pin-manager.js
```diff
@@ -0,0 +1,304 @@
...
+        }, (err, cid) => {
+          if (!err) {
+            root.multihash = cid.buffer
+          }
+          cb(err)
+        })
+      },
+
+      // hack for CLI tests
+      cb => this.repo.closed ? this.repo.open(cb) : cb(null, null),
```
hugomrdias: do we need this hack?

src/core/components/pin/pin-manager.js
```diff
@@ -0,0 +1,304 @@
...
+      if (err) { return callback(err) }
+      this.log(`Flushed pins with root: ${root}`)
+      return callback(null, root)
+    })
+  }
+
+  load (callback) {
+    waterfall([
+      // hack for CLI tests
+      (cb) => this.repo.closed ? this.repo.datastore.open(cb) : cb(null, null),
```
hugomrdias: do we need this hack?

test/cli/repo.js
```diff
@@ -18,4 +19,30 @@ describe('repo', () => runOnAndOff((thing) => {
...
+  it('should run garbage collection', async function () {
```
hugomrdias: This test should be moved to core or interface-core.
Here we should test these lines https://github.com/ipfs/js-ipfs/blob/8c21da09615dcc12ad91bc7eec96ca5567a37de7/src/cli/commands/repo/gc.js#L26-L32 using the same approach used here https://github.com/ipfs/js-ipfs/blob/master/test/cli/name.js

We will make some changes during this quarter to make tests like these even simpler, faster and cleaner (reducing mocking and those pesky `clearModule`).

test/core/gc.spec.js
```diff
@@ -0,0 +1,279 @@
...
+      await pinLockRequested
+      const gc = ipfs.repo.gc()
+      const deleted = (await gc).map(i => i.cid.toString())
+      await pin1
+
+      // TODO: Adding pin for removed block never returns, which means the lock
+      // never gets released
+      // const pin2 = ipfs.pin.add(cid2)
+
+      // Confirm second second block has been removed
```
hugomrdias:
```suggestion
      // Confirm second block has been removed
```

src/core/components/pin/pin-manager.js
daviddias: And if it is, explain what the hack is and why it was necessary

src/core/components/object.js
```diff
@@ -242,19 +242,21 @@ module.exports = function object (self) {
...
-          (cid) => {
-            if (options.preload !== false) {
-              self._preload(cid)
-            }
-
-            callback(null, cid)
-          },
-          (error) => callback(error)
-        )
+        self._gcLock.readLock((cb) => {
```
daviddias: I might have missed the explanation, can someone link me to it or explain here why does GC need another lock on top of the Repo lock?

src/core/components/pin/pin-manager.js
dirkmc: Good question - this was in the original code that I [refactored over from pin.js](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/pin.js#L159), do you know why it might be there @daviddias?

test/cli/repo.js
dirkmc: Thanks that's very helpful - I added CLI tests based on this example

src/core/components/object.js
dirkmc: The GC lock is used to prevent blocks being added/removed from the blockstore while GC is running. My understanding is that the Repo lock is used to prevent multiple instances of js-ipfs-repo from opening the same path, is that correct?

daviddias: So what happens to all inflight blocks coming in Bitswap? Is there a test to see if js-ipfs OOMs when performing a GC and someone sends a ton of blocks through bitswap that get queued in memory?

src/core/components/pin/pin-manager.js
dirkmc: It seems like if I remove it the tests work fine, so I think I will just go ahead and do that. Good catch @hugomrdias 👍

test/core/gc.spec.js
```diff
@@ -0,0 +1,279 @@
...
+    return super.readLock(lockedFn, cb)
+  }
+
+  writeLock (lockedFn, cb) {
+    this.emitter.emit('writeLock request')
+    return super.writeLock(lockedFn, cb)
+  }
+}
+
+describe('gc', function () {
```
daviddias: A couple of questions:
- What is the largest repo size that gc has been tested with? 
- How do these tests compare to go-ipfs gc tests?

daviddias: I'm asking this as I expect people to use GC when their repo gets of a significant size (10GB +++) and not when it is small.

src/core/components/object.js
dirkmc: That's a good question - in general that's the problem with stop-the-world GC, it blocks everything up while GC is being performed.

test/core/gc.spec.js
dirkmc: Good point 👍

I have been testing locally with repos a few hundred MB in size:
```
$ du -h ~/.jsipfs | tail -1
621M	/Users/dirk/.jsipfs

$ time js-ipfs repo gc
real	0m4.141s
user	0m0.858s
sys	0m0.174s
```

I haven't performed stress testing with repos of the size you're suggesting.

I had a look at go-ipfs, it seems like there are [some sharness tests](https://github.com/ipfs/go-ipfs/blob/777d306f6e66e31a05f43a337bce272050407386/test/sharness/t0082-repo-gc-auto.sh) for GC, but I didn't see any stress tests.

dirkmc: I just tried with an 11GB repo and it ran in a few seconds. I will add a sharness test for stress-testing

daviddias: 11GB repo and how large of a GC? half of it?

dirkmc: GC removes all blocks that are not pinned (same as go-ipfs). In this case I added 11GB of unpinned files and then GCed them.

dirkmc: I added a [sharness test for GC](https://github.com/ipfs/js-ipfs/pull/2022/commits/3a4912f84d92d187df8314e12cbd3ebb1d009610), in which the number of files and their size can be easily adjusted. It seems like GC performance is linear:

Screen Shot 2019-07-23 at 1 30 17 PM

[Spreadsheet](https://docs.google.com/spreadsheets/d/1A3hcRjVGHc1UVACM6CoCX6kz6dOkRZqBD_CnHnv_UHQ/edit?usp=sharing)

src/core/components/pin/pin-manager.js
```diff
@@ -0,0 +1,299 @@
...
+
+const createPinSet = require('./pin-set')
+
+// arbitrary limit to the number of concurrent dag operations
+const concurrencyLimit = 300
+const PIN_DS_KEY = new Key('/local/pins')
+
+function toB58String (hash) {
+  return new CID(hash).toBaseEncodedString()
+}
```
alanshaw: Use `cidToString` utility instead

src/utils/mutex.js
```diff
@@ -0,0 +1,65 @@
...
+  /**
+  * Request a read or write lock
+  *
+  * @param {String} type The type of lock: readLock / writeLock
+  * @param {function(releaseLock)} lockedFn A function that runs the locked piece of code and calls releaseLock when it completes
+  * @param {function(err, res)} [cb] A function that is called when the locked function completes
+  * @returns {void}
+  */
+  _lock (type, lockedFn, cb = noop) {
+    assert(typeof lockedFn === 'function', `first argument to CBLock.${type}() must be a function`)
```
alanshaw:
```suggestion
    assert(typeof lockedFn === 'function', `first argument to Mutex.${type}() must be a function`)
```

src/utils/mutex.js
```diff
@@ -0,0 +1,65 @@
...
+    assert(typeof cb === 'function', `second argument to CBLock.${type}() must be a callback function`)
```
alanshaw:
```suggestion
    assert(typeof cb === 'function', `second argument to Mutex.${type}() must be a callback function`)
```

src/utils/mutex.js
```diff
@@ -0,0 +1,65 @@
...
+    })
+
+    // Get a Promise for the lock
+    const lock = this.mutex[type](locked)
+
+    // When the locked piece of code is complete, the Promise resolves
+    return lock.then(
+      (res) => setImmediate(() => cb(null, res)),
+      (err) => setImmediate(() => cb(err))
+    )
```
alanshaw: No need to use `setImmediate` - promise fullfilment is always on a future tick.

achingbrain:
I've opened https://github.com/ipfs/js-ipfs/pull/2387 against this branch which should fix the conflicts.

PR #2044, opened 2019-05-11
[WIP] feat: integration of tool for repo migration

AuHau:
Integrating https://github.com/AuHau/js-ipfs-repo-migrations/ for repo's migration.

The PR adds a new config option `repoDisableAutoMigration` that controls if the migration should be automatically applied or not.

### TODO

* [x] Integrate ipfs-repo-migrations
* [x] Write tests
* [x] Documentation of migrations 
* [ ] Add proper package name when `js-ipfs-repo-migrations` is released on npm

### Problems

 1. As the migration is part of the booting sequence, if a user wants to disable the auto-migrations to prevent applying new migrations using CLI `js-ipfs config --bool repoDisableAutoMigration true`, then the migrations are applied before turning off the migration mechanism.

### Questions

 1. If repo's version is higher then expected, should initialization fail or just log this state (current approach)?


AuHau:
Should there be also an option for `const node = new IPFS(options)` to disable migrations or config value is sufficient?

AuHau:
So the PR is ready for review. 

The build is failing on `npm install` because of the dependency on the async/await refactors PRs.

package.json
```diff
@@ -118,6 +118,7 @@
     "ipfs-mfs": "~0.10.2",
     "ipfs-multipart": "~0.1.0",
     "ipfs-repo": "~0.26.5",
+    "ipfs-repo-migrations": "AuHau/js-ipfs-repo-migrations#dev",
```
daviddias: Can't this module just be part of `ipfs-repo`?

AuHau: Well, good question 😅 Maybe bit philosophical and so I am afraid long answer is in need.

**tl;dr: Yes, there is a relationship between the `ipfs-repo-migrations` and `ipfs-repo`, but not necessarily one that would lead to merging these two package.s**

I am not strongly against, but what I have discovered during writing the `ipfs-repo-migrations` is the delicate relationship between migration (& tooling around it) and the model migrated (`ipfs-repo`), where each migration is very much hard-linked to a specific version of the model (that is pretty much desired characteristic). But that relationship directly translates to a link between migration and specific version of the `ipfs-repo` package. 

If you take the main use-case "migrating to the last version of repo", then I don't see any problem there. But since the migration package currently supports actions like "migrate to specific, but not necessarily latest version" or "revert to specific version", these actions would lead to an inconsistent relationship "the current version of the repo vs. what current version of `ipfs-repo` package expects" and hence most probably not expected behavior. 

Another problem is a reversion of the repo. If you want to use some older version of `ipfs-repo` and revert your current repo to that version, you would have to install two versions of `ipfs-repo`. The "latest" one that has all the migrations up to your current repo's version (and hence which can revert the data to desired repo's version) and the actual version of `ipfs-repo` that can work with this version. Of course, having `ipfs-repo-migration` package separate does not solve this per se, as you still have to install a higher version of this package, but it just "makes more sense" in this use case. (At least for me).

That said these problems are, I think, quiet edge-cases. Merging it with `ipfs-repo` would definitely ease the chores around the integration of this tool, but that is definitely not a point that should be considered.

*After this whole write-up and thinking it through, I still believe that these packages should be left not-merged, but I am open to other opinions and/or guidance.*

AuHau:
I have updated the PR as I have moved the integration directly to `js-ipfs-repo` to https://github.com/ipfs/js-ipfs-repo/pull/202

AuHau:
Currently waiting for integration of async/await version of `js-ipfs-repo` as part of #1670. Not sure I will take a bite on that.

package.json
daviddias: Got it! :)

AuHau:
Replaced with https://github.com/ipfs/js-ipfs/pull/2527

PR #2083, opened 2019-05-21
fix(cli): make swarm addrs more resilient

jacobheun:
Dont fail if decapsulation of ipfs fails. Go node addresses typically wont contain their peerid

Fixes https://github.com/ipfs/js-ipfs/issues/1575

### DIFF
--- a/src/cli/commands/swarm/addrs.js
+++ b/src/cli/commands/swarm/addrs.js
@@ -21,7 +21,14 @@ module.exports = {
         print(`${peer.id.toB58String()} (${count})`)
 
         peer.multiaddrs.forEach((addr) => {
-          const res = addr.decapsulate('ipfs').toString()
+          let res
+          try {
+            res = addr.decapsulate('ipfs').toString()
+          } catch (_) {
+            // peer addresses dont need to have /ipfs/ as we know their peerId
+            // and can encapsulate on dial.
+            res = addr.toString()
+          }
           print(`\t${res}`)
         })
       })
--- a/src/cli/commands/swarm/addrs.js
+++ b/src/cli/commands/swarm/addrs.js
@@ -35,7 +35,7 @@ module.exports = {
       })
 
       // Return the output for printing
-      return { data: output.join('\n'), argv}
+      return { data: output.join('\n'), argv }
     })())
   }
 }
--- a/src/cli/commands/swarm/addrs.js
+++ b/src/cli/commands/swarm/addrs.js
@@ -1,7 +1,5 @@
 'use strict'
 
-const print = require('../../utils').print
-
 module.exports = {
   command: 'addrs',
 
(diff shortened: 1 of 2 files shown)


alanshaw:
Nice! Please can we have a test?

jacobheun:
@alanshaw sure, do you have a recommendation for testing the cli, it's been a while. I don't believe I can change addresses on the node via the api, so the options I can think of would be to add a go node to the test suite (this feels like it should go an interop though), or just mock and test the cli handler directly.

alanshaw:
> just mock and test the cli handler directly.

This please 😁

jacobheun:
I added a test for the cli handler. Since it's verifying process.stdout, the test cleanup isn't as clean as I'd like. If we just use the afterEach hook we'll lost the mocha output. We could stub the print function, but that would require a change to how it's being used in the cli file in order for the stub to work.

If we planned on adding more cli tests that just target the handlers directly (which would make cli tests much faster) we could build out some utilities to make the individual tests easier/cleaner to write.

alanshaw:
@jacobheun if you change the handler to build up a string for the output and then return it it'll get [printed automatically](https://github.com/ipfs/js-ipfs/blob/master/src/cli/bin.js#L46). It might save you from having to stub `process.stdout` and allows you to more easily verify the output.

In the future we might want to think about passing `print` in via `argv` for easier testing.

jacobheun:
>  if you change the handler to build up a string for the output and then return it it'll get printed automatically.

Ah, perfect! I updated the tests to check the returned output. I think testing the output returned from the handler is great. Then we only need to verify that bin is calling print appropriately.

Here is some truncated output running locally against an 0.36 rc daemon:
```log
$ node src/cli/bin.js swarm addrs
QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z (11)
        /ip4/104.236.176.52/tcp/4001
        /ip6/2604:a880:1:20::1f9:9001/tcp/4001
        /ip6/::1/tcp/4001
        /ip6/2604:a880:1:20::1f9:9001/tcp/4001
        /ip6/fce3:c53b:c3c5:2f54:8bb0:b6d9:898e:f140/tcp/4001
        /ip4/127.0.0.1/tcp/8081/ws
        /ip4/127.0.0.1/tcp/4001
        /ip4/104.236.176.52/tcp/4001
        /ip4/10.12.0.8/tcp/4001
        /ip4/172.17.0.1/tcp/4001
        /p2p-circuit
QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ (4)
        /ip4/104.131.131.82/tcp/4001
        /p2p-circuit
        /ip4/127.0.0.1/tcp/4001
        /ip4/104.131.131.82/tcp/4001
```

alanshaw:
Yeah it works for small non-streaming APIs, but print is still needed for commands like `ping`.

PR #2085, opened 2019-05-21
fix: use trickle builder in daemon mode too

achingbrain:
At the moment this doesn't work in daemon mode..

Depends on ipfs/js-ipfs-http-client#1015


achingbrain:
In the simplest case the balanced builder produces one unixfs `file` node with one or more unixfs `raw` or IPLD `raw` nodes as data leaves (depending on options passed) - when there's only one data leaf it will store the data on the `file` node instead - that is, collapse it into one node.

The trickle builder only ever stores data in leaves so the structure is different.

alanshaw:
I just [updated the dag vis to use 0.36](https://github.com/ipfs-shipyard/ipfs-dag-builder-vis/pull/10) and now I see.

Issue #1996, opened 2019-04-16
Add `ipfs repo stat --human`


alanshaw:
See https://github.com/ipfs/go-ipfs/pull/6223 and https://github.com/ipfs/go-ipfs/pull/6106 for details

jakehemmerle:
I'll take it!

alanshaw:
@jakehemmerle did you make any progress with this?

jakehemmerle:
@alanshaw sorry for the delay, just finished moving and I'm starting classes this week. I have made some progress but I'm still learning js-ipfs internals. Feel free to reopen this if it needs to get done within a few days

jakehemmerle:
@alanshaw it looks like the logic for changing the stats data to human readable format should be written in https://github.com/ipfs/js-ipfs-repo/blob/master/src/index.js inside the if block on line 314. I believe this ticket should get moved to that repo, unless I am missing something.

PR #2102, opened 2019-05-23
cli: add --human flag to repo stat & stats repo

chirag-shinde:
- Added a `humanize` object to /src/cli/utils with `Bytes` method
that takes bytes and returns Human readable size.
- Used `humanize.Bytes()` method in /commands/stats/repo and
/commands/repo/stat if `--human` flag is passed.

solves #1996 

License: MIT
Signed-off-by: Chirag Shinde <shinde.chirag2204@gmail.com>


chirag-shinde:
Screenshot of the output
Screenshot 2019-05-23 at 8 56 38 PM

src/cli/utils.js
```diff
@@ -135,3 +135,18 @@ exports.singleton = create => {
...
+    }
+    const units = ['kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+    let unit = -1
+    do {
+      bytes /= thresh
+      ++unit
+    } while (Math.abs(bytes) >= thresh && unit < units.length - 1)
+    return bytes.toFixed(1) + ' ' + units[unit]
+  }
+}
```
alanshaw: Can we please use a tested module on npm for this? `ipfs-http-response` already uses `filesize` - can we use the same module?

chirag-shinde:
I'd like some help with writing the unit test for handler please. 
As it is returning a string of stats I'm not sure how to go about testing for the `--human` flag as the only difference between the two strings would be with or without size units.
Thanks.

daviddias:

PR #2110, opened 2019-05-24
prepare for new aegir major version

hugomrdias:
electron tests
fixes linting 
one bug fix spotted in the electron tests
node 12 testing


package.json
```diff
@@ -196,10 +171,34 @@
...
+    "base64url": "^3.0.1",
+    "chai": "^4.2.0",
+    "delay": "^4.1.0",
+    "detect-node": "^2.0.4",
+    "dir-compare": "^1.4.0",
+    "dirty-chai": "^2.0.1",
+    "execa": "^1.0.0",
+    "form-data": "^2.3.3",
+    "hat": "0.0.3",
+    "interface-ipfs-core": "~0.103.0",
```
alanshaw: Is now `~0.104.0` 😄

src/core/components/object.js
```diff
@@ -333,7 +333,7 @@ module.exports = function object (self) {
         if (cid.codec === 'dag-cbor') {
           const links = findLinks(result)
 
-          callback(null, links)
+          return callback(null, links)
```
alanshaw: Good catch I've opened a separate PR for this here https://github.com/ipfs/js-ipfs/pull/2111

alanshaw:
Would you mind updating the circuit-relaying example at the same time?

.npmignore
```diff
@@ -1,36 +0,0 @@
...
-.lock-wscript
-
-build
-
-# Dependency directory
-# https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
-node_modules
-
-test
-examples
```
alanshaw: This file is needed so that the `dist` directory is published to npm - it also prevents the `test` and `examples` directories from being published.

package.json
```diff
@@ -49,47 +56,12 @@
...
-    "sinon": "^7.3.1",
-    "stream-to-promise": "^2.2.0"
-  },
   "dependencies": {
     "@hapi/ammo": "^3.1.0",
     "@hapi/hapi": "^18.3.1",
     "@hapi/joi": "^15.0.1",
     "async": "^2.6.1",
-    "async-iterator-all": "0.0.2",
+    "async-iterator-all": "1.0.0",
```
alanshaw: `^`?

.npmignore
hugomrdias: We have been removing this file on others repos in favor of the files property in package.json

package.json
hugomrdias:
```suggestion
    "async-iterator-all": "^1.0.0",
```

.npmignore
alanshaw: Got it, thanks - did not realise ["files" in package.json](https://docs.npmjs.com/files/package.json#files) did that.

PR #2114, opened 2019-05-24
fix: fixes rabin chunker truncating files

achingbrain:
Oops. As a bonus it also adds support for rabin chunking to the browser and Windows.


alanshaw:
Awesome! 👏

refs https://github.com/ipfs/js-ipfs-unixfs-importer/pull/30

PR #2127, opened 2019-05-29
refactor(add command, CLI): update a couple of minor help doc

Prabhakar-Poudel:
update a couple of minor help doc for add CLI command

License: MIT
Signed-off-by: Prabhakar-Poudel yuvrajzohan@gmail.com

### DIFF
--- a/src/cli/commands/add.js
+++ b/src/cli/commands/add.js
@@ -88,7 +88,8 @@ module.exports = {
     'wrap-with-directory': {
       alias: 'w',
       type: 'boolean',
-      default: false
+      default: false,
+      describe: 'Add a wrapping node'
     },
     'only-hash': {
       alias: 'n',


Prabhakar-Poudel:
I moved it from Draft PR. That could be the reason.
I will look for why that should happen anyway.
Meanwhile, can we force run it via Travis?

alanshaw:
Gah, I don't know how to do that - can you rebase and see if it picks it up?

Amorn-Kingthong:
ok

alanshaw:
@Prabhakar-Poudel this change has caused repeatable failures on CI - would you mind looking into why?

Amorn-Kingthong:
pull

Prabhakar-Poudel:
@alanshaw 
One of the tests is failing:
----
file ls
* prints a filename
----

Not sure why. Will look up the root cause.

Prabhakar-Poudel:
@alanshaw 
Just ran the tests locally.
All of them pass through.
Also, rebased and force pushed with the latest master. It did not trigger Travis.
Is there any way to debug these situations?

src/cli/commands/add.js
```diff
@@ -125,7 +126,8 @@ module.exports = {
     hash: {
       type: 'string',
       choices: Object.keys(mh.names),
-      describe: 'Hash function to use. Will set CID version to 1 if used. (experimental)'
+      describe: 'Hash function to use. Will set CID version to 1 if used. (experimental)',
+      default: 'sha2-256'
```
Prabhakar-Poudel: I was setting the default value to hashing function as `sha2-256`. This will set CID to v1.
But CID defaults to v0. And the tests that failed were looking up for v0 ID, which did not match.
Possible solutions:
- Fix tests to match with v1 ID.
- Remove default hash

@alanshaw Thoughts?

alanshaw: Yes I'd remove the default for now

PR #2132, opened 2019-05-30
feat (cli): show error message if minimum node version isn't met

### LINKED ISSUE #2126: Detect node version and show error message if t
Often our users try to run js-ipfs with a node version that is too low. (see for example https://github.com/ipfs/js-ipfs/issues/2122)
Currently they get an error like `Unexpected token *`
We should give them a nice error message instead, asking them to use the required minimum node version.

chirag-shinde:
- Added an error message asking the user to update their NodeJS to the minimum required version.
- Exits process with failure code 1.

resolves #2126 

License: MIT
Signed-off-by: Chirag Shinde <shinde.chirag2204@gmail.com>

### DIFF
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -23,8 +23,14 @@ const debug = require('debug')('ipfs:cli')
 const pkg = require('../../package.json')
 const parser = require('./parser')
 const commandAlias = require('./command-alias')
+const semver = require('semver')
 
 async function main (args) {
+  if (!semver.gt(process.versions.node, semver.minVersion(pkg.engines.node))) {
+    console.error(`Please update your NodeJS version to ${pkg.engines.node}`)
+    process.exit(1)
+  }
+
   const oneWeek = 1000 * 60 * 60 * 24 * 7
   updateNotifier({ pkg, updateCheckInterval: oneWeek }).notify()
 
--- a/src/cli/bin.js
+++ b/src/cli/bin.js
@@ -14,23 +14,24 @@ process.on('unhandledRejection', (err) => {
   throw err
 })
 
+const semver = require('semver')
+const pkg = require('../../package.json')
+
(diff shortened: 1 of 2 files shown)


chirag-shinde:
Screenshot of the output:
Screenshot 2019-05-30 at 3 35 52 PM

src/cli/bin.js
```diff
@@ -23,8 +23,14 @@ const debug = require('debug')('ipfs:cli')
 const pkg = require('../../package.json')
 const parser = require('./parser')
 const commandAlias = require('./command-alias')
+const semver = require('semver')
 
 async function main (args) {
+  if (!semver.gt(process.versions.node, semver.minVersion(pkg.engines.node))) {
```
alanshaw: Should this be `semver.gte(...)`?

alanshaw: How about using `satisfies` instead to simplify?

```suggestion
  if (!semver.satisfies(process.versions.node, pkg.engines.node)) {
```

src/cli/bin.js
```diff
@@ -23,8 +23,14 @@ const debug = require('debug')('ipfs:cli')
...
+    console.error(`Please update your NodeJS version to ${pkg.engines.node}`)
```
alanshaw:
```suggestion
    console.error(`Please update your Node.js version to ${pkg.engines.node}`)
```

src/cli/bin.js
alanshaw: This check needs to be the first thing we do in this file (after requiring `semver` and the `package.json`) because any one of the other requires could contain syntax that is incompatible with the node version we're compatible with.

chirag-shinde:
Thanks, I have made the requested changes. Please review.

alanshaw:
Normally I'd ask for a test but I don't know how you'd test this...

PR #2134, opened 2019-05-30
chore(package): update file-type to version 11.1.0

alanshaw:


alanshaw:
Updated in https://github.com/ipfs/js-ipfs/pull/2138

PR #2135, opened 2019-05-30
chore(package): update ipns to version 0.5.2

alanshaw:


alanshaw:
Updated in https://github.com/ipfs/js-ipfs/pull/2138

Issue #2148, opened 2019-05-31
RFC: js-ipfs profiles


dirkmc:
A profile is a shorthand name for a set of configuration options. This issue is to discuss which profiles we should have in js-ipfs and how each one should affect the behaviour of js-ipfs.

#### go-ipfs profiles

The profiles available for go-ipfs are documented in its [configuration README](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md) and implemented in [profile.go](https://github.com/ipfs/go-ipfs-config/blob/master/profile.go). Those that apply to js-ipfs are:

- `server ` / `local-discovery` (opposite of `server`)
  - Adds [private / local / unrouteable](https://github.com/ipfs/go-ipfs-config/blob/master/profile.go#L21) addresses to
    - `Addresses.NoAnnounce` - _prevents advertising these addresses to other peers_
    - `Swarm.AddrFilters` - _prevents dialing to these addresses_
  - Sets `Discovery.MDNS.Enabled` to `false`
  - Sets `Swarm.DisableNatPortMap` to `true`

- `test` / `default-networking` (opposite of `test`)
  - Sets `/ip4/127.0.0.1/tcp/0` for
    - `Addresses.API`
    - `Addresses.Gateway`
    - `Addresses.Swarm`
  - Sets `Discovery.MDNS.Enabled` to `false`
  - Sets `Swarm.DisableNatPortMap` to `true`
  - Sets `Bootstrap` to `[]` - _prevents bootstrapping with live nodes_

- `lowpower`
  - Sets `Routing.Type` to `dhtclient` - _puts DHT in ["client only"](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#routing) mode_
  - Sets `Reprovider.Interval` to `0` - _prevents node from advertising what content it has_
  - Sets `Swarm.ConnMgr` parameters - _maintains low connection count_
    - `Swarm.ConnMgr.LowWater`: `20`
    - `Swarm.ConnMgr.HighWater`: `40`
    - `Swarm.ConnMgr.GracePeriod`: `one minute`

- `randomports`
  - Sets `Addresses.Swarm` to listen on localhost at a random available port

#### js-ipfs profiles

[...]

PR #2165, opened 2019-06-06
Add config profile endpoint and CLI

dirkmc:
Fixes https://github.com/ipfs/js-ipfs/issues/2148

Depends on
- [x] https://github.com/ipfs/interface-js-ipfs-core/pull/488
- [x] https://github.com/ipfs/js-ipfs-http-client/pull/1030


dirkmc:
The go implementation stores a backup of the config every time a profile is applied. I'm not sure that's necessary, @alanshaw do you think we need it?

alanshaw:
FYI @dirkmc I have rebased and updated deps

src/core/components/config.js
```diff
@@ -17,6 +18,94 @@ module.exports = function config (self) {
...
+    config.Bootstrap = []
+    config.Discovery.MDNS.Enabled = false
+    config.Discovery.webRTCStar.Enabled = false
+  }
+}, {
+  name: 'default-networking',
+  description: 'Restores default network settings - inverse of "test" profile.',
+  transform: (config) => {
+    console.log('applying default-networking')
+    console.log('setting to', defaultConfig.Addresses)
```
alanshaw: Please remove :)

src/core/components/config.js
```diff
@@ -17,6 +18,94 @@ module.exports = function config (self) {
...
+    config.Swarm.ConnMgr.LowWater = 20
+    config.Swarm.ConnMgr.HighWater = 40
+  }
+}, {
+  name: 'default-power',
+  description: 'Inverse of "lowpower" profile.',
+  transform: (config) => {
+    config.Swarm = defaultConfig.Swarm
+  }
+}]
```
alanshaw: Key/value object instead?

```js
const profiles = {
  test: {
    description: '...',
    transform: () => { /* ... */ }
  },
  // ...
}
```

src/core/components/config.js
```diff
@@ -17,6 +18,94 @@ module.exports = function config (self) {
...
+        await self.config.replace(newCfg)
+      }
+
+      // Scrub private key from output
+      delete oldCfg.Identity.PrivKey
+      delete newCfg.Identity.PrivKey
+
+      callback(null, { oldCfg, newCfg })
+    } catch (err) {
+      callback(new Error(`Could not apply profile '${profileName}' to config: ${err.message}`))
```
alanshaw: Log this error using the debug module.

src/core/components/config.js
```diff
@@ -17,6 +18,94 @@ module.exports = function config (self) {
...
+    config.Addresses.Gateway = defaultConfig.Addresses.Gateway
+    config.Addresses.Swarm = defaultConfig.Addresses.Swarm
+    config.Bootstrap = defaultConfig.Bootstrap
+    config.Discovery.MDNS.Enabled = defaultConfig.Discovery.MDNS.Enabled
+    config.Discovery.webRTCStar.Enabled = defaultConfig.Discovery.webRTCStar.Enabled
+  }
+}, {
+  name: 'lowpower',
+  description: 'Reduces daemon overhead on the system - recommended for low power systems.',
+  transform: (config) => {
```
alanshaw: Lets require that these functions return the transformed config, then they can choose to mutate or recreate the config and it makes it easier to use these in a `reduce` when applying multiple configs:

```js
// Apply config profiles
config = profileNames.reduce((c, n) => Profiles[n].transform(c), config)
```

src/core/components/config.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 const promisify = require('promisify-es6')
+const defaultConfig = require('../runtime/config-nodejs.js')()
```
alanshaw: Please call this funciton every time you need `defaultConfig` because otherwise it is global, shared and can be mutated.

test/cli/config.js
```diff
@@ -97,4 +98,70 @@ describe('config', () => runOnAndOff((thing) => {
...
+      await ipfs('config profile apply --dry-run=false server')
+      const updated = updatedConfig()
+      expect(updated.Discovery.MDNS.Enabled).to.equal(false)
+    })
+
+    it('Private key does not appear in output', async () => {
+      const out = await ipfs('config profile apply server')
+      expect(out).not.includes('PrivKey')
+    })
+  })
```
alanshaw: Can these just be `interface-ipfs-core` tests?

alanshaw:
@hacdias @olizilla @lidel we're adding profiles to js-ipfs. I'd like to add a desktop profile - what would that look like? Is desktop the default? What changes to the browser config should we make? Should we also have a browser profile so you can switch to it in other environments? Are we missing any other profiles? Please help 🙏

lidel:
Prior art in [go-ipfs/docs/config.md#profiles](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#profiles): 
There is no `desktop` profile in go-ipfs (yet), for now we manually apply config changes via ipfs-desktop:
- [ipfs-desktop/src/common/store.js#L13](https://github.com/ipfs-shipyard/ipfs-desktop/blob/34077078d97d2528c8b28b76ff3731c71a92439e/src/common/store.js#L13) (dht client mode, enable gc)
- [ipfs-desktop/src/daemon/daemon.js#L71](https://github.com/ipfs-shipyard/ipfs-desktop/blob/5a32d083854a5b453fd0d59d5b754640221b71cc/src/daemon/daemon.js#L71) (connmgr limits to avoid killing consumer routers, mdns on)

I think we could scope initial js-ipfs profiles to:

- `server` - similar to one in go-ipfs, disables preload, disables default signaling server, no dialing to local networks, no gc and mdns is also off (not sure about relay, perhaps `relay` should be a separate profile)
- `desktop` - settings applied by ipfs-desktop (mdns, dht client, gc on), in future could be also a relay client mode
- `browser` - enable preload, enable signaling via -star (in future relay), enable gc 
- `test` - similar to one in go-ipfs: disables things like bootstrap, preload, mdns to speed up tests etc (would remove a lot of boilerplate in tests going forward)

achingbrain:
Merged master into this PR to fix the merge conflicts.  Still needs a bunch of work to address the PR comments and probably to integrate with #2428

achingbrain:
I wonder if we could have an `offline` profile that didn't start a libp2p node at all?  That would probably speed up the tests pretty dramatically as only a few of them actually require networking.

src/core/components/config.js
achingbrain: Done

src/core/components/config.js
achingbrain: Like it, done.

src/core/components/config.js
achingbrain: Done

src/core/components/config.js
achingbrain: Done

src/core/components/config.js
achingbrain: Done

test/cli/config.js
achingbrain: I'm hesitant for two reasons:

1. We test to see if the profile was correctly applied by inspecting the config which requires the tests to have knowledge of what the profiles are - as implemented this would involve adding `ipfs` as a dependency of the interface tests which it is not currently
2. There are CLI features that aren't in the interface definition (for better or for worse) like listing available profiles

achingbrain:
Anyone know why the method this adds is `ipfs.config.profile` and not `ipfs.config.applyProfile`?

To me at least, it's a bit more obvious what the second one is going to do.

achingbrain:
Or better yet:

```javascript
Promise<AsyncIterable<{ name: String, description: String }>> ipfs.config.profile.list()
Promise<void> ipfs.config.profile.apply(name, { dryRun: Boolean })
```

achingbrain:
Depends on:

- [ ] https://github.com/ipfs/js-ipfs-http-client/pull/1119
- [ ] https://github.com/ipfs/interface-js-ipfs-core/pull/536

test/cli/config.js
achingbrain: I've exposed profile definitions over the http interface so these can be moved now, hooray!

Issue #2160, opened 2019-06-04
Reprovider to reannounce blocks to the network


vasco-santos:
This issue is a child of [Connectivity Magic #1459](https://github.com/ipfs/js-ipfs/issues/1459), and is intended to discuss the implementation of the reprovider in `js-ipfs`.

The current implementation of reprovider in `go-ipfs` consists of iterating all blocks in blockstore and provide them every X hours (being X configurable through the config file). Other than that, the reprovider may have different strategies for deciding which blocks it should reprovide [ipfs/go-ipfs/docs/config.md#reprovider](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#reprovider).

However, `go-ipfs` team is currently working on improving their reprovider. Consequently, it would be great to have an idea of what is the plan, in order to decide if we should start with this simpler implementation or with the newest one.

Another aspect to take into consideration in the JS implementation is if we should create the concept of Provider. In this case, `js-ipfs-bitswap` would receive its reference and the Provider could abstract the providing logic from there. I don't know the current logic of `js-ipfs-bitswap` yet to have strong conclusions in this regard.

cc @michaelavila @Kubuxu @dirkmc @vmx

michaelavila:
@vasco-santos thanks for reaching out!
> However, go-ipfs team is currently working on improving their reprovider. Consequently, it would be great to have an idea of what is the plan, in order to decide if we should start with this simpler implementation or with the newest one.

The immediate next iteration of the reprovider gets simpler:

* We introduce a `Tracker` that is used by both the provider and the reprovider
* The `Provider` adds entries into the tracker
* All the `Reprovider` does is announce every block in the `Tracker`

Most of our effort is going into reworking the provider itself. Ultimately, the goal is to provide a subset of all the blocks we are needing to provide now, while retaining overall discoverability of blocks. This will necessitate things like backlinks or some kind of meta to determine which blocks we can call `FindProviders` on.

One additional small thing we've done, and intend to retain, is prioritizing announcing root blocks first.

> Another aspect to take into consideration in the JS implementation is if we should create the concept of Provider. In this case, js-ipfs-bitswap would receive its reference and the Provider could abstract the providing logic from there. I don't know the current logic of js-ipfs-bitswap yet to have strong conclusions in this regard.

[...]

PR #2184, opened 2019-06-17
feat: reprovider

vasco-santos:
In the context of `js-ipfs` reannouncing blocks to the network [ipfs/js-ipfs#2160](https://github.com/ipfs/js-ipfs/issues/2160), this PR aims to add a Provider layer controlled by `js-ipfs`. This provider will be used by bitswap.

Needs:

- [ ] Different strategies
- [ ] [ipfs/js-ipfs-bitswap#199](https://github.com/ipfs/js-ipfs-bitswap/pull/199)
- [ ] documentation - should be a part of [ipfs/js-ipfs#2204](https://github.com/ipfs/js-ipfs/issues/2204)

### DIFF
--- a/src/core/components/start.js
+++ b/src/core/components/start.js
@@ -1,11 +1,13 @@
 'use strict'
 
+const get = require('dlv')
 const series = require('async/series')
 const Bitswap = require('ipfs-bitswap')
 const setImmediate = require('async/setImmediate')
 const promisify = require('promisify-es6')
 
 const IPNS = require('../ipns')
+const Provider = require('../provider')
 const routingConfig = require('../ipns/routing/config')
 const createLibp2pBundle = require('./libp2p')
 
@@ -45,6 +47,8 @@ module.exports = (self) => {
           libp2p.start(err => {
             if (err) return cb(err)
             self.libp2p = libp2p
+
+            self._provider = new Provider(libp2p, self._repo.blocks, get(config, 'Reprovider'))
             cb()
           })
         })
@@ -56,8 +60,14 @@ module.exports = (self) => {
         self._bitswap = new Bitswap(
           self.libp2p,
           self._repo.blocks,
+          self._provider,
           { statsEnabled: true }
         )
+
+        if (!get(self._options, 'offline') &&
+          (get(self._options, 'libp2p.config.dht.enabled', false) || get(self._options, 'libp2p.modules.contentRouting', false))) {
+          self._provider.start()
+        }
 
         self._bitswap.start()
         self._blockService.setExchange(self._bitswap)
--- a/src/core/components/start.js
+++ b/src/core/components/start.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const get = require('dlv')
+const callbackify = require('callbackify')
 const series = require('async/series')
 const Bitswap = require('ipfs-bitswap')
 const setImmediate = require('async/setImmediate')
@@ -48,10 +49,22 @@ module.exports = (self) => {
             if (err) return cb(err)
             self.libp2p = libp2p
 
+            // create provider
             self._provider = new Provider(libp2p, self._repo.blocks, get(config, 'Reprovider'))
             cb()
           })
         })
+      },
+      (cb) => {
+        // start provider if libp2p routing enabled
+        if (!get(self._options, 'offline') &&
+          (get(self._options, 'libp2p.config.dht.enabled', false) || get(self._options, 'libp2p.modules.contentRouting', false))) {
+          const providerStart = callbackify(() => self._provider.start())
+
+          providerStart(cb)
+        } else {
+          cb()
+        }
       },
       (cb) => {
         const ipnsRouting = routingConfig(self)
@@ -63,11 +76,6 @@ module.exports = (self) => {
           self._provider,
           { statsEnabled: true }
         )
-
-        if (!get(self._options, 'offline') &&
-          (get(self._options, 'libp2p.config.dht.enabled', false) || get(self._options, 'libp2p.modules.contentRouting', false))) {
-          self._provider.start()
-        }
 
         self._bitswap.start()
         self._blockService.setExchange(self._bitswap)
--- a/src/core/provider/index.js
+++ b/src/core/provider/index.js
@@ -1,0 +1,93 @@
+'use strict'
+
+const errCode = require('err-code')
+const human = require('human-to-milliseconds')
+const promisify = require('promisify-es6')
+
+const CID = require('cids')
+
+const Reprovider = require('./reprovider')
+
+class Provider {
+  /**
+   * Provider goal is to announce blocks to the network.
+   * It keeps track of which blocks are provided, and allow them to be reprovided
+   * @param {Libp2p} libp2p
+   * @param {Blockstore} blockstore
+   * @param {object} options
+   * @memberof Provider
+   */
+  constructor (libp2p, blockstore, options) {
+    this._running = false
+
+    this._contentRouting = libp2p.contentRouting
+    this._blockstore = blockstore
+    this._options = options
+    this.reprovider = undefined
+  }
+
+  /**
+   * Begin processing the provider work
+   * @returns {void}
+   */
+  async start () {
+    // do not run twice
+    if (this._running) {
+      return
+    }
+
+    this._running = true
+
+    // handle options
+    const strategy = this._options.strategy || 'all'
+    const humanInterval = this._options.Interval || '12h'
+    const interval = await promisify((callback) => human(humanInterval, callback))()
+    const options = {
+      interval,
+      strategy
+    }
+
+    this.reprovider = new Reprovider(this._contentRouting, this._blockstore, options)
+
+    // Start reprovider
+    this.reprovider.start()
+  }
+
+  /**
+   * Stop the provider
+   * @returns {void}
+   */
+  stop () {
+    this._running = true
+
+    // stop the reprovider
+    this.reprovider.stop()
+  }
+
+  /**
+   * Announce block to the network and add and entry to the tracker
+   * Takes a cid and makes an attempt to announce it to the network
+   * @param {CID} cid
+   */
+  async provide (cid) {
+    if (!CID.isCID(cid)) {
+      throw errCode('invalid CID to provide', 'ERR_INVALID_CID')
+    }
+
+    await promisify((callback) => {
+      this._contentRouting.provide(cid, callback)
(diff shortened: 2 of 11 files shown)


vasco-santos:
@alanshaw @dirkmc could I get your feedback on this PR?

It is important to also look at the `bitswap`PR in this context.

FYI, for getting this to the finish line I only need to add the proper logic for the reprovide strategies. However, there is some code in Dirk's PR for the garbage collection [ipfs/js-ipfs#2022](https://github.com/ipfs/js-ipfs/pull/2022/) that would be great to use for this. So, it would be great to have that merged first.

src/core/provider/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+const Reprovider = require('./reprovider')
+
+class Provider {
+  /**
+   * Provider goal is to announce blocks to the network.
+   * It keeps track of which blocks are provided, and allow them to be reprovided
+   * @param {Libp2p} libp2p
+   * @param {Blockstore} blockstore
+   * @param {object} options
```
dirkmc: Could we add a description of each option here

src/core/provider/index.js
```diff
@@ -0,0 +1,93 @@
...
+    if (this._running) {
+      return
+    }
+
+    this._running = true
+
+    // handle options
+    const strategy = this._options.strategy || 'all'
+    const humanInterval = this._options.Interval || '12h'
+    const interval = await promisify((callback) => human(humanInterval, callback))()
```
dirkmc: It's a shame this code is complicated by human-to-milliseconds requiring a callback. It seems from [the source code](https://github.com/hacdias/human-to-milliseconds/blob/master/src/index.js) that it doesn't need to be a callback. If we can replace this with something synchronous then you don't need to worry about bitswap starting before provider in start.js above

src/core/provider/index.js
```diff
@@ -0,0 +1,93 @@
...
+   */
+  stop () {
+    this._running = true
+
+    // stop the reprovider
+    this.reprovider.stop()
+  }
+
+  /**
+   * Announce block to the network and add and entry to the tracker
```
dirkmc:
```suggestion
   * Announce block to the network and add an entry to the tracker
```

src/core/provider/index.js
```diff
@@ -0,0 +1,93 @@
...
+    if (!CID.isCID(cid)) {
+      throw errCode('invalid CID to provide', 'ERR_INVALID_CID')
+    }
+
+    await promisify((callback) => {
+      this._contentRouting.provide(cid, callback)
+    })()
+  }
+
+  async findProviders (cid, options) { // eslint-disable-line require-await
```
dirkmc: We should probably document the options here (even though they're the same as content routing `findProviders()`

src/core/provider/queue.js
```diff
@@ -0,0 +1,110 @@
+'use strict'
+
+const queue = require('async/queue')
+
+const debug = require('debug')
+const log = debug('ipfs:provider')
+log.error = debug('ipfs:provider:error')
+
+class WorkerQueue {
```
dirkmc: Can you use p-queue instead? It will take care of the Promise execution for you

src/core/provider/reprovider.js
```diff
@@ -0,0 +1,89 @@
+'use strict'
+
+const promisify = require('promisify-es6')
+const WorkerQueue = require('./queue')
+
+const { blockKeyToCid } = require('../utils')
+
+// const initialDelay = 15000
+const initialDelay = 3000
```
dirkmc: Let's add this to config

src/core/provider/reprovider.js
```diff
@@ -0,0 +1,89 @@
...
+
+class Reprovider {
+  /**
+   * Reprovider goal is to reannounce blocks to the network.
+   * @param {object} contentRouting
+   * @param {Blockstore} blockstore
+   * @param {object} options
```
dirkmc: Let's document the options

src/core/provider/reprovider.js
```diff
@@ -0,0 +1,89 @@
...
+    while (this._timeoutId) {
+      const blocks = await promisify((callback) => this._blockstore.query({}, callback))()
+
+      // TODO strategy logic here
+      if (this._options.strategy === 'pinned') {
+
+      } else if (this._options.strategy === 'pinned') {
+
+      }
+
```
dirkmc:
```suggestion
if (!this._timeoutId) return
```

src/core/provider/reprovider.js
```diff
@@ -0,0 +1,89 @@
...
+      await this._worker.execute(blocks)
+
```
dirkmc:
```suggestion
if (!this._timeoutId) return
```

src/core/provider/reprovider.js
```diff
@@ -0,0 +1,89 @@
...
+      // Each subsequent walk should run on a `this._options.interval` interval
+      await new Promise(resolve => {
```
dirkmc: I'd suggest using a timeout which recursively calls `_runPeriodically()`
That way you can cancel the timeout when the reprovider is stopped

dirkmc (changes requested):

dirkmc:
Looks like it's on the right track @vasco-santos 👍 
I left a few comments for minor things in the code.

I wonder if it would make sense to persist the time at which the last provide occurred, so that if a node restarts it doesn't wait the full 12 hours to reprovide?

vasco-santos:
Thanks @dirkmc ! I think that I addressed all your comments!

Regarding persisting the time of the last provide, the first reprovide will happen once the node starts, after a short delay. So, I think it should not be necessary

src/core/provider/index.js
```diff
@@ -0,0 +1,106 @@
...
+  async start () {
+    // do not run twice
+    if (this._running) {
+      return
+    }
+
+    this._running = true
+
+    // handle options (config uses uppercase)
+    const humanDelay = this._options.Delay || this._options.delay || '15s'
```
dirkmc: Is `0` a valid value for the delay?

src/core/components/start.js
```diff
@@ -56,9 +60,15 @@ module.exports = (self) => {
         self._bitswap = new Bitswap(
           self.libp2p,
           self._repo.blocks,
+          self._provider,
           { statsEnabled: true }
         )
 
+        if (!get(self._options, 'offline') &&
+          (get(self._options, 'libp2p.config.dht.enabled', false) || get(self._options, 'libp2p.modules.contentRouting', false))) {
+          self._provider.start()
```
dirkmc: Because `Provider.start()` is async, bitswap may start before provider finishes starting. It may be safer to wait till `Provider.start()` has completed before starting bitswap

src/core/provider/index.js
```diff
@@ -0,0 +1,106 @@
...
+    // Start reprovider
+    this.reprovider.start()
+  }
+
+  /**
+   * Stop the provider
+   * @returns {void}
+   */
+  stop () {
+    this._running = true
```
dirkmc:
```suggestion
    this._running = false
```

src/core/provider/reprovider.js
```diff
@@ -0,0 +1,95 @@
...
+
+const { blockKeyToCid } = require('../utils')
+
+class Reprovider {
+  /**
+   * Reprovider goal is to reannounce blocks to the network.
+   * @param {object} contentRouting
+   * @param {Blockstore} blockstore
+   * @param {object} options
+   * @param {string} options.delay reprovider initial delay in human friendly time
```
dirkmc:
```suggestion
   * @param {string} options.delay reprovider initial delay in milliseconds
```

src/core/provider/reprovider.js
```diff
@@ -0,0 +1,95 @@
...
+   * @param {string} options.interval reprovider interval in human friendly time
```
dirkmc:
```suggestion
   * @param {string} options.interval reprovider interval in milliseconds
```

src/core/provider/reprovider.js
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Do the reprovide work to libp2p content routing
+   * @param {Block} block
+   * @returns {void}
+   */
+  async _provideBlock (block) {
+    const cid = blockKeyToCid(block.key.toBuffer())
+
+    await promisify((callback) => {
```
dirkmc: Is there a way we can abort this Promise if the Reprovider is stopped?

dirkmc (changes requested):

src/core/provider/reprovider.js
vasco-santos: Once we do the DHT refactor, if we add the abortable signal to provide there, as well as to `js-libp2p` and to `delegated-content-routing`, we should be able to do that

src/core/components/start.js
vasco-santos: I switched the provider start to the previous function of the `series` array, and "callbackifyed" it to guarantee we wait for it

src/core/provider/index.js
vasco-santos: Ok, so I think we should not accept 0, as we should wait until we are connected to some peers. Do you agree to not allow 0?

As we are using a string, that validation is very annoying 😢

dirkmc: Yes I think that makes sense. The validation is a little tricky 🤔

src/core/components/start.js
dirkmc: The logic makes sense. I would suggest moving all the new code to another entry in the `series` array

vasco-santos: I can't move the Provider instantiation, as I would have to get the config again, or keeping it in a temporary constant. I would prefer to keep it there, and move the starting code to the new entry of the series array

dirkmc: Ok makes sense

alanshaw:
@vasco-santos what's the status of this PR - still draft?

vasco-santos:
I was waiting on [https://github.com/ipfs/js-ipfs/pull/2022](https://github.com/ipfs/js-ipfs/pull/2022) getting merged, which had a lot of useful code to integrate here without duplicating.

daviddias:

achingbrain:
I'm going to close this because I think since the `.provide` method lives on `libp2p`'s `ContentRouter` interface, re-provides should be handled by the various implementations of content routers since they all may have different rules about provider record validity, etc.

Some may call out to a remote service that does the re-provide for us, for example, or some may have shorter lived records than others, who knows!

Issue #2155, opened 2019-06-04
Add delegated routing to js-ipfs browser


jacobheun:
This issue is a child of [Connectivity Magic #1459](https://github.com/ipfs/js-ipfs/issues/1459), and is intended to track the completion of the **Delegated Routing** portion of that.

Until browsers have the support needed to handle DHT queries quickly (direct dialing capabilities), delegated peer and content routing will be an important part of ensuring nodes are able to find peers and content.

js-libp2p already has support for custom content and peer routing modules. There are currently two modules that can leverage the IPFS api to achieve both. An example of this is at https://github.com/libp2p/js-libp2p/tree/v0.25.3/examples/delegated-routing.

## TODOs
These are the necessary steps to adding delegated routing support to js-ipfs in the browser:
- [x] Deploy go-ipfs nodes that have support for the needed delegated routing endpoints. Context at https://github.com/ipfs/go-ipfs/pull/5565
- [x] Add [delegated content routing module](https://github.com/libp2p/js-libp2p-delegated-content-routing) to js-ipfs, libp2p browser config
- [x] Add [delegated peer routing module](https://github.com/libp2p/js-libp2p-delegated-peer-routing) to js-ipfs, libp2p browser config

Refs: 
* https://github.com/libp2p/js-libp2p/issues/120

lidel:
Some additional context:

### Enabling preload nodes as delegated routers

IIUC if we want to reuse preload nodes as delegated routers, we need to prepare them first:

-  expose `swarm/connect`, `dht/findprovs`, and `dht/findpeer` in addition to `refs` (which is already being used by existing preload logic)
   - Based on https://github.com/ipfs/go-ipfs-config/pull/10/ https://github.com/ipfs/go-ipfs/pull/5565 I think list of exposed commands can be tweaked via `Gateway.APICommands` config
- ensure preload nodes stay connected to gateways 

@mburns  is working on it

### Wiring up `*-delegated*-routing` modules to use preloads

I don't see `libp2p-delegated-content-routing` and `libp2p-delegated-routing` being wired up in js-ipfs to use existing preload servers yet. 
(Note that by default both modules will try `ipfs.io/api/v0/` and fail, as some of required commands are not exposed on `ipfs.io` gateways)

@alanshaw are you planning to PR and release this before IPFS Camp?
cc @hugomrdias as delegated DHT may be useful for IPNS as a fallback

### Tweak the way js-ipfs  picks preload server 

[...]

PR #2195, opened 2019-06-24
feat: add delegate routers to libp2p config

jacobheun:
Adds the delegate routing modules needed for https://github.com/ipfs/js-ipfs/issues/2155.

This enables js-ipfs nodes to leverage external nodes to provide content, find content, and find peers.

I added a crude randomizer for the preload nodes which are supporting delegate routing. Ideally this would be per request, but the modules will need to be updated to take multiple hosts to do so.

**A note**: If the DHT is enabled on your node, libp2p will attempt to perform the action using the DHT first. If that is not successful, it will then use the delegates to perform the request. This gives nodes the ability to fallback to the external nodes while they are building up their DHT records/peers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,18 @@ Enable and configure experimental features.
 
 Modify the default IPFS node config. This object will be *merged* with the default config; it will not replace it.
 
+###### Configuring Delegate Routers
+
+If you need to support Delegated Content and/or Peer Routing, you can enable it by specifying the multiaddrs of your delegate nodes in the config via `options.config.Addresses.Delegates`. If you need to run a delegate router we encourage you to run your own, with go-ipfs. You can see instructions for doing so in the [delegated routing example][:delegated-routing-example].
+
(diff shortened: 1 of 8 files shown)


jacobheun:
> Should the hostnames be in config?

Yeah, we should probably have the multiaddrs in there and then pull out the http endpoints.

This will cause the config to deviate from go, as we'll likely add a new address field like `Addresses.Delegates`. This should be fine as it's more of a js thing, but if we do that we should check to make sure go doesn't care about the extra field.

Does `Addresses.Delegates` sound like the right place for this to go?
```js
Addresses: {
  ...
  Delegates: [
    '/dns4/node0.preload.ipfs.io/tcp/443/https',
    '/dns4/node1.preload.ipfs.io/tcp/443/https'
  ]
}
```

dirkmc:
`Discovery` and `Bootstrap` are at the top level, so it may make sense for Delegates to be as well, and thereby less likely to conflict with go config

jacobheun:
I've updated this to only add delegated routing if the delegates multiaddrs are actually added to the Config `Addresses.Delegates` property. I also added docs and a couple tests.

alanshaw:
@jacobheun I fixed the link and rebased

PR #2203, opened 2019-06-27
fix: fix ipns routing test

hugomrdias:
This test 'should use the dht if enabled' was failing because the ipns routing was trying to get the wrong option.

### DIFF
--- a/src/core/ipns/routing/config.js
+++ b/src/core/ipns/routing/config.js
@@ -22,7 +22,7 @@ module.exports = (ipfs) => {
   }
 
   // DHT should not be added as routing if we are offline or it is disabled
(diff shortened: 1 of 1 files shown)


alanshaw:
I think @vasco-santos has the correct fix for this one https://github.com/ipfs/js-ipfs/pull/2205

hugomrdias:
You sure if you check libp2p config there is not "config" property

vasco-santos:
Have you really check the libp2p config?

[libp2p/js-libp2p#creating-your-own-libp2p-bundle](https://github.com/libp2p/js-libp2p#creating-your-own-libp2p-bundle)

PR #2206, opened 2019-06-28
use one cdn for easy copy&paste 

NukeManDan:
Copy paste implies you might use both if you don't read. :grin: Use one of the two.


alanshaw:
Thanks @NukeManDan 😁

PR #2208, opened 2019-06-30
Add test to name-pubsub: should handle event on publish correctly

Jonybang:
I think this test could help many people better understand and start using pubsub events about publishing and will increase coverage of properties of returned objects.


alanshaw:
Thanks @Jonybang

npopeka:
Great!

PR #2213, opened 2019-07-03
chore: upgrade human-to-milliseconds

vasco-santos:
Upgraded `human-to-milliseconds` dependency, since it got a new sync version that allows us to remove the callback logic.

### DIFF
--- a/src/core/components/stats.js
+++ b/src/core/components/stats.js
@@ -49,17 +49,26 @@ module.exports = function stats (self) {
     })
 
     if (opts.poll) {
+      let value
       try {
-        const value = human(opts.interval || '1s')
-
-        interval = setInterval(() => {
-          bandwidthStats(self, opts)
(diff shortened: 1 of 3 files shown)


src/core/components/stats.js
```diff
@@ -49,17 +49,17 @@ module.exports = function stats (self) {
...
-        if (err) {
-          return stream.end(errCode(err, 'ERR_INVALID_POLL_INTERVAL'))
-        }
+      try {
+        const value = human(opts.interval || '1s')
 
         interval = setInterval(() => {
           bandwidthStats(self, opts)
             .then((stats) => stream.push(stats))
             .catch((err) => stream.end(err))
```
dirkmc: This is not related to your changes, but maybe worth fixing: if we end the stream here should we also `clearInterval()`?

vasco-santos: Good point! I will do that

PR #2217, opened 2019-07-04
fix(gateway): keep pretty URL and return implicit index.html

lidel:
This cosmetic PR enables pretty URLs for directories with `index.html`
It follows what go-ipfs does: return implicit `index.html` without changing URL

**TL;DR:**

- BEFORE: http://127.0.0.1:9090/ipns/docs.ipfs.io → http://127.0.0.1:9090/ipns/docs.ipfs.io/index.html
- AFTER: http://127.0.0.1:9090/ipns/docs.ipfs.io → http://127.0.0.1:9090/ipns/docs.ipfs.io/ :ok_hand: 

It also makes logs more useful by showing both requested HTTP path and resolved IPFS one:

```
ipfs:http-gateway HTTP path  /ipns/tr.wikipedia-on-ipfs.org/ +0ms
ipfs:http-gateway IPFS path  /ipfs/QmT5NvUtoM5nWFfrQdVrFtvGfKFmG7AHE8P34isapyhCxX/index.html +0ms
```

### DIFF
--- a/src/http/gateway/resources/gateway.js
+++ b/src/http/gateway/resources/gateway.js
@@ -56,10 +56,11 @@ module.exports = {
     // so we convert /ipns/ to /ipfs/ before passing it to the resolver ¯\_(ツ)_/¯
     // This could be removed if a solution proposed in
     //  https://github.com/ipfs/js-ipfs-http-response/issues/22 lands upstream
-    const ipfsPath = decodeURI(path.startsWith('/ipns/')
+    let ipfsPath = decodeURI(path.startsWith('/ipns/')
       ? await ipfs.name.resolve(path, { recursive: true })
       : path)
 
+    let directory = false
     let data
     try {
       data = await resolver.cid(ipfs, ipfsPath)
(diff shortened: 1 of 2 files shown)


src/http/gateway/resources/gateway.js
```diff
@@ -56,10 +56,11 @@ module.exports = {
     // so we convert /ipns/ to /ipfs/ before passing it to the resolver ¯\_(ツ)_/¯
     // This could be removed if a solution proposed in
     //  https://github.com/ipfs/js-ipfs-http-response/issues/22 lands upstream
-    const ipfsPath = decodeURI(path.startsWith('/ipns/')
+    let ipfsPath = decodeURI(path.startsWith('/ipns/')
       ? await ipfs.name.resolve(path, { recursive: true })
       : path)
 
+    let directory = false
```
lidel: I went with explicit flag to make it easier to review code paths related to directories (with and without `index.html`)

src/http/gateway/resources/gateway.js
```diff
@@ -153,7 +159,7 @@ module.exports = {
           log.error(err)
           return reject(err)
         }
-        resolve({ peekedStream, contentType: detectContentType(path, streamHead) })
+        resolve({ peekedStream, contentType: detectContentType(ipfsPath, streamHead) })
```
lidel: `ipfsPath` instead of `path` because if we are returning implicit `index.html` we want to detect HTML content type of the file

PR #2227, opened 2019-07-08
fix(gateway): remove buffer-peek-stream

lidel:
This PR fixes "random" daemon crashes with `NodeError: Cannot call write after a stream was destroyed` (details in https://github.com/libp2p/js-libp2p/issues/374)

It removes `buffer-peek-stream` which caused uncaught errors inside of Hapijs when HTTP client closed HTTP connection before receiving entire payload from a Gateway port.

After falling into rabit hole of accounting for how Hapijs handles Node streams I decided to take step back, change approach and replace stream peeking with much simpler code that does not introduce intermediate streams.

**Steps to reproduce:** https://github.com/libp2p/js-libp2p/issues/374#issuecomment-508786129

Closes https://github.com/libp2p/js-libp2p/issues/374
Closes https://github.com/libp2p/pull-mplex/issues/13

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "bl": "^3.0.0",
     "boom": "^7.2.0",
     "bs58": "^4.0.1",
-    "buffer-peek-stream": "^1.0.1",
     "byteman": "^1.3.5",
     "callbackify": "^1.1.0",
     "cid-tool": "~0.3.0",
--- a/src/http/gateway/resources/gateway.js
+++ b/src/http/gateway/resources/gateway.js
@@ -9,7 +9,6 @@ const mime = require('mime-types')
 const { PassThrough } = require('readable-stream')
 const Boom = require('boom')
 const Ammo = require('@hapi/ammo') // HTTP Range processing utilities
-const peek = require('buffer-peek-stream')
 
 const multibase = require('multibase')
 const { resolver } = require('ipfs-http-response')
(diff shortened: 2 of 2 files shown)


jacobheun:
This looks like a good alternative. 

Do we have Gateway tests we can add for the early hangups from clients? If we're going to deploy any js-ipfs nodes we should get more tests around abusive client behavior.

lidel:
@jacobheun gateway tests in js-ipfs skip HTTP transport and inject request to hapijs using provided DSL:
https://github.com/ipfs/js-ipfs/blob/01419cb7539fbdbebd52c0c8f105e1ba151e802f/test/gateway/index.js#L342

I agree end-to-end tests like one you described are needed, but I'd argue its true not only for js-ipfs. 
We need "gateway interop tests" that ensure both go and js act the same way.

AFAIK we don't have gateway interop tests, but creating one could be Q3 OKR under "gateway as a product" or  web browser maintenance work. Having that kicked off, I'll be happy to seed the interop test suite with tests for issue at hand and other things (I've already started collecting potential issues around HTTP headers in https://github.com/ipfs/in-web-browsers/issues/132)

alanshaw:
Note: @lidel is setting up tests for this here https://github.com/ipfs/interop/issues/76

alanshaw:
@lidel is it possible to test this using the `inject` interface?

lidel:
@alanshaw I tried briefly, but could not see how reproduce it work without spawning real server. That is why I'd rather have it in https://github.com/ipfs/interop/issues/76

PR #2228, opened 2019-07-09
fix: pin type filtering in pin.ls

lidel:
### Summary

This PR fixes filtering, improves interop with go-ipfs and adds missing tests for `pin ls`

### Details

Old version returned indirect pin when `pin ls -t direct <path>` was executed.

This PR also tweaks error handling to match behavior from go-ipfs/js-ipfs-http-client and pass improved interop tests added in https://github.com/ipfs/interface-js-ipfs-core/pull/375

I've added some inline comments, hope it helps in review.

### Related

- Improved `pin ls` interop tests: https://github.com/ipfs/interface-js-ipfs-core/pull/375 
  (this PR is not blocked by interop tests, old ones were less strict)
- We need this fix for embedded js-ipfs in Brave :lion: https://github.com/ipfs-shipyard/ipfs-companion/issues/716


src/core/components/pin.js
```diff
@@ -327,7 +327,7 @@ module.exports = (self) => {
         // check the pinned state of specific hashes
         waterfall([
           (cb) => resolvePath(self.object, paths, cb),
-          (hashes, cb) => mapSeries(hashes, (hash, done) => pin._isPinnedWithType(hash, types.all, done), cb),
+          (hashes, cb) => mapSeries(hashes, (hash, done) => pin._isPinnedWithType(hash, type, done), cb),
```
lidel: The bug :beetle:

src/core/components/pin.js
```diff
@@ -348,12 +348,12 @@ module.exports = (self) => {
               })
 
             if (!results.length) {
-              return cb(new Error(`Path is not pinned`))
+              return cb(new Error(`path '${paths}' is not pinned`))
```
lidel: Changed to match error from go-ipfs
Interop test for this is added in https://github.com/ipfs/interface-js-ipfs-core/pull/375/files#diff-6a00eb1307f801bb37999815584589bdR190

src/http/api/resources/pin.js
```diff
@@ -57,7 +57,7 @@ exports.ls = {
     try {
       result = await ipfs.pin.ls(path, { type })
     } catch (err) {
-      throw Boom.boomify(err, { message: 'Failed to list pins' })
+      throw Boom.boomify(err)
```
lidel: Removed message prefix to match error from go-ipfs.
Interop test for this is added in https://github.com/ipfs/interface-js-ipfs-core/pull/375/files#diff-6a00eb1307f801bb37999815584589bdR199

test/core/pin.js
```diff
@@ -44,14 +45,22 @@ describe('pin', function () {
   let pin
   let repo
 
-  function expectPinned (hash, type, pinned = true) {
+  function expectPinned (hash, type = pinTypes.all, pinned = true) {
```
lidel: `expectPinned` was not checking if returned `type` match requested one, so I've added double-check for that.

test/core/pin.js
```diff
@@ -275,6 +285,78 @@ describe('pin', function () {
...
+            expect(out).to.have.deep.members([
+              { type: 'direct',
+                hash: pins.mercuryDir }
+            ])
+          )
+      })
+
+      it('direct for path (no match)', function (done) {
+        pin.ls(`/ipfs/${pins.root}/mercury/wiki.md`, { type: 'direct' }, (err, pinset) => {
+          expect(err).to.exist()
```
lidel: To make it easier to maintain the error message is not hardcoded here, but checked in interop tests: 
- https://github.com/ipfs/interface-js-ipfs-core/pull/375/files#diff-6a00eb1307f801bb37999815584589bdR190

lidel:
@alanshaw ok to merge this and https://github.com/ipfs/interface-js-ipfs-core/pull/375 before the fix gets old again? :)

PR #2243, opened 2019-07-12
fix: clean repo func on windows

alanshaw:
`rimraf.sync` does not retry when it encounters errors on windows. The async version retries a number of times before failing.

See https://github.com/isaacs/rimraf/issues/72 for context on why rimraf on windows might error.

Attempting to fix this error seeing quite often in CI now:

windows-error


alanshaw:
I re-ran the windows tests multiple times and did not see this error again.

PR #2254, opened 2019-07-12
Add CORS headers to gateway responses

nick:
This allows JSON content served from the gateway to be consumed without a CORS error.

I just tried upgrading from 0.34 to 0.36 and found this issue... adding this extra config seemed to fix it.

### DIFF
--- a/src/http/index.js
+++ b/src/http/index.js
@@ -125,7 +125,13 @@ class HttpApi {
   }
 
   async _createGatewayServer (host, port, ipfs) {
-    const server = Hapi.server({ host, port })
+    const server = Hapi.server({ 
+      host, 
+      port,
+      routes: {
+        cors: true
+      }
+    })
     server.app.ipfs = ipfs
 
     await server.register({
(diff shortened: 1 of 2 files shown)


alanshaw:
Any chance we can get a test for this @nick so that it doesn't happen again?

nick:
@alanshaw I tried adding a check for the headers in `test/gateway/index.js` but it seems 'access-control-origin' is not set in that context. Searching the test folder for 'cors' or 'access-control' didn't yield anything useful... do you have any pointers on where I could add an appropriate test?

alanshaw:
`test/gateway` looks like the right place to add it - I guess you need to add a test that sends an [OPTIONS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS) request so the server sends a response back with CORS headers?

nick:
Turns out CORS headers are only sent in the response if I set the 'Origin' header in the request. Have now added an appropriate test.

PR #2272, opened 2019-07-17
chore: fix 3 windows environments

alanshaw:
CI seems to be running tests on 3 windows envs instead of the expected 2.


dirkmc:
No wonder windows tests are so slow :)

PR #2281, opened 2019-07-19
[WIP] feat: allow /ipns/webui.ipfs.io on api port

lidel:
This PR enables user to try out the latest versions of the Web UI before they are officially released with js-ipfs (opt-in).

With this change in place, a user will be able to visit http://127.0.0.1:5001/ipns/webui.ipfs.io and have it fetch the latest version of the webui. JS IPFS can and should continue to ship with a hardcoded CID for a stable release of webui that it has been tested with, but including the ipns address in the list allows us to test out newer versions.

Context: https://github.com/ipfs-shipyard/ipfs-companion/issues/736
go-ipfs counterpart: https://github.com/ipfs/go-ipfs/pull/6530

### DIFF
--- a/src/http/api/routes/webui.js
+++ b/src/http/api/routes/webui.js
@@ -1,9 +1,25 @@
 'use strict'
 
 const Joi = require('@hapi/joi')
+const Boom = require('@hapi/boom')
 const resources = require('../../gateway/resources')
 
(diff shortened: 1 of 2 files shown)


alanshaw:
Holding off on this until the convo in https://github.com/ipfs/go-ipfs/pull/6530 is resolved.

lidel:
Pretty sure this is no longer the best way of doing this.
See https://github.com/ipfs/go-ipfs/pull/6530#issuecomment-608715653

PR #2287, opened 2019-07-22
chore: update err-code dependency

alanshaw:
`err-code` no longer takes a string as the first argument.

This PR also removes a bunch of useless DEBUG logs that were logging just the error message and adds logging where a caught error is not directly passed back to callbacks. This should help with debuggability.

### DIFF
--- a/src/core/ipns/republisher.js
+++ b/src/core/ipns/republisher.js
@@ -29,10 +29,7 @@ class IpnsRepublisher {
 
   start () {
     if (this._republishHandle) {
-      const errMsg = 'already running'
-
-      log.error(errMsg)
-      throw errcode(new Error(errMsg), 'ERR_REPUBLISH_ALREADY_RUNNING')
+      throw errcode(new Error('already running'), 'ERR_REPUBLISH_ALREADY_RUNNING')
     }
 
     // TODO: this handler should be isolated in another module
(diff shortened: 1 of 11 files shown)


src/core/ipns/republisher.js
```diff
@@ -29,10 +29,7 @@ class IpnsRepublisher {
 
   start () {
     if (this._republishHandle) {
-      const errMsg = 'already running'
-
-      log.error(errMsg)
-      throw errcode(new Error(errMsg), 'ERR_REPUBLISH_ALREADY_RUNNING')
+      throw errcode(new Error('already running'), 'ERR_REPUBLISH_ALREADY_RUNNING')
```
dirkmc:
```suggestion
      throw errcode(new Error('republisher is already running'), 'ERR_REPUBLISH_ALREADY_RUNNING')
```

src/core/ipns/republisher.js
```diff
@@ -78,10 +75,7 @@ class IpnsRepublisher {
     const republishHandle = this._republishHandle
 
     if (!republishHandle) {
-      const errMsg = 'not running'
-
-      log.error(errMsg)
-      return callback(errcode(new Error(errMsg), 'ERR_REPUBLISH_NOT_RUNNING'))
+      return callback(errcode(new Error('not running'), 'ERR_REPUBLISH_NOT_RUNNING'))
```
dirkmc:
```suggestion
      return callback(errcode(new Error('republisher is not running'), 'ERR_REPUBLISH_NOT_RUNNING'))
```

PR #2291, opened 2019-07-23
docs: update delegate node addrs

alanshaw:
Currently setup as aliases for the preload nodes. Allows us to not break application code when these get moved to their own nodes.


mburns:
```
;; ANSWER SECTION:
node0.delegate.ipfs.io.	3600	IN	TXT	"ALIAS for node0.preload.ipfs.io"
```

```
;; ANSWER SECTION:
node1.delegate.ipfs.io.	3600	IN	TXT	"ALIAS for node1.preload.ipfs.io"
```

lidel:
@mburns  we need certs for the new domain as well:

>

mburns:
fixed!

Issue #2000, opened 2019-04-18
IPNS over DNS tracking


hugomrdias:
todo:

- [ ] `ipfs dns` should be recursive by default to follow go-ipfs
- [ ] `ipfs resolve` should be recursive and support `/ipns/<fqdn>`
- [ ] `ipfs name resolve` should be recursive and support `/ipns/<fqdn>`
- [ ] implement dns resolver
- [ ] support cid v1 in subdomain ?? /cc @lidel 
- [ ] finish the RFC following https://github.com/ipfs/RFC/compare/rfc/5-ipns-over-dns

Ref:
https://github.com/ipfs/js-ipfs/issues/1918
https://github.com/ipfs/go-ipfs/issues/5287

lidel:
I'm dropping some notes about DNS over HTTPS:
- **Binary** DNS Queries over HTTPS (DoH): https://tools.ietf.org/html/rfc8484
  (same binary payload format as in regular UDP-based DNS, well understood)
  - good primer on how  `application/dns-message` looks on HTTP https://developers.cloudflare.com/1.1.1.1/dns-over-https/wireformat/
  - https://wiki.mozilla.org/Trusted_Recursive_Resolver
    - DoH endpoint used by Firefox: `https://mozilla.cloudflare-dns.com/dns-query`
- Representing DNS Messages in **JSON**: https://tools.ietf.org/html/rfc8427
  AFAIK this fairly young RFC is supported by DNS provided by Google and Cloudflare as `application/dns-json`:
  - https://developers.google.com/speed/public-dns/docs/dns-over-https
  - https://developers.cloudflare.com/1.1.1.1/dns-over-https/json-format/

If I had to pick which one use in our low level libraries, I am leaning towards binary one.
It not only gives us smaller payload sizes but also more servers support it.

momack2:
@hugomrdias - is there a spec for the js client implementation? the go-ipfs team wants to evaluate whether we can add a go client matching that interface in time for IPFS Camp

cc @Stebalien

momack2:
Is there a status update on this? I see the Q3 OKR has a grade of 0.7 - does that mean there is a finalized/implemented spec?

daviddias:
I'm also interested.

bmann:
Great to meet you @hugomrdias. @expede and @dholms are going to take a look at this and help with the spec / give our opinion.

achingbrain:
[...]

PR #2297, opened 2019-07-24
feat: add support for ipns and recursive to ipfs resolve

hugomrdias:
This PR adds IPNS support to resolve, makes the recursive option true by default and reworks the tests.

Jsdocs were added to the resolve method.

Two interface-core config profile tests needed to be skipped because js-ipfs doesn't support them yet

ref:
https://github.com/ipfs/js-ipfs/issues/2000


src/core/components/resolve.js
```diff
@@ -1,64 +1,82 @@
...
-    }
+/**
+ * @typedef { import("../index") } IPFS
+ */
 
-    opts = opts || {}
+/**
+ * @typedef {Object} ResolveOptions
+ * @prop {string} cidBase - Multibase codec name the CID in the resolved path will be encoded with
+ * @prop {boolean} [recursive=true] - Resolve until the result is an IPFS name
```
dirkmc: Should this be `Resolve until the result is an IPFS key`?

src/core/components/resolve.js
```diff
@@ -1,64 +1,82 @@
...
+
+/**
+ * IPFS Resolve factory
+ *
+ * @param {IPFS} ipfs
+ * @returns {ResolveWrapper}
+ */
+module.exports = (ipfs) => {
+  /**
+   * IPFS Resolve - Resolve the value of names to IPFS
```
dirkmc:
```suggestion
   * IPFS Resolve - Resolve names to IPFS keys
```

src/core/components/resolve.js
dirkmc: Is the recursive parameter implemented? It seems from the code like it's assumed to always be true

src/http/api/resources/resolve.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
   async handler (request, h) {
     const { ipfs } = request.server.app
     const name = request.query.arg
-    const recursive = request.query.r || request.query.recursive || false
+    const recursive = request.query.r || request.query.recursive || true
```
dirkmc: If `request.query.recursive` is explicitly set to `false`, will this code result in `recursive` being set to `true`?

src/core/components/resolve.js
hugomrdias: yeah probably or IPFS path, naming is broken in our code and docs :) 
right now this is just a copy paste from the docs for consistency

hugomrdias: right now this is just to be passed to ipns

src/http/api/resources/resolve.js
hugomrdias: hummm good catch

src/core/components/resolve.js
hugomrdias: right now this is just a copy paste from the docs for consistency

src/core/components/resolve.js
dirkmc: You're right IPFS path is more accurate 👍

package.json
```diff
@@ -74,7 +74,6 @@
     "bs58": "^4.0.1",
     "buffer-peek-stream": "^1.0.1",
     "byteman": "^1.3.5",
-    "callbackify": "^1.1.0",
```
alanshaw: `callbackify` is used here https://github.com/ipfs/js-ipfs/blob/813048ffb32a11cfefc51c2ec8634faaff2a924e/src/core/components/files-mfs.js so will need to be migrated to `promise-nodeify`. Would be rad if you also migrated `libp2p-kad-dht` also 😉🙏:

```
ipfs@0.37.0-rc.0 /Users/alan/Code/pl/ipfs/js-ipfs
├── callbackify@1.1.0 
└─┬ libp2p-kad-dht@0.15.3
  └── callbackify@1.1.0  deduped
```

alanshaw:
BREAKING CHANGE: `recursive` is now `true` by default in `ipfs resolve`

alanshaw:
@hugomrdias 🙏 please can we have this one fixed up?

alanshaw:
blocked by https://github.com/ipfs/js-ipfs/issues/2416

alanshaw:
Ok, DNS on ipfs.io has been fixed - @hugomrdias can you fix up the other test failures please? 🙏

hugomrdias:
ready

PR #2298, opened 2019-07-24
feat: gossipsub as default pubsub

vasco-santos:
Needs:

- [x] Add selection to config 
- [x] New release `js-libp2p` with [libp2p/js-libp2p#365](https://github.com/libp2p/js-libp2p/pull/365)
- [x] [ipfs/interface-js-ipfs-core#503](https://github.com/ipfs/interface-js-ipfs-core/pull/503)


jacobheun:
@vasco-santos fyi, I pushed up a version bump to the new libp2p RC and also fixed the pubsub methods to correct the breaking changes.

jacobheun:
Since we've changed how pubsub is registered, we'll need to release this with the update of libp2p

vasco-santos:
Thanks @jacobheun

vasco-santos:
We need to also change the gossipsub version and update the bundle size. I will do it

jacobheun:
I missed 1 subscribe call, that's been corrected. It's 🍏 !

vasco-santos:
thanks for looking into this @jacobheun 

@alanshaw this is ready for you to have a look

doc/config.md
```diff
@@ -168,6 +170,16 @@ Default:
 
 You can check the [parameter choice for pbkdf2](https://cryptosense.com/parameter-choice-for-pbkdf2/) for more information.
 
+## `Pubsub`
+
+Options for configuring the pubsub subsystem.
+
+### `Router`
```
alanshaw: Can we, not support this? It would mean we have to always bundle two implementations in the browser when 99.9% of people are very likely to use only gossipsub.

package.json
```diff
@@ -121,11 +121,13 @@
     "just-flatten-it": "^2.1.0",
     "just-safe-set": "^2.1.0",
     "kind-of": "^6.0.2",
-    "libp2p": "~0.25.4",
+    "libp2p": "~0.26.1",
     "libp2p-bootstrap": "~0.9.3",
     "libp2p-crypto": "~0.16.0",
     "libp2p-delegated-content-routing": "^0.2.4",
     "libp2p-delegated-peer-routing": "^0.2.4",
+    "libp2p-floodsub": "^0.17.0",
```
alanshaw: Remove?

src/core/components/libp2p.js
```diff
@@ -64,7 +64,9 @@ function defaultBundle ({ datastore, peerInfo, peerBook, options, config }) {
     peerBook,
     modules: {
       contentRouting,
-      peerRouting
+      peerRouting,
+      pubsub: get(config, 'Pubsub.Router', 'gossipsub') === 'floodsub'
+        ? require('libp2p-floodsub') : require('libp2p-gossipsub')
```
alanshaw: Just `pubsub: require('libp2p-gossipsub')`?

src/core/components/pubsub.js
```diff
@@ -16,65 +16,56 @@ module.exports = function pubsub (self) {
...
       if (!callback) {
-        return new Promise((resolve, reject) => {
-          self.libp2p.pubsub.subscribe(topic, options, handler, (err) => {
-            if (err) {
-              return reject(err)
-            }
-            resolve()
-          })
-        })
+        return self.libp2p.pubsub.subscribe(topic, handler, options)
```
alanshaw: 🎉

src/core/components/pubsub.js
```diff
@@ -16,65 +16,56 @@ module.exports = function pubsub (self) {
...
 
-      self.libp2p.pubsub.unsubscribe(topic, handler)
-
       if (!callback) {
-        return Promise.resolve()
+        return self.libp2p.pubsub.unsubscribe(topic, handler)
       }
 
-      setImmediate(() => callback())
+      self.libp2p.pubsub.unsubscribe(topic, handler, callback)
```
alanshaw: Can this be simplified? Instead of:

```js
      if (!callback) {
        return self.libp2p.pubsub.unsubscribe(topic, handler)
      }

      self.libp2p.pubsub.unsubscribe(topic, handler, callback)
```

Just:

```js
      return self.libp2p.pubsub.unsubscribe(topic, handler, callback)
```

As in, shouldn't promisify deal with a `null`/`undefined` `callback` and return a `Promise`?

src/core/config.js
```diff
@@ -68,6 +68,9 @@ const configSchema = s({
       }))
     })),
     Bootstrap: optional(s(['multiaddr-ipfs'])),
+    Pubsub: optional(s({
+      Router: 'string?'
+    })),
```
alanshaw: Remove?

src/core/runtime/config-browser.js
```diff
@@ -27,6 +27,9 @@ module.exports = () => ({
     '/dns4/node0.preload.ipfs.io/tcp/443/wss/ipfs/QmZMxNdpMkewiVZLMRxaNxUeZpDUb34pWjZ1kZvsd16Zic',
     '/dns4/node1.preload.ipfs.io/tcp/443/wss/ipfs/Qmbut9Ywz9YEDrz8ySBSgWyJk41Uvm2QJPhwDJzJyGFsD6'
   ],
+  Pubsub: {
+    Router: 'gossipsub'
+  },
```
alanshaw: Remove?

src/core/runtime/config-nodejs.js
```diff
@@ -40,6 +40,9 @@ module.exports = () => ({
     '/dns4/node0.preload.ipfs.io/tcp/443/wss/ipfs/QmZMxNdpMkewiVZLMRxaNxUeZpDUb34pWjZ1kZvsd16Zic',
     '/dns4/node1.preload.ipfs.io/tcp/443/wss/ipfs/Qmbut9Ywz9YEDrz8ySBSgWyJk41Uvm2QJPhwDJzJyGFsD6'
   ],
+  Pubsub: {
+    Router: 'gossipsub'
+  },
```
alanshaw: Remove?

test/core/libp2p.spec.js
```diff
@@ -149,8 +146,11 @@ describe('libp2p customization', function () {
               enabled: true
             }
           },
-          EXPERIMENTAL: {
-            pubsub: false
+          pubsub: {
+            enabled: false,
+            emitSelf: true,
+            signMessages: true,
+            strictSigning: true
```
alanshaw: Please would you [update the README](https://github.com/ipfs/js-ipfs#optionsexperimental) to reflect the movement of this config option out of `EXPERIMENTAL` and also document these new options? 🙏

alanshaw:
@vasco-santos `interface-ipfs-core@0.111` has been released.

.aegir.js
```diff
@@ -8,7 +8,7 @@ const ipfsdServer = IPFSFactory.createServer()
 const preloadNode = MockPreloadNode.createNode()
 
 module.exports = {
-  bundlesize: { maxSize: '689kB' },
+  bundlesize: { maxSize: '694kB' },
```
alanshaw: Hopefully this can stay the same if we're not bundling 2 implementations?

alanshaw:
FYI I have updated https://github.com/ipfs/js-ipfs-http-client/pull/1059 to support string data in `pubsub.publish` so don't worry about sending a separate PR!

doc/config.md
alanshaw: As discussed on IRC we're going to go with support this in Node.js and just throw "Router unavailable. Configure libp2p.modules.pubsub to use the ${config.Pubsub.Router} router." in the browser.

src/core/components/pubsub.js
vasco-santos: Let me call @jacobheun here. 

He added this, but I think there was an issue here with `promisify`

jacobheun: Promisify can't handle this because it adds extra args to the call. Promisify will see that the last argument is not a function, in this instance it is `null` or `undefined`. Promisify will then add a callback function and call unsubscribe. So you end up with the args `[topic, handler, undefined, promisifyCallback]`, and unsubscribe will internally call `undefined` thinking it's the callback.

README.md
```diff
@@ -506,6 +508,11 @@ You can see the bundle in action in the [custom libp2p example](examples/custom-
         - `kBucketSize` (number): bucket size (default `20`)
         - `randomWalk` (object): configuration for random walk
             - `enabled` (boolean): whether random DHT walking is enabled (default `false`)
+    - `pubsub` (object): Configuration options for Pubsub
+        - `enabled` (boolean): if pubbsub subsystem should be enabled (default: `true`)
+        - `emitSelf` (boolean): whether the node should emit to self on publish, in the event of the topic being subscribed (default: `true`)
```
vasco-santos: @alanshaw we are on the process of probably defaulting this to `false` on libp2p side [libp2p/js-libp2p-pubsub#22](https://github.com/libp2p/js-libp2p-pubsub/issues/22). There are tests who expect this to be `true` in `js-ipfs`. 

Anyway, this change will only occur in a next breaking release of `js-libp2p`. If by then, you would like to change this to `false` in ipfs, let me know

jacobheun: @vasco-santos we should explicitly set the js-ipfs config to `true` for emitSelf now, it's not doing that yet. That way if that changes in a breaking release, js-ipfs won't need an update until it wants to change.

vasco-santos:
@alanshaw I think that I addressed everything from your review

alanshaw:
ping @vasco-santos

vasco-santos:
I will have a look @alanshaw

package.json
```diff
@@ -191,8 +194,8 @@
     "execa": "^2.0.4",
     "form-data": "^2.5.1",
     "hat": "0.0.3",
-    "interface-ipfs-core": "^0.110.0",
-    "ipfsd-ctl": "^0.44.1",
+    "interface-ipfs-core": "^0.111.0",
+    "ipfsd-ctl": "ipfs/js-ipfsd-ctl#fix/pubsub-flag-defaults",
```
vasco-santos: @alanshaw @hugomrdias we have a circular dependency issue here with `ipfsd-ctl`.

vasco-santos:
Also, I am having issues with dns tests in browser. The same is occurring in master though

package.json
alanshaw: Why not just have `--enable-pubsub` as the official option name and create an [alias](https://yargs.js.org/docs/#api-aliaskey-alias) as `--enable-pubsub-experiment`?

vasco-santos: So, are you recommending to `alias` here, and use in the tests `--enable-pubsub-experiment` for now?

vasco-santos: Anyway, I need the ctl to do: https://github.com/ipfs/js-ipfsd-ctl/pull/363/files#diff-d06a3c390787c7b35d2ab649265df1bbR50

hugomrdias: should i merge/release that PR as is ?

vasco-santos: If we do, we cannot delete this branch. But it is the only way to get CI green. My main concern is that we need to wait for the `js-ipfs` release 😡

hugomrdias: we can point to master or a commit until the release

vasco-santos: I am ok with merging that, and take the AI of PR ctl to use master once this PR is merged

hugomrdias: cool

hugomrdias: ctl released as 0.45.0

alanshaw:
@vasco-santos I had to revert the upgrade of `ipfs-http-client` in master because it allows string messages so you'll need to update it here.

vasco-santos:
As talked with @alanshaw , I skipped the dns tests which are currently failing due to a gateway issue (not related to this). I will create a new PR to enable them as soon as this is merged

Issue #1436, opened 2018-07-12
Add examples of required setup/config to bundle IPFS for the browser using popular bundlers


alanshaw:
* [x] Create react app (see https://github.com/ipfs/js-ipfs/issues/1321)
* [x] Webpack
* [x] Browserify
* [x] Parcel (see https://github.com/ipfs/js-ipfs/issues/1344)
* [x] Vue (https://github.com/ipfs/js-ipfs/pull/2302)

PascalPrecht:
We should introduce documentation for that as well in case this doesn't exist yet. Those could then link to the examples.

alanshaw:
...and (on a slightly different but related topic) run our tests against our pre-bundled code.

mitra42:
Coincidentally ... I posted a comment in #1131 about this today, there is an example for webpack, however it ONLY works with the embedded webpack server, the normal `webpack --mode development` or `webpack --mode production` fail because of the "loaders" keyword that only seems to work for the embedded webpack server.   

Also ... I think there should be an example with and without react, to show that its actually relatively simple to include IPFS without react except for the nasty issues with UglifyJS.

alanshaw:
It would be better if IPFS just worked<sup>tm</sup> with uglify defaults.

mikeal:
> It would be better if IPFS just workedtm with uglify defaults.

I don't think that is practical. Uglify has given up on supporting versions of ES past ES5 natively and recommend transpiling with babel before using uglify. But, if you have to use babel anyway, you're better off using babel-minify. Even when the final output is ES5 the minification can be done a lot better.

It's just not practical to say that nowhere in the dep tree will someone not use a feature past ES5. We're sort of fighting against the tide on this one.

mitra42:
I don't know about Uglify, but I do know that IPFS is the only thing in any of the code we include that doesn't pass through webpack with standard unmodified options i.e. if there is any configuration then it is somehow included in the package itself.

mikeal:
Sorry, I should have been clearer. I was only referring to Alan's "just work" comment, not to using uglify with webpack (which is already running through babel).

Just looked through the other issue and it looks like we require the `unused` option to false, which turns off the removal of unused functions.

This means either:

* We're doing something very fancy/clever/horrible to address functions by a dynamic string value.
* There's a bug in UglifyJS.

[...]

PR #2302, opened 2019-07-26
docs: add Vue browser example

dbachko:
Example of how to use `js-ipfs` with `Vue`, one of the desired examples in #1436

Please add your feedback and/or suggestions. Thanks

ipfs-vue-screenshot

resolves https://github.com/ipfs/js-ipfs/issues/1436


examples/browser-vue/src/components/IpfsInfo.vue
```diff
@@ -0,0 +1,51 @@
...
+    this.getIpfsNodeInfo();
+  },
+  methods: {
+    async getIpfsNodeInfo() {
+      try {
+        // Await for ipfs node instance.
+        const ipfs = await this.$ipfs;
+        // Call ipfs `id` method.
+        // Returns the identity of the Peer.
+        const { agentVersion, id } = await pify(ipfs.id)();
```
alanshaw: All IPFS API methods return promises if you do not pass a callback - this is probably not necessary?

examples/browser-vue/src/plugins/vue-ipfs.js
```diff
@@ -0,0 +1,22 @@
+import IPFS from 'ipfs'
+
+function promisifyIpfs(opts) {
+  return new Promise((resolve, reject) => {
+    const node = new IPFS(opts)
+    node.once('ready', () => resolve(node))
+    node.once('error', err => reject(err))
+  })
```
alanshaw: With `0.37.0-rc.0` (the upcoming release) you do not need to do this any more.

You can just `await IPFS.create(opts)`. Can we use this new method in this example?

alanshaw:
Thank you @dbachko ❤️ this is rad! 🚀

PR #2304, opened 2019-07-26
fix: limit concurrent HTTP requests in browser

lidel:
This PR solves request floods like this:

> requests-due-to-the-lack-or-dns-cache-2019-07-23--14-37-58

It adds a limit of concurrent HTTP requests sent to remote API
by dns and preload calls when running in web browser contexts.

Browsers limit connections per host (~6). This change mitigates the problem of expensive and long running calls of one type exhausting connection pool for other uses.
This is similar to problems described in https://github.com/libp2p/js-libp2p-delegated-content-routing/issues/12

This PR additionally limits the number of DNS lookup calls by introducing time-bound cache with eviction rules following what browser already do.

TODO
- [x] add queue for remote HTTP API calls
- [x] add cache for remote `/api/v0/dns` in browser 
- [x] pass dnslink tests in browser
- [x] pass preload tests in browser
  - [x] disable unnecessary preloads when running interface-ipfs-core tests with `process.env.NODE_ENV === 'test'`
- [x] pass electron tests
- [x] figure out `addFromURL` tests 

cc https://github.com/ipfs-shipyard/ipfs-companion/issues/716


src/core/runtime/preload-browser.js
```diff
@@ -1,27 +1,31 @@
...
     .then(res => {
       if (!res.ok) {
         log.error('failed to preload', url, res.status, res.statusText)
         throw new Error(`failed to preload ${url}`)
       }
       return res.text()
     })
-    .then(() => callback())
-    .catch(callback)
+  ).then(() => callback(), callback)
```
hugomrdias: either call both callbacks inside a setImmediate (use .catch) or change everything to promises and use https://github.com/kevinoid/promise-nodeify to callbackify the top-level function

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,63 @@
...
-      } else {
-        return callback(new Error(response.Message))
-      }
+      cache.set(query, response, ttl)
+      return unpackResponse(domain, response, callback)
     })
     .catch((error) => {
       callback(error)
     })
+  )
```
hugomrdias: either call both callbacks inside a setImmediate or change everything to promises and use https://github.com/kevinoid/promise-nodeify to callbackify the top-level function

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,63 @@
...
+
+  // fallback to sending DNSLink query to ipfs.io
+  // TODO: replace this with generic DNS over HTTPS: https://github.com/ipfs/js-ipfs/issues/2212
+  let url = `https://ipfs.io/api/v0/dns?arg=${domain}`
   Object.keys(opts).forEach(prop => {
     url += `&${encodeURIComponent(prop)}=${encodeURIComponent(opts[prop])}`
   })
 
-  self.fetch(url, { mode: 'cors' })
+  _httpQueue.add(() => fetch(url, { mode: 'cors' })
```
hugomrdias: i would use https://github.com/sindresorhus/ky (check this discussion https://github.com/ipfs/js-ipfs-http-client/pull/1059#issuecomment-516378448) 

it will simplify the above lines too and you can specific retries

lidel: Took a stab at `ky` in https://github.com/ipfs/js-ipfs/pull/2304/commits/410ac101935099f2fae4fa39fd0e5621cc2a079d. Code is _much_ simpler now. 

We were using `fetch` for preload and dns in browser, however  `addFromURL` was handled via custom code on top of `node-fetch`. I removed custom code around `fetch` and used `ky-universal` everywhere for consistency.

All of them benefit from retries provided by `ky` out of the box. 
Question: should we hardcode `timeout` and `retry`, or go with defaults?

hugomrdias: we will probably create an ky instance to be used everywhere but for now just leave it hardcoded

src/core/components/files-regular/add-from-url.js
```diff
@@ -14,7 +14,10 @@ module.exports = (self) => {
 
     try {
       const parsedUrl = new URL(url)
```
hugomrdias: this is need for the 
`const path = decodeURIComponent(parsedUrl.pathname.split('/').pop())`
below

we should have the url in the response already in the form of an URL instance and there is no need to `decodeURIComponent` can you check that please ?

src/core/components/files-regular/add-from-url.js
```diff
@@ -14,7 +14,10 @@ module.exports = (self) => {
...
-      const res = await fetch(url)
+      const res = await ky(url, {
+        timeout: 15000,
+        retry: 3
+      })
 
       if (!res.ok) {
```
hugomrdias: you dont need this with `ky` https://github.com/sindresorhus/ky#usage

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,55 @@
...
     opts = {}
   }
-
   opts = opts || {}
-
   domain = encodeURIComponent(domain)
-  let url = `https://ipfs.io/api/v0/dns?arg=${domain}`
 
+  // `opts` impact returned value, so we cache per domain+opts
+  const query = `${domain}${JSON.stringify(opts)}`
```
hugomrdias: please use URLSearchParams toString

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,55 @@
...
+
+  // try cache first
+  if (!opts.nocache && cache.has(query)) {
+    const response = cache.get(query)
+    return unpackResponse(domain, response, callback)
+  }
+
+  // fallback to sending DNSLink query to ipfs.io
+  // TODO: replace this with generic DNS over HTTPS: https://github.com/ipfs/js-ipfs/issues/2212
+  let url = `https://ipfs.io/api/v0/dns?arg=${domain}`
```
hugomrdias: please use https://github.com/sindresorhus/ky#searchparams

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,55 @@
...
-        return callback(null, response.Path)
-      } else {
-        return callback(new Error(response.Message))
-      }
-    })
-    .catch((error) => {
-      callback(error)
-    })
+  _httpQueue.add(async () => {
+    const response = await ky(url, { mode: 'cors' }).json()
```
hugomrdias: why do set the mode to `cors`

src/core/runtime/preload-browser.js
```diff
@@ -1,27 +1,31 @@
...
-        log.error('failed to preload', url, res.status, res.statusText)
-        throw new Error(`failed to preload ${url}`)
-      }
-      return res.text()
-    })
-    .then(() => callback())
-    .catch(callback)
+  _httpQueue.add(async () => {
+    const res = await ky(url, { signal })
+    if (!res.ok) {
```
hugomrdias: no need for this with ky

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,55 @@
...
+    cache.set(query, response, ttl)
+    setImmediate(() => unpackResponse(domain, response, callback))
```
hugomrdias: you can setup a `ky` instance https://github.com/sindresorhus/ky#kycreatedefaultoptions and pre-setup hooks https://github.com/sindresorhus/ky#hooks to handle the cache and unpackResponse

lidel: Done: looks better. I'll take a stab at making this PR promise first.

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,62 @@
...
+const _httpQueue = new PQueue({ concurrency: 4 })
+
+// Delegated HTTP resolver sending DNSLink queries to ipfs.io
+// TODO: replace hardcoded host with configurable DNS over HTTPS: https://github.com/ipfs/js-ipfs/issues/2212
+const api = ky.create({
+  prefixUrl: 'https://ipfs.io/api/v0/',
+  hooks: {
+    afterResponse: [
+      async response => {
+        const query = new URLSearchParams(new URL(response.url).search).toString()
```
hugomrdias: the response doesn't have the searchParams already ?

hugomrdias:
damn ky doesnt support returning a Response from beforeRequest to skip the http request entirely and fetch from cache.

src/core/runtime/dns-browser.js
lidel: sadly no, when I checked it only had `.url` (`string`)

lidel:
Preload tests fail in browser because preload is enabled in browser context by default and while js-ipfs tests disable preload in tests, interface-ipfs-core does not.

This means interface-ipfs-core tests trigger dozens of slow/hanging requests to `nodeX.preload.iofs.io/api/v0/refs` and we execute only 4 at a time. By the time we do preload tests queue is long and it does not finish before timeout.

I believe we should detect js-ipfs is running in test environment (`process.env.NODE_ENV === 'test'` ?) and make preload opt-in in that environment, even in browser. That way we would still be able to tests preload (by explicitly providing preload nodes), but we won't have overhead when running interface-ipfs-core tests.

hugomrdias:
This basically is telling us to "really" solve this problem lol, with http2, http cache and making refs endpoint faster

Still if you want to do this workaround make sure NODE_ENV is properly set in browsers and electron if not file an issue in aegir and i'll fix it.

lidel:
76ae81c disables preload by default when `NODE_ENV` is `test`:

- We only care about preload in... preload tests: we use custom localhost server there and enable preload via explicit config.  
- This means there is no need to enable it by default in test environment. Running preload call in every test of interface-js-ipfs-core was just unecessary overhead. 
- `NODE_ENV` is set ok when running browser tests, so no aegir fix is necessary.

----

Remaining thing to fix is `addFromURL` test with HTTPS  under electron renderer.

lidel:
The problem was that `node-fetch` was not used in electron-renderer context:

`ky-universal` won't use `node-fetch` because electron-renderer already has
`global.fetch` defined, and we can't use the one from electron-renderer as
it returns different errors and makes HTTPS mocking impossible in tests

As a temporary workaround, https://github.com/ipfs/js-ipfs/pull/2304/commits/8f41347f940a16f2249bbc07554fe66548f547ea fixes the problem in userland by assigning `node-fetch` to `global.fetch` before `ky-universal` is imported.

We should solve this upstream somewhow. Options I see are:

- (A) PR `ky-universal` and ensure `node-fetch` is used in electron-renderer  context
- (B) PR `aegir` and ensure `node-fetch` overrides `global.fetch` provided by electron-renderer 
- (C) move workaround from 8f41347f940a16f2249bbc07554fe66548f547ea to `ipfs-utils/src/iso-ky` and use it instead of `ky-universal` directly?

@hugomrdias thoughts?

hugomrdias:
I would like to test with the new electron 6 before deciding

lidel:
Same errors in 6.0.0:

> 2019-08-02-151613_788x334_scrot

lidel:
### Updates

- rebased on top of v0.37.0
- electron tests got fixed for now by explicitly using `node-fetch` in electron-renderer

### On `addFromURL` tests

@hugomrdias  noted that js-ipfs already skips `addFromURL` tests in browsers:

https://github.com/ipfs/js-ipfs/blob/fc768750c5732a144fd1f40ead74ae4e32c0a628/test/core/interface.spec.js#L67-L68

This is because karma running `interface-js-ipfs-core` tests  in browser context is unable to start HTTP(S) server.  We may be able to make it work with something like [karma-server-side](https://github.com/featurist/karma-server-side/), but it sounds like something we need to add to aegir. 

When we have `addFromURL` tests working in web browser, then we can revisit electron errors.

hugomrdias:
for reference 


**hugomrdias:**  ok so we need to find a way to test https
or just skip the https tests
not the others
**lidel:** I'd skip https in browser
**hugomrdias:**  ok cool
remove this https://github.com/ipfs/js-ipfs/blob/fc768750c5732a144fd1f40ead74ae4e32c0a628/test/core/interface.spec.js#L67
and skip inside the tests with ipfs-utils/src/env.js
do you agree ?
**lidel:**  yup

alanshaw:
@lidel what's the current status of this PR - still draft?

lidel:
@alanshaw still a draft. Remaining work here is to figure out `addFromURL` tests in the browser, as noted by @hugomrdias

I took a break from this PR after my vacation, but will revisit it this week.

alanshaw:
Just incase this is helpful, you could spin up a server to serve some content via aegir hooks: https://github.com/ipfs/js-ipfs/blob/c67b8a5406178424092effc55cd8c29605baa393/.aegir.js#L27

Maybe the server accepts a querystring that is the fixture you want it to return? Then in tests you can call it and assert the correct content was added.

Won't work with HTTPS though, but only HTTP is better than nothing.

lidel:
Ok, took a stab at enabling `addFromURL` tests in browser. 
Two HTTPS tests are  skipped, but everything else works as expected.

@alanshaw @hugomrdias  Marking this PR ready for review.
Let me know if this can be improved somehow.

**Note:** this PR requires https://github.com/ipfs/interface-js-ipfs-core/pull/514 to be merged first.

hugomrdias:
we can disable browser ssl validation in the karma config, @lidel do you think we would still be testing https with that disabled ?

lidel:
@hugomrdias Yes, it would still be HTTPS, we just would not verify cert against know CAs in system/browser keystore. 

I imagine setting that in karma would be running chrome with `--ignore-certificate-errors` and `--ignore-urlfetcher-cert-requests` and something similar in Firefox.  Not sure about Electron (in theory it should respect chromium flags but you never know).

But honestly I would leave HTTPS it skipped in browsers. I don't know why we created tests for HTTPS in the first place. If anything, I would remove them from `addFromURL` test suite and  have one HTTP server less :upside_down_face:

src/core/components/files-regular/add-from-url.js
```diff
@@ -1,41 +1,22 @@
...
-      throw err
-    }
+const nodeify = require('promise-nodeify')
+const { default: ky } = require('ky-universal')
+
+module.exports = (ipfs) => {
+  const addFromURL = async (url, opts = {}) => {
+    const res = await ky.get(url)
+    const path = decodeURIComponent(new URL(res.url).pathname.split('/').pop())
+    const content = Buffer.from(await res.arrayBuffer())
```
alanshaw: One day we will use something like https://github.com/ipfs/js-ipfs-http-client/blob/master/src/lib/stream-to-iterable.js to stream the response.

src/core/runtime/preload-browser.js
```diff
@@ -1,27 +1,27 @@
...
-        log.error('failed to preload', url, res.status, res.statusText)
-        throw new Error(`failed to preload ${url}`)
-      }
-      return res.text()
-    })
-    .then(() => callback())
-    .catch(callback)
+  _httpQueue.add(async () => {
+    await ky.get(url, { signal })
+    setImmediate(callback)
```
alanshaw: No need to `setImmediate` already executed on a future tick.

```suggestion
```

src/core/runtime/preload-browser.js
```diff
@@ -1,27 +1,27 @@
...
+  }).catch((err) => setImmediate(() => callback(err)))
```
alanshaw: No need to `setImmediate` already executed on a future tick. Passing callback as second parameter to `.then` is safer (instead of a catch) because it won't be called again if it throws.

```suggestion
  }).then(() => callback(), callback)
```

alanshaw: This could just be:

```js
  _httpQueue.add(() => ky.get(url, { signal })).then(() => callback(), callback)
```

package.json
```diff
@@ -194,7 +196,7 @@
     "execa": "^2.0.4",
     "form-data": "^2.5.1",
     "hat": "0.0.3",
-    "interface-ipfs-core": "^0.111.0",
+    "interface-ipfs-core": "https://github.com/ipfs/interface-js-ipfs-core/tarball/a810adfcbb977257851a4be3aae158a350119810/interface-js-ipfs-core.tar.gz",
```
lidel: @alanshaw are you able to merge & release https://github.com/ipfs/interface-js-ipfs-core/pull/514, so this can be switched to real version ?

lidel:
(rebased & resolved conflicts + applied suggestions from  review)

src/core/runtime/preload-browser.js
lidel: @alanshaw  @hugomrdias raised concern that we need `setImmediate`, otherwise the callback will get caught inside the promise scope. Unless we are sure callback is "safe", i'd keep it

hugomrdias: yes the code inside the callback needs to break from the promise stack, because it handles stuff callback style. This is one of the reason we should always use packages to callbackify or promisify they handle this stuff.

lidel: Yikes. I will now be double paranoid about this. 
Restored in https://github.com/ipfs/js-ipfs/pull/2304/commits/c9c1080ce8b8b94b833ce437753919ce79d8829d

lidel:
Rebased again + switched to interface-js-ipfs-core v0.113.0

src/core/config.js
```diff
@@ -31,7 +31,10 @@ const configSchema = s({
     enabled: 'boolean?',
     addresses: optional(s(['multiaddr'])),
     interval: 'number?'
-  }, { enabled: true, interval: 30 * 1000 }),
+  }, { // defaults
+    enabled: process.env.NODE_ENV !== 'test', // preload by default, unless in test env
```
alanshaw: Process is not available in the browser. All users bundling IPFS will have to configure their bundler to shim a global process object. Can we postpone these changes to a separate PR?

.aegir.js
```diff
@@ -3,12 +3,23 @@
...
+const EchoHttpServer = require('interface-ipfs-core/src/utils/echo-http-server')
 
 const ipfsdServer = IPFSFactory.createServer()
 const preloadNode = MockPreloadNode.createNode()
+const httpEchoServer = EchoHttpServer.createServer() // used by addFromURL
+
+const batch = (call, done, ...srvs) => parallel(srvs.map(srv => cb => {
+  if (srv === ipfsdServer) {
+    srv[call]()
+    cb()
```
alanshaw: Wait for ipfsd server to start/stop before calling callback?

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,62 @@
...
+// Delegated HTTP resolver sending DNSLink queries to ipfs.io
+// TODO: replace hardcoded host with configurable DNS over HTTPS: https://github.com/ipfs/js-ipfs/issues/2212
+const api = ky.create({
+  prefixUrl: 'https://ipfs.io/api/v0/',
+  hooks: {
+    afterResponse: [
+      async (input, options, response) => {
+        const query = new URL(response.url).search.slice(1)
+        const json = await response.json()
+        cache.set(query, json, ttl)
```
alanshaw: Only cache when `response.ok`?

lidel: We can't: lack of DNSLink is returned as HTTP 500:
https://ipfs.io/api/v0/dns/arg=google.com

src/core/config.js
lidel: Can't postpone this without disabling preload tests.
If we enable preload by default in tests, then p-queue will get filled with requests for ephemeral CIDs that take time to timeout, and actual preload tests will fail.
 
Would it be ok to do:
```js
typeof process === 'undefined' || process && process.env && process.env.NODE_ENV !== 'test'
```
instead?

.aegir.js
lidel: done: https://github.com/ipfs/js-ipfs/pull/2304/commits/e7534ff624cf7171d1e1b352dc2eb6bf01b3acc7

src/core/config.js
lidel: Would the check proposed in https://github.com/ipfs/js-ipfs-utils/pull/10 be acceptable?

lidel: @alanshaw switched to ipfs-utils v0.2.0 in https://github.com/ipfs/js-ipfs/pull/2304/commits/ad65329253da333885b86c7927aa8f0a1e628551 should be ok to merge now

lidel:
@alanshaw switched to ipfs-utils v0.2.0 in https://github.com/ipfs/js-ipfs/pull/2304/commits/ad65329253da333885b86c7927aa8f0a1e628551 should be ok to merge now

alanshaw:
hmm, tests are failing in windows and both browsers and electron. Something is not right here! The `addFromURL` tests are failing in windows and they have been changed in `interface-ipfs-core` for _this_ PR.

lidel:
oh boy, this PR is cursed :eyes: 

@alanshaw electron was a random one, but remaining Windows issue was triggered by parallel `ipfs.add` of the same data. I fixed it in https://github.com/ipfs/interface-js-ipfs-core/pull/523 

Needs TODO to be merged & released.

lidel:
Resolved conflicts + re-enabled tests for https://github.com/ipfs/js-ipfs/pull/2379
Windows tests failed due to https://github.com/ipfs/interface-js-ipfs-core/pull/523#issuecomment-529388983 (should be fixed now, waiting for ci)

hugomrdias:
This should be ready to merge now.

src/core/components/files-regular/add-from-url.js
achingbrain: That day is today.  Well, yesterday.  Since we merged #2379 you should be able to pass iterables to `ipfs.add`.

src/core/runtime/dns-browser.js
```diff
@@ -1,33 +1,62 @@
...
+    ]
+  }
+})
+
+const ipfsPath = (response) => {
+  if (response.Path) return response.Path
+  throw new Error(response.Message)
+}
+
+module.exports = (fqdn, opts = {}, cb) => {
```
achingbrain: Can this be promise-first please?  E.g. make this an async method but export a nodeified version.  This will make future refactors to remove callbacks easier.

src/core/runtime/preload-browser.js
```diff
@@ -1,27 +1,25 @@
...
+const { default: PQueue } = require('p-queue')
+const { default: ky } = require('ky-universal')
 const debug = require('debug')
 
 const log = debug('ipfs:preload')
 log.error = debug('ipfs:preload:error')
 
+// browsers limit concurrent connections per host,
+// we don't want preload calls to exhaust the limit (~6)
+const _httpQueue = new PQueue({ concurrency: 4 })
```
achingbrain: Why the underscore here & in the `runtime/dns-browser.js` file?

src/core/components/files-regular/add-from-url.js
lidel: Seems that `ky` does not expose ReadableStream yet:
https://github.com/sindresorhus/ky/issues/3#issuecomment-502414532

To avoid blocking this PR, I'd keep this as-is and add streaming in separate PR, after support in `ky` lands.

src/core/runtime/dns-browser.js
lidel: I think its already done? the async method is inside: 
```js
const resolveDnslink = async (fqdn, opts = {}) => {
  ...
}
return nodeify(resolveDnslink(fqdn, opts), cb)
```

when we refactor to remove callbacks we just remove the wrapper.
Let me know if this should be handled some other way.

src/core/runtime/preload-browser.js
lidel: refactor scar tissue, fixed in https://github.com/ipfs/js-ipfs/pull/2304/commits/2e82b2a9b468db99937256f3d77261eaef9a8c0f

lidel:
Woohoo :tada: 

3HUQNgOl

typhu-xyz:
I'm encountering this problem in latest `ipfs-http-client` - has this been dropped during latest development?

hugomrdias:
> I'm encountering this problem in latest `ipfs-http-client` - has this been dropped during latest development?

Can you explain further ? this was a problem in js-ipfs not http-client

typhu-xyz:
@hugomrdias I was playing around with orbitdb + ipfs-http-client in web browser (connected to my local go-ipfs). I found that when I subscribe to a lot of datastores, orbitdb's periodic fetches / pubsub on each datastore effectively use up the concurrent HTTP request limit and stall the entire pipeline of requests, just like what's shown in the screenshot in the PR. This behavior seems to be a result from https://github.com/ipfs-shipyard/ipfs-pubsub-1on1.

When I saw this PR, I thought this might be something that addresses my issue, but you are right - I missed that this PR was for js-ipfs. It wouldn't make much sense for ipfs-http-client in fact.

hugomrdias:
can you make another issue with further info maybe a repo we can clone please ?

typhu-xyz:
I've decided to drop OrbitDB for now, so maybe won't see the issue anymore. But if I manage to reproduce it in a consistent manner will log a separate issue like you suggested. Thanks!

PR #2354, opened 2019-08-14
chore(package): update @hapi/boom to version 7.4.3

alanshaw:


alanshaw:
Tests are passing, bundlesize check is erroring becuase of permissions error

Issue #2335, opened 2019-08-06
Enable preload on MFS commands that accept IPFS paths


alanshaw:
Some MFS commands accept IPFS paths as input. When IPFS paths are passed, we should send a preload request so that the content can be retrieved.

Notably, the CID resolver in `ipfs-http-response` switched to using `ipfs.files.stat` instead of `ipfs.object.stat` (https://github.com/ipfs/js-ipfs-http-response/pull/19). This means that a JS IPFS gateway currently does not send preload requests and so is useful only for fetching content the node or the preload nodes already have in their repo.

cc @lidel @achingbrain @vasco-santos

lidel:
I am +1 for this: was experimenting and thinking about adding an explicit preload call for paths requested via [embedded gateway in Brave](https://github.com/ipfs-shipyard/ipfs-companion/issues/716), but  doing it at `ipfs.files.stat` is even better.

PR #2355, opened 2019-08-14
fix: enable preload on MFS commands that accept IPFS paths

alanshaw:
~~@achingbrain please could you verify these are the methods that accept IPFS paths? Also, do they accept CID instances or CID strings or "cid paths" as I've allowed here or is it _only_ IPFS paths?~~ Verified myself - it is IPFS paths (Strings) and CID _instances_.

TODO:

* [x] Add tests

fixes https://github.com/ipfs/js-ipfs/issues/2335

### DIFF
--- a/src/core/components/files-mfs.js
+++ b/src/core/components/files-mfs.js
@@ -35,19 +35,12 @@ module.exports = self => {
   })
 
   const withPreload = fn => (...args) => {
-    const cids = args.reduce((cids, p) => {
-      if (isIpfs.ipfsPath(p)) {
-        cids.push(p.split('/')[2])
-      } else if (isIpfs.cid(p)) {
-        cids.push(p)
-      }
-      return cids
-    }, [])
+    const paths = args.filter(arg => isIpfs.ipfsPath(arg) || isIpfs.cid(arg))
 
-    if (cids.length) {
+    if (paths.length) {
       const options = args[args.length - 1]
       if (options.preload !== false) {
-        cids.forEach(cid => self._preload(cid))
+        paths.forEach(path => self._preload(path))
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/core/components/files-mfs.js
```diff
@@ -32,15 +34,37 @@ module.exports = self => {
     repoOwner: self._options.repoOwner
   })
 
+  const withPreload = fn => (...args) => {
+    const cids = args.reduce((cids, p) => {
+      if (isIpfs.ipfsPath(p)) {
```
lidel: AFAIK `/ipns`/ paths are not supported by MFS, so HTTP gateway resolves `/ipns/` → `/ipfs/` before passing it to `files.stat` later:

https://github.com/ipfs/js-ipfs/blob/3cade672b1c3d2aa19394e2980cfad9cd94764dc/src/http/gateway/resources/gateway.js#L40-L46

That being said,  is it possible `/ipns/` paths will be supported in the future?

```suggestion
      if (isIpfs.path(p)) {
```

alanshaw: I think we'd also have to add the resolve step to this PR if we wanted to "future-proof" the check...right now it would end up sending Peer IDs to the preload nodes...

alanshaw:
Note: Tests are all passing on CI. The `bundlesize` check is currently failing due to a permissions error.

lidel:
While testing this PR in Brave I noticed [multiple requests for the same CID blocking browser's request queue](https://user-images.githubusercontent.com/157609/63433912-54acfa80-c424-11e9-9dfe-876e60b1180b.png). 

 Potential mitigations:
  - cache/skip subsequent preload calls of the same CID for some time 
    (probably the best way to handle this)
  - our preload nodes (`*.preload.ipfs.io`) should return HTTP headers to enable time-based [http caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching) in web browser and/or cache result at Nginx level for a few minutes to decrease load on go-ipfs

PR #2358, opened 2019-08-14
fix: preload addreses with trailing slash

### LINKED ISSUE #2333: Https address adds incorrectly to preload addreses
- **Version**: 0.35
- **Platform**: Browser
- **Subsystem**: Chrome

#### Description:
I'm trying to add https preload address with specific port as described in multiaddr tests:
https://github.com/multiformats/js-mafmt/blob/master/test/index.spec.js#L79
by following code:
``` 
const ipfsPromise = new Promise((resolve, reject) => {
    const ipfs = window['Ipfs'].createNode({
          preload: {
            enabled: true,
            addresses: ['/dnsaddr/ipfs.io/tcp/3456/https/']
         }
    });
   ipfs.once('ready', () => resolve(ipfs))
   ipfs.once('error', err => reject(err))
});
```

But when i'm requesting for ipld:
```
ipfsPromise.then((ipfs) => {
    ipfs.dag.get('bafyreicxal2jgaej7ihlzhlw34ficxhjcuw43omrtlw2jcymxj7lnptjs4');
});
``` 

I see in Network tab of developer console that ipfs trying to fetch:
```
http://https//ipfs.io:3456/api/v0/refs?r=true&arg=zdpuArH6ZmDNgCauNuoUCgzHurvCbZ9DQdoEAvftDNsAZM5Ui
```

With incorrect `http://https//` protocol.

lidel:
This PR:

- moves path utils to upstream utils
- closes #2333

### DIFF
--- a/src/core/preload.js
+++ b/src/core/preload.js
@@ -7,6 +7,7 @@ const debug = require('debug')
 const CID = require('cids')
 const shuffle = require('array-shuffle')
 const preload = require('./runtime/preload-nodejs')
+const { removeTrailingSlash } = require('../utils/path')
 
 const log = debug('ipfs:preload')
 log.error = debug('ipfs:preload:error')
@@ -86,7 +87,8 @@ module.exports = self => {
 }
 
 function apiAddrToUri (addr) {
-  if (!(addr.endsWith('http') || addr.endsWith('https'))) {
+  addr = removeTrailingSlash(addr)
+  if (!(addr.endsWith('/http') || addr.endsWith('/https') || addr.endsWith('/tls/http'))) {
     addr = addr + '/http'
   }
   return toUri(addr)
--- a/src/http/gateway/resources/gateway.js
+++ b/src/http/gateway/resources/gateway.js
@@ -12,7 +12,7 @@ const peek = require('buffer-peek-stream')
 
 const multibase = require('multibase')
 const { resolver } = require('ipfs-http-response')
-const PathUtils = require('../utils/path')
+const PathUtils = require('../../../utils/path')
 const { cidToString } = require('../../../utils/cid')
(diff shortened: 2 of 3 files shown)


alanshaw:
@lidel `multiaddr` already deals with this issue:

```js
multiaddr('/dns4/ipfs.io/tcp/3456/https/').toString()
// outputs: /dns4/ipfs.io/tcp/3456/https
```

The real issue is that we want to assume "http" when converting a multiaddr to a URI but `multiaddr-to-uri` does not.

I think it's a reasonable assumption and I reckon we can save ourselves a bunch of trouble and boilerplate by pushing this behaviour into `multiaddr-to-uri` (with a configuration option to disable it).

I think we should send a PR to `multiaddr-to-uri` to assume `http://` if `/http(s)` is not explicit in the multiaddr and add a configuration option to disable it. This would solve our trailing slash problem and make it easier to use `multiaddr-to-uri` due to less boilerplate appending `/http` for 99.9% of users.

So something like the following would work:

```js
toUri('/dns4/ipfs.io/tcp/3456')
// outputs: http://ipfs.io:3456

toUri('/dns4/ipfs.io/tcp/3456/https/')
// outputs: https://ipfs.io:3456

toUri('/dns4/ipfs.io/tcp/3456', { assumeHttp: false })
// outputs: tcp://ipfs.io:3456
```

What do you think?

lidel:
Makes sense. Will open PR shortly.

alanshaw:
I'm going to close this PR. Please open a new one that removes the `endsWith('/http')` boilerplate and updates the `multiaddr-to-uri` dependency when the time comes. 😘

lidel:
Continued in https://github.com/multiformats/js-multiaddr-to-uri/pull/7

PR #2368, opened 2019-08-17
tests: update ipfsd-ctl

vmx:
This new version of ipfsd-ctl is using async/await instead of callbacks.
Hence some rather mechanical changes are needed for tests.

I've tried to keep this change as focused as possible. The tests could
well need some more async/await love, but I didn't want to mix concerns.


vmx:
I've no clue why the tests fail on Windows, has anyone seen something like this before?

I also don't know why the bundle size is an issue, it seems fine locally (@hugomrdias do you perhaps know?).

test/cli/swarm.js
```diff
@@ -41,43 +40,30 @@ describe('swarm', () => {
...
+        })
+      ])
+      ipfsA = ipfsExec(res[0].repoPath)
+      const id = await res[1].api.id()
+      bMultiaddr = id.addresses[0]
+      nodes.push(...res)
     })
 
-    after((done) => parallel(nodes.map((node) => (cb) => node.stop(cb)), done))
+    after(() => Promise.all([nodes.map((node) => node.stop())]))
```
alanshaw:
```suggestion
    after(() => Promise.all(nodes.map((node) => node.stop())))
```

alanshaw:
🎉 thanks @achingbrain for fixing this!

PR #2372, opened 2019-08-20
fix: Do not load all of a DAG into memory when pinning

### LINKED ISSUE #2310: jsipfs add -Qr <directory-containing_1206 _files> doesn't work
Version: 0.35.0
Platform: Linux ip-XXX-aws #42-Ubuntu SMP Fri May 17 13:47:10 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

#### Description: I have a folder containing 1200+ files and takes 4.1G disk space. When I did
jsipfs add -Qr <folder_name>

It returns a Hash QmedigS67VJr71NZs7U1DBhrWXmBFwEHRVTWHQ6f624PMm. But
jsipfs ls /ipfs/QmedigS67VJr71NZs7U1DBhrWXmBFwEHRVTWHQ6f624PMm
doesn't return anything (I expect it to return a list of 1200+ files).
Local gateway also doesn't work.
 

See above

achingbrain:
Given a `CID`, the `dag. _getRecursive` method returns a list of all descendents of the node with the passed `CID`.  This can cause enormous memory usage when importing large datasets.

Where this method is invoked the results are either a) disgarded or b) used to calculate the `CID`s of the nodes which is then bad for memory *and* CPU usage.

This PR removes the buffering and `CID` recalculating for a nice speedup when adding large datasets.

In my (non-representative, may need all the other unfinished async/iterator stuff) testing, importing folder of 4MB files totalling about 5GB files with content from `/dev/urandom` into a fresh repo with a daemon running in the background is now:

```
go-ipfs
real	3m43.741s
user	0m31.955s
sys	0m31.959s
```

```
js-ipfs
real	3m40.725s
user	0m7.352s
sys	0m4.489s
```

Which is nice.

fixes #2310

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -8,7 +8,7 @@ const ipfsdServer = IPFSFactory.createServer()
 const preloadNode = MockPreloadNode.createNode()
 
 module.exports = {
-  bundlesize: { maxSize: '689kB' },
+  bundlesize: { maxSize: '756KB' },
   webpack: {
     resolve: {
       mainFields: ['browser', 'main'],
--- a/.aegir.js
+++ b/.aegir.js
@@ -8,7 +8,7 @@ const ipfsdServer = IPFSFactory.createServer()
 const preloadNode = MockPreloadNode.createNode()
 
 module.exports = {
-  bundlesize: { maxSize: '756KB' },
+  bundlesize: { maxSize: '1MB' },
   webpack: {
     resolve: {
       mainFields: ['browser', 'main'],
--- a/.aegir.js
+++ b/.aegir.js
@@ -8,7 +8,7 @@ const ipfsdServer = IPFSFactory.createServer()
 const preloadNode = MockPreloadNode.createNode()
 
 module.exports = {
-  bundlesize: { maxSize: '1MB' },
+  bundlesize: { maxSize: '756KB' },
   webpack: {
     resolve: {
       mainFields: ['browser', 'main'],
--- a/.aegir.js
+++ b/.aegir.js
@@ -8,7 +8,7 @@ const ipfsdServer = IPFSFactory.createServer()
 const preloadNode = MockPreloadNode.createNode()
 
 module.exports = {
-  bundlesize: { maxSize: '756KB' },
+  bundlesize: { maxSize: '689kB' },
   webpack: {
     resolve: {
       mainFields: ['browser', 'main'],
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ jobs:
   include:
     - stage: check
       script:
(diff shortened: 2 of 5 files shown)


src/core/components/pin.js
```diff
@@ -209,8 +230,10 @@ module.exports = (self) => {
             }
 
             // make sure we have the object
-            dag.get(new CID(multihash), { preload: options.preload }, (err) => {
-              if (err) { return cb(err) }
+            dag.get(cid, { preload: options.preload }, (err) => {
```
achingbrain: Would we ever want to preload here?  I was pinning a few 10s of 1000s of nodes during an add operation and eventually the preload servers start returning `429`s.

Also during an add operation we preload, then pin but this means pinning can preload too?

alanshaw:
Holy wow 3s faster than go-ipfs!

.aegir.js
```diff
@@ -8,7 +8,7 @@ const ipfsdServer = IPFSFactory.createServer()
 const preloadNode = MockPreloadNode.createNode()
 
 module.exports = {
-  bundlesize: { maxSize: '689kB' },
+  bundlesize: { maxSize: '756KB' },
```
alanshaw: I'm not sure this is change is needed, In this branch I'm getting the following:

```
⨎ npx aegir build --bundlesize
Child
       1466 modules
Child
       1466 modules
 PASS  ./dist/index.min.js: 682.3KB < maxSize 756KB (gzip) 
```

.travis.yml
```diff
@@ -25,7 +25,7 @@ jobs:
   include:
     - stage: check
       script:
-        - npx aegir build --bundlesize
+        # - npx aegir build --bundlesize
```
alanshaw: Note to self we need a sister PR that re-enables this after it is merged.

src/core/components/pin.js
```diff
@@ -52,30 +53,49 @@ module.exports = (self) => {
...
+      cb()
+    }
+    q.error = (err) => {
+      q.kill()
+      cb(err)
+    }
+    q.push({ cid })
+  }
+
+  function getIndirectKeys (callback) {
```
alanshaw: You need to pass the `preload` option from `pin.ls` here and then pass it to `walkDag`

.aegir.js
achingbrain: Will revert since we're disabling it anyway

achingbrain:
> Holy wow 3s faster than go-ipfs!

For certain datasets under certain conditions.  Not cracking out the 🍾 just yet...

src/core/components/pin.js
achingbrain: Done!

PR #2375, opened 2019-08-21
chore(package): update superstruct to version 0.6.2

alanshaw:


alanshaw:
Tests are passing, bundlesize check is failing due to permissions error.

PR #2379, opened 2019-08-21
feat: support adding async iterators

achingbrain:
Adds a `ipfs._addAsyncIterator` method for adding async iterators and refactors all add methods to call this, as when the Great Async Iterator Migration is complete this will become the one true method to add files to IPFS.

Purposefully does not add symmetrical `ipfs._catAsyncIterator` methods as this PR is big enough already.

Also fixes up the `jsipfs add` progress indicator so it works more like the `go-ipfs` one and prints out hashes as it goes instead of buffering them all up in one go, at the cost of not sorting the outout.

Also, also allows you to control preload behaviour from the command line.

Depends on:

* [x] https://github.com/ipfs/js-ipfs-http-client/pull/1078
* [x] https://github.com/ipfs/js-ipfs-mfs/pull/59
* [x] https://github.com/ipfs/js-ipfs-utils/pull/2


src/core/components/files-regular/add-async-iterator.js
```diff
@@ -0,0 +1,200 @@
...
+  return async function * (source) {
+    for await (const file of source) {
+      // Pin a file if it is the root dir of a recursive add or the single file
+      // of a direct add.
+      const pin = 'pin' in opts ? opts.pin : true
+      const isRootDir = !file.path.includes('/')
+      const shouldPin = pin && isRootDir && !opts.onlyHash && !opts.hashAlg
+
+      if (shouldPin) {
+        await ipfs.pin.add(file.hash, { preload: false })
```
achingbrain: We could probably do this asynchronously as well?

src/core/components/files-regular/add-async-iterator.js
```diff
@@ -0,0 +1,200 @@
...
+
+    const iterator = pipe(
+      source,
+      validateInput(),
+      normalizeInput(opts),
+      doImport(self, opts),
+      prepareFile(self, opts),
+      preloadFile(self, opts),
+      pinFile(self, opts)
+      // sort?
```
achingbrain: We used to buffer & sort all the output.  This can lead to OOM with really, really large imports (think: package manager scale datasets) and `go-ipfs` doesn't do it, so 🤷‍♂

src/cli/commands/add.js
```diff
@@ -166,32 +123,84 @@ module.exports = {
...
+          if (argv.quieter) {
+            finalHash = file.hash
+            continue
+          }
+
+          const cid = cidToString(file.hash, { base: argv.cidBase })
+          let message = cid
+
+          if (!argv.quiet) {
+            // print the hash twice if we are piping from stdin
```
achingbrain: Why does `go-ipfs` do this?

src/cli/commands/add.js
```diff
@@ -149,6 +101,11 @@ module.exports = {
       type: 'boolean',
       default: true,
       describe: 'Pin this object when adding'
+    },
+    preload: {
+      type: 'boolean',
+      default: true,
+      describe: 'Preload this object when adding'
```
achingbrain: You might want to disable this - servers, enormous datasets, etc.

src/core/components/files-regular/add-async-iterator.js
```diff
@@ -0,0 +1,200 @@
...
+  return async function * (source) {
+    for await (const file of source) {
+      let cid = file.cid
+      const b58Hash = cid.toBaseEncodedString()
+
+      if (opts.onlyHash) {
+        yield {
+          path: file.path === undefined ? b58Hash : (file.path || ''),
+          hash: b58Hash,
+          size: file.unixfs.fileSize()
```
achingbrain: This will be different to when you don't specify `onlyHash` as the node has been serialized so includes the protobuf overhead.  Typically we only show the user the hash when they've said `onlyHash` so I don't know how much of a problem this is.  I think we want to switch this over to only show file sizes eventually anyway so it might not be a big deal.

package.json
```diff
@@ -118,6 +118,8 @@
     "is-pull-stream": "~0.0.0",
     "is-stream": "^2.0.0",
     "iso-url": "~0.4.6",
+    "it-pipe": "^1.0.1",
+    "it-glob": "^0.0.1",
```
achingbrain: Can't help but mentally exclaim "it pipe!, it glob!" in childlike wonder when I read this.

src/core/components/files-regular/add-readable-stream.js
```diff
@@ -1,55 +1,45 @@
...
-    const s = pull(
-      p,
-      self.addPullStream(options)
-    )
+    const stream = new PassThrough({
+      objectMode: true
+    })
 
-    const retStream = new AddHelper(s, p)
+    const output = new AddHelper(stream, {
```
alanshaw: Can [`it-to-stream`](https://www.npmjs.com/package/it-to-stream) or [`stream-to-it`](https://www.npmjs.com/package/stream-to-it) help here to avoid buffering and respect backpressure?

src/core/components/files-regular/add-async-iterator.js
alanshaw: Probably needs to be done in the gc read lock for add though

src/core/components/files-regular/add-async-iterator.js
```diff
@@ -0,0 +1,200 @@
...
+    if (Buffer.isBuffer(source) || typeof source === 'string') {
+      source = [
+        source
+      ]
+    }
+
+    const iterator = pipe(
+      source,
+      validateInput(),
+      normalizeInput(opts),
```
alanshaw: I made a comprehensive normalise in https://github.com/ipfs/js-ipfs-http-client/pull/1063/files#diff-7cec0363d7e0ecfb1f2d5e04b55e2cac and https://github.com/ipfs/js-ipfs-http-client/pull/1063/files#diff-8913549ebf1a95e3151ad16513fa9a6f - it would be good to check this is supporting the same inputs (or come to an agreement on what inputs we support).

achingbrain: Nice, maybe we could move that into [`ipfs-utils`](https://github.com/ipfs/js-ipfs-utils)?

achingbrain:
I might actually break this up into three PRs - one for addAsyncIterator, one for parameterising `preload` and one for fixing the progress bar.  Should make it easier to digest.

achingbrain:
See #2384 and #2386.  I'll rebase this once those are in.

src/core/components/files-regular/add-readable-stream.js
achingbrain: Good shout, I'll pull those in.

src/core/components/files-regular/add-async-iterator.js
alanshaw: Yes, will do 👍

alanshaw:
@achingbrain those PRs are now merged 😇

achingbrain:
> Can we remove `glob` from dependencies?

Not yet - it's still used in `src/cli/commands/commands.js` and `src/core/components/init-assets.js`.

src/core/components/files-regular/add-async-iterator.js
achingbrain: I'm going to have a go at moving those utility functions over to `ipfs-utils` so they can be shared between the http client & core as it's going to simplify this a little.

achingbrain: I'm going to refactor some of it though, because it's full of `Buffer.from` which involves a buffer copy which is slow.

achingbrain:
There are a few `addFromUrl` interface tests new in `v0.113.0` failing on Windows but they will be fixed by ipfs/js-datastore-fs#27

achingbrain:
fireworks

PR #2384, opened 2019-08-22
feat: allow controlling preload behaviour from cli

achingbrain:
Pulled out from #2379

Depends on:

* [x] https://github.com/ipfs/js-ipfs/pull/2391

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ jobs:
   include:
     - stage: check
       script:
-        # - npx aegir build --bundlesize
+        - npx aegir build --bundlesize
         - npx aegir dep-check -- -i wrtc -i electron-webrtc
         - npm run lint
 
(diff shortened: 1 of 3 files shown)


alanshaw:
Should this be exposed in the HTTP API? Otherwise this won't work when talking to a daemon. Also, should `ipfs daemon --enable-preload=false` take precedence?

achingbrain:
> Should this be exposed in the HTTP API?

Good point, I have added this.

> Should `ipfs daemon --enable-preload=false` take precedence?

Not sure. I guess maybe if you've specified it explicitly?

achingbrain:
Exposed in `ipfs-http-client` here: https://github.com/ipfs/js-ipfs-http-client/pull/1079

alanshaw:
> > Should `ipfs daemon --enable-preload=false` take precedence?
> 
> Not sure. I guess maybe if you've specified it explicitly?

Just looked in the preload component and this happens already. If you start the daemon with `--enable-preload=false` then specifying `preload: true` here will have no effect.

PR #2387, opened 2019-08-22
fix: Do not load all of a DAG into memory when pinning

achingbrain:
Port of #2372 into gc branch to ease merging

### DIFF
--- a/src/core/components/pin/pin-manager.js
+++ b/src/core/components/pin/pin-manager.js
@@ -1,14 +1,14 @@
 /* eslint max-nested-callbacks: ["error", 8] */
 'use strict'
 
-const { DAGNode, DAGLink, util } = require('ipld-dag-pb')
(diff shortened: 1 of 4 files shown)


src/core/components/pin/pin-manager.js
```diff
@@ -51,30 +79,21 @@ class PinManager {
     return Array.from(this.recursivePins, key => new CID(key).buffer)
   }
 
-  getIndirectKeys (callback) {
+  getIndirectKeys ({ preload }, callback) {
```
dirkmc: Do we want to preload when getting indirect keys?

achingbrain: As implemented we leave it up to the caller and it's an undocumented option that defaults to false.

In general I think we want to be really careful with preloading because you can end up inadvertently sending huge amounts of data over the wire, maxing out the number of allowed connections, etc.

PR #2404, opened 2019-08-29
chore: re-enable bundlesize check

alanshaw:


achingbrain:
It's weird that the bundle size check [failed with a 500](https://travis-ci.com/ipfs/js-ipfs/jobs/229230954#L391) in the check step of the build but the stage still passed.

Issue #2396, opened 2019-08-27
⚡️ v0.38.0 RELEASE 🚀


alanshaw:
> TODO short tl;dr; of the release

# 🗺 What's left for release

* [x] feat: [Garbage Collection](https://github.com/ipfs/js-ipfs/pull/2022)
    * N.B. Initial implementation adding commands to core, CLI and HTTP API. No automatic or periodic collection.
* [x] fix: [`ipfs add` output fixes](https://github.com/ipfs/js-ipfs/pull/2386)
* [x] feat: [Gossipsub as default pubsub](https://github.com/ipfs/js-ipfs/pull/2298)
* ~~[ ] feat: [Config profiles](https://github.com/ipfs/js-ipfs/pull/2165)~~ (moved to v0.39.0)
* [x] perf: [Add via async iterators](https://github.com/ipfs/js-ipfs/pull/2379)
* [x] feat: [Support for IPNS and recursive option for `ipfs resolve`](https://github.com/ipfs/js-ipfs/pull/2297)

Future release:

* ~~feat: [Enable DHT in client mode](https://github.com/ipfs/js-ipfs/pull/1994)~~
    * ~~feat: [Reprovider](https://github.com/ipfs/js-ipfs/pull/2184)~~
* ~~feat: [Repo migration tool](https://github.com/ipfs/js-ipfs/pull/2044)~~
* ~~fix: [Locking for concurrent pin ops](https://github.com/ipfs/js-ipfs/pull/2174)~~
* ~~fix: [Offline blockstore for pinning](https://github.com/ipfs/js-ipfs/pull/2196)~~
* ~~perf: [Cache to improve pin performance](https://github.com/ipfs/js-ipfs/pull/2198)~~
* ~~fix: [Limit concurrent HTTP requests in browser](https://github.com/ipfs/js-ipfs/pull/2304)~~

# 🚢 Estimated shipping date

**30th September**

# 🔦 Highlights

<Top highlights for this release>

# 🏗 API Changes

<Any API changes breaking or otherwise that people should know of>

# ✅ Release Checklist

- [ ] **Stage 0 - Automated Testing**
  - [x] Feature freeze. If any "non-trivial" changes (see the footnotes of [doc/releases.md](https://github.com/ipfs/js-ipfs/tree/master/doc/releases.md) for a definition) get added to the release, uncheck all the checkboxes and return to this stage.
  - [x] Automated Testing (already tested in CI) - Ensure that all tests are passing, this includes:
[...]

Issue #1563, opened 2018-09-14
A viable alternative to go-ipfs?


alanshaw:
I would very much like for the js-ipfs implementation to be the most widely used IPFS implementation and I'm interested to know what the main blockers are for that.

_We realise you have a choice of IPFS provider and your choice in important to us._

Or in other words, why do you type `ipfs daemon` and not `jsipfs daemon`?

I've had some interesting insight already:

> @olizilla:
> ﻿﻿﻿﻿it's running on a different api port, so i have to config things to use it
> i've got a repo in ~/.ipfs that has some things in i want to share
> and the command is jsipfs rather than ipfs
> these are all things to make it easier to dev both, but they all make me feel like the go one is the one i should be running day-to-day
> also there is like 10 people on the js dht and 800+ on the go one

> @lidel
>  this may be a niche reason, but there is an official docker image with go version as well
> it may be important for devs, because historically, i used different docker images to quickly find regressions in IPFS APIs used by companion having one for jsipfs would be nice

lidel:
After giving it some additional thought, js-ipfs is missing:
- Full IPNS support (but we are getting there)
- DHT: interop with go and dhtclient mode  (to limit ambient traffic)
- [support for predefined configuration profiles](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#profiles). 
  - `server` one is a must-have IMO, without server profile people will get blocked by hosting providers: https://github.com/ipfs/go-ipfs/issues/1226 / https://github.com/ipfs/go-ipfs/issues/4343
- A hidden gem of go-ipfs  is p2p tunneling via `ipfs p2p`

alanshaw:
Would a post install step to symlink ipfs -> jsipfs if ipfs isn't already taken be useful or just complicate things?

hugomrdias:
What about a ipfsd-ctl like cli ? we would have only one jpfs bin with the option to choose from go or js, we could even have it work like nvm and provide the version to run. Makes sense ?

lidel:
I agree with @hugomrdias , CLI UX needs to be solved before we start playing with default symlink.

Perhaps it could be something like this:
```
USAGE
  ipfs provider ls      - print numbered list of installed providers (eg. go-ipfs, js-ipfs)
  ipfs provider set <x> - switch default provider (accepts  number from ls output or a path)
```

----

I've found another reason why people run `go-ipfs` instead of js one: debugging and metrics.
[...]

PR #2408, opened 2019-08-29
refactor: improvements to the release process

alanshaw:
This is an updated release process for JS IPFS that has been largely based on the new Go IPFS release process.

It adds a release document to explain the process, adds an early testers programme and updates the release issue template.

This isn't a huge change to what we do already but does more formally specify time frames and stages. I think it makes sense to keep these very similar between the two implementations even if there are some context specific tweaks.

~~17th September is 6 weeks since the `0.37` release. So we should have kicked off the process on the 27th August. I've started the [release issue](https://github.com/ipfs/js-ipfs/issues/2396) for `0.38` (and I'd also like your input into what to include in it) but I vote we kick off this new process as soon as we reach agreement on this PR and it is merged.~~ [Monday 9th September is day 0](https://github.com/ipfs/js-ipfs/pull/2408#issuecomment-527862830)


doc/releases.md
```diff
@@ -0,0 +1,120 @@
...
+
+1. Automated and internal testing (stage 0 and 1) will be compressed into a few hours - ideally less than a day.
+2. Stage 2 will be skipped.
+3. Community production testing will be shortened to 1-2 days of opt-in testing in production (early testers can choose to pass).
+
+Some patch releases, especially ones fixing one or more complex bugs, may undergo the full release process.
+
+## Performing a Release
+
+The release is managed by the "Lead Maintainer" for js-ipfs. It starts with the opening of an issue containing the content available on the [RELEASE_ISSUE_TEMPLATE](./RELEASE_ISSUE_TEMPLATE.md) not more than **48 hours** after the previous release.
```
alanshaw: Added the 48 hours which the go-ipfs release process doc currently doesn't have. This communicates expectations for when the new release is due and will help with planning of included features. The community could even suggest inclusions.

doc/releases.md
```diff
@@ -0,0 +1,120 @@
...
+
+This issue is pinned and labeled ["release"](https://github.com/ipfs/js-ipfs/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Arelease). When the cycle is due to begin the 5 stages will be followed until the release is done.
+
+## Release Version Numbers
+
+js-ipfs is currently pre-1.0. In semver terms this means [anything may change at any time](https://semver.org/#spec-item-4).
+
+However, js-ipfs follows semver "shifted by one". This means js-ipfs reserves MINOR version increments for BREAKING CHANGES and/or major feature additions and PATCH version increments for bug fixes and _backwards compatible_ minor feature additions.
```
alanshaw: This is what we do now. Is this working for everyone? Do we think that with the more strict release schedule we could actually use MINOR for **all** features (and breaks) and PATCH exclusively for bug fixes?

doc/releases.md
```diff
@@ -0,0 +1,120 @@
...
+
+Post `1.x.x` (future), MAJOR version number increments will contain BREAKING CHANGES, MINOR version increments will be reserved for backwards compatible new features and PATCH version increments for bug fixes.
+
+We do not yet retroactively apply fixes to older releases (no Long Term Support releases for now), which means that we always recommend users to update to the latest, whenever possible.
+
+### Pre-Releases and Release Candidates
+
+Prior to or during the early stages of a release cycle, js-ipfs may release a pre-release version for users who want to try out the "bleeding edge". This typically happens when a new major feature or breaking change lands in master. When this happens is entirely at the discretion of the Lead Maintainer.
```
alanshaw: Does this make sense? I've [been doing this already](https://www.npmjs.com/package/ipfs?activeTab=versions) for non-feature complete pre-releases but I realised I've not openly communicated this.

doc/releases.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Stage 0 - Automated Testing
+
+At this stage, we expect _all_ automated tests (unit, functional, integration, interop, testlab, performance, etc.) to pass.
+
+### Stage 1 - Internal Testing
+
+At this stage, we'll:
+
+1. Start a partial-rollout to our own infrastructure.
```
alanshaw: Again, this is a noop for js-ipfs right now but with the addition of `/refs` and now GC we're one step closer to being able to be our own preload node.

doc/RELEASE_ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,126 @@
...
+
+> short tl;dr; of the release
+
+# 🗺 What's left for release
+
+<List of items with PRs and/or Issues to be considered for this release>
+
+# 🚢 Estimated shipping date
+
+<Date this release will ship on if everything goes to plan>
```
alanshaw: I added this so that people can get a vague idea of when this release might be available... 👍 👎 ?

doc/releases.md
jacobheun: I think this is reasonable with a 6 week release cycle.

doc/RELEASE_ISSUE_TEMPLATE.md
jacobheun: I think specifying the estimated week would be sufficient and helpful, since this is designed to go in 6 week cycles. I don't think any more granular than that is necessary unless the release cycle ends up targeting specific days of the week.

doc/EARLY_TESTERS.md
```diff
@@ -0,0 +1,29 @@
...
+
+* When js-ipfs enters the second release stage, early testers will be asked to test js-ipfs on non-production infrastructure. This may involve things like:
+  - Running integration tests against the release candidate.
+  - Running simulations/benchmarks on the release candidate.
+  - Manually testing the release candidate to check for regressions.
+* When js-ipfs enters the third release stage (soft release), early testers will be asked to partially deploy the release candidate to production infrastructure. Release candidates at this stage are expected to be identical to the final release. However, this stage allows the js-ipfs team to fix any last-minute regressions without cutting an entirely new release.
+
+## Who has signed up?
+
+No early testers volunteers yet!
```
momack2:
```suggestion
No early tester volunteers yet!
```

doc/releases.md
momack2: @Stebalien not sure if you want to adopt this idea too (might help with writing release notes as we go?)

doc/RELEASE_ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,122 @@
...
+    # Minor prerelease (e.g. 0.33.1 -> 0.34.0-rc.0)
+    npx aegir release --type preminor --preid rc --dist-tag next
+
+    # Increment prerelease (e.g. 0.34.0-rc.0 -> 0.34.0-rc.1)
+    npx aegir release --type prerelease --preid rc --dist-tag next
+    ```
+  - Network Testing:
+    - test lab things - TBD
+  - Infrastructure Testing:
+    - TBD
```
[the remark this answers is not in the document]
daviddias: Not itself, but it does rely on *-star signalling servers, preload nodes, relay nodes and bootstrappers

doc/releases.md
alanshaw: > I think this is reasonable with a 6 week release cycle.

Just to clarify, you are voting for a change from what we do now to this?:

> MINOR for **all** features (and breaks) and PATCH exclusively for bug fixes?

jacobheun: > Just to clarify, you are voting for a change from what we do now to this?

Yes, I'm in favor of keeping the versioning simple and predictable. It also allows us to minimize the time investment on releases, and ensure that all new features have gone through the full release testing cycle, which should help reduce the need for patch releases.

doc/releases.md
```diff
@@ -0,0 +1,120 @@
...
+
+**Goals:**
+
+1. Test the release in some production environments with heavy workloads.
+2. Partially roll-out an upgrade to see how it affects the network.
+3. Retain the ability to ship last-minute fixes before the final release.
+
+### Stage 4 - Release
+
+At this stage, the release is "battle hardened" and ready for wide deployment. A new version is published to npm, announcements are made and a blog post is published to [ipfs.io/blog](https://ipfs.io/blog).
```
lidel: afaik subdomain is the canonical URL (path URL does not work with DNSLink):
```suggestion
At this stage, the release is "battle hardened" and ready for wide deployment. A new version is published to npm, announcements are made and a blog post is published to [blog.ipfs.io](https://blog.ipfs.io).
```

doc/RELEASE_ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,122 @@
...
+  - [ ] IPFS Application Testing - Run the tests of the following applications:
+    - [ ] [webui](https://github.com/ipfs-shipyard/ipfs-webui)
+    - [ ] [ipfs-desktop](https://github.com/ipfs-shipyard/ipfs-desktop)
+    - [ ] [ipfs-companion](https://github.com/ipfs-shipyard/ipfs-companion)
```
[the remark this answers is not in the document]
alanshaw: Yes I think we should but I think it needs a big update before we can include it here....

doc/releases.md
mikeal: This still isn’t going to conform to any of the tooling people have written around semver. Regardless of what we *say*, developers and the automation written by those developers will tend to assume that there aren’t intentional breaking changes between these releases — or they will assume that any version will break and may not adopt it because there isn’t any kind of guarantee.

I don’t see why we can’t just bump the major version number. Version numbers are essentially free, we have 2,147,483,647 of them available.

We like to talk about “1.0” as this big milestone, but we could just call that milestone something else and untangle it from version numbers. Version numbers, especially in JS, have a lot of expectations around them that we aren not currently conforming to. It’s easier to adjust to the community than to try and document/educate every JS developer who finds `js-ipfs`.

We’ve already done far more work to stabilize `js-ipfs` and ensure there aren’t breaking changes between these releases than the average package on npm, it seems silly that we wouldn’t just adjust the versioning along with it in order to gain more adoption given that we’re putting in all of this effort to avoid breaking changes anyway.

olizilla: I agree with @mikeal, but we got here as there was a desire to make 1.0 mean something about defining an API that we want to commit to. It feels like we did that a long time ago, but still https://github.com/ipfs/js-ipfs/issues/1563#issuecomment-527171589

Otherwise I am fine with patches for bug fixes, minor for everything else. Seems like it 'd be `0.40.0` then `0.40.1` when we spot the terrible regression, then `0.41.0` and `0.41.1` ~6 weeks later which seems nice and predictable

momack2: @mikeal - I don't think this is the right place to have that discussion. Feel free to start a separate issue, but I don't want to block merging this release process description on changing a very embedded IPFS standard - aka that 1.0 means interface reliability / forwards compatibility. Changing that current practice is a separate discussion.

I'm supportive of consistency between the js and go - aka patches should be for bug fixes or very scoped improvements - and minor should be for all new features (regardless of compatibility) such that they have testing.

alanshaw: I agree with @momack2, we should have that discussion elsewhere as it's not the question on the table right now and I don't want to further expand the scope of this PR.

It sounds as if there's at least agreement that we should switch to MINOR for all features (and breaks) and PATCH exclusively for bug fixes which I'm happy to do right now and would be consistent with go-ipfs so lets do that! 🚢🚢🚢

FWIW I have a strong desire to use the major version number and I dislike that 1.0 is this big milestone for stable, production ready, "finished" to a lot of people. We lose a lot of power of semver by not using the major version number and as @mikeal pointed out many people (and machines) are still going to view anything pre-1.0 as unstable and likely to change and break, whatever convention we use for pre-1.0 versioning.

achingbrain:
This all sounds good.  My feeling is that six weeks is a really long time for a release - we should be able to automate pretty much everything apart from the community testing/feedback bit.

For example, we should publish an RC to npm every time the head of the release branch changes (using GH actions?  I don't know - @olizilla has done a bunch of stuff with automating publishing with secrets).  I'm not above setting up a Jenkins box under my desk to do this if necessary.

For testing dependent projects, greenkeeper already PRs new versions against dependents - it's not massively visible though, maybe we need a dashboard of which PRs are failing/unmerged.  We could also split our examples into separate repos to do the same thing, or add a test suites to for them to js-IPFS itself.

@alanshaw points out his FileCoin tool for grepping contributors out of the git logs of all the various deps that change in a release, great, same thing for IPFS please.  We could then generate template blog posts etc (to be fleshed out by a human, obvs, but a bullet-point list of things to mention and people to thank would speed things along).

I think all of this fits within the framework proposed above, it's just taking as much manual work out as possible - it would be great to bring that 6 week window down over time.

achingbrain:
FWIW we're going to trial this in the next js-IPFS release and may make "improvements" to the process if it doesn't work in practice.  Monday is day 0, drawbridge up, feature freeze, etc.

From a practical perspective this means creating a release branch with a `package-lock.json`, all subsequently merged PRs will go into master so development can continue, but RCs will be published from the release branch and bug fixes will be cherry-picked between master and the release branch as necessary.

doc/releases.md
mikeal: sorry, wasn’t intending to block the merge, that’s why it’s just a comment and a not a full review. the change being discussed is a positive move in the right direction, i don’t see any need to block it.

alanshaw: No worries - it is a necessary point to make, I'm glad someone bought it up 🙂

Lets open an issue and pick up where we left off.

PR #2410, opened 2019-08-30
feat: encoding key's name to base32

### LINKED ISSUE #1937: Encoding key names with base32
#### Description:

As part of my PR https://github.com/ipfs/go-ipfs/pull/6012 go-ipfs will be encoding key's names into base32 in order to unify behavior across all platforms (case sensitivity). As a small bonus because of it, it will be able to use any characters for the key's names (like namespacing for example "websites/some-website").

I assume that since js-ipfs has a goal to have interoperable repositories with go-ipfs, this behavior should be propagated from go-ipfs to js-ipfs as well.

I am happy to provide PR for js-ipfs as well. 

I have a few questions.
Is this desired from your side? Will you accept the PR? 
My idea of implementation of it would be to create something like `EncodedDatastore` wrapper that would be used in `js-ipfs-repo` for the `keys` property and would do the translation from/to base32. 
I believe the base32 encoding won't be necessary for the `datastore-level`, should I hardwire exception for it or also encode it?

AuHau:
This PR tackles #1937. Yet I am not so sure if it is actually needed. 

Since how Keystore works (eq. has stored metadata in `/info`), it could be easier to just lowercase the filenames instead of encoding them into base32. Yet, if there is a desire to have the same IPFS Repo like with `go` then this will be needed. Although I guess it should be actually `go` who should adopt the same pattern for Keystore as is in `js` (eq. encrypt keys with a passphrase and have metadata aside). Not sure if it is in plan in the near future, though.

Blocking PRs:
 - [ ] integration of migration tool https://github.com/ipfs/js-ipfs-repo/pull/202
 - [ ] migration for encoding existing keys https://github.com/ipfs/js-ipfs-repo-migrations/pull/2

### DIFF
--- a/src/core/components/pre-start.js
+++ b/src/core/components/pre-start.js
@@ -8,6 +8,8 @@ const mergeOptions = require('merge-options')
 const NoKeychain = require('./no-keychain')
 const callbackify = require('callbackify')
 const promisify = require('promisify-es6')
+const KeyTransformDatastore = require('datastore-core').KeytransformDatastore
+const keychainTransformer = require('../../utils/keychain-encoder')
 
 /*
  * Load stuff from Repo into memory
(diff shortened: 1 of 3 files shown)


daviddias:

tinytb:
Is this actually in progress?

achingbrain:

I don't think this is required any more since Kubo adopted Badger and broke repo compatibility, js-IPFS is free to do what it wants.

PR #2428, opened 2019-09-05
feat: allow daemon to init and start in a single cmd

hugomrdias:
This PR aligns `ipfs init` and `ipfs daemon` with go-ipfs.

https://github.com/ipfs/go-ipfs/pull/6489

`ipfs init` changed to accept a file path as an argument

`ipfs daemon` changed to support `--init` and `--init-config` options.

Now we can do `ipfs daemon --init --init-config /path/to/custom-config`

refs:
https://github.com/ipfs/js-ipfsd-ctl/pull/303

Missed the breaking change in the commit msg...

BREAKING CHANGE: `ipfs init` first positional argument is a path to a file instead of a JSON string. Although the previous version was completely broken.


achingbrain:
It would be nicer if it just `init`'d the repo if it needed to without having to pass flags or anything else.

The new user experience is:

```console
$ ipfs
/// blah blah blah blah long thing don't make me read
$ ipfs daemon
// NO! ERROR!  Run this other command
$ ipfs init
// THANK YOU HUMAN
$ ipfs daemon
...
```

achingbrain:
What I'm saying is, I think we should just `init` when required instead of adding more flags like this.

Running `ipfs init` would then be a no-op, unless maybe we want to let the user re-init, in which case something like `ipfs init --force` might be appropriate?

alanshaw:
This is really cool and I'm excited to see the speed increase in our CLI tests.

However, I'm struggling to think of a reason why we wouldn't want `ipfs daemon` to auto-init if not initialised already. Is it a barrier for new users for no reason? @Stebalien @djdv what do you think?

So, how about: remove `--init` (just auto initialise) but retain `--init-config` so that we can still pass config?

Stebalien:
> However, I'm struggling to think of a reason why we wouldn't want ipfs daemon to auto-init if not initialised already.

I agree it should be initialized by default. However, the _one_ reason not to do this is that the current behavior prevents the daemon from being run against the _wrong_ repo. In my case, I always set an explicit repo with `IPFS_PATH`. My user's repo contains an `api` file (for my system-wide daemon) and that's it.

But yeah, there's really no reason to not auto initialize.

hugomrdias:
Should we get this in as is or change it ?

achingbrain:
I think we should change this to initialise by default instead of adding extra flags.

hugomrdias:
done can we get this merged ?

src/cli/commands/init.js
```diff
@@ -1,15 +1,17 @@
...
 module.exports = {
-  command: 'init [config] [options]',
+  command: 'init [default-config] [options]',
   describe: 'Initialize a local IPFS node',
   builder (yargs) {
     return yargs
       .epilog(ipfsPathHelp)
-      .positional('config', {
-        describe: 'Node config, this should JSON and will be merged with the default config. Check https://github.com/ipfs/js-ipfs#optionsconfig',
+      .positional('default-config', {
```
daviddias: Whats the different between `init-config` and `default-config` other than one appears on `ipfs init` and the other on `ipfs daemon`?

test/cli/daemon.js
```diff
@@ -9,10 +9,35 @@ const os = require('os')
 const path = require('path')
 const hat = require('hat')
 const fs = require('fs')
+const tempWrite = require('temp-write')
 const pkg = require('../../package.json')
 
 const skipOnWindows = isWindows() ? it.skip : it
+const daemonReady = (daemon, fn) => {
```
daviddias: what is this fn? A callback? mind calling it a callback?

hugomrdias: its a function to be called after ready not a callback

src/cli/commands/init.js
hugomrdias: its the same thing, this just mirror go-ipfs

daviddias: I see.

Won't bikeshed then, although I'm curious if there was a reason in go-ipfs, @Stebalien ?

test/cli/daemon.js
daviddias: and checking if it is ready, is an async operation because it has to do IO, hence it is a callback to return to the original routine once it is ready

src/cli/commands/init.js
Stebalien: Ah. The issue is that go-ipfs already _has_ a `--config` option (for specifying an out of repo config). It doesn't _work_ but it's already defined...

Stebalien: Wait, actually, I'm confused. This is just a positional param so it doesn't really matter.

test/cli/daemon.js
```diff
@@ -265,4 +282,34 @@ describe('daemon', () => {
       expect(err.stdout).to.include(`Node.js version: ${process.versions.node}`)
     }
   })
+
+  it('should init by default if not already', async function () {
```
achingbrain: How does this test assert that the daemon auto-initialised it's own repo?

hugomrdias: ipfs daemon cmd throws if there is no repo initialised. So if this test passes we should be good.

achingbrain: It should also assert that the repo was created after the command has run, otherwise it doesn't actually test anything other than that the daemon process was killed.  Check the repo directory exists or something like that.

hugomrdias: My reasoning is to not over test stuff, here we are testing the `daemon` command and logic, `init` was added to that logic but we should only test that it gets ran not that `init` works correctly we have other tests for that.
This test will fail if init isn't ran or if init fails internally, which is the scope i wanted when i wrote the test. 
Makes sense ?

hugomrdias: I guess its a bit hidden because init is on by default and if init doesn't work daemon also fails completely but still we have other tests that make sure init creates a repo.

achingbrain: If this test doesn't test anything not covered by other tests it should be removed.

If it does test something not covered by other tests, it should contain assertions to prove the functionality being tested works as expected.

Too many of our tests don't actually test anything, they just ensure that nothing on the happy path explodes so aren't all that useful.

daviddias: @achingbrain is right. Tests should be both valuable and informative. If it requires one to know that the test is testing for the program not to blow up in some weird situation, at least that needs to be well documented and not just live in one persons brain.

hugomrdias: the assertion is this `if (stdout.includes('Daemon is ready')) {` i agree this isn't the best way to do things and i will improve this in a near future, but this is the way we assert a daemon ran successfully in the others tests. 

this is the first test to work without calling `ipfs(init)`.

either i remove the test or change the name to something more informative like 'should start the daemon and init automatically without calling init first' or something else but i don't think testing for the repo is useful here. 
im open to suggestions

src/cli/commands/init.js
daviddias: Got it. The name does not matter but it would have been great if the name was the same. (no need to block )

test/cli/daemon.js
daviddias: Suggestions were:
- Document it with code comments
- Assert that the repo was created as part of the test

PR #2431, opened 2019-09-05
fix(tests): remove Math.random from  tests

lidel:
This PR:

- replaces Math.random() with [hat()](https://www.npmjs.com/package/hat)
- increases timeout of bitswap tests that fail often

### Motivation

CI often fails in tests that use Math.random.
I suspect Math.random is not random enough on some platforms.
Let's see if this PR pass without a problem


lidel:
I've  re-run rests about three times and it seems bitswap tests don't fail (or don't fail that often).

@alanshaw  as this is a cosmetic change, should be safe to merge.

PR #2440, opened 2019-09-10
chore: run interop tests as part of the build

achingbrain:
This will need updating to only run on master & add `npx aegir update-last-successful-build` from https://github.com/ipfs/aegir/pull/419 as a post-build step

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -10,7 +10,7 @@ const preloadNode = MockPreloadNode.createNode()
 const echoServer = EchoServer.createServer()
 
 module.exports = {
-  bundlesize: { maxSize: '692kB' },
+  bundlesize: { maxSize: '696kB' },
   webpack: {
     resolve: {
       mainFields: ['browser', 'main'],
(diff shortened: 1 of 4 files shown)


hugomrdias:
can you use https://github.com/hugomrdias/connect-deps instead of npm link please ?

achingbrain:
> can you use https://github.com/hugomrdias/connect-deps instead of npm link please ?

Sure.  I copied the commands from https://github.com/ipfs/interop - can you PR into the README there to update them?

achingbrain:
I guess something I don't like about this is that a PR could pass CI and be merged, then break master due to an interop test failing.  Maybe we should run them on every branch?

achingbrain:
We should probably include them as a devDep at that point, otherwise specifying which version of the interop tests should run becomes hard (e.g. breaking changes, new tests, etc).

achingbrain:
I'm going to merge this because the new interop bits of the build are passing, the failing bits are the electron-render related stuff @hugomrdias is fixing in ipfs/js-ipfs-http-client#1105

PR #2442, opened 2019-09-10
chore: update packages-list

daviddias:
The list that keeps on growing (look at the package-list.json to avoid rolling your eyes)


achingbrain:
That's a lot of packages.

PR #2495, opened 2019-09-26
chore: convert tests to async/await

achingbrain:
Builds on the work in #2390

 - [x] cli
 - [ ] ~core~ @hugomrdias is doing these
 - [ ] ~gateway~ These were already async await
 - [x] http - most of these were deleted as they are duplicates of tests in `interface-js-ipfs-core` - if not duplicates they were moved to the interface tests


achingbrain:
I think most of the core tests should be moved to either `ipfs/interface-js-ipfs-core` (for the ones that assert things about function inputs and outputs) or `ipfs/interop` (for ones that assert things about network operations).

That'd leave us with just cli, gateway and http tests in this repo, and the http tests should all be converted to `inject` style tests as they are much faster to run.

What do you think @hugomrdias @alanshaw @jacobheun @vasco-santos

hugomrdias:
> I think most of the core tests should be moved to either `ipfs/interface-js-ipfs-core` (for the ones that assert things about function inputs and outputs) or `ipfs/interop` (for ones that assert things about network operations).
> 
> That'd leave us with just cli, gateway and http tests in this repo, and the http tests should all be converted to `inject` style tests as they are much faster to run.
> 
> What do you think @hugomrdias @alanshaw @jacobheun @vasco-santos

yes im working on that and moving interface-core to async also

achingbrain:
> yes im working on that and moving interface-core to async also

Well I'll just sit right back then...

...and convert the gateway & http tests to async/await.  Will leave the core ones alone.

achingbrain:
@hugomrdias please de-dupe with extreme prejudice!

jacobheun:
Ideally the cli tests should only need to verify that the interface is called correctly and that interface outputs are printed properly. We shouldn't need to run ipfs at all for the cli tests and it would make them take seconds instead of minutes.

hugomrdias:
> Ideally the cli tests should only need to verify that the interface is called correctly and that interface outputs are printed properly. We shouldn't need to run ipfs at all for the cli tests and it would make them take seconds instead of minutes.

yes i already changed a couple of sub systems to work like that, and doing the rest is on my todo list

achingbrain:
I've pulled out all the http tests that start a daemon in the background.  Most were duplicates of tests in the `interface-js-ipfs-core` suite but a few tested unhappy paths not otherwise covered.

I've added them to the interface tests here: https://github.com/ipfs/interface-js-ipfs-core/pull/534

PR #2502, opened 2019-10-01
docs: the keyword is docs, not doc

daviddias:
The usage of `doc` rather than `docs` has been making me twitch every time I see it. It is a nitpick for the sake of consistency everywhere :) 

Update: and now I know why it was selected `doc` over `docs`, because there was a .gitignore for `docs` because of the long gone effort to document everything with documentation.js


achingbrain:
I have no strong opinion on `doc` vs `docs` but I have been putting out links directly to `doc/EARLY_TESTERS.md` to encourage a bit of community testing help so maybe we could sit on this for a few weeks so people don't immediately get 404s?

daviddias:
The longer we wait, the more links to `doc` there will be. My advice is to then have a redirect README on /doc that points to /docs and delete that in one or two quarters

achingbrain:
Not sure a redirect README would help, if people follow a link to https://github.com/ipfs/js-ipfs/blob/master/doc/EARLY_TESTERS.md they'll see the GH 404 page, not the README.

daviddias:
Ah right! Ouch.. well, if it stays `doc` isn't the end of the world. Feel free to close if you think it is not worth it.

alanshaw:
I don't really have strong feels on this but if most people are expecting "doc**s**" then lets use it. Also, if it'll make @daviddias happy then it's definitely worth it. It'll be consistent with the go-ipfs directory name as well as our commit message prefix for documentation changes (`docs:`).

Lets get this moved, shall we leave stub files in `doc/*.md` with a link to the updated file temporarily and remove them in 1 month?

TODO: **We also need to update aegir release scripts to add `--no-docs` since it outputs to the `docs` directory.**

@daviddias do you have a sec to fix up this PR?

achingbrain:
> shall we leave stub files in doc/*.md with a link to the updated file temporarily and remove them in 1 month?

I'd say no - it's been a couple of weeks, no new early testers have surfaced for a little while so let's just rename it (once the change to aegir has been made).

daviddias:
> @daviddias do you have a sec to fix up this PR?

Caaaaaan doooo

Issue #2509, opened 2019-10-04
The one true API, the ipfsx upgrade path


achingbrain:
> Land #1670 🤯 🎊 😅
> https://github.com/ipfs/js-ipfs/issues/2506#issuecomment-538307520

But what does that mean though?

Something that bothers me is that we already support promises & callbacks in the API, so removing callbacks doesn't really give our users anything except a whole bunch of refactoring for those who *do* used the callbackified API.  That can't be the end goal of #1670.

I think we should get on and implement [`ipfsx`](https://github.com/ipfs-shipyard/ipfsx).

We've seen the API grow organically over time and it doesn't present the most streamlined interface.  The key point is that IPFS is all about files, yet loads of the file operations are not top-level operations.  Plus we have confusing caveat-filled weirdness with things like `ipfs.ls` and `ipfs.files.ls`.

The `ipfsx` approach is to finally [unify the files API](https://github.com/ipfs/interface-js-ipfs-core/issues/284) and move them all to the top level front and centre and have the low level stuff grouped by domain (e.g. `block`, `dag`, etc).  It also removes readable streams, pull streams, etc giving us a smaller, more focused API surface area and (hopefully) smaller bundle size.  I think we're all agreed that this is where we're heading.

As the internals follow the path of [`ipfs._addAsyncIterator`](https://github.com/ipfs/js-ipfs/blob/master/src/core/components/files-regular/add-async-iterator.js) we can start to implement ipfsx on top of them, then eventually split out the current API as a separate module to ease upgrading for people.

### Step 1

Allow the user to expose the new API as an experimental flag.  The old API will not be exposed - this may mean that some operations are impossible until they are implemented.  Welcome to the bleeding edge.

```javascript
const IPFS = require('ipfs')

const ipfs = await IPFS.create({
  EXPERIMENTAL: {
    ipfsx: true
  }
  // ... other options
})

// call the new API
await ipfs.mkdir('/somedir')
```

### Step 2

Once the API is implemented, remove the old API entirely, deps and all.  This should give us a leaner bundle and small surface area for those who have kept up to date.

Then create a wrapper for the new API that implements the old by calling new API methods and is callbackified/promisified/promise-nodified.  This gives existing users a smooth upgrade path.

We should be able to verify it using today's `interface-ipfs-core` suite, though we should make tomorrow's version will only test the `ipfsx` interface.

```javascript
const IPFS = require('ipfs')
const legacy = require('ipfs-legacy')

[...]

Issue #2446, opened 2019-09-11
Repos, repos, repos


achingbrain:
Making changes in `js-ipfs` typically involves changing this repo, the `ipfs-http-client`, `interface-js-ipfs-core` and also `ipfds-ctl` (because of it's dependency on `ipfs-http-client`.

We have these as separate modules but they are really modular in name only due to the cross-dependencies indicated above.

This has had a number of consequences:

1. Developers have to coordinate PRs across all of these repos which is painful for experienced contributors and completely bewildering for new contributors
1. It's impossible to have one PR that encapsulates a changeset - this makes it hard to test, hard to review, hard to integrate, hard to release and hard to roll back
1. There is no one build that says "Yes, this change is ok" which results in a lot of manual work and makes automation harder because you have to juggle multiple versions of multiple modules
1. There is a large amount of duplication across these repos, from tests to utility methods (`js-ipfs-utils` is helping with the utility methods but is also compounding the problem).
1. Coordinating shared dependency updates across these modules becomes a Byzantine task (e.g. see how out of date libp2p, ipld, are at the moment etc)

Since this project does not include the libp2p and IPLD codebases it should be a relatively thin layer on top of them, yet development is very complex.

We've started talking about re-integrating some core modules in #2222 but we should go further than that.

## ⚒️ Break the `ipfsd-ctl`/`ipfs-http-client` dependency

Possible solutions:

a. Make `ipfs-http-client` a peer dependency of `ipfsd-ctl`
b. Require an implementation to be passed in the same way as with in-proc `js-ipfs` daemons

## 💍 One PR to rule them all

Combine the repos.  One way to do this might be adding the client/ipfsd to core:

```javascript
const IPFS = require('ipfs')
const IPFSClient = require('ipfs/client')

...
```

✅ Pro: Simple to use
✅ Pro: Works with existing unspecified tooling that assumes one-module-per-repo
❌ Con: If you only consume the client you'll end up installing lots of dependencies you don't need
❌ Con: Separate PRs to `interface-js-ipfs-core` still required unless you add that in too
❌ Con: Poor separation of concerns
❌ Con: All-or-nothing releases

Another way might be a [lerna-style mono-repo](https://github.com/lerna/lerna) a la [babel](https://github.com/babel/babel):

```console
$ ls -l
js-ipfs/
├── package.json
├── lerna.json
└── packages/
    ├── ipfs
    ├── ipfsd-ctl
    ├── ipfs-http-client
[...]

PR #2514, opened 2019-10-05
feat: support block.rm over http api

achingbrain:
Also, validate that a block is not pinned before removing it.

Might, uh, might want to do that.

### DIFF
--- a/src/http/api/resources/block.js
+++ b/src/http/api/resources/block.js
@@ -8,6 +8,7 @@ const Boom = require('@hapi/boom')
 const { cidToString } = require('../../../utils/cid')
 const debug = require('debug')
 const all = require('async-iterator-all')
+const { PassThrough } = require('readable-stream')
 const log = debug('ipfs:http-api:block')
 log.error = debug('ipfs:http-api:block:error')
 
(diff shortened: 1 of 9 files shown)


src/http/api/resources/block.js
```diff
@@ -102,20 +103,66 @@ exports.put = {
...
+      })
+      .catch(err => {
+        request.raw.res.addTrailers({
+          'X-Stream-Error': JSON.stringify({
+            Message: err.message,
+            Code: 0
+          })
+        })
+      })
+      .then(() => {
```
hugomrdias: Finally instead ?

achingbrain: Done

achingbrain:
Needs https://github.com/ipfs/interop/pull/89 for CI to pass.

AuHau:
Hmmm this broke few things 😅  

First of all the change of taking multiple CIDs into `rm()` should be also propagated to SPEC into `interface-js-ipfs-core`, if I am not mistaken. Maybe you are already working on it?

Second of all quiet, some tests were expecting for `ipfs.block.rm()` to raise some exceptions at some point. For example https://github.com/ipfs/js-ipfs/blob/master/test/core/gc.spec.js#L193 . I will fix this one as I am working on GC related stuff, but you should make sure that other parts of tests are not silently broken because they might not be correctly written (not expecting not to throw like in the GC tests). Not sure if it can be the case somewhere else...

Lastly one general question about programming attitude. When you have bulk operations (like bulk remove of blocks), do you fail-fast or do as much as possible? From the perspective of the backward compatibility of this change, it would make sense to fail-fast and don't catch the exceptions, but not sure what is the general approach about this across IPFS JS-land...

@achingbrain

achingbrain:
Looks like that test should be refactored to enforce that the exception is thrown, e.g:

```javascript
// Second rm should fail because GC has already removed that block
await rm2.then(
  () => expect.fail('should have thrown when trying to remove already removed block'),
  (err) => expect(err.code).eql(Errors.dbDeleteFailedError().code)
)
```

I think in general we try to fail fast, but there are a few exceptions.  `block.rm` is one of them as `go-IPFS` doesn't fail fast in this case and we need to maintain compatibility.

AuHau:
I see, alright make sense.

Sure, I will do that. But at the same time, looking on diff of this PR, it is bit suspicious that you did not change any tests which would depend on `block.rm()` either throw exceptions or fail at some point, which I would think should be also somewhere else then just at the `gc` part.

achingbrain:
All the tests in this repo passed, plus most of the tests are in the [`interface-ipfs-core`](https://github.com/ipfs/interface-js-ipfs-core/blob/master/src/block/rm.js) repo.

In general we want to add tests there and not here where possible so they get run against `ipfs-http-client` (and as such, `go-IPFS`) as well.

Side note: I find this sort of development flow hard to work with, maybe you do too, in which case chime in on #2446

AuHau:
Hmmm well, then I found an inconsistency. 

https://github.com/ipfs/js-ipfs/blob/master/test/core/block.spec.js#L56 expects to throw on invalid CID. Which is happening and correctly passing. But with your introduced change it should not throw but return an array: 
```
[
 { error: "Invalid CID" }
] 
```

Right? So currently validation of CID fails fast, but deletion errors do not...

AuHau:
Also maybe https://github.com/ipfs/js-ipfs/blob/master/test/core/pin.js#L208 could be un-skipped?

AuHau:
Also, I assume that there is a plan to go all "async" and hence support generators at some point? Maybe it would be worth to track the points where currently generators are degraded to arrays because of this, so the final refactor switch could be easier?

achingbrain:
> So currently validation of CID fails fast, but deletion errors do not...

This is intentional.  Sort of.  It's what `go-IPFS` does and we are bound by that contract.

> Also maybe https://github.com/ipfs/js-ipfs/blob/master/test/core/pin.js#L208 could be un-skipped?

Yes, though better to rewrite it so it doesn't affect other tests in the suite so they can all be run in isolation as well as together.

> I assume that there is a plan to go all "async" and hence support generators at some point?

Yes!  It's happening in two stages, the first is #1670 to convert everything internally, then #2509 to expose it to the wider world in a more refined API that reflects what we've learnt over the past few years.

> Maybe it would be worth to track the points where currently generators are degraded to arrays

This should really only be happening at the very last moment - e.g. in the functions in `src/core/components`

PR #2517, opened 2019-10-07
chore: converts remaining file api methods to async iterators

achingbrain:
Depends on https://github.com/ipfs/js-ipfs-http-client/pull/1124


src/core/components/files-regular/add.js
```diff
@@ -5,7 +5,7 @@ const all = require('async-iterator-all')
 module.exports = function (self) {
   // can't use callbackify because if `data` is a pull stream
   // it thinks we are passing a callback. This is why we can't have nice things.
-  return (data, options, callback) => {
+  return function add (data, options, callback) {
```
daviddias: Why move from arrow func to named func here?

test/core/files.spec.js
```diff
@@ -49,6 +49,7 @@ describe('files', function () {
       const invalidPath = null
       const stream = ipfs.getReadableStream(invalidPath)
 
+      stream.on('data', () => {})
```
daviddias: Interesting. Something changed for now it requiring to start pumping for the error to be emitted?

src/core/components/files-regular/add.js
achingbrain: To have function names in any stack traces emitted making them a teensy bit easier to read.

test/core/files.spec.js
achingbrain: In the old implementation, converting the pull stream to a readable stream calls [`s.sink(read)`](https://github.com/pull-stream/pull-stream-to-stream/blob/master/index.js#L85) in `pull-stream-to-stream` which calls [`drain()`](https://github.com/pull-stream/pull-stream-to-stream/blob/master/index.js#L97) on the next tick which causes data to start flowing, potentially before the consuming code has set up pipes or data event listeners.

I've opened https://github.com/pull-stream/pull-stream-to-stream/pull/7 as this looks like a bug, at least, it's not consistent with the node docs.

The new implementation doesn't do this conversion so doesn't have the same bug.

daviddias: good catch! 👍

src/core/components/files-regular/ls-async-iterator.js
```diff
@@ -0,0 +1,54 @@
...
+        child = mapFile(child, options)
+        child.depth--
+
+        yield child
+      }
+
+      return
+    }
+
+    throw errCode(new Error(`Unknown UnixFS type ${file.unixfs.type}`), 'EUNKNOWNUNIXFSTYPE')
```
alanshaw:
```suggestion
    throw errCode(new Error(`Unknown UnixFS type ${file.unixfs.type}`), 'ERR_UNKNOWN_UNIXFS_TYPE')
```

src/core/components/files-regular/ls-async-iterator.js
```diff
@@ -0,0 +1,54 @@
...
+    const pathComponents = path.split('/')
+
+    if (options.preload !== false) {
+      self._preload(pathComponents[0])
+    }
+
+    const file = await exporter(ipfsPath, self._ipld, options)
+
+    if (!file.unixfs) {
+      throw errCode(new Error('dag node was not a UnixFS node'), 'ENOTUNIXFS')
```
alanshaw:
```suggestion
      throw errCode(new Error('dag node was not a UnixFS node'), 'ERR_NOT_UNIXFS')
```

src/core/components/files-regular/refs-async-iterator.js
```diff
@@ -0,0 +1,157 @@
...
+    } catch (err) {
+      if (err.code === ERR_NOT_FOUND) {
+        err.message = `Could not find object with CID: ${parent.cid}`
+      }
+
+      throw err
+    }
+  }
+
+  yield * await traverseLevel({ cid: rootCid }, 0)
```
alanshaw: If I'm reading this right you don't need the await here, `traverseLevel` returns a generator which you can `yield *` from.

```suggestion
  yield * traverseLevel({ cid: rootCid }, 0)
```

src/core/components/files-regular/refs-async-iterator.js
```diff
@@ -0,0 +1,157 @@
...
+          parent: parent,
+          node: link,
+          isDuplicate: uniqueOnly && seen.has(link.cid.toString())
+        }
+
+        if (uniqueOnly) {
+          seen.add(link.cid.toString())
+        }
+
+        yield * await traverseLevel(link, nextLevelDepth)
```
alanshaw: If I'm reading this right you don't need the await here, `traverseLevel` returns a generator which you can `yield *` from.

```suggestion
        yield * traverseLevel(link, nextLevelDepth)
```

src/http/api/resources/files-regular.js
```diff
@@ -330,75 +328,71 @@ exports.refs = {
...
+    .then(() => fn(output))
+    .catch(err => {
+      request.raw.res.addTrailers({
+        [errorTrailer]: JSON.stringify({
+          Message: err.message,
+          Code: 0
+        })
+      })
+    })
+    .finally(() => {
```
alanshaw: Wow, I did not realise promises had a finally...

alanshaw:
ping @achingbrain 🙏

src/http/api/resources/files-regular.js
achingbrain: Putting the Java into Javascript 😆

PR #2520, opened 2019-10-08
feat: web ui latest

hacdias:
Relevant release notes can be found on [v2.5.5](https://github.com/ipfs-shipyard/ipfs-webui/releases/tag/v2.5.5).


achingbrain:
Do you think this could be an IPNS name instead of a CID?  That way you wouldn't have to keep submitting these PRs that just change the hash.

hacdias:
The latest released/stable is always on `/ipns/webui.ipfs.io` and the latest commit to master on `/ipns/dev.webui.ipfs.io` so I guess we can change it to the former.

achingbrain:
Feature freeze was on Monday.  This missed the deadline and will go out in the next release.

achingbrain:
Hmm, actually having just seen https://github.com/ipfs/js-ipfs/pull/2281 and the discussion on https://github.com/ipfs/go-ipfs/pull/6530 maybe we should switch back to the `CID` before release.

PR #2521, opened 2019-10-08
feat: add mssing `dag put` and `dag resolve` cli commands

achingbrain:
Also allows passing input to `ipfs-exec` to simulate piping in cli tests.

### DIFF
--- a/src/cli/commands/dag/put.js
+++ b/src/cli/commands/dag/put.js
@@ -1,0 +1,127 @@
+'use strict'
+
+const mh = require('multihashes')
+const multibase = require('multibase')
+const dagCBOR = require('ipld-dag-cbor')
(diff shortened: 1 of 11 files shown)


src/cli/commands/dag/put.js
```diff
@@ -0,0 +1,127 @@
...
+
+      const cid = await ipfs.dag.put(source, {
+        format,
+        hashAlg,
+        version: cidVersion,
+        onlyHash,
+        preload
+      })
+
+      if (pin) {
```
alanshaw: We should push the pinning functionality into core so that it can take a gc lock to avoid being collected between adding and pinning.

alanshaw: Or back out of pinning in this PR...

PR #2528, opened 2019-10-11
Test examples

achingbrain:
Adds automated tests for all examples so we can run them as part of CI instead of doing everything manually.

They ran as part of [this build](https://travis-ci.com/ipfs/js-ipfs/builds/131472658) and passed (though other jobs may have failed).


alanshaw:
👏👏👏 this is awesome!

PR #2531, opened 2019-10-13
chore: remove try-catch from tests where functions are async

achingbrain:
We had a few instances in the tests where assertions may be missed due to functions not throwing where we thought they would so this PR refactors that code to use `.then(onResult, onReject)` instead.

Follows on from https://github.com/ipfs/js-ipfs/pull/2514#issuecomment-539945452


src/http/api/resources/files-regular.js
```diff
@@ -336,6 +336,10 @@ exports.refs = {
     const unique = request.query.unique
     const maxDepth = request.query['max-depth']
 
+    if (maxDepth <= 0) {
+      return h.response()
+    }
```
achingbrain: This is because `ndjson.serialize` writes a single `\n` into the response even if the stream was empty - in the client we strip the `\n` from response and try to parse what's left as JSON which then explodes, so if we know we're not going to send anything back, shortcut the streaming of refs.

This is noted here and not in the code because it's all going to be removed when #2517 lands.

test/cli/daemon.js
```diff
@@ -104,14 +104,13 @@ describe('daemon', () => {
...
-
-      expect(stdout).to.include('Daemon is ready')
-    }
+    await daemon.then(
+      () => expect.fail('Did not kill process'),
+      (err) => {
+        expect(err.killed).to.be.true()
+        expect(stdout).to.include('Daemon is ready')
+      }
+    )
```
alanshaw: These look equivalent to me - am I missing something?

achingbrain: No, you're not missing anything - it was more an attempt at consistency with other tests that have been refactored to be more rigorous.

achingbrain:
The `await something().then(failOnResolve, expectOnReject)` approach was something @jacobheun recommended [here](https://github.com/ipfs/js-datastore-pubsub/pull/15#discussion_r327660247).

I don't mind either way, but if we're going to go back & forth on this, something should be added to [`CONTRIBUTING_JS.md`](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md) to settle the matter.

jacobheun:
I'm not a huge fan of mixing the two either, but I think it has it's place. Tests where you expect failures are a place I find them useful, because IMO it helps prevent false positive tests. It's very easy to forget to throw or add the `expect.fail` after a try/catch (I've already seen this happen quite a bit).

alanshaw:
Ah ok now I understand - would an assertion like https://github.com/avajs/ava/blob/master/docs/03-assertions.md#throwsasyncthrower-expected-message be the solution then?

achingbrain:
We could do this sort of thing instead?

```javascript
await expect(promise)
  .to.be.rejectedWith(Error, /optionally match message/)
  .that.eventually.has.property('code', 'ERR_A_CODE')
```

PR #2540, opened 2019-10-16
fix: trailing slash in ls crash

AuHau:
When executing command like "jsipfs ls /ipfs/.../" result was a crash.
This patch introduces removing of traling slash

### DIFF
--- a/src/cli/commands/ls.js
+++ b/src/cli/commands/ls.js
@@ -48,11 +48,7 @@ module.exports = {
       const multihashWidth = Math.max.apply(null, links.map((file) => file.hash.length))
       const sizeWidth = Math.max.apply(null, links.map((file) => String(file.size).length))
 
-      let pathParts = key.split('/')
-
-      if (key.startsWith('/ipfs/')) {
-        pathParts = pathParts.slice(2)
-      }
+      let pathParts = key.replace(/\/ipfs\/|\/$/g, '').split('/')
 
       links.forEach(link => {
         const fileName = link.type === 'dir' ? `${link.name || ''}/` : link.name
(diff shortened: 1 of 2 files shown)


test/cli/ls.js
```diff
@@ -40,7 +40,7 @@ describe('ls', () => runOnAndOff((thing) => {
 
   it('adds a header, -v', async function () {
     this.timeout(20 * 1000)
-    const out = await ipfs('ls /ipfs/Qmaj2NmcyAXT8dFmZRRytE12wpcaHADzbChKToMEjBsj5Z -v')
+    const out = await ipfs('ls /ipfs/Qmaj2NmcyAXT8dFmZRRytE12wpcaHADzbChKToMEjBsj5Z/ -v')
```
achingbrain: Could you make this an additional test please?

src/cli/commands/ls.js
```diff
@@ -48,11 +48,7 @@ module.exports = {
       const multihashWidth = Math.max.apply(null, links.map((file) => file.hash.length))
       const sizeWidth = Math.max.apply(null, links.map((file) => String(file.size).length))
 
-      let pathParts = key.split('/')
-
-      if (key.startsWith('/ipfs/')) {
-        pathParts = pathParts.slice(2)
-      }
+      let pathParts = key.replace(/\/ipfs\/|\/$/g, '').split('/')
```
achingbrain: What would happen when we try to list paths that contain `ipfs` inside UnixFS DAGs?  E.g. `jsipfs ls /ipfs/Qmfoo/ipfs/bar`?

AuHau: Good catch!

achingbrain: Regular expressions, now you have two problems, etc, etc 🤣

PR #2569, opened 2019-10-29
docs: offline option

vasco-santos:
We are missing documentation for the `offline` option, as discussed in [ipfs/go-ipfs#3860#issuecomment-547094569](https://github.com/ipfs/go-ipfs/issues/3860#issuecomment-547094569)


achingbrain:
Good news, the bundle size thing has been resolved by https://github.com/ipfs/js-ipfs/pull/2571

PR #2588, opened 2019-11-06
refactor: switch to async peer-id

lidel:
This PR switches to async `"peer-id": "^0.13.4"`

I originally made this for (#2566) before I realized it does not work with other deps that use non-async `peer-id`. 

cc @alanshaw  @vasco-santos @jacobheun to be aware of this

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,10 +302,9 @@ Instead of a boolean, you may provide an object with custom initialization optio
     ```js
     // Generating a Peer ID:
     const PeerId = require('peer-id')
-    PeerId.create({ bits: 2048 }, (err, peerId) => {
-      // Generates a new Peer ID, complete with public/private keypair
-      // See https://github.com/libp2p/js-peer-id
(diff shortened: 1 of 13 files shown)


lidel:
What would be the next step here? `peer-info`? `peer-book`?
(i don't want to step on anyone's toes)

achingbrain:
Can please we hold off on merging this until `libp2p` upgrades it's dependency on peer id?

I also performed this upgrade but backed it out after finding that previously equal peer IDs were no longer equal because one module was using v0.12.x and another was using v0.13.x, hence the promisification of the v.0.12.x peer IDs in the codebase currently.

lidel:
yes, :100:, forgot to make it clear: 
this should be parked until `libp2p` and is deps use async `peer-id`

lidel:
Superseded by https://github.com/ipfs/js-ipfs/pull/2680

PR #2590, opened 2019-11-07
refactor: convert preload to async/await

alanshaw:
...and removes `async` from dependencies!

And half the code:

Screenshot 2019-11-07 at 12 13 14


.aegir.js
```diff
@@ -30,13 +30,13 @@ module.exports = {
     node: {
       pre: (cb) => {
         parallel([
-          (cb) => preloadNode.start(cb),
+          (cb) => preloadNode.start().then(cb, cb),
```
achingbrain: We should avoid mixing callback/promise error handling (eh, @hugomrdias ?) so should callbackify this.

Alternatively these aegir hooks support promises if you [don't declare any parameters to the function](https://github.com/ipfs/aegir/blob/988d40aa190be0cf68bb47446956d2dfb26aca63/src/config/user.js#L30), can the echoServer be refactored to be promised based and then we can `return Promise.all` here?

PR #2591, opened 2019-11-07
feat: pass libp2pOptions to the bundle function

### LINKED ISSUE #2579: Feature request: Allow extending libp2p.transports/libp2p.discovery instead of overriding it via flag
- **Version**: next
- **Platform**: *
- **Subsystem**: libp2p component

#### Description: Feature request: Allow extending libp2p.transports/libp2p.discovery instead of overriding it via flag

It would be good if there would be a flag to add custom transports _in addition to the defaults_, instead of fully overriding them.

This flag could be named `libp2p.extendOnly` or `libp2p.extends`

alanshaw:
This allows bundle creation by merging/overriding option defaults instead of creating them in their entirety from scratch. This is for convenience.

The `libp2pOptions` passed to the bundle function include any options defined in the required runtime files (for Node.js/browser config).

To allow this, there is a slight change to `runtime/libp2p-nodejs.js` and `runtime/libp2p-browser.js` to return their options instead of a libp2p instance. This actually removes some repeated boilerplate.

e.g.

```js
const ipfs = await IPFS.create({
  libp2p: ({ libp2pOptions, peerInfo, ...rest }) => {
    // Merge, extend, override etc. etc.
    libp2pOptions.transports.push(...)
    return new Libp2p(libp2pOptions)
  }
})
```

resolves #2579

### DIFF
--- a/src/core/components/libp2p.js
+++ b/src/core/components/libp2p.js
@@ -13,15 +13,20 @@ module.exports = function libp2p (self, config) {
   const options = self._options || {}
   config = config || {}
 
-  // Always create libp2p via a bundle function
-  const createBundle = typeof options.libp2p === 'function'
-    ? options.libp2p
-    : defaultBundle
-
   const { datastore } = self._repo
   const peerInfo = self._peerInfo
   const peerBook = self._peerInfoBook
-  const libp2p = createBundle({ options, config, datastore, peerInfo, peerBook })
+
+  const libp2pOptions = getLibp2pOptions({ options, config, datastore, peerInfo, peerBook })
+  let libp2p
+
+  if (typeof options.libp2p === 'function') {
+    libp2p = options.libp2p({ libp2pOptions, options, config, datastore, peerInfo, peerBook })
+  } else {
+    // Required inline to reduce startup time
+    const Libp2p = require('libp2p')
+    libp2p = new Libp2p(mergeOptions(libp2pOptions, get(options, 'libp2p', {})))
+  }
 
   libp2p.on('stop', () => {
     // Clear our addresses so we can start clean
@@ -39,7 +44,7 @@ module.exports = function libp2p (self, config) {
   return libp2p
 }
 
-function defaultBundle ({ datastore, peerInfo, peerBook, options, config }) {
+function getLibp2pOptions ({ options, config, datastore, peerInfo, peerBook }) {
   // Set up Delegate Routing based on the presence of Delegates in the config
   let contentRouting
   let peerRouting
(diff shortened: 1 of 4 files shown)


src/core/components/libp2p.js
```diff
@@ -13,15 +13,20 @@ module.exports = function libp2p (self, config) {
...
+  const libp2pOptions = getLibp2pOptions({ options, config, datastore, peerInfo, peerBook })
+  let libp2p
+
+  if (typeof options.libp2p === 'function') {
+    libp2p = options.libp2p({ libp2pOptions, options, config, datastore, peerInfo, peerBook })
+  } else {
+    // Required inline to reduce startup time
+    const Libp2p = require('libp2p')
+    libp2p = new Libp2p(mergeOptions(libp2pOptions, get(options, 'libp2p', {})))
+  }
```
jacobheun: So, options is getting passed into `getLibp2pOptions`, but then it is still passed into libp2p creation. Should `getLibp2pOptions` just handle merging all of that so options doesn't need to be passed again?

> options.libp2p({ libp2pOptions, options, config, datastore, peerInfo, peerBook })

Along those lines, why is `options` being passed here? It feels like the result of `getLibp2pOptions` should just be able to get passed directly to both `new Libp2p` and `options.libp2p`. No other params should be needed right?

alanshaw: > Should getLibp2pOptions just handle merging all of that so options doesn't need to be passed again?

There may be some other IPFS option you want to use in your custom bundle function. It's "passed again" because that's what currently happens, and I'm not trying to avoid a breaking change.

> > options.libp2p({ libp2pOptions, options, config, datastore, peerInfo, peerBook })
>
> Along those lines, why is `options` being passed here? It feels like the result of `getLibp2pOptions` should just be able to get passed directly to both `new Libp2p` and `options.libp2p`. No other params should be needed right?

`options` is IPFS options, passed to the IPFS constructor. It's passed here because that's currently what happens when you provide a custom bundle function. If we removed it, we'd have a breaking change.

`getLibp2pOptions` is passed IPFS `options` and returns _libp2p options_ (I would call them defaults, but they're not really because they're based off the IPFS `options` and the IPFS config from the repo).

jacobheun: Okay, so `options` and `config` are for IPFS? If so, would it be easier to understand if in the scope of the custom bundler `options` were the libp2p options and the IPFS ones were prefixed? If those are going to be exposed I think it might be helpful to make that distinction.

jacobheun: nvm, that would be a breaking change. I think that would be clearer, but since the goal is to avoid the breaking change I think this is fine.

alanshaw: > I think that would be clearer

FWIW I agree!

PR #2625, opened 2019-11-22
chore: change interface tests to async node creation

PedroMiguelSS:
This PR allows `interface-core` tests to create and spawn nodes asynchronously.

Related to:
https://github.com/ipfs/interface-js-ipfs-core/pull/562

### DIFF
--- a/test/utils/interface-common-factory.js
+++ b/test/utils/interface-common-factory.js
@@ -16,7 +16,7 @@ const DEFAULT_FACTORY_OPTIONS = {
 function createFactory (options) {
   options = options || {}
 
-  options.factoryOptions = options.factoryOptions || DEFAULT_FACTORY_OPTIONS
+  options.factoryOptions = options.factoryOptions || { ...DEFAULT_FACTORY_OPTIONS }
   options.spawnOptions = mergeOptions({
     initOptions: { bits: 512 },
     config: {
@@ -78,8 +78,9 @@ function createAsync (options = {}) {
     const nodes = []
     const setup = async (setupOptions = {}) => {
       options.factoryOptions = mergeOptions(
+        options.factoryOptions ? {} : { ...DEFAULT_FACTORY_OPTIONS },
         setupOptions.factoryOptions,
-        options.factoryOptions || DEFAULT_FACTORY_OPTIONS
+        options.factoryOptions
       )
 
       // When not an in proc daemon use the http-client js-ipfs depends on, not the one from ipfsd-ctl
--- a/test/utils/interface-common-factory.js
+++ b/test/utils/interface-common-factory.js
@@ -8,10 +8,15 @@ const callbackify = require('callbackify')
 const mergeOptions = require('merge-options')
 const IPFS = require('../../src')
 
+const DEFAULT_FACTORY_OPTIONS = {
+  type: 'proc',
+  exec: IPFS
+}
+
 function createFactory (options) {
   options = options || {}
 
-  options.factoryOptions = options.factoryOptions || { type: 'proc', exec: IPFS }
+  options.factoryOptions = options.factoryOptions || DEFAULT_FACTORY_OPTIONS
   options.spawnOptions = mergeOptions({
     initOptions: { bits: 512 },
     config: {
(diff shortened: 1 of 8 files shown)


test/utils/interface-common-factory.js
```diff
@@ -8,10 +8,15 @@ const callbackify = require('callbackify')
...
+const DEFAULT_FACTORY_OPTIONS = {
+  type: 'proc',
+  exec: IPFS
+}
+
 function createFactory (options) {
   options = options || {}
 
-  options.factoryOptions = options.factoryOptions || { type: 'proc', exec: IPFS }
+  options.factoryOptions = options.factoryOptions || DEFAULT_FACTORY_OPTIONS
```
alanshaw: I don't know if ipfsd mutates these, it it does you'll get some weird things happening...

```suggestion
  options.factoryOptions = options.factoryOptions || { ...DEFAULT_FACTORY_OPTIONS }
```

test/utils/interface-common-factory.js
```diff
@@ -68,25 +73,23 @@ function createFactory (options) {
...
-          type: 'proc',
-          exec: IPFS
-        },
-        factoryOptions,
-        createFactoryOptions
+    const setup = async (setupOptions = {}) => {
+      options.factoryOptions = mergeOptions(
+        setupOptions.factoryOptions,
+        options.factoryOptions || DEFAULT_FACTORY_OPTIONS
       )
```
alanshaw: The order here has changed?

PedroMiguelSS: No. `factoryOptions` is now `setupOptions.factoryOptions` and `createFactoryOptions` is now `options.factoryOptions`.
So the order is still the same!

alanshaw: Right but the defaults are now last. Should this be:

```js
      options.factoryOptions = mergeOptions(
        { ...DEFAULT_FACTORY_OPTIONS },
        setupOptions.factoryOptions,
        options.factoryOptions
      )
```

test/utils/interface-common-factory.js
PedroMiguelSS: I think we don't want to spread `DEFAULT_FACTORY_OPTIONS`. If you check [`createFactory()`](https://github.com/ipfs/js-ipfs/blob/e9ae0a4d2c790a14c11142692cbce75f06f1df39/test/utils/interface-common-factory.js#L19) you will see that we either have `factoryOptions` if passed or `DEFAULT_FACTORY_OPTIONS`. I did try to keep the daemon config exactly as it was previously.

alanshaw: You've pulled them out into a `const DEFAULT_FACTORY_OPTIONS` object which is possibly used multiple times. It's similar but not exactly the same as it was before.

If we call `createFactory()` with no options then `options.factoryOptions` will be `DEFAULT_FACTORY_OPTIONS` which is `{ type: 'proc', exec: IPFS }`.

If `options.factoryOptions` gets mutated later in this function or by `ipfsd-ctl` then the next call to `createFactory()` will get `{ type: 'proc', exec: IPFS, some: 'mutation' }` which are possibly no longer the intended defaults.

I'm suggesting spreading the defaults so each call to `createFactory()` gets a new set of defaults.

PedroMiguelSS: Yes, you're right. I wasn't answering you considering the real problem here. I got your point, nice catch! Thanks.

test/utils/interface-common-factory.js
PedroMiguelSS: Got it. The thing here is: we only want to use `DEFAULT_FACTORY_OPTIONS` if there's no `options.factoryOptions`.

Should we do something like:
```js
options.factoryOptions = mergeOptions(
     options.factoryOptions ? { ...DEFAULT_FACTORY_OPTIONS } : {},
     setupOptions.factoryOptions,
     options.factoryOptions
)
```

alanshaw: yeah that works 👍

alanshaw:
Can we get ci green @PedroMiguelSS?

PR #2632, opened 2019-11-25
perf: debounce pin flush

achingbrain:
When importing a directory of files, after every block is added we pin it and flush pins to the repo.  This is really slow and if we are adding lots of files, the pinset is out of date almost immediately so this PR debounces flushing pins until after we've stopped adding things.
    
A call to `pinManager.flushPins` will return a promise, and that promise will resolve once the pins have been flushed, but it will first wait 100ms and see if any subsequent flushes have been requested in that time - if so that promise will end early and return the promise created by the later flush request.

### DIFF
--- a/src/core/components/pin.js
+++ b/src/core/components/pin.js
@@ -59,7 +59,7 @@ module.exports = (self) => {
         results.forEach(key => pinset.add(key))
 
         // persist updated pin sets to datastore
-        await pinManager.flushPins()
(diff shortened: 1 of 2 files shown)


alanshaw:
Closed?

achingbrain:
I reopened it as a draft, can’t switch an open PR to be a draft, silly GitHub..

Issue #2197, opened 2019-06-24
js-ipfs pinning performance


dirkmc:
`ipfs.add()` performance degrades severely once the number of pins exceeds `8192`

#### Background

Users can pin a file or a block to prevent it from being garbage collected.

The pinning module maintains two sets of pins:
- direct: the CID of the block that is pinned
- recursive: the CID of the root node of a DAG of blocks

These pin sets are stored in the block store with the following structure:
- if the number of pins is less than `8192`, create a node with
  - `256` links pointing to an empty block
  - a link pointing to each pinned block
- if the number of pins is greater than `8192`
  - distribute pins deterministically amongst `256` buckets
  - each bucket is a node with one or more pins, with the same structure described above (ie buckets with `> 8192` pins distribute them into sub-buckets etc)

#### Performance

A pin set with less than `8192` pins is stored in a single DAG node. Once there are more than `8192` pins, they are distributed between `256` buckets, each with its own DAG node. Each time a new pin is added to the set, the distribution across the group of buckets is calculated and written to the block store. The distribution is deterministic, so in reality only one bucket changes each time a new pin is added.

For example, if we simplify and say there are 8 buckets, with 5 pins (`A` - `E`):
`[] [D] [] [EA] [] [C] [] [B]`
When we add pin `F` only one bucket changes:
`[] [D] [] [EA] [] [C] [] [BF]`

We can improve performance by adding a cache that remembers the structure of the pin sets, and only write nodes that change to the block store (instead of writing all nodes each time a pin is added or removed). This improves `ipfs.add()` performance dramatically once we exceed `8192` pins:

<img width="607" alt="pinning-perf" src="https://user-images.githubusercontent.com/169124/60051646-75b1d300-96a1-11e9-8ef2-29cbd0a765bd.png">

#### Memory usage

The pinner uses [fnv1a](https://github.com/sindresorhus/fnv1a) to distribute pins. fnv1a outputs a number (8 bytes) so if we also use it for cache keys each key will be 8 bytes. Each pin is represented by a DAG link pointing to the pinned CID. The DAG link has
- a name (the empty string)
- a size (number)
- a cid
  - version (number)
  - codec (eg 'sha2-256')
  - multihash (the hash itself)
  - multibaseName (eg 'base58btc')

So rounding up, a DAG link requires about 128 bytes of memory. eg 10k pins requires about 1MB memory for the cache.
[...]

PR #2634, opened 2019-11-25
perf: debounce pin flush

achingbrain:
When importing a directory of files, after every block is added we pin it and flush pins to the repo.  This is really slow and if we are adding lots of files, the pinset is out of date almost immediately so this PR debounces flushing pins until after we've stopped adding things.

A call to `pinManager.flushPins` will create a promise that first waits 100ms to see if any subsequent flushes have been requested in that time - if so that promise will end early and the later flush request will flush the pins.


src/core/components/pin/pin-manager.js
```diff
@@ -106,43 +107,65 @@ class PinManager {
...
-      format: multicodec.DAG_PB,
-      hashAlg: multicodec.SHA2_256,
-      preload: false
-    })
+    const controller = new AbortController()
+
+    this._flushingPins = {
+      controller,
+      promise: delay(100)
+        .then(async () => {
```
alanshaw: How come this call has a `.then().catch()`? We should stick to `await` as much as possible.

```js
promise: (async () => {
  await delay(100)
  if (controller.signal.aborted) return
  // ...
})()
```

src/core/components/pin/pin-manager.js
```diff
@@ -106,43 +107,65 @@ class PinManager {
...
+            hashAlg: multicodec.SHA2_256,
+            preload: false
+          })
+
+          // save root to datastore under a consistent key
+          await this.repo.datastore.put(PIN_DS_KEY, rootCid.buffer)
+
+          this.log(`Flushed pins with root: ${rootCid}`)
+        })
+        .catch(err => this.log(`Could not flush pins: ${err}`))
```
alanshaw: Please let us not smother the error!

Secondly, if the flush does fail, any debounced calls succeed...and they succeed before the flush has been performed so there's no way to know if an error happens.

Could we use something like [`debounce-promise`](https://www.npmjs.com/package/debounce-promise)? It would mean the debounced calls would have to wait for the debounce run so perhaps this wouldn't exhibit as much perf improvement but would still be good.

src/core/components/pin/pin-manager.js
achingbrain: Personal preference I guess?  I've always found IIFEs a bit, well iffy to look at.

src/core/components/pin/pin-manager.js
achingbrain: I don't think so.  If I am importing files A, B and C, and we block on flushing pins after importing A, then B will not start to import until A has finished and the pins have flushed, debounced or otherwise so there's no benefit to be gained.

This PR works by doing the flushing in another async context but as you point out, there's no way to know if an error occurred, because the only way to know is to block until the operation has completed, which means the next file in the list will not get imported until the pins have flushed.

achingbrain:
Thinking about this more, I'm not sure it's a great idea - the errors aren't returned to the calling context and I don't know what will happen if the repo is closed while the debounce timeout is running.

I'm going to have a go at storing pins in the datastore (see https://github.com/ipfs/js-ipfs/issues/2197), that way we won't have to flush them or do any of these complicated DAG manipulations.

achingbrain:
Going to close this as refactoring how pins are stored is a better way forwards.

Issue #2276, opened 2019-07-18
Improve overall state of js-ipfs tests


hugomrdias:
Goals:

1. Improve developer productivity
2. Reduce the complexity of adding new tests for core and community devs
3. Increase overall coverage and quality of the tests

Tasks:
- [x] P0 `ipfsd-ctl` needs to be cleaned up https://github.com/ipfs/js-ipfsd-ctl/pull/308
- [x] P0 `ipfsd-ctl` **Complete Rewrite** more info here https://github.com/ipfs/js-ipfsd-ctl/pull/395#issue-327228792 
    - [x] https://github.com/ipfs/go-ipfs/pull/6489
    - [x] https://github.com/ipfs/js-ipfs/pull/2428
    - [x] ~~https://github.com/ipfs/js-ipfsd-ctl/pull/303~~
    - [x] ~~https://github.com/ipfs/js-ipfsd-ctl/pull/369~~
    - [x] https://github.com/ipfs/js-ipfsd-ctl/pull/395 (complete rewrite)
    - [x] `ipfsd-ctl 1.0.0` released
        -  Highlights:
            - All issues and prs closed
            - Much cleaner codebase to work on
- [x] P0 simpler, faster and async/await for setup for tests using the new ctl 
  - [x] https://github.com/ipfs/interface-js-ipfs-core/pull/541
  - [x] https://github.com/ipfs/js-ipfs-http-client/pull/1127
  - [x] https://github.com/ipfs/js-ipfs/pull/2529
    - Many tests were skipped in the browser and now **almost all** are running and passing
  - [ ] ~~interop~~
  - [ ] ~~ipfs-desktop~~
- [x] P0 All interface-core tests in js-ipfs run and pass in the browser
- [ ] P0 improve errors handling https://github.com/ipfs/js-ipfs/pull/1746
	- [ ] https://github.com/ipfs/js-ipfs-repo/pull/203#issuecomment-520413184
- [ ] P0 js-ipfs tests need some love
    - [ ] CLI tests needs to test cli logic and not core logic
        - [ ] cli needs to allow dependency injection to reduce mocks and spy (plus weird hacks to the node js require system)
        - [ ] internal yargs handling of promises needs to be rewritten
        - [x] improve CLI reliability to avoid zombie processes  https://github.com/ipfs/js-ipfs/pull/1981
- [ ] P1 Deduplication of tests from core tests and interface-core tests
[...]

PR #2636, opened 2019-11-26
chore: split test jobs into stages

alanshaw:
The current test stage uses _all_ the workers we have available in Travis and starves other projects from running their tests. This change separates interop, externals and examples into stages that only execute if the previous was successful. This means that we use fewer workers concurrently and fewer workers overall since a failure in the main tests will not process and run interop/externals and examples tests. This is at the expense of slower overall CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ branches:
 stages:
   - check
   - test
+  - interop
+  - externals
+  - examples
   - tag
 
 node_js:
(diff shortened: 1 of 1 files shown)


achingbrain:
Assuming zero flaky tests this will extend CI time to almost two hours (sum of the longest build time in each stage) which starts to impede how many builds we can do in a day.  We really need to make our tests faster!

daviddias:
> We really need to make our tests faster!

We got an @hugomrdias working on just that! :D https://github.com/ipfs/js-ipfs/issues/2276

alanshaw:
Perhaps limiting concurrent jobs is a better option? https://docs.travis-ci.com/user/customizing-the-build/#limiting-concurrent-jobs

alanshaw:
I've limited to 20, so that there's always at least 5 workers available to service other projects.

alanshaw:
Closing this for now. @daviddias please shout if you still feel js-ipfs is still not being a good CI citizen.

daviddias:
I'll avoid policing and let the team find the optimal point on ways it share the CI. My only remarks are:

- An artificial number will always be arbitrary and non optimal
- If all repos take a similar approach (in which they have to run 20+ jobs), using a 20 jobs ceiling would not scale.
- An observation is that if the primary tests of JS IPFS do not capture all the breakage over time, then they are probably not through enough. 
- Another observation, If the tests still take what it feels like _too much time_, it is probably because they are not parallel enough

Remarks said, as long as the current situation works for everyone, this doesn't feel urgent, so prioritize accordingly :)

PR #2637, opened 2019-11-27
perf: expose importer concurrency controls when adding files

achingbrain:
Adds two new arguments to the cli & http interface:

`--file-import-concurrency` and `--block-write-concurrency`

See https://github.com/ipfs/js-ipfs-unixfs-importer/pull/41 for further discussion.

Depends on:

- [x] https://github.com/ipfs/js-ipfs-http-client/pull/1187


alanshaw:
Is this adding 10kb to the bundle size?

achingbrain:
I had to increase the bundle size by 1kb so something has changed.

PR #2652, opened 2019-12-05
chore: revert #2647 upgrade to libp2p-delegated-content-routing

achingbrain:
Reverts ipfs/js-ipfs#2647 to see if it fixes the examples.

PR #2653, opened 2019-12-05
fix: add charset to script tag

achingbrain:
Maybe might fix the build?

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -11,7 +11,7 @@ const preloadNode = MockPreloadNode.createNode()
 const echoServer = EchoServer.createServer()
 
 module.exports = {
-  bundlesize: { maxSize: '650kB' },
+  bundlesize: { maxSize: '651kB' },
   webpack: {
     resolve: {
       mainFields: ['browser', 'main'],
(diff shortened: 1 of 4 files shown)


achingbrain:
\*shakes fist at bundlesize*

achingbrain:
Seems to fix breakage caused by [this line](https://github.com/jkroso/parse-duration/blob/master/index.js#L14).  Though the breakage was introduced by a `libp2p-delegated-content-routing` upgrade which doesn't seem related but it does go away if you [downgrade](https://github.com/ipfs/js-ipfs/pull/2652) that dep.  But that's not a solution.  Only forward, eh? 🤷‍♂

alanshaw:
urgh, that dep was added to the client in v40 which is now being used in the delegated routing module.

achingbrain:
I noticed that, but this module has had [builds pass](https://github.com/ipfs/js-ipfs/runs/323381337) with the v40 rpc client which would have pulled that dep into the tree.  I guess it could it have been dead code elimination in the browser bundle or something?

achingbrain:
A better fix: https://github.com/jkroso/parse-duration/pull/11

alanshaw:
Unnecessary now :)

PR #2661, opened 2019-12-11
refactor: convert bootstrap API to async/await

alanshaw:
⚠️ Note: this will be merged into https://github.com/ipfs/js-ipfs/pull/2547


src/core/components/bootstrap/rm.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+const isMultiaddr = require('mafmt').IPFS.matches
+
+function isValidMultiaddr (ma) {
```
dirkmc: Should we move this function to some common file?

src/core/components/bootstrap/rm.js
```diff
@@ -0,0 +1,39 @@
...
+    options = options || {}
+
+    if (multiaddr && !isValidMultiaddr(multiaddr)) {
+      throw new Error(`${multiaddr} is not a valid Multiaddr`)
+    }
+
+    let res = []
+    const config = await repo.config.get()
+
+    if (options.all) {
```
dirkmc: IIUC this function is doing two different things, it feels like maybe it should be two different functions:
- if the `all` option is not set it deletes a single multiaddress
- if the `all` option is set it deletes the whole bootstrap list
  However you still have to supply a null / undefined value as the first parameter. Feels a little weird.

alanshaw: Yes this really needs to be `ipfs.bootstrap.rm.all()` like https://docs.ipfs.io/reference/api/http/#api-v0-bootstrap-rm-all

One for a future refactoring :)

PR #2660, opened 2019-12-11
refactor: convert block and pin APIs to async/await

alanshaw:
⚠️ Note: this will be merged into https://github.com/ipfs/js-ipfs/pull/2547

`pin.ls` now returns an async iterable but internals were not refactored to not buffer output. Saved for whoever picks up https://github.com/ipfs/js-ipfs/issues/2244 ;)

PR #2664, opened 2019-12-11
refactor: convert dag API to async/await

alanshaw:
⚠️ Note: this will be merged into https://github.com/ipfs/js-ipfs/pull/2660 since it depends on the pin API.

In the future we might need a `DAGService` like go-ipfs to break the circular dependency.

TODO: fix in `ipfs-http-client` https://github.com/ipfs/js-ipfs-http-client/pull/1183 - BREAKING CHANGE: `dag.resolve` and `dag.tree` now return an async iterable


src/core/components/dag/put.js
```diff
@@ -0,0 +1,71 @@
+'use strict'
+
+const multicodec = require('multicodec')
+
+module.exports = ({ ipld, pin, gcLock, preload }) => {
+  return async function put (dagNode, options) {
+    options = options || {}
+
+    if (options.cid && (options.format || options.hashAlg)) {
```
dirkmc: It seems like we do this hash algorithm and format selection in a couple of places in the code, I wonder if we can unify it into one central place?

PR #2673, opened 2019-12-13
fix: skipped object.stat test

PedroMiguelSS:
This PR unskips `should respect timeout option` test of `object.stat()` over http-api as timeout is currently implemented as you can see [here](https://github.com/ipfs/js-ipfs-http-client/blob/master/src/object/stat.js#L22-L27). There's no need to be skipped.


PedroMiguelSS:
CI is fully green 🎉

PR #2674, opened 2019-12-13
refactor: convert repo API to async/await

alanshaw:
Depends on:

* [ ] https://github.com/ipfs/js-ipfs/pull/2658
* [ ] https://github.com/ipfs/js-ipfs/pull/2660


src/core/components/repo/gc.js
```diff
@@ -0,0 +1,114 @@
...
+    } finally {
+      release()
+    }
+  }
+}
+
+// Get Set of CIDs of blocks to keep
+async function createMarkedSet ({ pin, pinManager, refs, repo }) {
+  const pinsSource = async function * () {
+    for await (const { hash } of pin.ls()) {
```
dirkmc: Can we use streaming-iterables `map` here?

PR #2676, opened 2019-12-16
refactor: convert stats API to async/await

alanshaw:
Depends on:

* [ ] libp2p 0.27
* [ ] https://github.com/ipfs/js-ipfs/pull/2659
* [ ] https://github.com/ipfs/js-ipfs/pull/2674


src/core/components/stats/bw.js
```diff
@@ -0,0 +1,63 @@
...
+    rateIn: new Big(movingAverages.dataReceived[60000].movingAverage() / 60),
+    rateOut: new Big(movingAverages.dataSent[60000].movingAverage() / 60)
+  }
+}
+
+module.exports = ({ libp2p }) => {
+  return async function * (options) {
+    options = options || {}
+
+    if (!options.poll) {
```
dirkmc: Having a `poll` option feels slightly strange to me. I wonder if it would make sense to have two different methods:
- `stats()` that returns a single value
- `poll()` that is async iterable

alanshaw: I agree...

I'm not sold on the suggestion yet...`ipfs.stats.bw.stats()` doesn't read well to me! It would be nice to have a convention for APIs where we want to expose a one shot request/response but also a live streaming version...

```js
// Single shot:
const stat = await ipfs.stats.bw()

// Live feed:
for await (const stat of ipfs.stats.bw.feed()) {
}
```

dirkmc: ha yes that sounds much better :)

PR #2681, opened 2019-12-16
refactor: convert swarm API to async/await

alanshaw:


src/core/components/swarm/peers.js
```diff
@@ -0,0 +1,31 @@
...
+const CID = require('cids')
+
+module.exports = ({ libp2p }) => {
+  return async function peers (options) { // eslint-disable-line require-await
+    options = options || {}
+
+    const verbose = options.v || options.verbose
+    const peers = []
+
+    for (const [peerId, peerInfo] of libp2p.peerStore.peers.entries()) {
```
jacobheun: I created a PR in libp2p to make this simpler, https://github.com/libp2p/js-libp2p/pull/522. In theory you could have a large number of peers and only a few connections. Additionally, `peerInfo.isConnected()` is unreliable so we're moving away form using it. This will enable to the code to be:

```js
for (const [peerId, connections] of libp2p.connections) {
  const connectedAddr = connections[0].remoteAddr
  const tupple = {
    addr: connectedAddr,
    peer: new CID(peerId)
  }

  if (verbose) {
    tupple.latency = 'n/a'
  }

  peers.push(tupple)
}
```

This will only get you 1 of the possible connection addresses. If you want to see each connected address you can just iterate over the connections array.

PR #2718, opened 2020-01-15
chore: remove rogue j

olizilla:
i did not expect a j there.


achingbrain:
I mean, no-one ever does.

PR #2725, opened 2020-01-21
Add additional preloaders

coryschwartz:
Adding more preloaders, see the following issue.
https://github.com/protocol/bifrost-infra/issues/233

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -52,6 +52,8 @@ class IPFS extends EventEmitter {
         addresses: [
           '/dnsaddr/node0.preload.ipfs.io/https',
           '/dnsaddr/node1.preload.ipfs.io/https'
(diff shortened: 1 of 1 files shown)


alanshaw:
I'm not clear from the issue, are these up and running @coryschwartz or still in the process of being setup?

coryschwartz:
Still in the process, but I think they will be up and running by today.

alanshaw:
Nice, these also need to be added to the bootstrap lists here:

* https://github.com/ipfs/js-ipfs/blob/master/src/core/runtime/config-nodejs.js
* https://github.com/ipfs/js-ipfs/blob/master/src/core/runtime/config-browser.js

achingbrain:
These were added in a different PR

PR #2729, opened 2020-01-24
docs: updated examples to new api

achingbrain:
BREAKING CHANGE:

Where `PeerID`s were previously [CID]s, now they are Strings

- `ipfs.bitswap.stat().peers[n]` is now a String (was a CID)
- `ipfs.dht.findPeer().id` is now a String (was a CID)
- `ipfs.dht.findProvs()[n].id` is now a String (was a CID)
- `ipfs.dht.provide()[n].id` is now a String (was a CID)
- `ipfs.dht.put()[n].id` is now a String (was a CID)
- `ipfs.dht.query()[n].id` is now a String (was a CID)
- `ipfs.id().id` is now a String (was a CID)
- `ipfs.id().addresses[n]` are now [Multiaddr]s (were Strings)

Depends on:

- [x] https://github.com/ipfs/interface-js-ipfs-core/pull/581
- [x] https://github.com/ipfs/js-ipfs-http-client/pull/1226
- [x] https://github.com/libp2p/js-libp2p/pull/545

[CID]: https://www.npmjs.com/package/cids
[Multiaddr]: https://www.npmjs.com/package/multiaddr

PR #2730, opened 2020-01-29
chore: update deps

alanshaw:
Also adds `go-ipfs-dep` to `devDependencies`. See https://github.com/ipfs/interface-js-ipfs-core/issues/582#issuecomment-580215033

Also ~~cherry picks~~ copy pastes a fix from https://github.com/ipfs/js-ipfs/pull/2729/commits/2b6f0e9116da773dbca6a6523a3811e1a2552c55 to ensure our peer ID is not double suffixed onto our swarm addresses.


.travis.yml
```diff
@@ -81,6 +81,7 @@ jobs:
         - export IPFS_JS_EXEC=`pwd`/src/cli/bin.js
         - export IPFS_REUSEPORT=false
         - cd node_modules/ipfs-interop
+        - npm install go-ipfs-dep
```
achingbrain: Since this is in the `devDependencies` map, these lines can be removed, right?

alanshaw: they can now yes

alanshaw:
Closing as dependencies are all being upgraded here https://github.com/ipfs/js-ipfs/pull/2729

Issue #2696, opened 2019-12-25
ipfs.get() with CIDv1 does not fetch content


AuHau:
- **Version**: 0.40.0
- **Platform**: macos
- **Subsystem**:

#### Description:

Retrieving CIDv1 returns `type: dir` and no content.

#### Steps to reproduce the error:

```javascript
  it('js-ipfs should work', async () => {
    const cids = await ipfs.add(Buffer.from('hello world'), { cidVersion: 1 })
    console.log('All CIDs: ', cids)

    const result = await ipfs.get(cids[cids.length - 1].hash)
    console.log(result)

    const fetchedFromIpfs = result[0]
    expect(fetchedFromIpfs.content && fetchedFromIpfs.content.toString()).to.equal('hello world')
  })
```

Output of the former test:

```
All CIDs:  [
  {
    path: 'bafkreifzjut3te2nhyekklss27nh3k72ysco7y32koao5eei66wof36n5e',
    hash: 'bafkreifzjut3te2nhyekklss27nh3k72ysco7y32koao5eei66wof36n5e',
    size: 11
  }
]
[
  {
    hash: 'bafkreifzjut3te2nhyekklss27nh3k72ysco7y32koao5eei66wof36n5e',
    path: 'bafkreifzjut3te2nhyekklss27nh3k72ysco7y32koao5eei66wof36n5e',
    name: 'bafkreifzjut3te2nhyekklss27nh3k72ysco7y32koao5eei66wof36n5e',
    depth: 1,
    size: 0,
    type: 'dir'
  }
]
```

AuHau:
Hmm from tests in `interface-ipfs-core` I found out that with `rawLeaves: false` option this test passes. Why? Is this then a bug or not?

kuabhish:
Hey everyone, I have a similar issue.

I was going through the js-ipfs/example/ipfs-101. In the example ipfs.get is not discussed. Still I tried to write get like this.

Hello, how are you today? Welcome to the Distributed Web! : is written in hello.txt. How can i obtain it with and set a different name for saving?

`const IPFS = require('ipfs')

async function main () {
const node = await IPFS.create()
const filesAdded = await node.add([{
path: 'hello.txt',
content: 'Hello World 101'
}])
console.log('Added file:', filesAdded[0].path, filesAdded[0].hash) ## adding the file
console.log('Added file:', filesAdded )

const fileBuffer = await node.cat(filesAdded[0].hash) ## viewing the file
[...]

PR #2747, opened 2020-02-05
chore: update importer to only output unixfs things

achingbrain:
BREAKING CHANGE:

If you specify `raw-nodes` to be true or `cid-version` to be `1` and your data fits in one chunk, you would previously get a `CID` that resolves to an `ipld-raw` node.

Now you will get a `CID` that resolves to a `dag-pb` node, so the output of every import will be a `unixfs` object.


achingbrain:
Refs #2696

alanshaw:
Just double checking: `rawNodes` => `rawLeaves`?

I think this is a good idea. It will cause an interop issue with go-ipfs because CIDs will be different for the same data (when `cidVersion` is `1` or `rawLeaves` is `true`). Someone will probably raise this as a problem if go-ipfs doesn't catch up quickly. So, I just wanted to check there's consensus on this with the go-ipfs team and that there is an issue to track it that we can point people at when it is reported?

Will you also check there is an interop test for this (albeit skipped for now)? 🙏

achingbrain:
> check there's consensus on this with the go-ipfs team

This change is in line with @Stebalien's comment here: https://github.com/ipfs/js-ipfs-unixfs-importer/issues/44#issuecomment-573745942 so I believe there is consensus 😉 

> and that there is an issue to track it that we can point people at when it is reported?

Right here: https://github.com/ipfs/go-ipfs/issues/6940

> Will you also check there is an interop test for this (albeit skipped for now)? 

Here: https://github.com/ipfs/interop/pull/105

Kind of an aside but I think we should move most of those interop tests to the interface suite - I think the interop tests should be focused on networking type stuff between go, js, rust etc.  E.g. bitswap, the dht, pubsub, etc.

PR #2748, opened 2020-02-05
chore: enable caching, disable on windows

achingbrain:
From what I remember, we disabled npm caching in this build because windows builds were timing out when it tried to restore the cache due to no output being received.

This change re-enables caching but turns it off on windows so at least the other builds should be faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+cache: npm
 
 branches:
   only:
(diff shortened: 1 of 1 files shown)


achingbrain:
Bleugh, npm caching has been on by default in Travis since last July, I should read the docs before pushing: https://docs.travis-ci.com/user/caching/#npm-cache

PR #2759, opened 2020-02-07
feat: metadata for ipfs.get

alanshaw:
Pack file `mtime` and `mode` if set

Depends on:

* [ ] https://github.com/ipfs/js-ipfs/pull/2758


src/http/api/resources/files-regular.js
```diff
@@ -11,16 +11,20 @@ const Boom = require('@hapi/boom')
...
-const concat = require('it-concat')
 const ndjson = require('iterable-ndjson')
 const { map } = require('streaming-iterables')
 const streamResponse = require('../../utils/stream-response')
 
+const toBuffer = async function * (source) {
+  for await (const chunk of source) {
+    yield chunk.slice()
+  }
+}
```
hugomrdias: this pattern is begging for a `it-transform`

alanshaw: Could use this: https://www.npmjs.com/package/it-buffer, it's already used in libp2p.

alanshaw:
I guess the CLI should set the `mode` when writing the file...and then...change it's `mtime` when done?

src/http/api/resources/files-regular.js
```diff
@@ -95,7 +95,9 @@ exports.get = {
       async function * (source) {
         for await (const file of source) {
           const header = {
-            name: file.path
+            name: file.path,
+            mtime: file.mtime ? new Date(file.mtime.secs * 1000) : null,
```
achingbrain: Similar to my comment on https://github.com/ipfs/js-ipfs-http-client/pull/1235 - can this support the `nsecs` mtime component too?

achingbrain: I think this might be more complicated than I'd previously thought - `it-tar` emits regular tar files, right?  As I understand it, regular tar files only support seconds for mtimes, higher resolution requires GNU Tar files so `it-tar` may need to be updated to support producing these.

alanshaw: Correct https://github.com/alanshaw/it-tar/blob/master/pack-headers.js#L96

Issue #2751, opened 2020-02-07
docs: how do you create an in-memory offline ipfs node for testing?


mikeal:
I’ve been trying to create a purely in-memory offline ipfs node and it doesn’t appear to be working. The node appears to function but blocks I’ve just added aren’t actually available. https://github.com/mikeal/ipfs-for-car/blob/master/index.js#L1-L31

alanshaw:
If `key` is a CID you need to pass it as `cid` in the options object.

```js
await ipfs.block.put(value, { cid: key })
```

https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/BLOCK.md#blockput

mikeal:
Thanks, I’m sure that helps but it’s still not working.

https://github.com/mikeal/ipfs-for-car/blob/master/index.js

I get `Not Found` when trying to pull out a block I’ve just added. I’m assuming something is wrong with how I’m configuring the inmemory storage because I doubt basic block storage is broken 🤷‍♂️

```
first bafyreiec5aigscqhrvctqugyrox5hsjyar624aegy3ooq75usntmffz2hq
(node:4997) UnhandledPromiseRejectionWarning: Error: Not Found
    at Object.module.exports.notFoundError (/root/ipfs-for-car/node_modules/interface-datastore/src/errors.js:21:
16)
    at MemoryDatastore.get (/root/ipfs-for-car/node_modules/interface-datastore/src/memory.js:22:31)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Object.get (/root/ipfs-for-car/node_modules/ipfs-repo/src/blockstore.js:49:21)
(node:4997) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwi
ng inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch
(). (rejection id: 1)
(node:4997) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

alanshaw:
Does the CID you’re retrieving it with match the CID you got back from block.put?

If yes then can you share the car file and I can take a look?

alanshaw:
[...]

PR #2760, opened 2020-02-09
fix: block put with CID as string

alanshaw:
resolves https://github.com/ipfs/js-ipfs/issues/2751

Conform to [documented block.put API](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/BLOCK.md#blockput).

Depends on:

* [x] https://github.com/ipfs/interface-js-ipfs-core/pull/588


alanshaw:
Merging this now so it can be included in the next RC

PR #2765, opened 2020-02-10
chore: remove ipfs-pubsub-room from CI

alanshaw:
It does not depend on IPFS or the HTTP client.

https://github.com/ipfs-shipyard/ipfs-pubsub-room

Looks like this was a recent change https://github.com/ipfs-shipyard/ipfs-pubsub-room/pull/73


achingbrain:
Turns out it wasn't actually using any IPFS features...

PR #2769, opened 2020-02-11
docs: improve browserify example readme

davideicardi:
Just tried to explain better how to run the example.

### DIFF
--- a/examples/browser-browserify/README.md
+++ b/examples/browser-browserify/README.md
@@ -1,8 +1,10 @@
 # Bundle js-ipfs with Browserify!
 
-> In this example, you will find a boilerplate you can use to guide yourself into bundling js-ipfs with browserify, so that you can use it in your own web app!
+In this example, you will find a boilerplate you can use to guide yourself into bundling js-ipfs with [browserify](http://browserify.org/), so that you can use it in your own web app!
 
 ## Run this example
+
+First install dependencies in the project root directory with `npm install`, then build the browser bundle with `npm run build`. Then in this directory run:
 
 ```bash
 > npm install


examples/browser-browserify/README.md
```diff
@@ -1,9 +1,11 @@
 # Bundle js-ipfs with Browserify!
 
-> In this example, you will find a boilerplate you can use to guide yourself into bundling js-ipfs with browserify, so that you can use it in your own web app!
+In this example, you will find a boilerplate you can use to guide yourself into bundling js-ipfs with [browserify](http://browserify.org/), so that you can use it in your own web app!
 
 ## Run this example
 
+First install dependencies and build main ipfs package on the parent directory. Then run:
```
alanshaw: Parent directory is `/path/to/js-ipfs/examples` but we actually need deps to be installed in `/path/to/js-ipfs`, also you shouldn't need to build the browser bundle for this.

```suggestion
First install dependencies in the project root directory with `npm install`. Then in this directory run:
```

davideicardi: Thanks @alanshaw .
But from what I understand this example uses `dist` directory. I see this in `package.json`:
```
  "browser": {
    "ipfs": "ipfs/dist"
  }
```
I think that this is a feature of browserify: https://github.com/browserify/browserify#browser-field

So we need to first build the `dist` directory. Correct?

alanshaw: Yes, correct, I missed that.

alanshaw: How about this instead then?

```suggestion
First install dependencies in the project root directory with `npm install`, then build the browser bundle with `npm run build`. Then in this directory run:
```

alanshaw:
Thanks @davideicardi ❤️

PR #2772, opened 2020-02-12
chore: upgrade to latest ipfsd-ctl

achingbrain:
Pulled out of #2762 because it's needed elsewhere.


achingbrain:
https://github.com/libp2p/js-libp2p/pull/558 is breaking the custom libp2p example, I'm going to merge this and then patch that up in a subsequent PR.

PR #2773, opened 2020-02-12
fix: await on things that need awaiting on

achingbrain:
These are async methods and can result in unhandledPromiseRejection events :(


alanshaw:
I think [this](https://github.com/libp2p/js-libp2p/commit/a317a8b0118ea03b76cdc430c9e28e2451b84532) has broken the libp2p example.

achingbrain:
Wot, more breaking changes in a patch release?

achingbrain:
I'm going to merge this and fix up the custom libp2p example in a subsequent PR.

PR #2784, opened 2020-02-18
chore: update testing externals

achingbrain:
I've generalised the aegir test-externals command so it's not js-ipfs specific any more and can be used by our other projects, hopefully by aegir itself too to catch breakage.  This change updates our invocation of the command.


achingbrain:
Merged because the new command went out in `aegir@21.1.0` instead of `aegir@22.0.0`.  Oops.

PR #2785, opened 2020-02-20
perf: remove asn1.js from dependencies

alanshaw:
Current master: 579.82KB
This PR: 558.25KB


achingbrain:
This was released in `libp2p-crypto@0.17.3` so we should get it thanks to semver magic!

PR #2793, opened 2020-02-24
chore: windows build fixes

achingbrain:
Stop things that were started and checkout things with the right line endings.

Must move gateway fixtures to fixtures dir, or preferably generate data during the tests.


achingbrain:
Hmm, not convince git will re-checkout files with the line endings specified in `.gitattributes` when switching between branches so am going to merge this as it's resolved some of the windows problems for me locally.

PR #2810, opened 2020-03-03
feat: remove ky from http-client and utils

### LINKED ISSUE #2801: Tracking issue for `ky` removal
ky has problems with node 13, clones responses https://github.com/ipfs/js-ipfs-http-client/blob/master/src/lib/error-handler.js#L12

Related:
https://github.com/ipfs/js-ipfs-http-client/issues/1117#issuecomment-589011968
https://github.com/ipfs/js-ipfs-http-client/pull/1224#issuecomment-589220724 this needs to be patched in react-native not here
https://github.com/ipfs/js-ipfs-http-client/issues/1219#issuecomment-582846315
https://github.com/ipfs/js-ipfs-http-client/pull/1224

hugomrdias:
This PR removes ky and simplifies a couples things searchParams handling, baseUrl normalization and native fetch options usage.

Native fetch options can be passed into the contructor or methods and are merged properly, methods now use the third argument to receive these options. 

left some comments in the code to explain some of the ideas, please review so i can continue refactoring if approved.

No support for retries or hooks will implement as needed.

closes #2801

related
https://github.com/ipfs/interface-js-ipfs-core/issues/58


packages/ipfs-http-client/src/index.js
```diff
@@ -8,14 +8,109 @@ const multicodec = require('multicodec')
...
+    headers: config.headers,
+    base: normalizeURL(config).toString(),
+    handleError: errorHandler,
+    // apply two mutations camelCase to kebad-case and remove undefined/null key/value pairs
+    // everything else either is a bug or validation is needed
+    transformSearchParams: (obj) => {
+      const out = {}
+
+      for (const [key, value] of obj) {
+        if (value !== 'undefined' && value !== 'null') {
```
achingbrain: Should these be strings?

hugomrdias: yes the input is a URLSearchParams instance

packages/ipfs-http-client/src/add/index.js
```diff
@@ -2,56 +2,48 @@
...
 const toIterable = require('stream-to-it/source')
 const { toFormData } = require('./form-data')
 const toCamel = require('../lib/object-to-camel')
+const merge = require('merge-options')
 
-module.exports = configure(({ ky }) => {
-  return async function * add (input, options) {
-    options = options || {}
+module.exports = api => {
+  return async function * add (input, options = {}, fetchOptions = {}) {
```
achingbrain: Should we stick to the interface api here and not add extra arguments?

I see `timeout`, `signal` and `headers` - `signal` will become a thing with https://github.com/ipfs/interface-js-ipfs-core/issues/58, `timeout` might then be redundant I guess as you can just race them and use the signal to do the timeout.

For headers, I think these are used for authentication mainly when people put a proxy up in front of their nodes, we could allow the user to specify headers added to all requests in the http-client constructor?

packages/ipfs-http-client/src/index.js
hugomrdias: inside the API i do a double wrap in URLSearchParams so the user can input anything the URLSearchParam constructor accepts and here its already a URLSearchParam instance that can in turn be transformed into anything the constructor accepts. 
this way its easy for the user and inside the transformer there is no surprises because its already an URLSearchParams

packages/ipfs-http-client/src/add/index.js
hugomrdias: constructor already supports these 3 and methods 3rd arg merges with the constructor input.

also timeout is already internally connected to any signal passed in to avoid races.

this 3rd argument is totally optional i added it for advanced use cases where you want to customize fetch native options, this is why i didn't re-use the 2nd argument. This should support all these https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch not just headers, timeout and signal.

this should be http-client specific, and if we add signal to the Core API we can easily merge the two.

hugomrdias: we can even make the constructor consistent with methods adding a `fetchOptions` arg and not mix Core API options with fetch options.

achingbrain: I think merging the two would be preferable, otherwise we have embrace-and-extend style implementations of the core interface.

We can think about the fetch args we want to allow through - it's supposed to be an RPC client, exposing too many transport concerns makes the abstraction leak.

achingbrain:
Looks good so far 👍

packages/ipfs-http-client/src/add/index.js
```diff
@@ -2,56 +2,43 @@
...
-    if (options.quiet != null) searchParams.set('quiet', options.quiet)
-    if (options.quieter != null) searchParams.set('quieter', options.quieter)
-    if (options.rawLeaves != null) searchParams.set('raw-leaves', options.rawLeaves)
-    if (options.shardSplitThreshold) searchParams.set('shard-split-threshold', options.shardSplitThreshold)
-    if (options.silent) searchParams.set('silent', options.silent)
-    if (options.trickle != null) searchParams.set('trickle', options.trickle)
-    if (options.wrapWithDirectory != null) searchParams.set('wrap-with-directory', options.wrapWithDirectory)
-    if (options.preload != null) searchParams.set('preload', options.preload)
-    if (options.fileImportConcurrency != null) searchParams.set('file-import-concurrency', options.fileImportConcurrency)
-    if (options.blockWriteConcurrency != null) searchParams.set('block-write-concurrency', options.blockWriteConcurrency)
```
pcowgill: It's nice that this is much DRYer now

pcowgill:
@hugomrdias What work remains before this PR will leave the draft stage? Thanks!

packages/ipfs-http-client/src/index.js
```diff
@@ -8,14 +8,114 @@ const multicodec = require('multicodec')
...
+}
+
+function ipfsClient (config = {}) {
+  const api = new API({
+    timeout: config.timeout || 60000 * 20,
+    signal: config.signal,
+    headers: config.headers,
+    base: normalizeURL(config).toString(),
+    handleError: errorHandler,
+    // apply two mutations camelCase to kebad-case and remove undefined/null key/value pairs
```
pcowgill:
```suggestion
    // apply two mutations camelCase to kebab-case and remove undefined/null key/value pairs
```

packages/ipfs-http-client/src/index.js
```diff
@@ -8,14 +8,114 @@ const multicodec = require('multicodec')
...
+
+  // If we managed to extract a message from the response, use it
+  if (msg) {
+    error.message = msg
+  }
+
+  throw error
+}
+
+var KEBAB_REGEX = /[A-Z\u00C0-\u00D6\u00D8-\u00DE]/g
```
pcowgill: Might be helpful to insert a comment above this line explaining in plain text what this regex is intended to catch beyond A-Z

packages/ipfs-http-client/src/index.js
```diff
@@ -8,14 +8,114 @@ const multicodec = require('multicodec')
...
+
+  url.pathname = config.apiPath || 'api/v0'
+  if (!api) {
+    if (isBrowser || isWebWorker) {
+      url.protocol = config.protocol || location.protocol
+      url.host = config.host || location.hostname
+      url.port = config.port || location.port
+    } else {
+      url.host = config.host || 'localhost'
+      url.port = config.port || '5001'
```
pcowgill: Defaults to `go-ipfs` rather than `js-ipfs` (5002)?

packages/ipfs-http-client/src/add/index.js
pcowgill: We can resolve this comment thread

pcowgill:
Once `dag.put` is refactored to use the new `api.post` function I can check whether https://github.com/ipfs/js-ipfs/issues/2825 is still an issue in this feature branch.

packages/ipfs-http-client/src/add/index.js
```diff
@@ -1,51 +1,36 @@
...
+      options,
+      {
+        'stream-channels': true,
+        progress: Boolean(progressFn),
+        hash: options.hashAlg // TODO fix this either is hash or hashAlg
+      }
+    )
 
-    const res = await ky.post('add', {
+    const res = await api.ndjson('add', {
```
achingbrain: This is something to tidy up in a subsequent PR, but I think given the `fetch` pattern is:

```javascript
const res = await fetch('url')
const body = await res.json()
// ... do something with body
```

a more idiomatic approach here might be:

```javascript
const res = await api.post({...})

for await (const file of res.ndjson()) {
  //... do something with file
}
```

pcowgill: I’d vote for having the res.ndjson put in a variable first in a separate line so that stack traces are easier to read

pcowgill: So that would look like this:
```js
const res = await api.post({...})

const ndjsonRes = res.ndjson()

for await (const file of ndjsonRes) {
  //... do something with file
}
```

packages/ipfs-http-client/src/version.js
```diff
@@ -1,19 +1,24 @@
...
 
-    const res = await ky.post('version', {
+/**
+ * Version
+ * @param {API} api
+ * @returns {function(Object): Promise<Object>}
+ */
+const version = (api) => {
+  return async (options = {}) => {
+    const res = await (await api.post('version', {
```
pcowgill: The double await here - and also in places like `bitswap/wantlist` - would be more readable split out into separate lines

packages/interface-ipfs-core/src/dht/put.js
```diff
@@ -12,7 +13,7 @@ module.exports = (common, options) => {
   const describe = getDescribe(options)
   const it = getIt(options)
 
-  describe('.dht.put', function () {
+  describe.skip('.dht.put', function () {
```
hugomrdias: missing a todo here 
// TODO fix when go-ipfs 0.5.0 ships, currently doesn't even work properly

packages/ipfs-http-client/src/lib/core.js
```diff
@@ -0,0 +1,137 @@
...
+   */
+  constructor (options = {}) {
+    /** @type {ClientOptions} */
+    const opts = normalizeInput(options)
+    super({
+      timeout: parseTimeout(opts.timeout) || 60000 * 20,
+      headers: opts.headers,
+      base: normalizeInput(opts.url).toString(),
+      handleError: errorHandler,
+      // apply two mutations camelCase to kebad-case and remove undefined/null key/value pairs
```
pcowgill:
```suggestion
      // apply two mutations camelCase to kebab-case and remove undefined/null key/value pairs
```

achingbrain:
> Let's get it in and address any problems with follow up PRs.

I didn't mean merge while CI was red 🤣

hugomrdias:
fixing it, for some reason i just so green  🤣

packages/ipfs-http-client/src/version.js
pcowgill: @hugomrdias Curious for your thoughts on this when you get a chance. Thanks!

packages/ipfs-http-client/src/add/index.js
pcowgill: @hugomrdias @achingbrain Curious for your thoughts on this when you get a chance. Thanks!

packages/ipfs-http-client/src/index.js
pcowgill: @hugomrdias Curious for your thoughts on this when you get a chance. Thanks!

packages/ipfs-http-client/src/version.js
hugomrdias: i will fix this in another PR later

packages/ipfs-http-client/src/add/index.js
hugomrdias: i would prefer to keep this. ty
alex comment will be improved later.

pcowgill: When you say keep this, do you mean your current approach as opposed to Alex's suggestion?

Or do you mean that in general you prefer the style where multiple operations happen in a single line?

PR #2811, opened 2020-03-03
chore: move mfs and multipart files into core

achingbrain:
I've converted the mfs tests to interface tests and they are a lot more thorough than the previous ones.

Turns out a whole bunch of stuff didn't work over http.  Also, some of our interface tests now make go-ipfs fail, mostly around mfs operations on hamt shards - removing files from shards in particular.  I've skipped all of these tests when we run against go-ipfs but we'll need some follow up issues.

BREAKING CHANGES:

`ipfs.files.stat(path)` run on a hamt sharded directory used to report a `type` of `hamt-sharded-directory`, now it reports `directory` to bring js-ipfs in line with go-ipfs.


achingbrain:
Blocked on https://github.com/ipfs/js-ipfs/pull/2810

achingbrain:
Fixes #2863 and #2838

packages/ipfs-utils/src/http.js
```diff
@@ -126,6 +123,18 @@ class HTTP {
...
+    if (opts.body && (opts.body.on || opts.body.addEventListener)) {
+      const on = (opts.body.on || opts.body.addEventListener).bind(opts.body)
+
+      // TODO: not sure this will work after the high water mark is hit and the request
+      // starts streaming
+      on('error', (err) => {
+        bodyErr = err
+      })
+    }
+
```
hugomrdias: can you explain why we need this ?

packages/ipfs-utils/src/http.js
```diff
@@ -292,3 +306,38 @@ HTTP.ndjson = ndjson
...
+ * @returns {Promise<Response>}
+ */
+module.exports.delete = (resource, options) => new HTTP().delete(resource, options)
+
+/**
+ * @param {string | URL | Request} resource
+ * @param {APIOptions} options
+ * @returns {Promise<Response>}
+ */
+module.exports.options = (resource, options) => new HTTP().options(resource, options)
```
hugomrdias:
```suggestion
/**
 * @param {string | URL | Request} resource
 * @param {APIOptions} options
 * @returns {Promise<Response>}
 */
HTTP.post = (resource, options) => new HTTP().post(resource, options)

/**
 * @param {string | URL | Request} resource
 * @param {APIOptions} options
 * @returns {Promise<Response>}
 */
HTTP.get = (resource, options) => new HTTP().get(resource, options)

/**
 * @param {string | URL | Request} resource
 * @param {APIOptions} options
 * @returns {Promise<Response>}
 */
HTTP.put = (resource, options) => new HTTP().put(resource, options)

/**
 * @param {string | URL | Request} resource
 * @param {APIOptions} options
 * @returns {Promise<Response>}
 */
HTTP.delete = (resource, options) => new HTTP().delete(resource, options)

/**
 * @param {string | URL | Request} resource
 * @param {APIOptions} options
 * @returns {Promise<Response>}
 */
HTTP.options = (resource, options) => new HTTP().options(resource, options)

module.exports = HTTP
```

its easier to do typescript stuff like this ...

packages/ipfs/src/core/components/files/index.js
```diff
@@ -1,10 +1,88 @@
 'use strict'
 
-const mfs = require('ipfs-mfs/core')
+const assert = require('assert')
```
hugomrdias: please lets just throw, assert is node global package and is big.

packages/ipfs-http-client/src/files/cp.js
```diff
@@ -3,20 +3,21 @@
...
       signal: options.signal,
-      searchParams
+      searchParams: toUrlSearchParams(
+        sources.map(src => CID.isCID(src) ? `/ipfs/${src}` : src),
+        options
+      )
     })
-    return res.text()
+
+    await res.text()
```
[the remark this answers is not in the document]
achingbrain: If we return `res.text()` the promise resolves to an empty string, it should resolve to undefined.

packages/ipfs-utils/src/http.js
achingbrain: Less good for tree shaking though, you know, when import/export is a thing.

hugomrdias: yeah but here bundlers wouldn't tree shake much anyway actually only rollup would potentially treeshake something so.. with esm typescript understands exports better and we can do whatever, with cjs we have some limitations sadly.

packages/ipfs-utils/src/http.js
achingbrain: If the body is a stream, and that stream errors the error is not surfaced to the calling code.

This will throw any error thrown until the response headers arrive.  I'm not sure if we should keep it TBH as it stops working once the response headers have been received, after that any error is smothered, same as without the `bodyErr` lines in the PR.

It makes this test pass:

```javascript
it('should handle errors in streaming bodies', async function () {
  const err = new Error('Should be caught')
  const body = async function * () {
    yield Buffer.from([0, 1, 2])

    // await delay(1000) <-- uncomment this line and the test fails

    throw err
  }()

  await expect(HTTP.post('http://localhost:3000', {
    body: toStream.readable(body)
  })).to.eventually.be.rejectedWith(err)
})
```

PR #2831, opened 2020-03-06
chore: add stardust to libp2p config and updated instructions on browser example

vasco-santos:
This PR adds `libp2p-stardust` transport to the ipfs browser runtime configuration.

Moreover, the stardust was added to the `exchange-files` example, in order to testing and provide an example for the community on how to use it.

Needs:

- [x] [libp2p/js-libp2p-stardust#14](https://github.com/libp2p/js-libp2p-stardust/pull/14)
- [ ] final stardust release
- [ ] Add hosted stardust server address to swarm config


vasco-santos:
cc @mkg20001  (cannot ask you review on ipfs org)

examples/exchange-files-in-browser/package.json
```diff
@@ -7,14 +7,18 @@
...
+    "server-stardust": "stardust-server"
   },
   "license": "MIT",
   "devDependencies": {
     "browserify": "^16.2.3",
     "execa": "^3.2.0",
     "http-server": "^0.11.1",
-    "ipfs-http-client": "^42.0.0"
+    "ipfs-http-client": "^42.0.0",
+    "libp2p-stardust": "libp2p/js-libp2p-stardust#refactor/async-await",
```
mkg20001: It's released `0.2.0-rc.0`

packages/ipfs/README.md
```diff
@@ -942,6 +942,29 @@ const node = await IPFS.create({
 
 The code above assumes you are running a local `signaling server` on port `9090`. Provide the correct values accordingly.
 
+**Note:** We host a signalling server at star-signal.cloud.ipfs.team that can be used for practical demos and experimentation, it should not be used for apps in production. A libp2p-webrtc-star address, using the signalling server we provide, looks like:
+
+/dns4/star-signal.cloud.ipfs.team/wss/p2p-webrtc-star/p2p/<your-peer-id>
```
jacobheun: This address is going to end up changing when we update the deploy, which is in progress. This server is currently not stable and is old.

packages/ipfs/README.md
```diff
@@ -942,6 +942,29 @@ const node = await IPFS.create({
...
+
+Note: The address above indicates WebSockets Secure, which can be accessed from both http and https.
+
+#### Is there a more stable alternative to webrtc-star that offers a similar functionality?
```
jacobheun: I don't think this is an accurate statement to make. It is an alternative but unless I am mistaken we have no performance measurements between the two to say stardust is more/less stable. Ideally, we should do some load testing on the rendezvous servers to make these comparisons.

packages/ipfs/README.md
```diff
@@ -942,6 +942,29 @@ const node = await IPFS.create({
...
+
+Yes, stardust! A WebSockets based transport that uses a Relay to route the messages. To enable it, just do:
+
+```JavaScript
+const node = await IPFS.create({
+  config: {
+    Addresses: { 
+      Swarm: [
+        '/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star'
```
jacobheun: This address is wrong and is to websocket-star, it will need to change once there is a public test server to point to.

packages/ipfs/README.md
vasco-santos: Yeah, I will add this once we have the new server deployed

packages/ipfs/README.md
vasco-santos: I will put the own `signalling` instructions while we don't have this deployed

packages/ipfs/test/core/libp2p.spec.js
```diff
@@ -8,6 +8,8 @@ const Libp2p = require('libp2p')
 const EE = require('events')
 const libp2pComponent = require('../../src/core/components/libp2p')
 
+const isBrowser = require('is-browser')
```
lidel: No need for external lib, `isBrowser` is already implemented by `ipfs-utils`:

```suggestion
const { isBrowser } = require('ipfs-utils/src/env')
```

packages/ipfs/README.md
```diff
@@ -942,6 +942,25 @@ const node = await IPFS.create({
...
+    Addresses: { 
+      Swarm: [
+        `/ip4/0.0.0.0/tcp/5892/ws/p2p-stardust/p2p/${stardustServerId}`
+      ]
+    }
+  }
+})
+// your instance with stardust is ready
+```
+
```
lidel: Good place to mention self-hosting and Docker images:

```suggestion

While the libp2p infra team host stardust servers for users to test with, these servers should **not be relied on for Production applications**.

If you wish to leverage stardust servers for your application, we recommend running your own.
Docker images can be found at [libp2p Docker Hub](https://hub.docker.com/u/libp2p).

```

packages/ipfs/README.md
```diff
@@ -942,6 +942,25 @@ const node = await IPFS.create({
...
+Yes, stardust! A WebSockets based transport that uses a Relay to route the messages. To enable it, you will need to execute its `signaling server` (available on [libp2p-stardust](https://github.com/libp2p/js-libp2p-stardust)) and include the correct `stardustServerId` as follows:
+
+```JavaScript
+const stardustServerId = 'QmY7t45zkBxzw5rtWQp3oAjzoJGTjh49f7LNz4YPBtTpqy'
+
+const node = await IPFS.create({
+  config: {
+    Addresses: { 
+      Swarm: [
+        `/ip4/0.0.0.0/tcp/5892/ws/p2p-stardust/p2p/${stardustServerId}`
```
lidel: Just noticed that `Address.Swarm` notation is pretty different from one in webrtc-star and legacy ws-star where user just points at anonymous endpoint.

- Is this because the `stardust` protocol requires prior knowledge of `stardustServerId`?
- This is the main README, are we going to replace `/ip4/0.0.0.0/tcp/5892` with a real server?
  (`0.0.0.0` makes sense in examples where we run one on localhost, but here it may be confusing to users who try to copy&paste just to try it out)

packages/ipfs/package.json
```diff
@@ -184,6 +185,7 @@
     "interface-ipfs-core": "^0.132.0",
     "ipfs-interop": "^1.0.0",
     "ipfsd-ctl": "^3.0.0",
+    "is-browser": "^2.1.0",
```
vasco-santos:
```suggestion
```

packages/ipfs/README.md
vasco-santos: When the `multiaddrs` are added to libp2p peer for listening, `js-ipfs` adds a `/p2p/{own_peer_id}`. 

In `webrtc-star`, it gets the valid `multiaddr`, however `stardust` multiaddr to listen must be `/p2p/${stardustServerId}/p2p/{own_peer_id}`. This is because `stardust` server is a libp2p node, while `webrtc-star` is not a libp2p node.

I thought about creating a fixture with the `peer-id` first, so that we have always the same id and not need this change. However, I think that this is not the best solution. 

Anyway, the plan is to go as you asked. Once we have some servers deployed for experimenting, the server id issue will not be a problem anymore and we can replace this by the server address. The same will happen for `webrtc-star`.

vasco-santos:
We can close this as we will not move on with startdust

PR #2832, opened 2020-03-06
fix: tag stdin with mtime and mode when piping to cli 'add'

### LINKED ISSUE #2763: mtime and mode not set when piping from stdin
Stdin stream passed directly to `ipfs.add` https://github.com/ipfs/js-ipfs/blob/39d87c5a59c8187d849088e06ca9d9224d7397e7/src/cli/commands/add.js#L239

e.g.

```console
$ echo "data" | node src/cli/bin add --mtime=100                        
added QmNtocSs7MoDkJMc1RkyisCSKvLadujPsfJfSdJ3e1eA1M QmNtocSs7MoDkJMc1RkyisCSKvLadujPsfJfSdJ3e1eA1M

$ jsipfs files stat /ipfs/QmNtocSs7MoDkJMc1RkyisCSKvLadujPsfJfSdJ3e1eA1M
QmNtocSs7MoDkJMc1RkyisCSKvLadujPsfJfSdJ3e1eA1M
Size: 5
CumulativeSize: 13
ChildBlocks: 0
Type: file
Mode: -rw-r--r--
Mtime: -
```

kanej:
Converts the stdin stream to a async iterator of one file and tags that file with the mtime and
mode.

fixes #2763

### DIFF
--- a/packages/ipfs/src/cli/commands/add.js
+++ b/packages/ipfs/src/cli/commands/add.js
@@ -236,7 +236,11 @@ module.exports = {
         mode: argv.mode,
         mtime
       })
-      : getStdin() // Pipe directly to ipfs.add
+      : {
+        content: getStdin(),
+        mode: argv.mode,
+        mtime
+      } // Pipe to ipfs.add tagging with mode and mtime
 
     let finalCid
 
--- a/packages/ipfs/test/cli/files-regular.js
+++ b/packages/ipfs/test/cli/files-regular.js
@@ -157,12 +157,36 @@ describe('files', () => {
   it('add from pipe', async () => {
     const cid = new CID('QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB')
 
-    ipfs.add.withArgs(matchIterable(), defaultAddArgs()).returns([{
+    ipfs.add.withArgs(sinon.match({
+      content: matchIterable()
+    }), defaultAddArgs()).returns([{
       cid,
       path: 'readme'
     }])
 
     const proc = cli('add', {
+      ipfs,
+      getStdin: function * () {
+        yield Buffer.from('hello\n')
+      }
+    })
+
+    const out = await proc
+    expect(out).to.equal(`added ${cid} ${cid}\n`)
+  })
+
+  it('add from pipe with mtime=100', async () => {
+    const cid = new CID('QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB')
+
+    ipfs.add.withArgs(sinon.match({
+      content: matchIterable(),
+      mtime: { secs: 100 }
+    }), defaultAddArgs()).returns([{
+      cid,
+      path: 'readme'
+    }])
+
+    const proc = cli('add --mtime=100', {
       ipfs,
       getStdin: function * () {
         yield Buffer.from('hello\n')


achingbrain:
Neat! Could you add some tests around this please?  The CLI handlers add take a `getStdin` function as a context argument so it should be relatively straightforward.

See the [`add from pipe`](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/test/cli/files.js#L157-L174) test for an example.

kanej:
Hi @achingbrain, I have added in a test to cover the piping with the mtime argument. The build failed shows a failure on the `ipfs-http-client` tests though the changes are in the cli package. Is this likely to be a gremlin in the test suite or an indicator of my capacity for truly epic bugs?

packages/ipfs/src/cli/commands/add.js
```diff
@@ -236,7 +243,10 @@ module.exports = {
         mode: argv.mode,
         mtime
       })
-      : getStdin() // Pipe directly to ipfs.add
+      : tagWithMetadata(getStdin(), {
+        mode: argv.mode,
+        mtime
+      }) // Pipe to ipfs.add tagging with mode and mtime
```
alanshaw: Would this work, without having to use `normaliseAddInput`?:

```suggestion
      : {
        content: getStdin(),
        mode: argv.mode,
        mtime
      } // Pipe to ipfs.add tagging with mode and mtime
```

kanej: Just gave it a go, but it seems `ipfs.add` is expecting an iterator:

> TypeError: Cannot read property 'Symbol(Symbol.asyncIterator)' of undefined

We are able to get the job done if we turn the object into an async iterator

```javascript
: (function * () {
   yield {
     content: getStdin(),
     mode: argv.mode,
     mtime
   }
 })()
```

Is there a iterator util for turning an element into a async iterator?

alanshaw: Weird, that should be supported. You should also be able to do:

```javascript
: [{
     content: getStdin(),
     mode: argv.mode,
     mtime
   }]
```

kanej: That worked. Should I open a bug for the add function when passed:

```javascript
{
   content: getStdin(),
   mode: argv.mode,
   mtime
}
```

achingbrain: I think this is because the test stubs `ipfs.add` to only return something when passed an iterable (async or otherwise) as the first argument:

```javascript
ipfs.add.withArgs(matchIterable(), defaultAddArgs()).returns([{
  cid,
  path: 'readme'
}])
```

If you pass:

```javascript
{
   content: getStdin(),
   mode: argv.mode,
   mtime
}
```

You're not passing an iterable so the stub doesn't match and `undefined` is returned from `ipfs.add` because JavaScript which causes the reported `TypeError` in the `for await..of` loop.

The test needs to be updated to this:

```javascript
it('add from pipe with mtime=100', async () => {
  const cid = new CID('QmPZ9gcCEpqKTo6aq61g2nXGUhM4iCL3ewB6LDXZCtioEB')

  ipfs.add.withArgs(sinon.match({
    content: matchIterable(),
    mtime: { secs: 100 }
  }), defaultAddArgs()).returns([{
    cid,
    path: 'readme'
  }])

  const proc = cli('add --mtime=100', {
    ipfs,
    getStdin: function * () {
      yield Buffer.from('hello\n')
    }
  })

  const out = await proc
  expect(out).to.equal(`added ${cid} ${cid}\n`)

  const input = ipfs.add.getCall(0).args[0]
  expect(input).to.have.nested.property('mtime.secs', 100)
})
```

achingbrain: TBH it could even lose the last assertion (for the `mtime.secs` property) as it'll only fail if the `ipfs.add` stub doesn't match, which will cause the `TypeError` issue before it gets to that assertion.

kanej: @achingbrain I have updated the usage and test as suggested. The mocking had to be changed in 'add from pipe' as well.

achingbrain:
Please could you rebase this on top of master?  Then once CI is green it should be good to go.

kanej:
@achingbrain I have rebased off of master. The build has failed on 'interop - electron-renderer', which I suspect is a build gremlin, can you kick off another build?

Issue #2727, opened 2020-01-23
Use mafmt module to only append peer ids to certain multiaddrs


achingbrain:
As part of startup we [append our peer id](https://github.com/ipfs/js-ipfs/blob/f8862bc24761b6704d435e7fb6ba9b9d5fedb695/src/core/components/pre-start.js#L65) to every multiaddr that contains a peer id.

This should only happen to `circuit` addresses.  We should use the [`mafmt`](https://www.npmjs.com/package/mafmt) module to match on those.

PR #2833, opened 2020-03-06
fix: multiaddr validation to add peer id for listening

vasco-santos:
When we add the own peer id to a multiaddr, we should verify if it already has it, or if it is an address of `circuit-relay` / `stardust`. Otherwise, we might end up with multiaddrs like:  `/ip4/127.0.0.1/tcp/5892/ws/p2p/PEER_ID/p2p/PEER_ID`

### DIFF
--- a/packages/ipfs/src/core/components/start.js
+++ b/packages/ipfs/src/core/components/start.js
@@ -39,6 +39,9 @@ module.exports = ({
       config.Addresses.Swarm.forEach(addr => {
         let ma = multiaddr(addr)
 
+        // multiaddrs that go via a signalling server or other intermediary (e.g. stardust,
+        // webrtc-star) can have the intermediary's peer ID in the address, so append our
+        // peer ID to the end of it
         const maId = ma.getPeerId()
         if (maId && maId !== peerInfo.id.toB58String()) {
           ma = ma.encapsulate(`/p2p/${peerInfo.id.toB58String()}`)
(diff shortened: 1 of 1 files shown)


achingbrain:
Do you think this would be better done by only appending the peer id to certain types of multiaddrs, a la https://github.com/ipfs/js-ipfs/issues/2727 ?

vasco-santos:
I haven't considered that since I did not know about the issue, but this is a good question!

This also needs to be considered for `stardust`, and possibly others in the future. This might make the validation of a growing list of different multiaddrs, which is my biggest concerns.

FYI, we currently consider multiaddr without `p2p` as valid: https://github.com/multiformats/js-mafmt/blob/master/src/index.js#L77-L96

packages/ipfs/src/core/components/start.js
```diff
@@ -39,7 +39,8 @@ module.exports = ({
       config.Addresses.Swarm.forEach(addr => {
         let ma = multiaddr(addr)
 
-        if (ma.getPeerId()) {
+        const maId = ma.getPeerId()
```
achingbrain:
```suggestion
        // multiaddrs that go via a signalling server or other intermediary (e.g. stardust,
        // webrtc-star) can have the intermediary's peer ID in the address, so append our
        // peer ID to the end of it
        const maId = ma.getPeerId()
```

PR #2826, opened 2020-03-05
docs: recommend jsdelivr to users and add download counts to the README

daviddias:
This PR:
- Adds download counts
- Recommend jsdelivr (faster CDN) and gives us the download counts

PR #2914, opened 2020-03-11
fix: reinstate hlsjs-ipfs-loader dep for video streaming example

achingbrain:
Otherwise the test times out. We have to specify the dist file in the url too, otherwise the test times out, not 100% sure why.

Follows on from #2826


achingbrain:
I think it's missing a browser entry in the package.json or something..

PR #2915, opened 2020-03-11
chore: only run jobs in packages that have changed since master

achingbrain:
Should make PR builds much faster. Might have to be a bit smarter about coverage reports over time though.


hugomrdias:
Can you explain further this change, maybe an example.
Doesn't lerna run tests only in packages with changed files?

achingbrain:
> Doesn't lerna run tests only in packages with changed files?

Sort of, if you don't specify [`--since [ref]`](https://github.com/lerna/lerna/blob/92f3c6691b8dd87227e85acdefef66662d49340d/core/filter-options/README.md#--since-ref), it'll run scripts in everything that's changed since the last release.

For example #2914 only changed an example, which nothing depends on 
so only the tests for the example should have run, yet in [the build for that PR](https://travis-ci.com/github/ipfs/js-ipfs/builds/152764913) it ran everything - interop, interface tests, the lot.

When specifying `--since` to `lerna run` it does still work out the dependency graph and run scripts in packages that haven't changed but are dependant on ones that have.

Issue #2924, opened 2020-03-13
Track: remove usage of node globals


hugomrdias:
Bundlers are starting to remove automatic polyfilling of node globals so we need to keep up.

- [x] multibase version: `v0.6.1` size: `8.92KB -> 8.9KB` https://github.com/multiformats/js-multibase/pull/51
- [x] multiaddr version: `7.4.1` size: `26.36KB -> 23.95KB` https://github.com/multiformats/js-multiaddr/pull/113, https://github.com/multiformats/js-multiaddr/pull/115
- [x] cid version `0.7.4` https://github.com/multiformats/js-cid/pull/104
- [x] multicodec version: `1.0.1` https://github.com/multiformats/js-multicodec/pull/58
- [x] protons version `1.0.2` https://github.com/ipfs/protons/pull/11
- [x] multihashing-async https://github.com/multiformats/js-multihashing-async/pull/66
- [x] multihash version: `0.4.16` https://github.com/multiformats/js-multihash/pull/71
  - https://github.com/multiformats/js-multihash/pull/72
  - https://github.com/multiformats/js-multihash/issues/73
- [x] libp2p-crypto-secp256k1 version `v0.4.2` https://github.com/libp2p/js-libp2p-crypto-secp256k1/pull/25
- [x] libp2p-crypto size: `155KB -> 122.65KB` https://github.com/libp2p/js-libp2p-crypto/pull/170
  - `node: false` working 🎉
- [x] bl https://github.com/rvagg/bl/pull/81
- [x] it-tar version `1.2.2` https://github.com/alanshaw/it-tar/pull/2
- [x] alanshaw/it-length-prefixed version `3.0.1` https://github.com/alanshaw/it-length-prefixed/pull/9
- [x] js-libp2p-mplex version `v0.9.5` https://github.com/libp2p/js-libp2p-mplex/pull/106
- [x] peer-id version `v0.13.10` needs libp2p-crypto https://github.com/libp2p/js-peer-id/pull/120
- [x] it-* https://github.com/achingbrain/it/pull/6
- [x] it-pb-rpc https://github.com/mkg20001/it-pb-rpc/pull/7
- [x] libp2p-secio https://github.com/libp2p/js-libp2p-secio/pull/118
- [x] https://github.com/mkg20001/it-buffer/pull/1
- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/39
- [x] https://github.com/libp2p/js-libp2p-websockets
[...]

PR #2932, opened 2020-03-20
fix: remove node globals

hugomrdias:
This PR main goal is to remove node globals and is the final PR of this awesome endeavour https://github.com/ipfs/js-ipfs/issues/2924

I tried to split the commits to make it easier to digest https://github.com/ipfs/js-ipfs/pull/2932/commits

Running `aegir test -t browser --node false` still fails in a couple of tests because im still waiting on merges in `readable-stream` and `simple-peer` but its ready for review.

todo: 
- write an example for using other ipld formats


hugomrdias:
looks like libp2p-crypto is working

packages/interface-ipfs-core/src/add.js
```diff
@@ -467,12 +470,12 @@ module.exports = (common, options) => {
       expect(expectedResult.err).to.not.exist()
       expect(result[0].cid.toString()).to.equal(expectedResult[0].cid.toString())
       expect(result[0].size).to.equal(expectedResult[0].size)
-      expect(result[0].path).to.equal(text)
+      // expect(result[0].path).to.equal(text)
     })
 
     it('should add from a HTTP URL with redirection', async () => {
       const text = `TEST${Math.random()}`
-      const url = echoUrl(text) + '?foo=bar#buzz'
```
achingbrain, line 475: Why did the extra stuff at the end of the url get removed?

packages/interface-ipfs-core/src/add.js
```diff
@@ -483,7 +486,7 @@ module.exports = (common, options) => {
       expect(expectedResult.err).to.not.exist()
       expect(result[0].cid.toString()).to.equal(expectedResult[0].cid.toString())
       expect(result[0].size).to.equal(expectedResult[0].size)
-      expect(result[0].path).to.equal(text)
+      // expect(result[0].path).to.equal(text)
```
achingbrain, line 489: Should this be uncommented?

packages/interface-ipfs-core/src/add.js
```diff
@@ -499,12 +502,12 @@ module.exports = (common, options) => {
...
     it('should add from a URL with wrap-with-directory=true', async () => {
       const filename = `TEST${Date.now()}.txt` // also acts as data
-      const url = echoUrl(filename) + '?foo=bar#buzz'
+      const url = echoUrl(filename)
       const addOpts = { wrapWithDirectory: true }
 
       const [result, expectedResult] = await Promise.all([
         all(ipfs.add(urlSource(url), addOpts)),
-        all(ipfs.add([{ path: filename, content: Buffer.from(filename) }], addOpts))
+        all(ipfs.add([{ path: 'download', content: Buffer.from(filename) }], addOpts))
```
achingbrain, line 510: Why has `filename` changed to `'download'`?

packages/interface-ipfs-core/src/files/write.js
```diff
@@ -174,21 +169,17 @@ module.exports = (common, options) => {
     })
 
     it('writes a small file using a Node stream (Node only)', async function () {
-      if (!isNode) {
-        return this.skip()
-      }
-
+      if (!isNode) this.skip()
```
achingbrain, line 172: Was changing this to a one-line if necessary?  I can't help but think of [this bug](https://www.imperialviolet.org/2014/02/22/applebug.html) every time I see a one-line if statement.

packages/ipfs/package.json
```diff
@@ -40,21 +36,21 @@
...
+    "test:electron": "aegir test -t electron-main  -t electron-renderer",
+    "test:electron-main": "aegir test -t electron-main",
+    "test:electron-renderer": "aegir test -t electron-renderer",
+    "test:cli": "aegir test -t node -f test/cli/index.js",
+    "test:node:core": "aegir test -t node -f test/core/**/*.js",
+    "test:node:http": "aegir test -t node -f test/http-api/index.js",
+    "test:node:gateway": "aegir test -t node -f test/gateway/index.js",
+    "test:interface": "aegir test -f test/core/interface.spec.js",
+    "test:interface:http": "aegir test -f test/http-api/interface.js",
+    "test:bootstrapers": "IPFS_TEST=bootstrapers aegir test -t browser -f test/bootstrapers.js",
```
achingbrain, line 53: Much better!

packages/ipfs/test/core/preload.spec.js
```diff
@@ -2,9 +2,10 @@
 'use strict'
 
 const { nanoid } = require('nanoid')
+const { Buffer } = require('buffer')
 const { expect } = require('interface-ipfs-core/src/utils/mocha')
 const all = require('it-all')
-const MockPreloadNode = require('../utils/mock-preload-node')
+const MockPreloadNode = require('../utils/mock-preload-node-utils')
```
achingbrain, line 8: Should this not be `const MockPreloadNode = require('../utils/mock-preload-node')` ?

packages/interface-ipfs-core/src/files/write.js
hugomrdias: ofc not... changing it back

packages/ipfs/test/core/preload.spec.js
hugomrdias: i split the server from the helpers so we don't include the server stuff in the tests bundle

packages/interface-ipfs-core/src/add.js
hugomrdias: this has to do with the echoserver download endpoint, the previous server was simpler and one could do just any path and it would respond, but the new has defined endpoints for different stuff.
So urlSource uses the pathname of the url for the path (http://localhost:3000/download?data=helloworld) we get "download" in file path, i could do a catch-all endpoint just for this use-case so we can build the file path with the url pathname but i didn’t see the value because the test still is testing whats needed so i left it like that.

packages/interface-ipfs-core/src/add.js
hugomrdias: its not needed and the new echoserver uses searchParams, i will remove the comments

packages/interface-ipfs-core/src/add.js
hugomrdias: same reason as below, path is `download` now and not the text, i will remove the comment

PR #2939, opened 2020-03-23
docs(ipfs.files.ls): type is a number

zebateira:
Each object contains a property `type` which is a `number`, not a string.
0-file
1-directory


achingbrain:
Fixed in #3352

achingbrain:
Thanks for opening this but the fix is to have it return strings consistent with `ipfs.files.stat`

zebateira:
@achingbrain no worries, thank you for getting back to this 🙏

PR #2949, opened 2020-03-25
feat: allow `--debug`

bluelovers:
https://github.com/visionmedia/debug#set-dynamically

### DIFF
--- a/packages/ipfs/src/cli/bin.js
+++ b/packages/ipfs/src/cli/bin.js
@@ -48,6 +48,11 @@ async function main () {
 
   try {
     const data = await cli(command, async (argv) => {
+
+      if (argv.debug) {
+        debug.enable('ipfs:cli')
+      }
+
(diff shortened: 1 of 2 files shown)


achingbrain:
There is no need for this, you can just do:

```console
$ DEBUG=ipfs* jsipfs add ./somefile.txt
```

bluelovers:
i think use `--debug` is more easy

why **only** allow user debug when set env var?

achingbrain:
`--debug` only sets `DEBUG=ipfs:cli` which doesn't help with diagnosing problems with libp2p or other system components.

The nice thing about `DEBUG=ipfs*` is it presents a uniform way of getting debug logs out of IPFS in the browser, on the CLI, within your app, etc - I don't think adding special-case code to the CLI is the way forward.

bluelovers:
@achingbrain  it just need update code for keep it
https://github.com/ipfs/js-ipfs/pull/2967

Issue #2745, opened 2020-02-05
feat: import/export a DAG from/to a CAR file


mikeal:
Across IPLD and Filecoin we’re starting to really zero in on CAR files being the preferred import/export format for DAGs.

It would be great if IPFS had an easy way to export the full graph for any CID to a CAR file and to load a CAR file’s blocks into its local blockstore.

It’s not very hard. I was able to write a proof of concept in less than an hour for the exporter https://github.com/mikeal/export-ipld-graph

achingbrain:
This might be better as a separate command?

```console
$ cat ./path-to-input.car | read-car | ipfs add
```

```console
$ ipfs get $CID | write-car > ./path-to-output.car
```

In the API a `carSource` module similar to [`globSource`](https://github.com/ipfs/js-ipfs-http-client/blob/f30031163b9ac4ce2cff34ad4854f24b23cbff0b/README.md#glob-source) or [`urlSource`](https://github.com/ipfs/js-ipfs-http-client/blob/f30031163b9ac4ce2cff34ad4854f24b23cbff0b/README.md#url-source) could be created?

```javascript
for await (const entry of ipfs.add(carSource('path or stream'))) {
  console.log(entry)
}
```

Then for outputting, use something like:

```javascript
await carWriter('output-path', ipfs.get(carCid))
```

mikeal:
The car file contains all the blocks for the entire graph, so you have to traverse through all of the links and it’s not quite as simple as this. It could possibly still be a separate command but you’d need to output in a multi-part structure that distinguishes the blocks, which is basically a car file.

The car file library shouldn’t be too large, the POC for export may look large because the default codecs and hash functions used in the new Block API, but js-ipfs wouldn’t need to use that and already includes relevant codecs and hashing functions.

rvagg:
proposal / WIP @ #2953 to match what go-ipfs is getting

ribasushi:
Just a poke on this issue: what we are shipping in go-ipfs 0.5 is the final version. It is missing only:
- Import of zero-root `.car` files (discussion: https://github.com/ipld/go-car/issues/26#issue-588032760)
- Progress events on import (no functional changes)

The [test fixture](https://github.com/ipfs/go-ipfs/tree/master/test/sharness/t0054-dag-car-import-export-data) and the [tests themselves](https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0054-dag-car-import-export.sh#L70-L138) are complete. 

[...]

PR #2953, opened 2020-03-26
(WIP) feat: dag import and export to and from CAR files

rvagg:
Closes: #2745
Ref: https://github.com/ipfs/go-ipfs/pull/7011
Ref: https://github.com/ipfs/go-ipfs/issues/6870

_Disclaimer: this is my first hack directly on this repo, so I've had a steep learning curve today figuring this out and I wouldn't be surprised if I have some things very wrong! Guidance gratefully accepted, or perhaps someone else would like to take ownership of this?_

I thought it would be nice to widen the conversation about the new `ipfs dag export` and `ipfs dag import` commands being added in go-ipfs. This is approximate parity, minus some minor details (mostly noted in the code). No tests here, it would be good to share the same fixture suite as @ribasushi is building for go-car.

Because [datastore-car](https://github.com/ipld/js-datastore-car) uses the newer [`@ipld/block`](https://github.com/ipld/js-block), that gets imported here (export uses a different `Block` than import). That creates a bit of space to have duplicated dependencies so that'd be worth checking on and trying to minimise.

Notes:

* `export` supports only a single root CID and applies a full DAG walk to it. In the future it's expected that you'll also apply a selector to it (default would be like a `*`) as well and CARv2 would even store that selector with the root. It produces a "well-formed" CAR, that's deterministic in theory. We need to test the limits of that determinism but assuming we're walking the graph in the same order then it should be identical each time you run it with the same root in js-ipfs or go-ipfs.
* `import` is lax, by design, it accepts a CAR as a "bundle of blocks", perhaps with no root, perhaps with many roots, perhaps with roots that don't even exist in the CAR body. It just dumps the blocks into your store. Where there's a root specified that's found in the body, it'll pin that root in your store. I suspect some details here will be refined in go-car and need to be synced here. For one: go-car currently doesn't accept zero-root CAR files but JS does.

Examples:

export:

```
$ ./src/cli/bin.js dag export QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm > xkcd.car
```

Will result in a 108M file named xkcd.car containing a Unixfs mirror of XKCD from some point in time.

import:

```
$ ./src/cli/bin.js dag import xkcd.car
importing from xkcd.car...
pinned QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm
imported 7342 blocks
```

lazy simulation of multi file import:

```
$ ./src/cli/bin.js dag import xkcd.car xkcd.car
importing from xkcd.car...
pinned root QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm
importing from xkcd.car...
pinned root QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm
imported 14684 blocks
```

import from stdin:

```
$ cat xkcd.car | ./src/cli/bin.js dag import
importing from stdin...
pinned root QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm
imported 7342 blocks
```


ribasushi:
Regarding the xkcd archive I get the following in go-ipfs:
```
cmd/ipfs/ipfs dag export QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm | sha256sum 
Exported .car size:	112277914
5a49df60c4842e7fe71849410025a8a5a98e2d0e6e71f90b4913635f7df7250d  -
```

Would be nice to double-check we converge bit-to-bit

rvagg:
👍 

```
$ shasum -a 256 xkcd.car
5a49df60c4842e7fe71849410025a8a5a98e2d0e6e71f90b4913635f7df7250d  xkcd.car
$ ls -al xkcd.car
-rw-r--r-- 1 rvagg staff 112277914 Mar 26 20:55 xkcd.car
```

rvagg:
`export` and `import` got merged into go-ipfs today: https://github.com/ipfs/go-ipfs/pull/7036 & https://github.com/ipfs/go-ipfs/pull/7038

Can someone in charge of js-ipfs tell me whether it's worth pursuing this PR further to match functionality (I think it's pretty close) and reproduce the excellent tests that Peter has included over there? There's objections in #2745 about code bloat and I don't want to waste time on something that might be rejected on that basis.

@achingbrain maybe?

rvagg:
I've updated datastore-car upstream to handle this, and I've imported the sharness test from go-ipfs that covers this but it seems like we're not set up to run sharness fully compatible with go-ipfs. `iptb` seems to be missing and I'm not sure how far down this rabbit hole I should try and go because I have zero idea how deep it might be. @achingbrain or someone else, can you background me one sharness and the compatibility between here and go-ipfs? Is it a subset here for a reason, is it just a matter of fast-forwarding some functionality to match, or is it not possible, or not reasonable to get compatibility? Maybe I need to rewrite the tests entirely in JS to avoid sharness?

packages/ipfs/test/sharness/t0054-dag-car-import-export.sh
```diff
@@ -0,0 +1,204 @@
...
+          pipe_testnet \
+          pipe_devnet \
+          ../t0054-dag-car-import-export-data/combined_naked_roots_genesis_and_128.car \
+        > basic_fifo_import_actual
+        result=$?
+
+        wait || true	# work around possible trigger of a bash bug on overloaded circleci
+        exit "$result"
+    )
+  '
```
ribasushi, line 130: @rvagg the exercise of the GC-lock and the import of FIFOs may not be something you are too inteerested in testing within js-ipfs. Just raising it here as it was a very important part of making 🗡️ viable.

packages/ipfs/src/cli/commands/dag/import.js
```diff
@@ -0,0 +1,73 @@
...
+      rootStatus[key] = true
+    }
+    count++
+  }
+
+  // A well-formed CAR will have roots that reference blocks that actually exist
+  // in the CAR body, but that may not be the case. Don't try to pin if the root
+  // didn't exist in the body.
+  // TODO: ^ go-car currently attempts to pin roots even if they don't exist in
+  // the CAR body, need to align behaviour
```
ribasushi, line 58: @rvagg this is done in `go-ipfs` to allow a copy-less "transactional" operation:
```
<some source> | stream-dagger <many options> --emit-stdout=car-v0-fifos-xargs | xargs -0 ipfs dag import
```
What that mode does is print 2 fifo names on stdout. The first fifo contains all the data. The second contains the roots only ( because we can derive the roots only once we streamed everything over ). The full `dag import` context serves as a "transaction" of sorts, keeping GC at bay between the lengthy data stream and the pin at the very end.

achingbrain:
> Maybe I need to rewrite the tests entirely in JS to avoid sharness?

Yes please - sorry for the misdirection. The sharness tests here aren't run and should really be deleted.

The testing strategy is (I really must put this in a doc):

### CLI

Tests live in [/packages/ipfs/test/cli](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs/test/cli).

All interactions with IPFS core are stubbed so we just ensure that the correct arguments are passed in

### HTTP API

Tests live in [/packages/ipfs/test/http-api](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs/test/http-api) and are similar to the CLI tests in that we stub out core interactions and inject requests with [shot](https://www.npmjs.com/package/@hapi/shot).

### Core

Anything non-implementation specific should be considered part of the 'Core APIs'.  For example node setup code is not Core, but anything that does useful work, e.g. network/repo/etc interactions would be Core.

All Core APIs should be documented in [/docs/core-api](https://github.com/ipfs/js-ipfs/tree/master/docs/core-api).

All Core APIs should have comprehensive tests in [/packages/interface-ipfs-core](https://github.com/ipfs/js-ipfs/tree/master/packages/interface-ipfs-core).

`interface-ipfs-core` should ensure API compatibility across implementations. Tests are run:

1. Against [/packages/ipfs/src/core](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs/src/core) directly
1. Against [/packages/ipfs/src/http](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs/src/http) over HTTP via `ipfs-http-client`
1. Against `go-ipfs` over HTTP via `ipfs-http-client`

### Non-Core

Any non-core API functionality is tested in [/packages/ipfs-http-api/tests](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-http-client/test) and [/packages/ipfs/tests](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs/test) for `ipfs-http-client` and `ipfs` respectively.

packages/ipfs/src/cli/commands/dag/import.js
mikeal: i’d wait to support this until a later date if `js-ipfs` wants to prioritize it.

jacobheun:
What's the status of this? Do we intend to finish it?

rvagg:
We have a replacement CAR library now, https://github.com/ipld/js-car, so that needs to be integrated here. It uses the new js-multiformats though, so it'll start bringing in some new stack pieces.

Is this work needed though? I don't have a feel for whether this kind of parity with go-ipfs is even the goal these days?

rvagg:
had to start this again because so much has moved on, both here and in ipld/multiformats

PR #2963, opened 2020-03-31
chore: be explicit about running interface tests

achingbrain:
Instead of running unit, interface, interface over http, etc tests in each build step, split the running of interface tests out into their own commands and run them independently on travis, hopefully the build should be a bit faster.


hugomrdias:
the files should still have the .spec.js in  the name or else the browser run will ignore them

achingbrain:
That's kind of the point - I don't want the interface tests to run unless I specify the setup file with the `-f` argument.

hugomrdias:
so running `npm test` or aegir release will not run interface tests ?

achingbrain:
`npm test` will run unit tests, our interface tests are more like integration tests anyway.

TBH we have so many test suites and so many target environments it's not really feasible to run anything other than a subset of these on your laptop as smoke tests.

The total CI time when you take running the tests on two flavours of node, chrome, firefox, their webworkers, electron main, electron renderer all on Mac OS, Linux and Windows, interop tests, interface tests to core, over http, over http to go-ipfs, over http in node, over http in firefox, chrome, etc, etc, etc is something like six hours.  We've parallelised that in CI to bring it down to 45 minutes or so, this PR brings it down to about 30.  If we can speed up the interop tests that'll fall further.

On the `aegir release` thing, we're using `lerna` for releases which doesn't run the tests but we don't do releases unless we have green CI and you are on the head of master.

Really, we want to do releases from CI at the end of the pipeline.  That the tests pass on a developer's machine doesn't prove much - how many times have you said 'well it works locally'?  I know I've said it a lot.

achingbrain:
We could make `npm test` in the repo root run unit + interface-core/interface-http tests?

PR #2965, opened 2020-04-02
fix: dht.findPeer API endpoint returns ndjson

achingbrain:
It doesn't appear to be [documented](https://docs.ipfs.io/reference/api/http/#response-30), but `POST /api/v0/dht/findpeer` can return multiple results, so treat it as `ndjson`.

Otherwise we see these sorts of test failures:

```
1) DelegatedPeerRouting
       findPeers
         should be able to find peers via the delegate with a peer id string:
     FetchError: invalid json response body at http://127.0.0.1:42421/api/v0/dht/findpeer?search-params=%5Bobject+Object%5D&arg=QmcvRwsBkiEfLFPC6NAXmbpt1f9oKDT9dmmgAcEsGd9ZuW reason: Unexpected token { in JSON at position 93
```

### DIFF
--- a/packages/ipfs-http-client/src/dht/find-peer.js
+++ b/packages/ipfs-http-client/src/dht/find-peer.js
@@ -9,23 +9,23 @@ module.exports = configure(api => {
   return async function findPeer (peerId, options = {}) {
     options.arg = `${Buffer.isBuffer(peerId) ? new CID(peerId) : peerId}`
 
-    const res = await api.post('dht/findpeer', {
+    const res = await api.ndjson('dht/findpeer', {
       timeout: options.timeout,
       signal: options.signal,
       searchParams: options
     })
 
-    const data = await res.json()
+    for await (const data of res) {
+      if (data.Type === 3) {
+        throw new Error(data.Extra)
+      }
 
-    if (data.Type === 3) {
-      throw new Error(data.Extra)
-    }
-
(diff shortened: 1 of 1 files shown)


packages/ipfs-http-client/src/dht/find-peer.js
```diff
@@ -9,23 +9,23 @@ module.exports = configure(api => {
   return async function findPeer (peerId, options = {}) {
     options.arg = `${Buffer.isBuffer(peerId) ? new CID(peerId) : peerId}`
 
-    const res = await api.post('dht/findpeer', {
+    const res = await api.ndjson('dht/findpeer', {
```
jacobheun, line 12: Since this was changed from a `.post` it needs to include `method: 'POST'` in the options. 
See https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/dht/find-provs.js#L10-L15

achingbrain: Good point - I think @lidel was going to PR the client to change all methods to post.

achingbrain: I can do that now though..

achingbrain: Taking shape here: https://github.com/ipfs/js-ipfs/pull/2977

PR #2967, opened 2020-04-02
Pr/debug 001

bluelovers:
https://github.com/ipfs/js-ipfs/pull/2949


achingbrain:
Thanks for opening this PR, but how is it different to #2949?

bluelovers:
because u close it, it can't get new commit

achingbrain:
Thank you for your PR but as discussed in #2949 this isn't how we want to go with enabling debug output.

PR #2968, opened 2020-04-02
docs: fix ipfs.cat example

Joncom:
Old example was printing an AsyncGenerator object to the console

Updated code comes straight from https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#cat


achingbrain:
Thanks!

PR #2972, opened 2020-04-07
Refactor refs tests

koivunej:
Minor changes to refs to stop using `ipfs.add` and `ipfs.object.put` in favor of `ipfs.block.put` with test-local formatting. Tested against `go-ipfs-0.4.23` the CIDs do not change. This is result of discussions at https://github.com/ipfs-rust/rust-ipfs/issues/92#issuecomment-602686606 and at least another PR for `refs-local.js` should be coming later.

Sorry for the semicolons, I don't know the semicolon insertion rules by heart :)

### DIFF
--- a/packages/interface-ipfs-core/src/refs.js
+++ b/packages/interface-ipfs-core/src/refs.js
@@ -10,6 +10,8 @@ const dagPB = require('ipld-dag-pb')
 const DAGNode = dagPB.DAGNode
 const DAGLink = dagPB.DAGLink
 
+const UnixFS = require('ipfs-unixfs')
+
 /** @typedef { import("ipfsd-ctl/src/factory") } Factory */
 /**
  * @param {Factory} common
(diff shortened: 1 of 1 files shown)


koivunej:
Is the `aegir lint` documented somewhere? For me post `npm install -g aegir` it'll just:

```
$ aegir lint
.../.npm-packages/lib/node_modules/aegir/node_modules/yargs/yargs.js:1242
      else throw err
           ^

Error: Failed to load config "standard" to extend from.
Referenced from: BaseConfig
```

The same happens if I try to run `lerna run lint "--since" "master"` as executed by CI.

achingbrain:
If you run `npm run lint` in the root of the repo it will print out all the linting errors.  No need to install aegir globally.

koivunej:
The build is now failing for a dht test failure, something failed with chrome but the related tests passed, interop failure looks transient (failed between two go-ipfs nodes).

koivunej:
@achingbrain thanks for the previous tip of `npm run lint`; I had only tried `npm lint` and then decided to go ahead with aegir but your hint made me realized I had not executed `npm install` so there's that :)

PR #2977, opened 2020-04-09
fix: make http api only accept POST requests

achingbrain:
This is to prevent people maliciously controlling your local node by injecting images into webpages with URLs of API endpoints.

BREAKING CHANGE:

Where we used to accept all and any HTTP methods, now only POST is accepted.  The API client will now only send POST requests too.


jacobheun:
There are still several calls to `api.ndjson` that don't include `method: 'POST'`. Hopefully the new tests will show those failures.

achingbrain:
`http.ndjson` calls `http.iterator` which calls `http.fetch` which is overridden to only send `POST` requests so it *should* be handled.

jacobheun:
> http.ndjson calls http.iterator which calls http.fetch which is overridden to only send POST requests so it should be handled.

I think it would be helpful to just remove the method specification in the existing calls then to clean things up. Having most of them with the declaration and some without is confusing, and if it's already doing a post we should just remove the extraneous code.

achingbrain:
I agree - I'd much rather have something like:

```javascript
const res = await api.post(...)

for await (const thing of res.ndjson()) {
  ...
}
```

I suggested having a non-general-purpose http client for this sort of thing here: https://github.com/ipfs/js-ipfs-utils/issues/32

achingbrain:
@jacobheun: https://github.com/ipfs/js-ipfs-utils/pull/35

packages/ipfs/test/http-api/inject/bootstrap.js
```diff
@@ -13,77 +13,118 @@ module.exports = (http) => {
...
-        method: 'GET',
-        url: '/api/v0/bootstrap/list'
+    describe('/list', () => {
+      it('only accepts POST', async () => {
+        const res = await api.inject({
+          method: 'GET',
+          url: '/api/v0/bootstrap/list'
+        })
+
+        expect(res.statusCode).to.equal(404)
```
lidel, line 28: This is not interop test, but for what its worth, go-ipfs 0.5.0-rc1 returns [`405`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405):

```console
$ curl -I -X GET http://127.0.0.1:5001/api/v0/id
HTTP/1.1 405 Method Not Allowed
Allow: HEAD
Allow: OPTIONS
Allow: POST
Content-Type: text/plain; charset=utf-8
Vary: Origin
X-Content-Type-Options: nosniff
Date: Thu, 09 Apr 2020 23:48:22 GMT
Content-Length: 25
```

achingbrain: That's the correct code but Hapi does not make this easy: 

https://github.com/hapijs/hapi/issues/1534
https://github.com/hapijs/hapi/issues/3238

PR #2980, opened 2020-04-10
Refactor refs-local to not use ipfs.add

aphelionz:
This PR refactors `ipfs.add` away from the refs-local test, relying instead on the more basic `block.put`. This is based on the need for `ipfs-rust/rust-ipfs` to work through the conformance testing in the order planned in the dev grant.

Basing this work off of the techniques used in #2972

PR #2982, opened 2020-04-14
test: make refs tests iteration order independent

koivunej:
Proposing here a change which will make the `refs.js` tests dag node iteration order independent by sorting the expected and actual refs list. In the soon-to-be polished rust-ipfs refs implementation the refs are iterated breadth-first. Looking at the js-ipfs implementation I am not 100% sure but looks like the implementation is recursive depth-first (I could be wrong).

The iteration orders differ on almost half of the tests, but only big difference is with the test modified on the second commit: recursive+unique test with formatting `<linkname>` where a document has two names and breadth-first and depth-first offer the different names as a result.

The proposed changes are similar to what is already done for the dag-cbor nodes (sorted).

### DIFF
--- a/packages/interface-ipfs-core/src/refs.js
+++ b/packages/interface-ipfs-core/src/refs.js
@@ -199,19 +199,19 @@ function getRefsTests () {
     },
 
     'should get refs with recursive and unique option': {
-      params: { format: '<linkname>', recursive: true, unique: true },
+      params: { format: '<dst>', recursive: true, unique: true },
       expected: [
-        'animals',
-        'land',
-        'african.txt',
-        'americas.txt',
-        'australian.txt',
-        'sea',
-        'atlantic.txt',
-        'indian.txt',
-        'fruits',
-        'tropical.txt',
-        'mushroom.txt'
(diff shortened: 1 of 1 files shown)


koivunej:
I think the travis failure is transient. This time I did run `npm run lint` for sure :)

koivunej:
@hugomrdias this is continuation to #2972 and Marks #2980 but this change, while small in lines changed is somewhat larger on another level, the iteration order. I couldn't find any specs for `api/v0/refs` that would had locked in any iteration order, so when implementing the rust-ipfs one I just went with the breadth-first one which is more familiar for me as I always try to stay clear of recursion.

achingbrain:
Master had a temporary GH url for one of the dependencies - the branch the url referenced has since been merged & deleted.

The GH url has since been removed from master so this PR just needs rebasing for CI to run.

koivunej:
Test failures look unrelated but similar to last time.

aphelionz:
Hi @achingbrain just pinging on this - is the CI something on our end or is it just a flaky run?

Issue #2971, opened 2020-04-05
POST-only HTTP API


lidel:
### Problem

go-ipfs 0.5  will block `GET` commands on the API port (https://github.com/ipfs/go-ipfs/pull/7097), requiring every command to be sent as HTTP POST request (makes it a true RPC + solves some security issues).

### :mag:  Ensure `ipfs-http-client` works with API without `GET`

- [x] I think it does: interop tests run by go-ipfs CI passed, 
  but would appreciate second pair of eyes to confirm I did not miss anything
- [x] We should have tests that fail if GET is accepted by `/api/v0` on API port 
  unsure where to add those, any ideas?

###  :sweat_smile:  Ship a patch release for old JS API?

Blocking `GET` broke _Files_ screen in ipfs-webui as noted in https://github.com/ipfs-shipyard/ipfs-webui/issues/1429#issuecomment-609071306

ipfs-webui is using older version of `ipfs-http-client`, one before huge refactor into async iterables, which means switching to the latest version won't be a quick task. 

I suspect there are third party projects which use older version of `ipfs-http-client` but have no time to refactor entire codebase to work with the latest version. We could make it easier if we shipped a patch release with literally 7 line diff (`ky.get` → `ky.post`, as seen in seen in [`ipfs-http-client@39.0.2.patch`](https://github.com/ipfs-shipyard/ipfs-webui/blob/5d3c6b39c24728e38bec93612d9ee32e60973e65/patches/ipfs-http-client%2B39.0.2.patch))

I am unsure how to do this given we moved to monorepo recently.
I think v41.0.0 of `ipfs-http-client` was the last release before Async Await and Async Iterables.

@achingbrain is it possible to release something like `ipfs-http-client@41.0.1`? 
Any way I can help?

**Update:** `ipfs-http-client@41.0.1`  is ok for use (if you are stuck at old API)

bluelovers:
this?
https://github.com/lerna/lerna/tree/master/commands/version#--allow-branch-glob

achingbrain:
> I am unsure how to do this given we moved to monorepo recently.

The history from merged repos has been maintained so it should just be a case of making a branch from the last ipfs-http-client release commit, making the changes, doing the release and pushing the updated branch back to the repo.

lidel:
[...]

PR #2984, opened 2020-04-14
feat: webui v2.7.5, with feeling

lidel:
This PR is a bit bigger than expected, but I am running on robusta so completed some side quests:

- fix: `/webui` opened in browser returned HTTP 405 because browser sends GET and not a POST :-)
- `/webui` now loads from API port and works with POST-only RPC API
  - matching behavior of go-ipfs 0.5.0-rc1
  - switches to the latest [ipfs-webui v2.7.5](https://github.com/ipfs-shipyard/ipfs-webui/releases/tag/v2.7.5)
  - `/webui` is loaded from the same Origin as the RPC API itself
     - Why? right now it does not matter, but it will when we add subdomains and CORS hardening around `/api/v0` on API port. When that happens, pages loaded from Gateway won't have access to the API port without being whitelisted, and we want /webui work out of the box without the need for any manual configuration.
- UX nit: both `/webui` and `/webui/` work now

cc #2971


packages/ipfs/src/http/api/routes/webui.js
```diff
@@ -1,36 +1,31 @@
 'use strict'
 
-const multiaddr = require('multiaddr')
 const debug = require('debug')
+const { gateway } = require('../../gateway/resources')
 const log = debug('ipfs:webui:info')
 log.error = debug('ipfs:webui:error')
 
+const webuiCid = 'bafybeigxqbvc6qxk2wkdyzpkh7mr7zh5pxbvpjb6a6mxdtpwhlqaf4qj5a' // v2.7.4
```
lidel, line 8: @jessicaschilling  in the future, when we release new ipfs-webui we will need to update only this one line :crossed_fingers:

jessicaschilling: @lidel Is this the new version of the (broken link) js-ipfs file that appears here? https://github.com/ipfs-shipyard/ipfs-webui#releasing-a-new-version-of-the-webui

lidel: @jessicaschilling  Yes. I'll update that README when I finish the release dance for v2.7.4

achingbrain:
It would be really nice to get a smoke test for the webui - it could use [nightwatch](https://www.npmjs.com/package/nightwatch) to spin up a node, open a browser and just assert on the revision in the UI.

PR #2987, opened 2020-04-17
refactor: replace ipfs.add in dag.get tests

koivunej:
this is similar to other similar changes, we didn't start by implementing `ipfs.add` in `rust-ipfs`.

Related PR's: #2972, #2980, #2982, #2983


achingbrain:
This looks fine, but go-ipfs is missing a `version` parameter from the /block/put API so cannot add blocks with a V1 CID :(

We can generate the CID on the client and send it to the block API though - I've added a second commit to #3005 which does this so it would supersede this PR.

achingbrain:
Fixed in #3005

PR #2989, opened 2020-04-18
fix: allow either cidv{0,1} in refs-local test

koivunej:
Rationale: simplest way to support using cidv0 and cidv1 interchangeably in an ipfs implementation is to just convert all cids to cidv1 transparently. this has the miniscule setback of `refs/local` not being able to list cids in their original version. this fix allows the implementation to use either version internally.

I don't know all of the lore in the cidv0 -> cidv1 migration story but the `refs/local` is the only place I've seen there being an issue with doing the cidv0 -> cidv1 migration in the simplest way described above. Looking at the test I assumed the cidv0 string was expected because of "ease of test implementation" and not a specification of "any ipfs impl must be able to recover the original cid format for stored blocks in `refs/local` output."

I did think about opening up an issue of allowed cidv0 <-> cidv1 strategies but perhaps this PR is better suited for that, so please don't hesitate to let me know of the things I haven't found out yet! Could be that this approach has been tried already with bad results but I haven't found about it.

This might have a small conflict with #2980 so I'd recommend merging that first.

### DIFF
--- a/packages/interface-ipfs-core/src/refs-local.js
+++ b/packages/interface-ipfs-core/src/refs-local.js
@@ -7,6 +7,7 @@ const all = require('it-all')
 const importer = require('ipfs-unixfs-importer')
 const drain = require('it-drain')
 const testTimeout = require('./utils/test-timeout')
+const CID = require('cids')
 
 /** @typedef { import("ipfsd-ctl/src/factory") } Factory */
 /**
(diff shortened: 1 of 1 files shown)


koivunej:
Looking into getting this updated. Looks like I need to make the pinning conditional as the rust side doesn't have it yet.. It's a bit unobvious requirement which was probably hidden by previous use of `add` pinning by default?

packages/interface-ipfs-core/src/refs-local.js
```diff
@@ -39,15 +40,45 @@ module.exports = (common, options) => {
 
       const imported = await all(importer(dirs, ipfs.block))
 
-      // otherwise go-ipfs doesn't show them in the local refs
-      await Promise.all(
-        imported.map(i => ipfs.pin.add(i.cid))
-      )
+      // rust-ipfs doesn't yet have pinning api, it'll just list all local cids
+      // in /refs/local
+      if (common.opts.type !== 'rust') {
```
koivunej, line 45: Per https://github.com/ipfs/js-ipfsd-ctl/compare/master...ipfs-rust:add_rust_ipfs_http#diff-d4165b1e3527a04dd7e9d370e18c2a1aR47 (also published as draft PR to js-ipfsd-ctl) which we are using in https://github.com/ipfs-rust/ipfs-rust-conformance

aphelionz:
Thanks for looking at #2982 and re-running the tests. Can we try this again here? This and #2982 will be important in the next Rust IPFS phase - sorry for not mentioning this earlier on the call.

hugomrdias:
can you please update the branch ?

packages/interface-ipfs-core/src/refs-local.js
```diff
@@ -47,15 +48,46 @@ module.exports = (common, options) => {
...
+        // allow the cids to be either original or in later cid version.
+        let removed = false
+        for (const version of alt) {
+          const index = cids.indexOf(version)
+          if (index === -1) {
+            continue
+          }
+          removed = true
+          delete cids[index]
+          break
```
koivunej, line 79: Reading this again and noticed this was break was missing, added this while rebasing.

koivunej:
Test failures look transient and unrelated.

koivunej:
Updating as we have discussed this privately with @hugomrdias and/or @achingbrain. I understand the current situation with regards to this PR is that `rust-ipfs` should first start supporting the `pin/add ` so that the special casing could be thrown out. `pin/add` is still in progress but this should be ok to leave open until we have one, even a mocked API.

koivunej:
I am pretty sure this is good to close, #3124 already fixed this test case to check the multihashes instead of Cids.

koivunej:
Confirmed! Only thing we need to patch is the pinning away.

PR #2990, opened 2020-04-18
test: stub out ipfs interactions in http injection tests

achingbrain:
This makes running the tests faster as we do no network operations, nor do we wait for ports to open, etc.

Functionality covered by the tests is also covered by the interface tests so there is no testing gap.

Also adds some tests where there were no tests previously.


achingbrain:
Build times

Type | Before | After
-----|--------|-------
Linux | 25mins | 22mins
Mac OS | 45mins | 34 mins
Windows | 37mins | 32 mins

PR #3001, opened 2020-04-24
fix: do not abort dht operation on error responses

### LINKED ISSUE #2991: ipfs-http-client dht functions throw exceptions too early
Ping @alanshaw

- **Package**: `ipfs-http-client`

- **Version**:  `>41.0.0`

#### Description:

Calls to DHT functions e.g. `findProvs()` fail too quickly on encountering a `type: 3` message from the ipfs node. This seems to be naive, as type 3 error messages are __not__ fatal to the dht functions; very often a valid response is obtained __after__ a `type: 3` message. The example code below more often than not fails to discover any providers at all, with error messages such as `Error: dial backoff`  or  `Error: failed to dial : all dials failed` or `Error: failed to dial : no good addresses`. These are to be expected and should __not__ abort the operation.
The effect is that the dht functions are rendered unusable because of the highly erratic nature of the errors and the unreliability of getting any results.

This change was implemented in version 41.0.1,
PR: https://github.com/ipfs/js-ipfs-http-client/pull/1183

Code e.g.
```
const IPFS = require('ipfs-http-client')
const hashes = [
  "zdpuAuSAkDDRm9KTciShAcph2epSZsNmfPeLQmxw6b5mdLmq5",
  "zdpuAmm8YdN9eHmQ2WHNShrdLF58SYEsqzDRFsx5GrCDVmNLr",
  "zdpuAwo4v8f1NYbUpoPSmCCondF2UPeB7fEqmWPQZD9Ss9Kts",
  "zdpuAyJcWDGajL3XMJf5iRmMRwM1zwAp3u2d928gZgr3ED8EB",
  "zdpuAyx3MUm6ZtkvabuMqbBZpnoPqsnAhnDpcCPYj9tBCmXff"
]

const ipfs = IPFS()

async function findProvs(hash) {
  const providerList = []
  const find = ipfs.dht.findProvs(hash)  
  try {
    for await (const provider of find) {
      providerList.push(provider)
      console.log(provider)
    }
  } catch (err) {
    console.error(err)
  }
  return providerList
}

hashes.forEach(async function (h) {
  console.log(h, await findProvs(h))
})

```

achingbrain:
Response streams from DHT operations contain lots of message types, one of which is 'Error', but they are emitted for all sorts of reasons, failing to dial peers, etc and valid responses can arrive after several errors.

The change here is to ignore error messages sent by the remote node as they do not indicate that the request has failed.

Fixes #2991

### DIFF
--- a/packages/ipfs-http-client/src/dht/find-peer.js
+++ b/packages/ipfs-http-client/src/dht/find-peer.js
@@ -5,6 +5,7 @@ const CID = require('cids')
 const multiaddr = require('multiaddr')
 const configure = require('../lib/configure')
 const toUrlSearchParams = require('../lib/to-url-search-params')
+const { FinalPeer } = require('./response-types')
 
 module.exports = configure(api => {
   return async function findPeer (peerId, options = {}) {
@@ -18,14 +19,10 @@ module.exports = configure(api => {
     })
 
     for await (const data of res.ndjson()) {
-      if (data.Type === 3) {
-        throw new Error(data.Extra)
-      }
-
-      if (data.Type === 2 && data.Responses) {
+      if (data.Type === FinalPeer && data.Responses) {
         const { ID, Addrs } = data.Responses[0]
         return {
-          id: ID,
+          id: new CID(ID),
           addrs: (Addrs || []).map(a => multiaddr(a))
         }
       }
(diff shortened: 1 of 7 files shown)


alanshaw:
Just checking, I was under the impression all Peer IDs were CIDs atm?

achingbrain:
Good point, I've updated the PR

phillmac:
LGTM 👍

PR #3003, opened 2020-04-24
chore: upgrade ipfs-unixfs-importer to use block api only

achingbrain:
Removes use of ipld by the importer, instead using the block api directly.

This allows use by implementations that have only completed low-level APIs and also makes importing ever so slightly faster by not serializing nodes twice while adding them to the block store.


packages/ipfs/src/core/components/files/utils/hamt-utils.js
```diff
@@ -69,7 +69,7 @@ const addLinksToHamtBucket = async (links, bucket, rootBucket) => {
       }
 
       return (rootBucket || bucket).put(link.Name.substring(2), {
-        size: link.TSize,
+        size: link.Tsize,
```
achingbrain, line 72: 😱

PR #3004, opened 2020-04-24
feat(ipfs): add libp2p-noise as fallback support for libp2p-secio

jacobheun:
The target release of this is in conjunction with the go-ipfs 0.6 release (scheduled for June with an RC in in May).

* This adds support for the Noise security transport, but does not change the default from Secio
* A future release should prioritize Noise once the network has had time to Upgrade

### Rollout Plan
- Add support for Noise
- [Time allotted for network to upgrade]
- Prioritize Noise over Secio
- [Time allotted for network to upgrade]
- Remove support for Secio

required by https://github.com/libp2p/js-libp2p/issues/615


jacobheun:
Duplicate of #3074

PR #3005, opened 2020-04-24
refactor: use low-level block api in refs-local interface test

achingbrain:
Follows on from #2980 but uses the `unixfs-importer` which has been refactored to only use the block API, allowing implementations to be written from low-level APIs upwards.

CC @aphelionz


aphelionz:
Works for me! Can we close the other PRs?

achingbrain:
Will do, just waiting for green CI

PR #3007, opened 2020-04-27
fix: skip gc tests over http on windows

achingbrain:
A temporary measure until go-ipfs 0.5.0 is released

Refs: https://github.com/ipfs/go-ipfs/issues/7225


achingbrain:
I don't think this is the answer :(

achingbrain:
For more discussion & analysis see https://github.com/ipfs/js-ipfs/issues/2999

PR #3012, opened 2020-04-29
chore: upgrade go-ipfs-dep to 0.5.0

achingbrain:
Upgrades to the latest go-ipfs release

Depends on:

- [x] https://github.com/ipfs/js-ipfsd-ctl/pull/504


achingbrain:
Fixed in #3013

PR #3013, opened 2020-04-29
feat: sync with go-ipfs 0.5

hugomrdias:
- [x] ipfs.ls remove all options in core api, but keep `long` in http api and cli same as go-ipfs
	- remove sort from the cli and ignore go-ipfs `U` option
	- enable http api mfs tests
- [x] key.gen defaults, to rsa and 2048
- [x] pin.list no need to support non stream responses
- [x] understand the `U` args in `files.ls` doesn't seem to sort (ignore)
- [x] rewrite dht get/put tests
- [x] unskip dht get/put tests in http client
- [x] dht get accepts string or buffer
- [x] dht put uses body to send data instead of query string 
- [x] http-client dht get returns a buffer to match core api
- [x] 'should get the wantlist by peer ID for a different node' test should pass
	- needs https://github.com/ipfs/go-bitswap/pull/390 to be released skipped for now
- [x] add test for floodsub
- [x] fix gossipsub test, needs a delay after seeing the peers to be able to delivers msgs to others
- [x] pubsub publish
  - add undefined msg test
  - ipfs api supports msg in the body
- [x] pubsub http server tests for data in the body

needs:
https://github.com/ipfs/interop/pull/113

Tracking issue https://github.com/ipfs/js-ipfs/issues/3030


aphelionz:
Thanks a lot @hugomrdias. This will be super helpful in getting the next release of OrbitDB out

aphelionz:
Just checking on this - `go-ipfs@0.5.0` seems to be working now. Is this still necessary?

hugomrdias:
> Just checking on this - `go-ipfs@0.5.0` seems to be working now. Is this still necessary?

yes

PR #3014, opened 2020-04-30
docs: remove Tech Lead section

daviddias:
I've been effectively not contributing as Tech Lead for a several months now, so not worth having my name there and mislead folks.

I see this is a natural transition, specially after all the great work from @alanshaw and @achingbrain as the [Lead Maintainers](https://github.com/ipfs/team-mgmt/blob/master/LEAD_MAINTAINER_PROTOCOL.md). You've made the project so much better, thank you for taking really great care of it, as you can imagine, it means a lot to me ❤️


hugomrdias:
Thank you for creating all of this, we do our best to improve on what you have done. 🙏

achingbrain:
😢 I've we've seen further, it's because we've stood on the shoulders of giants!

PR #3016, opened 2020-05-01
chore: add codecov for all builds

achingbrain:


hugomrdias:
karma is not configured for coverage, theres a issue in aegir for that but i will probably fix that when i change to playwright-test and mocha-electron i dunno if it works well with nyc/istanbul

PR #3019, opened 2020-05-04
chore: use libp2p 0.28.x

vasco-santos:
This PR updates libp2p usage to `0.28.x`. 

The biggest impact in `js-ipfs` is the deprecation of `PeerInfo`, which creates breaking changes in several parts of the libp2p API.

For more information you can check the libp2p migration guide and release notes:

- [libp2p/js-libp2p/doc/migrations/v0.27-v.28.md](https://github.com/libp2p/js-libp2p/blob/master/doc/migrations/v0.27-v.28.md)
- [libp2p/js-libp2p#603](https://github.com/libp2p/js-libp2p/issues/603)

Needs:

- [x] https://github.com/libp2p/js-libp2p-kad-dht/pull/194
- [x] `libp2p@0.28`
- [x] [ipfs/js-ipfs-bitswap#217](https://github.com/ipfs/js-ipfs-bitswap/pull/217)


packages/ipfs/package.json
```diff
@@ -93,10 +93,10 @@
     "hapi-pino": "^6.1.0",
     "hashlru": "^2.3.0",
     "interface-datastore": "^1.0.2",
-    "ipfs-bitswap": "^1.0.0",
+    "ipfs-bitswap": "ipfs/js-ipfs-bitswap#chore/use-libp2p-0.28",
     "ipfs-block-service": "^0.17.1",
     "ipfs-core-utils": "^0.2.3",
-    "ipfs-http-client": "^44.1.0",
+    "ipfs-http-client": "44.0.3",
```
vasco-santos, line 99: This needs to be changed! When I rebased this PR with the `js-ipfs` repo changes from the last week, I started getting this error:

```
Error: Cannot find module 'ipfs-http-client/src/dht/find-peer'
Require stack:
- /Users/vsantos/work/pl/gh/ipfs/js-ipfs/node_modules/libp2p-delegated-peer-routing/src/index.js
- /Users/vsantos/work/pl/gh/ipfs/js-ipfs/packages/ipfs/src/cli/daemon.js
```

From my analysis  so far, using the `"ipfs-http-client": "^44.1.0"` will install `ipfs-http-client` in `js-ipfs/packages/ipfs/node_modules/ipfs-http-client` as a link to its folder, which should be correct! However, the require in [libp2p/js-libp2p-delegated-peer-routing/master/src/index.js#L4](https://github.com/libp2p/js-libp2p-delegated-peer-routing/blob/master/src/index.js#L4) does not work with the above error and the tests will not start to run. If I put this fixed version, the dependency will be installed in `js-ipfs/node_modules/ipfs-http-client` and the above error will not happen. I am not familiar with lerna dependency mgr and I could not understand yet the issue. @achingbrain @hugomrdias do you have any idea on the issue?

packages/interface-ipfs-core/src/bitswap/wantlist.js
```diff
@@ -32,7 +32,11 @@ module.exports = (common, options) => {
       // Add key to the wantlist for ipfsB
       ipfsB.block.get(key).catch(() => { /* is ok, expected on teardown */ })
 
-      await ipfsA.swarm.connect(ipfsB.peerId.addresses[0])
+      // Add peers to addressBook
+      ipfsA.libp2p.peerStore.addressBook.set(ipfsB.libp2p.peerId, ipfsB.libp2p.multiaddrs)
+      ipfsB.libp2p.peerStore.addressBook.set(ipfsA.libp2p.peerId, ipfsA.libp2p.multiaddrs)
```
achingbrain, line 37: Will this work when `ipfsA` or `ipfsB` are go-ipfs nodes?

achingbrain: This is why the `ipfs-http-client` tests are failing in CI..

vasco-santos:
FYI CI is green now, this is pending a dht fix [libp2p/js-libp2p-kad-dht#194](https://github.com/libp2p/js-libp2p-kad-dht/pull/194) to be merged/released and the final release of `js-libp2p`.

packages/ipfs/package.json
achingbrain: For posterity, the problem here was that the delegate-* modules depend on a module that resolves to one in the `/packages` folder, which other modules in that folder also depend on.  So lerna creates symlinks for those modules but then fails to hoist the symlink to the root `node_modules` folder.

This is a bit of an edge case as our dependency tree is a little wonky in this case but it also seems like a bug in lerna.  The short-term solution is to add the delegate-* modules to the `nohoist` list in `lerna.json`.

PR #3020, opened 2020-05-04
New preload nodes

thattommyhall:
I'd hoped to DRY up the tests so we didnt have to edit them to add new preload nodes, but I could not figure out how requires work at that point (rusty at JS and was never that great at it)


thattommyhall:
Don't merge, problem with config on a node and I think I'll need to change its id

PR #3023, opened 2020-05-06
fix: typeof bug when passing timeout to dag.get

mistakia:
Passing `options` without a `path` to dag.get will throw an error.
```
ipfs.dag.get(cidPath, { timeout: 2000 })

TypeError: Cannot read property 'timeout' of null
```

Cause by the lovely JS issue of `typeof null === 'object'`.

### DIFF
--- a/packages/interface-ipfs-core/src/dag/get.js
+++ b/packages/interface-ipfs-core/src/dag/get.js
@@ -204,5 +204,19 @@ module.exports = (common, options) => {
       const result = await ipfs.dag.get(cid)
       expect(result.value).to.deep.equal(buf)
     })
+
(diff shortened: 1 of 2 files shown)


achingbrain:
Great spot, thanks!  Could you please add a test to catch this sort of thing in future?

The tests for `ipfs.dag.get` are here: [packages/interface-ipfs-core/src/dag/get.js](https://github.com/ipfs/js-ipfs/blob/master/packages/interface-ipfs-core/src/dag/get.js)

achingbrain:
Fixed in #3035 with a `Co-authored-by` so you get git history credit.  Thanks for opening this PR!

Issue #2945, opened 2020-03-24
Track: Typescript types support


hugomrdias:
We will use JSDoc to declare types for the top level API first and internally we will add types incrementally where we feel it adds value.

A TS config will be added to the repo to enable type declaration generation from JSDoc comments. 

Type declaration tests will be added as we see fit and we will add a ts type check job to our CI.

## Repo track list
Done = ✅ 
In Progress = 🚧 
TODO = ⛔ 
| Status | Repo/PR | Owner |
| --- | --- | --- |
|✅  | aegir https://github.com/ipfs/aegir/pull/671 https://github.com/ipfs/aegir/pull/701 | - |
|✅  | ipfs-utils https://github.com/ipfs/js-ipfs-utils/pull/89 | - |
|✅  | interface-datastore https://github.com/ipfs/interface-datastore/pull/56 | - |
|✅  | datastore-core https://github.com/ipfs/js-datastore-core/pull/39 | - |
|✅  | datastore-level https://github.com/ipfs/js-datastore-level/pull/53 | - |
|✅ | datastore-pubsub https://github.com/ipfs/js-datastore-pubsub/pull/60 | @achingbrain |
|✅  | datastore-fs https://github.com/ipfs/js-datastore-fs/pull/62 | - |
|✅ | ipfs-repo https://github.com/ipfs/js-ipfs-repo/pull/275 | @hugomrdias  | 
|✅ | ipfs-unixfs https://github.com/ipfs/js-ipfs-unixfs/pull/114 https://github.com/ipfs/js-ipfs-unixfs/pull/122 | @achingbrain |
|✅ | ipfs-bitswap https://github.com/ipfs/js-ipfs-bitswap/pull/261 | @Gozala |
|✅ | is-ipfs https://github.com/ipfs-shipyard/is-ipfs/pull/39 | @achingbrain |
|✅ | ipns https://github.com/ipfs/js-ipns/pull/106 | @achingbrain |
|✅ | ipfs-block-service https://github.com/ipfs/js-ipfs-block-service/pull/136 | @achingbrain |
|✅  | [cids](https://github.com/multiformats/js-cid) https://github.com/multiformats/js-cid/pull/131 | - |
|✅ | multihash https://github.com/multiformats/js-multihash/pull/104 | - |
|✅  | multibase https://github.com/multiformats/js-multibase/pull/72 | - |
[...]

PR #3046, opened 2020-05-22
WIP: Adding JSDoc supporrted  types

Gozala:
I have spend time trying to make progress on #2945. This is not complete TS right now reports `3211` errors which is a lot but it was 100x that when I started. I do not know how much more time I can afford spending on this, but I'll try some over the weekend.

Few high level notes / overview

- This does add `tsconfig.json` which I beleieven technically isn't necessary, however it is virtually impossible to otherwise understand what TS will infer. E.g. there weer cases here and dependencies where jsdoc comments existed by TSC was not happy due to various issues.
- I have created `types/*` directory with typings for supporting libraries & TS know to look there first before trying to look into `node_modules`.
- There are problems with how ES modules are interpreted by CommonJS modules and vise versa mostly because `default export` does not correspond to `module.exports`. That is also why I changed `B = require(...)` to `{B} = require(...)` here and there as it resolved some of those issues.
- I have created `types/ipfs-interface` with bunch of helper types and things so they could be pulled as needed throughout.
- Dependency injection complicates things, because usually type of actual component ends enclosed in function so you can't refer to types to it from anywhere (without retyping). I do not yet know what would be a better strategy . So far my hope was that TS would be enable to infer the whole type. But even then problem is that some of that needs to be injected elsewhere and ideally dependencies would be typed.
- There are cases where current API is fundamentally incompatible with static typing. E.g. `ApiManager` has a field `api` who's type dependents on later call to `update` meaning it would start of with type `T` and later change to type `D`. That is not possible I'm afraid although there are some alternatives that might be worth exploring that would enable that e.g. instead of using `apiManager.api` use `{api} = apiManager.update(...)`. That can be inferred because return type could be derived from type of input.
- There are other cases of changing types e.g. `log.error = debug(...)`. To address this specific one I have created helper `debug` module that just does that inform TS about return type.
- All over the place I had to change the way decorators functions are applied because JSDoc comments on top of e.g. `withTimeoutOption(async function put (block, options) {})` do not get associated with `put` function being decorated. Solution is to first define than decorate.
- There are many cases where types like `*` are used. Which essentially tells TS to infer the type rather than complain about knowing. That is workaround to get to state where TS can successful type check. Once there we can incrementally replace those with actual types.
- Some types are probably wrong. There is a lot of new code for me a I did my best to represent interfaces consulting docs were available to get things right, but I'm I made mistakes, in fact I discovered many such mistakes as TS was able to make more sense of it all. I also end fixing them where I could. 

https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html


achingbrain:
Nice work, this looks to have been pretty painful to put together.

I'm not 100% on including types for our dependencies as we don't want to be guardians of the APIs of those modules, but I don't know enough about TS to know if that's a dealbreaker or not.

> I'll try some over the weekend.

Please don't! Take some time to unwind.

packages/ipfs/src/core/components/pin/add.js
```diff
@@ -66,9 +93,11 @@ module.exports = ({ pinManager, gcLock, dag }) => {
     const release = await gcLock.readLock()
 
     try {
-      await pinAdd()
+      return await pinAdd()
```
Gozala, line 96: I think this `return` here is intended behavior, given that on line 90 there's `return pinAdd()`. At least that is what TS identifying as type mismatch.

packages/ipfs/src/core/ipns/publisher.js
```diff
@@ -44,11 +63,13 @@ class IpnsPublisher {
...
     }
 
+    // @ts-ignore - Type defs for PeerId hav no _pubKey
     const publicKey = peerId._pubKey
     const embedPublicKeyRecord = await ipns.embedPublicKey(publicKey, record)
+    /** @type {any} */
     const keys = ipns.getIdKeys(peerId.toBytes())
 
-    await this._publishEntry(keys.routingKey, embedPublicKeyRecord || record, peerId)
+    await this._publishEntry(keys.routingKey, embedPublicKeyRecord || record)
```
Gozala, line 72: `_publishEntry` below takes just two arguments. I changed this because TS is rightfully points that out.

packages/ipfs/src/core/ipns/routing/utils.js
```diff
@@ -3,14 +3,34 @@
...
+   * @param {Buffer} buf
+   * @returns {string}
+   */
   encodeBase32: (buf) => {
     const m = multibase.encode('base32', buf).slice(1) // slice off multibase codec
 
     return m.toString().toUpperCase() // should be uppercase for interop with go
   },
   validator: {
-    func: (key, record, cb) => ipns.validator.validate(record, key, cb)
```
Gozala, line 13: validator doesn't take `cb` (anymore) so I changed this to reflect current state.

packages/ipfs/src/core/preload.js
```diff
@@ -4,29 +4,58 @@ const toUri = require('multiaddr-to-uri')
 const debug = require('debug')
 const CID = require('cids')
 const shuffle = require('array-shuffle')
-const AbortController = require('abort-controller')
+const { AbortController } = require('abort-controller')
```
Gozala, line 7: This was necessary because:

1. TS does not treat ES default export as `module.exports` in CommonJS. Which on it's own would not have being issues, but for a fact that abort-controller module provides `.d.ts` file that provides type signature that tells TS it is an ES module. In the end just using destracturing here addresses everything as module happens to export same value as `AbortController` as well.

Gozala:
> I'm not 100% on including types for our dependencies as we don't want to be guardians of the APIs of those modules, but I don't know enough about TS to know if that's a dealbreaker or not.

I would frame it as TS wants you to encode invariants you happen to assume, but it's up to you to choose the level of details.  In fact types for most dependencies were generated and are really loose (have `any` type), which is basically tells TS don't worry about that it's all cool. There are however few modules that I've deliberately typed manually for one of the following reasons:

1. It's library part of IPFS which happend to live in different repo. Ideally I would just added JSDoc comments there, however to reduce coordination overhead I've just created type definition in hope that later they could be moved to actual repo as JSDoc comments after which type definition could be removed.
2. It is an essential library e.g. Without knowledge of `it-pipe` type signature TS (and if I dare say myself) has little clue of what are the things passed around (as illustrated by the screenshot below). This also means I need to add jsdoc comments to the closure.
    image
    With type information for `pipe` however TS can infer the type of `source` and tools like VSCode can provide helpful tips as shown in screenshot below:
   image
    I should say it is possible to suppress TS from complaining about implicit `any` type, however it would not be able to help (someone new to the code base, like myself) to understand what are the things being passed around without taking side trips to `pipe`, `normaliseAddInput`, `importer` etc...

Gozala:
I'm getting inconsistent results from testing. Is there some shared knowledge I could draw from ?

Gozala:
I had discussed this work with @hugomrdias on the call. We have decided that I'll leave this as is (even though it is tempting to resolve last 81 errors). He will incorporate work from here into #2945 some form.

Issue #2779, opened 2020-02-15
Error: no valid addresses were provided for transports


icidasset:
👋 Thanks for the 0.41.0 update! 

I'm having an issue with upgrading from 0.38
I'm getting the following error:
```
Error: no valid addresses were provided for transports [WebSockets,WebRTCStar,Circuit]
```

I'm setting it up as follows, which worked before the upgrade:
```js
const ipfs = await Ipfs.create();
await ipfs.swarm.connect("/dns4/ipfs.runfission.com/tcp/4003/wss/ipfs/QmVLEz2SxoNiFnuyLpbXsH6SvjPTrHNMU88vCQZyhgBzgw")
```

Any thoughts on what changed?

alanshaw:
Are you running in Node.js or the browser or somewhere else? If in Node.js, what swarm addresses do you have configured? ipfs.config.get(‘Addresses.Swarm’). I’m guessing you have a websocket-star address or something in there.

icidasset:
I found the "issue" 😅 

I'm connecting to a peer (swarm) after I called `create`. But, I got the error immediately after calling `create` without any configuration (ie. just `Ipfs.create()`)

If I pass an empty swarm list to create, as follows, it does work:
```js
Ipfs.create({ config: { Addresses: { Swarm: [] } } })
await ipfs.swarm.connect(...)
```

Previously I didn't have to pass this empty list to create. What do you think @alanshaw? Is the new behaviour better, is this confusing, is this just edge-case, or am I not using this as intended?

Thanks ✌️ 

PS. I'm running this in the browser using the minified unpkg distribution.

npfoss:
I was in the same situation except I did want to pass addresses to the swarm config, and I just had to follow the setup instructions here to add webrtc support: https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs#is-there-webrtc-support-for-js-ipfs-with-nodejs

alanshaw:
I think by using that config you removed the websocket-star address you previously had saved in `Addresses.Swarm` so there was no longer an address without a transport.

@jacobheun do you think this startup error should be downgraded to a warning?

jacobheun:
@alanshaw yes, I think so, this shouldn't be a fatal error. The available startup addresses don't have any transports so it will fail to connect, however, performing a manual connect will still work.

The important thing to note is that your node will be dial only.

[...]

PR #3051, opened 2020-05-27
chore: throw error no swarm on multiaddrs using websocket-star

vasco-santos:
On `js-ipfs@0.41`, we removed the `websocket-star` module from the libp2p default configuration for browser nodes in `js-ipfs`. We removed it because this module was not refactored during https://github.com/ipfs/js-ipfs/issues/1670, since the libp2p goal is to [sunset the star protocols](https://github.com/libp2p/js-libp2p/issues/385) and we preferred to use the time refactoring `websocket-star` into improving the browser support.

In the refactor, the `webrtc-star` module was refactored so that browser users could discover other peers in the network. In addition, `circuit-relay` may be used for establishing connections between browser nodes using `websockets`.

However, this migration has not been smooth for `js-ipfs` users. Users previously using `websocket-star` swarm addresses, did not have enough visibility of this change, since the removal of `websocket-star` was transparent for them and on the default `js-ipfs` release and not technically a programmable breaking change. 

With the above in mind, once `js-ipfs` users updated they started getting errors from `js-libp2p` in scenarios where they were providing only `websocket-star` swarm addresses. When `js-libp2p` receives listening multiaddrs, it throws an error if it cannot use any to listen on the configured transports, which was the case here (For instance https://github.com/ipfs/js-ipfs/issues/2779).

In `js-libp2p`, we added an option to tolerate these errors [libp2p/js-libp2p#643](https://github.com/libp2p/js-libp2p/pull/643), which was also mentioned for some other cases earlier.

After syncing about this issue, we decided to temporarily throw an error when `js-ipfs` receives `websocket-star` multiaddrs. This aims to help users who still need to migrate to identify the problem faster.

### DIFF
--- a/packages/ipfs/src/core/components/start.js
+++ b/packages/ipfs/src/core/components/start.js
@@ -11,6 +11,8 @@ const Components = require('./')
 const createMfsPreload = require('../mfs-preload')
 const { withTimeoutOption } = require('../utils')
 
+const WEBSOCKET_STAR_PROTO_CODE = 479
+
 module.exports = ({
   apiManager,
   options: constructorOptions,
@@ -40,6 +42,12 @@ module.exports = ({
     if (config.Addresses && config.Addresses.Swarm) {
       config.Addresses.Swarm.forEach(addr => {
         let ma = multiaddr(addr)
+
+        // Temporary error for users migrating using websocket-star multiaddrs for listenning on libp2p
+        // websocket-star support was removed from ipfs and libp2p
+        if (ma.protoCodes().includes(WEBSOCKET_STAR_PROTO_CODE)) {
+          throw new Error('websocket-star swarm addresses are not supported. See https://github.com/ipfs/js-ipfs/issues/2779')
+        }
 
         // multiaddrs that go via a signalling server or other intermediary (e.g. stardust,
         // webrtc-star) can have the intermediary's peer ID in the address, so append our
@@ -133,7 +141,6 @@ module.exports = ({
     apiManager.update(api, () => undefined)
   } catch (err) {
     cancel()
-    startPromise.reject(err)
     throw err
   }
 
--- a/packages/ipfs/src/core/components/start.js
+++ b/packages/ipfs/src/core/components/start.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const log = require('debug')('ipfs:components:start')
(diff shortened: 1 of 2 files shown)


packages/ipfs/test/core/create-node.spec.js
```diff
@@ -250,4 +250,28 @@ describe('create node', function () {
...
+        },
+        Bootstrap: []
+      },
+      preload: { enabled: false }
+    })
+
+    try {
+      await node.start()
+    } catch (err) {
+      expect(err).to.exist()
```
achingbrain, line 271: Please assert the correct error was thrown.

Better yet, do something like

```js
await expect(node.start()).to.eventually.be.rejectedWith(/regex to use on error message/)
```

or

```js
await expect(node.start()).to.eventually.be.rejected().with.property('code', 'ERR_SOME_CODE')
```

then you don't need the try/catch and the following throw.

packages/ipfs/src/core/components/start.js
```diff
@@ -130,7 +138,6 @@ module.exports = ({
     apiManager.update(api, () => undefined)
   } catch (err) {
     cancel()
-    startPromise.reject(err)
```
achingbrain, line 133: Why did this get removed?  If the user calls `node.start()` twice, the second promise will never resolve/reject.

I'm not sure how common this is, but I'd like to re-evaluate the use of the api proxy as part of #2762.

vasco-santos: Oh, I did not understand that this was used to restart.

So, what happens with the `startPromise.reject(err)` is that this promise rejection is not being handled anywhere. I looked closely on this now, and it seems that the `ApiManager` should take care of handling promise rejections. Is that fair to assume? I am not sure on how to actually do this though.

achingbrain: It's not used to restart, it's if the user calls `node.start()` on a node that's already starting up.

vasco-santos: sorry, I understood, but misspelled it: parallel starts

So, how to you think we should go on handling the promise rejections on the APIManager

vasco-santos: I added a catch, but I am not sure this is the best approach here. Let me know

PR #3055, opened 2020-05-28
fix: reflect actual return value of `files.write`

Gozala:
As far as I can tell documentation is incorrect here given this is what write returns

https://github.com/ipfs/js-ipfs/blob/b2dc47c18a0c5b307a8c9da4bf6fe0100096228f/packages/ipfs/src/core/components/files/write.js#L222-L225

### DIFF
--- a/docs/core-api/FILES.md
+++ b/docs/core-api/FILES.md
@@ -807,7 +807,12 @@ An optional object which may have the following keys:
 
 | Type | Description |
 | -------- | -------- |
-| `Promise<void>` | If action is successfully completed. Otherwise an error will be thrown |
+| `Promise<Object>` | If action is successfully completed. Otherwise an error will be thrown |
+
+The returned object has the following keys:
+
+- `cid` a [CID][cid] instance
+- `size` is an integer with the file size in Bytes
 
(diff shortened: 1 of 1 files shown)


achingbrain:
Can you cross reference with what go-ipfs returns?  From the looks of the [go http api docs](https://docs.ipfs.io/reference/http/api/#api-v0-files-write), the api docs here are correct and the js implementation should not be returning anything.

Gozala:
> Can you cross reference with what go-ipfs returns? From the looks of the [go http api docs](https://docs.ipfs.io/reference/http/api/#api-v0-files-write), the api docs here are correct and the js implementation should not be returning anything.

I would rather argue that HTTP API should return something.

achingbrain:
In that case please open an issue on ipfs/go-ipfs for discussion of HTTP API responses to reach an agreement - we're largely trying to replicate the go-ipfs HTTP API here, warts and all.

Gozala:
Closing this until we decide between:

1. Changing js-ipfs API to match docs.
2. Changing HTTP API to match implementation of JS-IPFS.

PR #3057, opened 2020-05-31
refactor: changing another ipfs.add to importer

aphelionz:
This [WIP] PR modifies the `pin/*.js` tests to use `importer` instead of `ipfs.add`. This will be developed in tandem with https://github.com/rs-ipfs/rust-ipfs/pull/152

- [x] pin.add
- [ ] pin.rm
- [ ] pin.ls

### DIFF
--- a/packages/interface-ipfs-core/src/pin/add.js
+++ b/packages/interface-ipfs-core/src/pin/add.js
@@ -6,6 +6,7 @@ const { getDescribe, getIt, expect } = require('../utils/mocha')
 const all = require('it-all')
 const testTimeout = require('../utils/test-timeout')
 const CID = require('cids')
+const importer = require('ipfs-unixfs-importer')
 
 /** @typedef { import("ipfsd-ctl/src/factory") } Factory */
 /**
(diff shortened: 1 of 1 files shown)


packages/interface-ipfs-core/src/pin/add.js
```diff
@@ -23,7 +24,7 @@ module.exports = (common, options) => {
     before(async () => {
       ipfs = (await common.spawn()).api
       await Promise.all(fixtures.files.map(file => {
-        return all(ipfs.add(file.data, { pin: false }))
+        return all(importer(file.data, ipfs.block, { pin: false }))
```
achingbrain, line 27:
```suggestion
        return all(importer([{ content: file.data }], ipfs.block, { pin: false }))
```

koivunej:
I don't think this will be needed by rust-ipfs anymore. Also I am not sure if this would be very important in the API leveling sense: before add there can of course be trees created manually however add is the first big feature with the creation directory and other unixfs trees so that there is something to be pinned (recursively).

aphelionz:
Sounds good. I will close

PR #3058, opened 2020-06-01
Denpo satoshi patch 1

denpo-satoshi:


achingbrain:
I think you may have opened this by accident. The PRs you have merged into this branch are not ready and will be merged into master individually.  Also I'm not sure why you have added GitHub actions, they are not required.

PR #3059, opened 2020-06-01
Revert "docs: fixed adherence typo error"

denpo-satoshi:
Reverts ipfs/js-ipfs#3049


achingbrain:
I think you've opened this by mistake, we don't want to revert a typo fix.

PR #3061, opened 2020-06-02
chore: use ipfs npm registry in travis

### LINKED ISSUE #2997: Update .travis to install from registry.js.ipfs.io
To stop CI from polluting the npm download stats, we should pull from `registry.js.ipfs.io` instead.

tarunbatra:
Have added a step to set npm registry in the travis build process to solve #2997.

Another way to do this would have been maintaining a `.npmrc` file in the root directory, but that would have lead to installs happening during development also to use ipfs npm registry which I feel is more than the scope defined.


achingbrain:
Thanks for submitting this, I will merge it but not immediately as the service behind `registry.js.ipfs.io` needs updating to be more stable.  The work is about 80% done, I just need to find the time to finish it off.

tinytb:
What is the service behind registry.js.ipfs.io? How (un)stable is it?

achingbrain:

The registry.js.ipfs.io service was retired so this isn't going to get done.  It (still) really needs a bit of an overhaul tbh.

PR #3067, opened 2020-06-04
fix: fix ipfs config show

hugomrdias:
This PR makes `ipfs config show` work again and `config.get` first arg required.

Also this makes ipfsd-ctl work again.

### DIFF
--- a/docs/core-api/CONFIG.md
+++ b/docs/core-api/CONFIG.md
@@ -26,7 +26,7 @@
   - [Returns](#returns-4)
   - [Example](#example-4)
 
-## `ipfs.config.get([key,] [options])`
+## `ipfs.config.get(key, [options])`
 
 > Returns the currently being used config. If the daemon is off, it returns the stored config.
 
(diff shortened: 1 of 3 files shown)


achingbrain:
Can we change the API to be something like:

```js
ipfs.config.getAll()
ipfs.config.get(key)
```

instead of

```js
ipfs.config.get(undefined)
ipfs.config.get(key)
```

Isomorphism in js is such a hack.

hugomrdias:
Maybe `ipfs.config.show ` to keep it consistent with http api and cli ?

achingbrain:
The CLI then literally shows the config though, which is not what we are doing, and the HTTP API is generated from the CLI commands.

I think we can give the API commands more straightforward names, though they may call HTTP API endpoints that are named less descriptively.

achingbrain:
Could do `ipfs.config.read`?  Then it's similar to `ipfs.files.read`?

hugomrdias:
Makes sense, lets go with `getAll` its already there in ipfs-repo and also other apis i researched use the same method.

hugomrdias:
closed in favor of https://github.com/ipfs/js-ipfs/pull/3071

Issue #3022, opened 2020-05-05
Sharing IPFS node across browsing contexts (tabs) on same origin


Gozala:
> cc https://github.com/ipfs/in-web-browsers/issues/158

When site with origin e.g. `foo.com` uses JS-IPFS, new node is bootstrapped for every browser context (tab, iframe) this implies:

- All the network connections need to be reestablished.
- General memory / CPU overhead is multiplied by number of browsing contexts.
- Create a possibility for race condition (two nodes read / write data concurrently).

There is an opportunity to improve this by offloading most of JS-IPFS work into [`SharedWorker`](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker) in browsers where API is available and fallback to a dedicated [`Worker`](https://developer.mozilla.org/en-US/docs/Web/API/Worker) elsewhere (basically Safari).
<details>
We can explore <code>ServiceWorker</code> as a better fallback mechanism in the future. Unilke dedicated worker they can be shared, but they also come with enough challenges to justify keeping it out of scope initially. 
</details>

However use of `SharedWorker` implies some trade-offs:

#### Worker script distribution

Worker needs to be instantiated with a separate JS script. Which leaves us with the following options:

1. Create a separate "worker" dist as opt-in for shared node. That still has two more options:
    1. Spit out two scripts on for main context and other for worker context.
    1. Use something like a [worker-loader](https://webpack.js.org/loaders/worker-loader/) to bundle worker script as a blob embedded in the the main script.
2. Self load main script into worker and context detect to choose between main and worker operational modes.

#### Lack of WebRTC in workers

This is a major problem as WebRTC is largely replacing WebSocket transport and requires more research. Pragmatic approach could be to tunnel WebRTC traffic into a worker, however that would still imply reestablishing existing connections which would be a major drawback.

Gozala:
Another thing I'm recognizing now is that two different browsing contexts (tabs) could start IPFS node with different configurations. Which needs to be handled in some way.

Gozala:
I have started hacking on the `ipfs-message-port-client` that in the spirit is very similar to `ipfs-http-client` except it wraps around [MessagePort](https://developer.mozilla.org/en-US/docs/Web/API/MessagePort).

[...]

PR #3070, opened 2020-06-05
chore: update dag-pb dep

Gozala:
This is an attempt to asses if changes from https://github.com/ipld/js-ipld-dag-pb/pull/184 can be integrated into js-ipfs without breaking things, which in turn would address some problems I've run into https://github.com/ipfs/js-ipfs/issues/3022#issuecomment-638008545

### DIFF
--- a/packages/interface-ipfs-core/src/object/links.js
+++ b/packages/interface-ipfs-core/src/object/links.js
@@ -59,7 +59,7 @@ module.exports = (common, options) => {
       const node1bCid = await ipfs.object.put(node1b)
 
       const links = await ipfs.object.links(node1bCid)
-      expect(node1b.Links[0]).to.eql({
+      expect(node1b.Links[0]).to.containSubset({
         Hash: links[0].Hash,
         Tsize: links[0].Tsize,
         Name: links[0].Name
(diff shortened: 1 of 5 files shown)


Gozala:
This was mostly to verify that proposed changes to dag-pb were not introducing regressions. However since dag-pb was released it would be good to land this to avoid divering #3081 too much.

Gozala:
Never mind, I this mixed with dag-cbor instead.

packages/interface-ipfs-core/src/object/links.js
```diff
@@ -60,7 +60,7 @@ module.exports = (common, options) => {
       const node1bCid = await ipfs.object.put(node1b)
 
       const links = await ipfs.object.links(node1bCid)
-      expect(node1b.Links[0]).to.eql({
+      expect(node1b.Links[0]).to.containSubset({
```
achingbrain, line 63: This has broken because [this PR](https://github.com/ipld/js-ipld-dag-pb/pull/184) changed `DagPBNode.Links` to make it [return DAGLinks](https://github.com/ipld/js-ipld-dag-pb/commit/0f11d5c1f42df0f6f9e6ec850f1c2acf0d52adf0#diff-a144e7bd8c520628cd0cdc5279f2f900R37) where it used to [return plain objects](https://github.com/ipld/js-ipld-dag-pb/commit/0f11d5c1f42df0f6f9e6ec850f1c2acf0d52adf0#diff-a144e7bd8c520628cd0cdc5279f2f900L108-L116), making it a breaking change.  Thankfully it was released as `0.x.0` so hasn't been pulled in by people ahead of merging this PR 😥 

```suggestion
      expect(node1b.Links[0]).to.deep.equal(links[0])
```

Issue #3062, opened 2020-06-03
`ipfs.dag` API docs missmatch implementation.


Gozala:
So I have implemented `ipfs.dag` API as per documentation. E.g. [`ipfs.dag.get`](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/DAG.md#ipfsdaggetcid-path-options) per docs take `CID` instance.

However following tests exercise string and buffer inputs instead:

https://github.com/ipfs/js-ipfs/blob/5e0a18aa07c1852d2a7589899661761db50eeda9/packages/ipfs/test/core/dag.spec.js#L22-L32

We should update either docs or an implementation. I have also looked at [`/api/v0/dag/get`](https://docs.ipfs.io/reference/http/api/#api-v0-dag-get) but I am afraid it did not help me clarify here.

---

I am personally biased towards doing what API docs say, because otherwise in the shared worker use case we end up with double normalization / validation more specifically it would imply that:

1. Client needs to assess `cid` argument as either string, buffer, or CID.
2. Server piece would need to asses received `cid` input and decode it to one of those representations.
3. Server will pass decoded args to `ipfs.dag.get` which in turn will again asses `cid` argument.

It is also worth noting that if variadic `cid` is desired, it would make a lot more sense to analyze it in the main thread and error without messaging worker.

achingbrain:
This is because the API previously specified strings, buffers and CIDs as arguments.  Going forward we only want to take CIDs, but the implementation still supports strings and buffers to avoid breaking existing code, though there needs to be a time limit on this.

New implementations of the core API can just support CIDs as per the API docs and skip the interface tests that test with non-CIDs.

Gozala:
Cool, thanks for clarifying

Issue #3036, opened 2020-05-14
Flaky CI tests


achingbrain:
There are several intermittent test failures in CI.  Here are the most common, their causes and the issues which when closed will resolve them:

```
ipfs:   1) interface-ipfs-core over ipfs-http-client tests
ipfs:        .pubsub.peers
ipfs:          "after all" hook for "should return peers for a topic - multiple peers":
ipfs:      FetchError: request to http://127.0.0.1:44199/api/v0/shutdown failed, reason: read ECONNRESET
ipfs:       at ClientRequest.<anonymous> (/home/travis/build/ipfs/js-ipfs/node_modules/node-fetch/lib/index.js:1455:11)
ipfs:       at Socket.socketErrorListener (_http_client.js:401:9)
ipfs:       at emitErrorNT (internal/streams/destroy.js:91:8)
ipfs:       at emitErrorAndCloseNT (internal/streams/destroy.js:59:3)
ipfs:       at process._tickCallback (internal/process/next_tick.js:63:19)
```

```
ipfs: FAILED TESTS:
ipfs:   interface-ipfs-core tests
ipfs:     .pubsub.peers
ipfs:       ✖ "after all" hook for "should return peers for a topic - multiple peers"
ipfs:         Chrome Headless 81.0.4044.138 (Linux x86_64)
ipfs:       Error: Uncaught TypeError: Cannot assign to read only property 'code' of object '' (file:/home/travis/build/ipfs/js-ipfs/node_modules/simple-peer/index.js:26)
ipfs:  Command failed with exit code 1: karma start /home/travis/build/ipfs/js-ipfs/node_modules/aegir/src/config/karma.conf.js --files-custom --log-level error --bail true --timeout 5000
```

### What's happening here?

We are running an interface test over HTTP to a remote js-IPFS node.  While the test is being torn down, a libp2p operation is still in progress on the remote node, which throws.

See: https://github.com/libp2p/js-libp2p-tcp/issues/130
And the browser version: https://github.com/libp2p/js-libp2p-webrtc-star/issues/222
Spurious simple-peer error: https://github.com/feross/simple-peer/pull/660

----

```
ipfs:   1) circuit
ipfs:        js-go-go
ipfs:          connect:
ipfs:      HTTPError: 
[...]

PR #3081, opened 2020-06-10
feat: share IPFS node between browser tabs

### LINKED ISSUE #3022: Sharing IPFS node across browsing contexts (tabs) on same origin

Gozala:
## High level overview

This pull request adds 3 (sub)packages:

1. `ipfs-message-port-client` - Provides an API to an IPFS node over the [message channel][MessageChannel].
2. `ipfs-message-port-server` - Provides an IPFS node over [message channel][MessageChannel].
3. `ipfs-message-port-protocol` - Shared code between client / server mostly related to wire protocol encoding / decoding.

## Status

I consider implementation mostly complete, with some caveats (see below). I think it is ready  

1. All interface tests pass except:
  - Some were deliberately disabled because they deprecated API that was not implemented (See https://github.com/ipfs/js-ipfs/issues/3062#issuecomment-645333274)
  - Some tests depend on the other APIs e.g. `ipfs.object.*` that were falling out of scope. 
2. I'll be doing another pass to clean up things further & add doc / comments as needed.
3. Unrelated tests are disabled due to intermittent failures that occur on master (See https://github.com/ipfs/js-ipfs/issues/3036). This changes aren't intended to be part of this pull.
4. I have increased bundle size of ipfs-http-client (I just can't figure out what the size is changing, as I make no changes to it and no library seems to get bundled twice). This change isn't intended to be part of this pull.
5. Some dependencies point to my forks, until changes in corresponding libraries propagate
    - [x] `Uint8Array` support in dag-pb https://github.com/ipld/js-ipld-dag-pb/pull/184
      - [x] `Uint8Array` support in protons https://github.com/ipfs/protons/pull/13
        - [x] Exposing `TextEncoder` from ipfs-utils https://github.com/ipfs/js-ipfs-utils/pull/44
    - [x] `Uint8Array` support in dag-cbor https://github.com/ipld/js-ipld-dag-cbor/pull/129
    - [x] `Uint8Array` support in ipld-block https://github.com/ipld/js-ipld-block/pull/53
    - ~~[Structure clonable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) CID representation  https://github.com/multiformats/js-cid/issues/109~~
    - [x] Consensus on exposed API https://github.com/ipfs/js-ipfs/issues/3022#issuecomment-641542193

Fixes #3022

[MessageChannel]:https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel


packages/ipfs/test/http-api/inject/block.js
```diff
@@ -60,7 +60,9 @@ describe('/block', () => {
       return testHttpMethod('/api/v0/block/put')
     })
 
-    it('returns 400 if no node is provided', async () => {
+    // Intermittent timeouts
+    // https://github.com/ipfs/js-ipfs/issues/3105
+    it.skip('returns 400 if no node is provided', async () => {
```
achingbrain, line 65: This was failing because hapi doesn't support node 14 yet. Ugh. I've disabled node 14 in master, it should be safe to unskip now.

Gozala: @achingbrain Here is the result of previous commit, which was merge with master
https://github.com/ipfs/js-ipfs/runs/788842730

Gozala: Never mind I had to refresh browser

Gozala:
Oh no bundle size error is back again 😭 I really don't get what's the deal with it. On my device bundle size seems to go down:

```
ipfs-http-client: Use http://webpack.github.io/analyse/ to load "./dist/stats.json".
ipfs-http-client: Check previous sizes in https://bundlephobia.com/result?p=ipfs-http-client@44.2.0
ipfs-http-client: 87.82KB (▼-1.18KB / 89KB)
lerna success run Ran npm script 'build' in 1 package in 20.2s:
lerna success - ipfs-http-client
``` 

But in CI it goes up:

```
ipfs-http-client: > ipfs-http-client@44.2.0 build /home/travis/build/ipfs/js-ipfs/packages/ipfs-http-client
ipfs-http-client: > aegir build "--bundlesize"
ipfs-http-client: Use http://webpack.github.io/analyse/ to load "./dist/stats.json".
ipfs-http-client: Check previous sizes in https://bundlephobia.com/result?p=ipfs-http-client@44.2.0
ipfs-http-client: 89.48KB (▲491B / 89KB)
ipfs-http-client: npm ERR! code ELIFECYCLE
ipfs-http-client: npm ERR! errno 1
ipfs-http-client: npm ERR! ipfs-http-client@44.2.0 build: `aegir build "--bundlesize"`
ipfs-http-client: npm ERR! Exit status 1
ipfs-http-client: npm ERR! 
ipfs-http-client: npm ERR! Failed at the ipfs-http-client@44.2.0 build script.
ipfs-http-client: npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
ipfs-http-client: npm ERR! A complete log of this run can be found in:
ipfs-http-client: npm ERR!     /home/travis/.npm/_logs/2020-06-19T21_33_08_085Z-debug.log
lerna ERR! npm run build -- --bundlesize exited 1 in 'ipfs-http-client'
```

Gozala:
@achingbrain this change https://github.com/ipfs/js-ipfs/commit/87bff06e5bea51853e3ea152d1cfe69f58660f9d introduced a regression in chrome because error instances sub-classing `Error` posted across message channel loose their prototype chain and come out as `Error` instances. So if Timeout occurs in the worker thread (which is surprising because timer is also set on the client too, but sometimes worker seems to report timeout sooner, I am guessing when main thread is stressed)  on the client it will appear as `Error` instance with name "Error" instead of "TimeoutError".

It is certainly possible to encode error types differently to overcome this, that being said I am wondering if it would be better to just not rely on custom error names.

achingbrain:
we should be able to use the `.code` property found elsewhere in the codebase instead, right? If so, can you submit a PR to fix please? You may need to add the `.code` to [TimeoutError](https://github.com/ipfs/js-ipfs-utils/blob/master/src/http.js#L14-L19) I guess, and/or ensure it's persisted across the message channel.

Gozala:
> we should be able to use the `.code` property found elsewhere in the codebase instead, right? If so, can you submit a PR to fix please? You may need to add the `.code` to [TimeoutError](https://github.com/ipfs/js-ipfs-utils/blob/master/src/http.js#L14-L19) I guess, and/or ensure it's persisted across the message channel.

I'm afraid that is not going to help, as per [StructuredSerialize](https://html.spec.whatwg.org/multipage/structured-data.html#structuredserialize) spec (quoting inline below

> Otherwise, if value has an [[ErrorData]] internal slot and value is not    a [platform object](https://heycam.github.io/webidl/#dfn-platform-object), then:
>
> 1. Let name be ? [Get](https://tc39.es/ecma262/#sec-get-o-p)(value, "name").
>
> 2. If name is not one of "Error", "EvalError", "RangeError", "ReferenceError",     "SyntaxError", "TypeError", or "URIError", then set name to "Error".
>
> 3. Let valueMessageDesc be ? value.[[GetOwnProperty]]("`message`").
>
> 4. Let message be undefined if     [IsDataDescriptor](https://tc39.es/ecma262/#sec-isdatadescriptor)(valueMessageDesc) is false, and     ? [ToString](https://tc39.es/ecma262/#sec-tostring)(valueMessageDesc.[[Value]]) otherwise.
>
> 5. Set serialized to { [[Type]]: "Error", [[Name]]: name,     [[Message]]: message }.
>
> 6. User agents should attach a serialized representation of any interesting accompanying      data which are not yet specified, notably the `stack` property, to      serialized.
>
>    See the Error Stacks proposal for in-progress work on      specifying this data. [[JSERRORSTACKS\]](https://html.spec.whatwg.org/multipage/references.html#refsJSERRORSTACKS)

Error `.name` is normalized to one of the known error names,  and no custom property like `.code` is considered. I have raised following issue https://github.com/whatwg/html/issues/5665. In the meantime I can just do manual serialization / deserialization. Unless we choose to not depend on `.name` & `.code` properties.

achingbrain:
Can you removed the test skips please?  Master is a lot more stable than it was a couple of weeks ago.

packages/ipfs-message-port-client/README.md
```diff
@@ -0,0 +1,141 @@
...
+  }
+}
+
+window.onload = main
+window.onmessage = ({ports}) => {
+  IPFSClient.attach(ports[0])
+}
+```
+
+### Additional Options
```
lidel, line 87: Can we rename this section to not bury the lead, and add it to TOC at the top for easy discovery? 
```suggestion
### Notes on Performance
```

packages/ipfs-message-port-client/README.md
```diff
@@ -0,0 +1,141 @@
...
+- [Contribute](#contribute)
+- [License](#license)
+
+## Install
+
+```bash
+$ npm install --save ipfs-message-port-client
+```
+
+## Usage
```
lidel, line 30: Missing from TOC?

packages/ipfs-message-port-client/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+[![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](http://protocol.ai)
+[![](https://img.shields.io/badge/project-IPFS-blue.svg?style=flat-square)](http://ipfs.io/)
+[![](https://img.shields.io/badge/freenode-%23ipfs-blue.svg?style=flat-square)](http://webchat.freenode.net/?channels=%23ipfs)
+[![Travis CI](https://flat.badgen.net/travis/ipfs/js-ipfs)](https://travis-ci.com/ipfs/js-ipfs)
+[![Codecov branch](https://img.shields.io/codecov/c/github/ipfs/js-ipfs/master.svg?style=flat-square)](https://codecov.io/gh/ipfs/js-ipfs)
+[![Dependency Status](https://david-dm.org/ipfs/js-ipfs/status.svg?path=packages/ipfs-message-port-client)](https://david-dm.org/ipfs/js-ipfs?path=packages/ipfs-message-port-client)
+[![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat-square)](https://github.com/feross/standard)
+
+> A client library for the IPFS API over [message channel][]. This client library provides (subset) of [IPFS API](https://github.com/ipfs/js-ipfs/tree/master/docs/api) enabling applications to work with js-ipfs running in the different JS e.g. [SharedWorker][].
```
lidel, line 11:
```suggestion
> A client library for the IPFS API over [message channel][]. This client library provides (subset) of [IPFS API](https://github.com/ipfs/js-ipfs/tree/master/docs/core-api) enabling applications to work with js-ipfs running in the different JS e.g. [SharedWorker][].
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+[![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](http://protocol.ai)
+[![](https://img.shields.io/badge/project-IPFS-blue.svg?style=flat-square)](http://ipfs.io/)
+[![](https://img.shields.io/badge/freenode-%23ipfs-blue.svg?style=flat-square)](http://webchat.freenode.net/?channels=%23ipfs)
+[![Travis CI](https://flat.badgen.net/travis/ipfs/js-ipfs)](https://travis-ci.com/ipfs/js-ipfs)
+[![Codecov branch](https://img.shields.io/codecov/c/github/ipfs/js-ipfs/master.svg?style=flat-square)](https://codecov.io/gh/ipfs/js-ipfs)
+[![Dependency Status](https://david-dm.org/ipfs/js-ipfs/status.svg?path=packages/ipfs-message-port-server)](https://david-dm.org/ipfs/js-ipfs?path=packages/ipfs-message-port-server)
+[![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat-square)](https://github.com/feross/standard)
+
+> A library for providing IPFS node over [message channel][]. This library enables
+applications running in the different JS context to use [IPFS API](https://github.com/ipfs/js-ipfs/tree/master/docs/api) (subset) via `ipfs-message-port-client`.
```
lidel, line 12:
```suggestion
applications running in the different JS context to use [IPFS API](https://github.com/ipfs/js-ipfs/tree/master/docs/core-api) (subset) via `ipfs-message-port-client`.
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+  for (const port of ports.splice(0)) {
+    server.connect(port)
+  }
+}
+
+main()
+```
+
+
+### Additional Consideration
```
lidel, line 74: This section should be in TOC, for easy discovery when skimming (similar note in `README` for `-client`)
```suggestion
### Performance Consideration
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```js
+const IPFS = require('ipfs')
+const { IPFSService } = require('ipfs-message-port-server')
+const { Server } = require('ipfs-message-port-server/src/server')
+
+const main = async () => {
+  const ports = []
+  // queue connections that occur while node was starting.
+  self.onconnect = ({ports}) => connections.push(...ports)
```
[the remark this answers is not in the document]
Gozala: Oh that was an error, it meant to push into `ports` above. Thanks for catching I'll fix it.

Gozala:
> Can you removed the test skips please? Master is a lot more stable than it was a couple of weeks ago.

To be clear those changes never meant for landing, but rather to get around intermittent failures in this pull.

packages/ipfs-message-port-client/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+Client can be instantiated from the [`MessagePort`][] instance
+
+
+```js
+const IPFSClient = require('ipfs-message-port-client')
+
+
+const main = async () => {
+  const worker = new SharedWorker(IPFS_SERVER_URL)
```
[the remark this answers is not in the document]
Gozala: It depends on how you bundle things I'm afraid. I will add a section about it in the readme, because there's a lot to be said there.

packages/ipfs-message-port-protocol/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+```bash
+$ npm install --save ipfs-message-port-protocol
+```
+
+## Usage
+
+## Wire protocol codecs
+
+Library provides encode / decode functions for types that are not supported by [structured cloning algorithm][] and therefore need to be encoded before posted over [message channel][] and decoded on the other end.
```
achingbrain, line 39:
```suggestion
This module provides encode / decode functions for types that are not supported by [structured cloning algorithm][] and therefore need to be encoded before being posted over the [message channel][] and decoded on the other end.
```

packages/ipfs-message-port-protocol/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+All encoders take optional `transfer` array. If provided, encoder will add all `Transferable` fields of the given value so the they could be moved across threads without copying.
```
achingbrain, line 41:
```suggestion
All encoders take an optional `transfer` array. If provided, the encoder will add all `Transferable` fields of the given value so they can be moved across threads without copying.
```

packages/ipfs-message-port-protocol/README.md
```diff
@@ -0,0 +1,214 @@
...
+Codecs for [CID][] implementation in JavaScript.
+
+```js
+const { CID, encodeCID, decodeCID } = require('ipfs-message-port-protocol/src/cid')
+
+const cid = new CID('bafybeig6xv5nwphfmvcnektpnojts33jqcuam7bmye2pb54adnrtccjlsu')
+
+const { port1, port2 } = new MessageChannel()
+
+// Will copy underyling memory
```
achingbrain, line 54:
```suggestion
// Will copy underlying memory
```

packages/ipfs-message-port-protocol/README.md
```diff
@@ -0,0 +1,214 @@
...
+```js
+const { Block, encodeBlock, decodeBlock } = require('ipfs-message-port-protocol/src/block')
+
+const data = Buffer.from('hello')
+const cid = new CID('QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n')
+const block = new Block(data, cid)
+
+const { port1, port2 } = new MessageChannel()
+
+// Will copy underyling memory
```
achingbrain, line 81:
```suggestion
// Will copy underlying memory
```

packages/ipfs-message-port-protocol/README.md
```diff
@@ -0,0 +1,214 @@
...
+```js
+const { encodeNode, decodeNode } = require('ipfs-message-port-protocol/src/dag')
+
+
+const cid = CID('QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n')
+const dagNode = { hi: 'hello', link: cid }
+
+const { port1, port2 } = new MessageChannel()
+
+// Will copy underyling memory
```
achingbrain, line 109:
```suggestion
// Will copy underlying memory
```

packages/ipfs-message-port-protocol/README.md
```diff
@@ -0,0 +1,214 @@
...
+// On the receiver thread
+port2.onmessage = ({data}) => {
+  const dagNode = decodeNode(data)
+  dagNode.link instanceof CID // true
+}
+```
+
+### AsyncIterable
+
+Encoder allows producer to encode [async iterables][] such that it can be transferred across threads and decoded by a consumer on the other end and take care of all the IO coordination between two. Unlike other encoders `transfer` argument is mandatory (because value is encoded to a [MessagePort][] that can only be transferred). Additionally encoder / decoder take item encoder / decoder functions to encode each item of the async iterable.
```
achingbrain, line 126:
```suggestion
This encoder encodes [async iterables][] such that they can be transferred across threads and decoded by a consumer on the other end while taking care of all the IO coordination between two. Unlike other encoders the `transfer` argument is mandatory (because value is encoded to a [MessagePort][] that can only be transferred). Additionally encoder / decoder can take item encoder / decoder functions to encode each item of the async iterable.
```

packages/ipfs-message-port-protocol/README.md
```diff
@@ -0,0 +1,214 @@
...
+{
+  const transfer = []
+  port1.postMessage(
+    encodeIterable(content, chunk => chunk, transfer),
+    transfer
+  )
+}
+
+
+// Will transfer each chunk to the reciever thread (corrupting it on this thread)
```
achingbrain, line 146:
```suggestion
// Will transfer each chunk to the receiver thread (corrupting it on this thread)
```

packages/ipfs-message-port-protocol/README.md
achingbrain: If encoder/decoder functions are optional, please put them in an options argument.

packages/ipfs-message-port-protocol/src/dag.js
```diff
@@ -0,0 +1,95 @@
...
+ * the CID instances into an array so they could be turned back into CIDs
+ * without traversal on the other end.
+ *
+ * If `transfer` array is provided all the encountered `ArrayBuffer`s within
+ * this node will be added to transfer so they are moved across without copy.
+ * @param {DAGNode} dagNode
+ * @param {Transferable[]} [transfer]
+ * @returns {EncodedDAGNode}
+ */
+const encodeNode = (dagNode, transfer) => {
```
achingbrain, line 55: I still think in the first instance it would have been much simpler to just use ipld on the UI thread and serialize the node to a ArrayBuffer and then transfer it with a codec that says how to interpret the data on the other end - in the same way we do with the http client.

It might not be as fast as a no-copy memory transfer but it would have made all of this a lot more straightforward.  Serialization speed is not where we've had performance bottlenecks in the past.

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Install
+
+```bash
+$ npm install --save ipfs-message-port-server
+```
+
+## Usage
+
+This library can wrap JS IPFS node and expose it over the [message channel][].
```
achingbrain, line 35:
```suggestion
This library can wrap a JS IPFS node and expose it over the [message channel][].
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+It assumes `ipfs-message-port-client` on the other end, however it is not
+strictly necessary anything compling with wire protocol will do.
```
achingbrain, line 37:
```suggestion
strictly necessary anything complying with the wire protocol will do.
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+It provides following API subset:
+
+- [`ipfs.dag`](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/DAG.md)
+- [`ipfs.block`](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/BLOCK.md)
+- [`ipfs.add`](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#ipfsadddata-options)
+- [`ipfs.cat`](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#ipfscatipfspath-options)
+- [`ipfs.files.stat`](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#ipfsfilesstatpath-options)
+
+Server is designed to run in the [SharedWorker][] (although it is possible to
```
achingbrain, line 47:
```suggestion
The server is designed to run in a [SharedWorker][] (although it is possible to
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+run it in the other JS contexts). Example below illustrates running js-ipfs
```
achingbrain, line 48:
```suggestion
run it in the other JS contexts). The example below illustrates running a js-ipfs
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+node in [SharedWorker][] and exposing it to all connected ports
```
achingbrain, line 49:
```suggestion
node in a [SharedWorker][] and exposing it to all connected ports
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+main()
+```
+
+
+### Notes on Performance
+
+Since the data over [message channel][] is copied via
```
achingbrain, line 77:
```suggestion
Since the data sent over the [message channel][] is copied via
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+[structured cloning algorithm][] it may lead to suboptimal
```
achingbrain, line 78:
```suggestion
the [structured cloning algorithm][] it may lead to suboptimal
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+results (espacially with large binary data). In order to avoid unecessary
```
achingbrain, line 79:
```suggestion
results (especially with large binary data). In order to avoid unnecessary
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+copying server will transfer all the [Transferable][] which will be emptied
```
achingbrain, line 80:
```suggestion
copying the server will transfer all passed [Transferable][]s which will be emptied
```

packages/ipfs-message-port-server/README.md
```diff
@@ -0,0 +1,103 @@
...
+on the server side. This should not be a problem in general as IPFS node itself
+does not retain references to returned values, but is something to keep in mind
+when doig something custom.
```
achingbrain, line 83:
```suggestion
when doing something custom.
```

packages/ipfs-message-port-server/test/basic.spec.js
```diff
@@ -0,0 +1,46 @@
...
+const { expect } = require('interface-ipfs-core/src/utils/mocha')
+
+describe('dag', function () {
+  this.timeout(10 * 1000)
+
+  describe('Server', () => {
+    it('IPFSService', () => {
+      expect(IPFSService).to.be.a('function')
+      const service = new IPFSService()
+      expect(service).to.be.an.instanceOf(IPFSService)
```
achingbrain, line 15: How can this assertion fail?

packages/ipfs-message-port-server/test/basic.spec.js
```diff
@@ -0,0 +1,46 @@
...
+      expect(service)
+        .to.have.nested.property('dag.tree')
+        .be.a('function')
+    })
+    it('Server', () => {
+      expect(Server).to.be.a('function')
+      const service = new IPFSService()
+      const server = new Server(service)
+
+      expect(server).to.be.an.instanceOf(Server)
```
achingbrain, line 32: How can this assertion fail?

packages/ipfs-message-port-server/src/core.js
```diff
@@ -0,0 +1,368 @@
...
+
+  /**
+   * @typedef {Object} AddAllResult
+   * @property {RemoteIterable<AddedEntry>} data
+   * @property {Transferable[]} transfer
+
+   * @param {AddAllQuery} query
+   * @returns {AddAllResult}
+   */
+  addAll (query) {
```
achingbrain, line 133: Does this need documenting in the readme?  It says it only supports `ipfs.add`.

achingbrain:
Can you please:

1. Merge master into this branch
2. Ensure all dependency versions are aligned
    * To expand the monorepo to temporarily include the examples:
        1. Run [these commands](https://github.com/ipfs/js-ipfs/blob/master/.travis.yml#L134-L137) from the repo root
        2. Look at the end of the output from lerna, fix any warnings about dep versions differing from the hoisted version
        3. Run `npm run reset && npm i` from the repo root again until no warnings remain
        4. Ensure all examples depending on modules in the `/packages` directory depend on the same version as the one in the `/packages` directory
        5. Once everything is aligned, revert your changes to `lerna.json`

examples/browser-sharing-node-across-tabs/lerna.json
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
Gozala: It was just a simpler way to get it bootstrap independent of other examples. I'll remove it.

packages/interface-ipfs-core/src/add.js
```diff
@@ -242,7 +245,7 @@ module.exports = (common, options) => {
       const url = echoUrl(text)
 
       const [result, expectedResult] = await Promise.all([
-        ipfs.add(urlSource(url)),
+        ipfs.add(await urlSource(url)),
```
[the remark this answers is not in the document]
Gozala: Wait is `ipfs.add` supposed to take promise ? I think you're confused by the fact that I imported `urlSource` as `urlSources` (which returns async iterable) and defined local `urlSource` that turns it into promise.

I don't believe `ipfs.add` supports promise input.

packages/interface-ipfs-core/src/object/links.js
```diff
@@ -59,7 +59,7 @@ module.exports = (common, options) => {
       const node1bCid = await ipfs.object.put(node1b)
 
       const links = await ipfs.object.links(node1bCid)
-      expect(node1b.Links[0]).to.eql({
+      expect(node1b.Links[0]).to.containSubset({
```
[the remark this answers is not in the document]
Gozala: I'm not sure I understand what you want me to do here. We did discuss this here https://github.com/ipld/js-ipld-dag-pb/pull/184#discussion_r436880297 when that change was reviewed and per your request I have created following issue https://github.com/moxystudio/js-class-is/issues/27

Can you please elaborate what do here ?

packages/ipfs-message-port-server/src/core.js
Gozala: Right, I forgot to update docs after I incorporated changes from `addAll`.

packages/interface-ipfs-core/src/add.js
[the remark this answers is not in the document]
Gozala: Discussed this with @achingbrain on higher bandwidth channel. We've decided to skip these tests and figure out how to address this as a followup. I have created issue https://github.com/ipfs/js-ipfs/issues/3195 for that.

packages/ipfs-message-port-client/.aegir.js
```diff
@@ -0,0 +1,48 @@
+'use strict'
+
+const EchoServer = require('aegir/utils/echo-server')
+const echoServer = new EchoServer()
+
+module.exports = {
+  bundlesize: { maxSize: '89kB' },
```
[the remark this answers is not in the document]
Gozala: It's 74 on my machine, I'll reduce it to 80kb

packages/ipfs-message-port-client/README.md
```diff
@@ -0,0 +1,146 @@
...
+}
+```
+
+### Notes on Performance
+
+Since client works with IPFS node over [message channel][] all the data passed
+is copied via [structured cloning algorithm][], which may lead to suboptimal
+results (especially with large binary data). In order to avoid unnecessary
+copying all API options have being extended with optional `transfer` property
+that can be supplied [Transferable][]s which will be used to move corresponding
```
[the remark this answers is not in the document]
Gozala: I think boolean would be too coarse, because:

- This would force all or nothing choice, while in practice very likely it would be some.
- Things that contain transferrables include CIDs, binary data, etc... Caller of this API may not **own** some of those references (as they were passed to them by other caller) so they won't be capable of making the decision (which likely will push toward safer side of always copy).
- `{transfer: true}` would require internally arbitrary depth traversal to collect all the transferables.
- I'm not fan of this API, but it follows approach used by web APIs like [`postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage).

I would like to explore ways to improve this, because it's not ideal, but I would prefer to do that in the future.

packages/ipfs-message-port-client/src/client.js
```diff
@@ -0,0 +1,348 @@
...
+ * RPC Transport over `MessagePort` that can execute queries. It takes care of
+ * executing queries by issuing a message with unique ID and fullfilling a
+ * query when corresponding response message is received. It also makes sure
+ * that aborted / timed out queries are calcelled out as needed.
+ *
+ * It is expected that there will be at most one transport for a message port
+ * instance.
+ * @class
+ */
+class Transport {
```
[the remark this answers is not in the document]
Gozala: I've considered ripping `client.js` / `server.js` pieces into separate library. I think it may make sense but given already too many moving pieces, I would much rather land this and then extract some more general pieces later on.

Please let me know if you agree, if not I'll rip them out.

packages/ipfs-message-port-client/src/client.js
```diff
@@ -0,0 +1,348 @@
...
+ */
+class Client {
+  /**
+   * @param {string} namespace
+   * @param {ProcedureNames<T>} methods
+   * @param {Transport} transport
+   */
+  constructor (namespace, methods, transport) {
+    /** @type {RemoteService<T>} */
+    this.remote = (new Service(namespace, methods, transport))
```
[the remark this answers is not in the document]
Gozala: I'm afraid they are necessary to make TS recognize this as type casting per directive on line 323. I know that's annoying but I think it's still worth it because in return clicking methods on client takes you to implementations on sever, and stuff like that.

packages/ipfs-message-port-client/src/core.js
```diff
@@ -0,0 +1,479 @@
...
+
+/**
+ * Encodes input passed to the `ipfs.add` via the best possible strategy for the
+ * given input.
+ *
+ * @param {AddInput} input
+ * @param {Transferable[]} transfer
+ * @returns {EncodedAddInput}
+ */
+const encodeAddInput = (input, transfer) => {
```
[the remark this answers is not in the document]
Gozala: It would remove first 4 conditions, but the at the cost of: 

1. `normaliseInput` will turn everything into `Blob`, so things that could have being transferred  would be copied.
2. Introduce artificial asynchrony.
3. TS will no longer be helpful here because `normaliseInput` is untyped.

Considering cost / benefit ration I would much rather keep it as is & attempt to unify two as the followup.

packages/ipfs-message-port-client/src/core.js
```diff
@@ -0,0 +1,479 @@
...
+}
+
+/**
+ * Pattern matches given input as `Iterable<I>` and returns back either matched
+ * iterable or `null`.
+ * @template I
+ * @param {Iterable<I>|AddInput|AddAllInput} input
+ * @returns {Iterable<I>|null}
+ */
+const asIterable = (input) => {
```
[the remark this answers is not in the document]
Gozala: That's fare! It was attempting to resemble `isXXX` but maybe that is a stretch ? It is basically a poor mans patter matching found in other languages that usually manifests es:

```
const b = match a {
  Iterable =>  ...
  _ => ...
}
```
In languages where exhaustiveness isn't required (`_` case here) result is optional / maybe / nullable. 

Maybe `matchIterable` is more clear name here ?

packages/ipfs-message-port-client/test/dag.spec.js
```diff
@@ -0,0 +1,87 @@
...
+/* eslint max-nested-callbacks: ["error", 8] */
+
+'use strict'
+
+const { Buffer } = require('buffer')
+const { expect } = require('interface-ipfs-core/src/utils/mocha')
+const { DAGNode } = require('ipld-dag-pb')
+const { activate } = require('./util/client')
+
+describe('dag', function () {
```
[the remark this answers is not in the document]
Gozala: No, I had those before i figured out how to run interface tests and then forgot to remove.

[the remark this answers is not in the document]
Gozala: Will remove this.

packages/ipfs-message-port-protocol/README.md
Gozala: they aren't optional.

Gozala: Changed wording to be more clear about that.

packages/ipfs-message-port-protocol/src/dag.js
Gozala: Please note, this will become unnecessary once I'll get around to backporting this https://github.com/multiformats/js-multiformats/pull/23 into js-cids as per https://github.com/multiformats/js-cid/issues/111

If this is not a blocker, I would prefer to go with a current approach and address https://github.com/multiformats/js-cid/issues/111 to remove this encoding entirely instead of:

- Having to include codecs in both client and server bundles
- And going with a more sub-optimal approach.

packages/ipfs-message-port-client/src/block.js
```diff
@@ -0,0 +1,154 @@
...
+   * running requests started as a result of this call
+   * @param {Transferable[]} [options.transfer] - References to transfer to the
+   * worker if passed.
+   * @returns {Promise<Block>}
+   */
+  async get (cid, options = {}) {
+    const { transfer } = options
+    const { block } = await this.remote.get({
+      ...options,
+      cid: encodeCID(cid, transfer)
```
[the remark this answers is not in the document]
Gozala: I still need `encodeCID` for sending stuff from the server side. There we know everything can be transferred. And since I do have `encodeCID / decodeCID` it's actually easier to just use them across the board than use different encoding / decoding options.

packages/ipfs-message-port-client/src/client.js
```diff
@@ -0,0 +1,348 @@
...
+  /**
+   * Connects this transport to the given message port. Throws `RangeError` if
+   * transport is already connected. All the pending queries will be executed
+   * as connection occurs.
+   *
+   * @param {MessagePort} port
+   */
+  connect (port) {
+    if (this.port) {
+      throw new RangeError('Transport is already open')
```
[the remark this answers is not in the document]
Gozala: Year should probably just be an `Error`. Thinking was it's out of range for the given state of the state machine, but as I explain it, I see how little sense it makes.

packages/ipfs-message-port-client/src/client.js
```diff
@@ -0,0 +1,348 @@
...
+   * @param {Query<I, O>} query
+   * @returns {Promise<O>}
+   */
+  execute (query) {
+    const id = `${this.id}@${this.nextID++}`
+    this.queries[id] = query
+
+    // If query has a timeout is a timer.
+    if (query.timeout > 0 && query.timeout < Infinity) {
+      setTimeout(Transport.timeout, query.timeout, this, id)
```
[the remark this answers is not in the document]
Gozala: Thanks for catching this!

packages/ipfs-message-port-client/src/client.js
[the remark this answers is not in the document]
achingbrain: No rush, it can be extracted later.

packages/ipfs-message-port-client/src/client.js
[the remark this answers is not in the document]
achingbrain: This example is relatively benign so it's not going to block merging this PR, but please let's not make our code twist itself into weird contortions to satisfy the arbitrary requirements of some external tool.

PR #3083, opened 2020-06-12
docs: fix circuit relaying example

### LINKED ISSUE #2925: circuit-relaying example+tests don't work
**Version**: [bfba76748b8aabc0d66768009ff6f4adb24ff439]
**Type**: bug
**Severity**: medium
**Description**:

Following the steps and building from this example:
https://github.com/ipfs/js-ipfs/tree/master/examples/circuit-relaying

In the UI, the "Addresses:" section is empty and `info.addresses` is an empty list.  Connecting to other peers works, but not dialing `/p2p-circuit/<peerid>`.

achingbrain:
Replaces `/ipfs/Qmfoo` with `/p2p/Qmfoo`, shows available circuit relay addresses again and does more validation in the tests.

Fixes #2925

### DIFF
--- a/examples/circuit-relaying/src/helpers.js
+++ b/examples/circuit-relaying/src/helpers.js
@@ -6,6 +6,8 @@ const $message = document.querySelector('#message')
 const $msgs = document.querySelector('#msgs')
 const $addrs = document.querySelector('#addrs')
 const $peers = document.querySelector('#peers')
+const $pAddrs = document.querySelector('#peers-addrs')
+const delay = require('delay')
 
 const NAMESPACE = 'ipfs-quick-msg'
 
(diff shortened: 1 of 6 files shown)


examples/circuit-relaying/src/helpers.js
```diff
@@ -53,22 +55,64 @@ module.exports = (ipfs, peersSet) => {
...
+
+      const cons = connections.get(peerId)
+
+      for (let j = 0; j < cons.length; j++) {
+        const con = cons[j]
+        const remoteAddr = con.remoteAddr.toString()
+
+        for (let k = 0; k < 5; k++) {
+          // protocols is undefined until the connection is negotiated so try a few times
+          const protocols = ipfs.libp2p.peerStore.protoBook.get(con.remotePeer)
```
achingbrain, line 88: This was a bit weird, we've called `ipfs.swarm.connect` at this point, so we're connected, but the protobook returns undefined for a little while after connecting so we can't actually use the connection yet.

I guess the connection has not been fully negotiated - might be be worth having `libp2p.dial` only return after the connection is ready to be used?

jacobheun: You can listen for protocol changes on a peer now, https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#known-protocols-for-a-peer-change. This prevents the need to iterate over your connections as you can just wait for the relay protocol, but this is going to trigger for every peer (STOP is on by default). Currently there's no way to know if the peer is a relay without asking it if it supports `HOP`, which is currently done by trying to dial through it, AutoRelay needs to be added to js-libp2p for this, or at least the ability to check the HOP status.

```js
ipfs.libp2p.peerStore.on('change:protocols', ({ peerId, protocols}) => {})
```

The connection is ready, but identify is being run in parallel as to not block. There's nothing prohibiting you from using the connection once dial has been completed, you just might not know their supported protocols yet.

examples/circuit-relaying/src/helpers.js
```diff
@@ -53,22 +55,64 @@ module.exports = (ipfs, peersSet) => {
...
+            await delay(500)
+            continue
+          }
+
+          const isRelay = protocols.find(proto => proto.toString().includes('/circuit/relay'))
+
+          // only add addresses for peers that are relays and that we aren't already
+          // connected to them via another relay
+          if (isRelay && !remoteAddr.includes('p2p-circuit') && !remoteAddr.includes(info.id)) {
+            relayAddrs.push(`${remoteAddr}/p2p-circuit/p2p/${info.id}`)
```
jacobheun, line 100: A peer that supports the `/libp2p/circuit/relay/0.1.0` protocol isn't necessarily a relay and in the majority of cases won't be one.

achingbrain:
I've fixed this up in the [examples repo](https://github.com/ipfs-shipyard/js-ipfs-examples/tree/master/examples/circuit-relaying), following the suggestions here it gets the user to create the circuit relay address instead of trying to guess it for them.

PR #3084, opened 2020-06-12
Fixed wrong folder

Neurone:


achingbrain:
Thanks for opening this.  The instructions are correct, you should run `npm i` and `npm run build` in the root of the repo.  This will install all the deps and symlink correct versions of the modules everywhere.

Have you run into a problem with this?

Neurone:
Examples are separated node projects, so if you want to execute them you need to be in those root folders. This is the case for all examples, not only browser-browserify, so maybe you should extend this fix to all those readme.md also.

achingbrain:
You are of course free to run `npm i` in individual example folders, but following the instructions as written will have the main ipfs package in this module built and symlinked into the `node_modules` folder which is better for experimentation.

Again, have you encountered an error or some situation where this isn't working?

Neurone:
It works, but it takes a lot of time. I created a new PR with a better modification and wording.

PR #3088, opened 2020-06-16
chore: support current and active LTS node

achingbrain:
Version numbers taken from https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions


achingbrain:
Refs: https://github.com/ipfs/community/pull/466

PR #3092, opened 2020-06-17
chore: update libp2p runtime config

vasco-santos:
This PR removes the websocketStar discovery configuration, since it is not used anymore.

In addition, I have changed the discovery keys to use the modules tag from the module. This has created some confusion to libp2p users, and we would like people to use the tag instead of magic keys that might also be changed in the future: https://github.com/libp2p/js-libp2p/issues/664#issuecomment-642249349

### DIFF
--- a/packages/ipfs/src/core/runtime/libp2p-browser.js
+++ b/packages/ipfs/src/core/runtime/libp2p-browser.js
@@ -39,7 +39,8 @@ module.exports = () => {
         bootstrap: {
           enabled: true
         },
-        [WebRTCStar.tag]: {
+        // [WebRTCStar.discovery.tag]
+        webRTCStar: {
           enabled: true
         }
       },
--- a/packages/ipfs/src/core/runtime/libp2p-browser.js
+++ b/packages/ipfs/src/core/runtime/libp2p-browser.js
@@ -35,8 +35,6 @@ module.exports = () => {
     config: {
       peerDiscovery: {
         autoDial: true,
-        // Optimization
-        // Requiring bootstrap inline in components/libp2p to reduce the cli execution time
         // [Bootstrap.tag] = 'bootstrap'
         bootstrap: {
           enabled: true
--- a/packages/ipfs/src/core/runtime/libp2p-browser.js
+++ b/packages/ipfs/src/core/runtime/libp2p-browser.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const Bootstrap = require('libp2p-bootstrap')
 const WS = require('libp2p-websockets')
 const WebRTCStar = require('libp2p-webrtc-star')
 const Multiplex = require('libp2p-mplex')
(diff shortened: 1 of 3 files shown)


packages/ipfs/src/core/runtime/libp2p-browser.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+const Bootstrap = require('libp2p-bootstrap')
```
vasco-santos, line 3: I noticed that Bootstrap is only required in `components/libp2p`. Is this a bundle optimization?

vasco-santos: cc @achingbrain

achingbrain: AFAIK the inline requires are to speed up node start-up time.

vasco-santos: ok, so I will add a comment for that and keep the Bootstrap as it was and only use the tag for the other one

jacobheun: iirc the inline requires were done to reduce the cli execution time, not startup time (ie: if it's not needed for interacting with the running daemon, avoid requiring it). Not sure if this is still a problem with how things are setup internally.

vasco-santos: Should I change the comment to: `Requiring bootstrap inline in components/libp2p to reduce the cli execution time`?

jacobheun: I don't think that's really accurate though. In reality the whole libp2p component should be ignored in CLI, as I don't think it's used at all (unless I'm mistaken). I'd leave it for now, and we can clean that all up when we update libp2p configuration.

packages/ipfs/src/core/runtime/libp2p-browser.js
```diff
@@ -35,13 +35,13 @@ module.exports = () => {
     config: {
       peerDiscovery: {
         autoDial: true,
+        // Optimization
+        // Requiring bootstrap inline in components/libp2p to reduce the cli execution time
```
achingbrain, line 39: This is browser config, the reference to cli execution time isn't relevant here.

achingbrain:
Tests are failing:

```
ipfs: FAILED TESTS:
ipfs:   libp2p customization
ipfs:     options
ipfs:       ✖ should use options by default
ipfs:         Chrome Headless 84.0.4147.89 (Linux x86_64)
ipfs:       AssertionError: expected { Object (dht, peerDiscovery, ...) } to have deep property 'peerDiscovery' of { Object (autoDial, bootstrap, ...) }, but got { Object (autoDial, bootstrap, ...) }
ipfs:       + expected - actual
ipfs:          }
ipfs:          "mdns": {
ipfs:            "enabled": false
ipfs:          }
ipfs:       -  "undefined": {
ipfs:       -    "enabled": true
ipfs:       -  }
ipfs:          "webRTCStar": {
ipfs:            "enabled": false
ipfs:          }
ipfs:        }
```

Is the `.tag` property missing from one of the peer discovery modules?  Does a dep need updating?

vasco-santos:
Yeah, we cannot use the tag in webrtc after all 😞 
https://github.com/libp2p/js-libp2p-webrtc-star/blob/master/src/index.js#L57
We need the instance of webrtc to access it. All the other discovery have it exported, but as webrtc is also a transport it is inside the discovery scope.

I will have this stuff properly addressed when we improve the libp2p config. For now we can include a comment to let users know where the tag comes from

PR #3093, opened 2020-06-17
test: replace ipfs.add with unixfs importer in get.js

koivunej:
Continuing the same theme, though this might be coming to an end as next up we should start implementing the `/add` :)


achingbrain:
I think this is a good exercise - it'll help the next implementation too!

PR #3106, opened 2020-06-19
fix: examples to match current implementation

Gozala:
Discovered some inconsistency between docs and an implementation of `ipfs.block` API


achingbrain:
Good spot, thanks!

PR #3118, opened 2020-06-24
fix: optional arguments go in the options object

achingbrain:
We have a few older APIs that take multiple optional arguments, which makes our code more complicated as it has to guess the users' intent, sometimes by inspecting properties on the passed args to see if they happen to correspond with properties on the actual options object.

The options object was recently added to all API calls and is the right place for optional arguments to go, so the change here is to move all optional arguments into the options object, except where the presence of an optional argument dramatically changes the behaviour of the call (`ipfs.bootstrap` I'm mostly looking at you), in which case the methods are split out into multiple versions that do distinct things.

Only the programatic API is affected, the CLI and HTTP APIs do not change.

BREAKING CHANGES:

- `ipfs.bitswap.wantlist([peer], [options])` is split into:
  - `ipfs.bitswap.wantlist([options])`
  - `ipfs.bitswap.wantlistForPeer(peer, [options])`
- `ipfs.bootstrap.add([addr], [options])` is split into:
  - `ipfs.bootstrap.add(addr, [options])` - add a bootstrap node
  - `ipfs.bootstrap.reset()` - restore the default list of bootstrap nodes
- `ipfs.bootstrap.rm([addr], [options])` is split into:
  - `ipfs.bootstrap.rm(addr, [options])` - remove a bootstrap node
  - `ipfs.bootstrap.clear([options])` - empty the bootstrap list
- `ipfs.dag.get(cid, [path], [options])` becomes `ipfs.dag.get(cid, [options])`
  - `path` is moved into the `options` object
- `ipfs.dag.tree(cid, [path], [options])` becomes `ipfs.dag.tree(cid, [options])`
  - `path` is moved into the `options` object
- `ipfs.dag.resolve(cid, [path], [options])` becomes `ipfs.dag.resolve(cid, [options])`
  - `path` is moved into the `options` object
- `ipfs.files.flush([path], [options])` becomes `ipfs.files.flush(path, [options])`
- `ipfs.files.ls([path], [options])` becomes `ipfs.files.ls(path, [options])`
- `ipfs.object.new([template], [options])` becomes `ipfs.object.new([options])`
  - `template` is moved into the `options` object
- `ipfs.pin.ls([paths], [options])` becomes `ipfs.pin.ls([options])`
  - `paths` is moved into the `options` object


achingbrain:
Follows on from https://github.com/ipfs/js-ipfs/pull/3071

docs/core-api/BITSWAP.md
```diff
@@ -16,22 +16,21 @@
...
-| peerId | [PeerId][], [CID][], `String` or `Buffer` | An optional peer ID to return the wantlist for |
+None
 
 ### Options
 
 An optional object which may have the following keys:
 
 | Name | Type | Default | Description |
 | ---- | ---- | ------- | ----------- |
+| peer | [PeerId][], [CID][], `String` or `Buffer` | A peer ID to return the wantlist for |
```
hugomrdias, line 33: This is an array of peers ? and the default can probably be more descriptive like Self Peer ID or something.

docs/core-api/BOOTSTRAP.md
```diff
@@ -71,6 +78,48 @@ console.log(res.Peers)
...
+| Name | Type | Default | Description |
+| ---- | ---- | ------- | ----------- |
+| timeout | `Number` | `undefined` | A timeout in ms |
+| signal | [AbortSignal][] | `undefined` |  Can be used to cancel any long running requests started as a result of this call |
+
+### Returns
+
+| Type | Description |
+| -------- | -------- |
+| `Promise<Object>` | An object that contains an array with all the added addresses |
```
hugomrdias, line 102:
```suggestion
| `Promise<{ Peers: Array<MultiAddr> }>` | An object that contains an array with all the added addresses |
```

docs/core-api/BOOTSTRAP.md
```diff
@@ -113,27 +162,64 @@ console.log(res.Peers)
...
+## `ipfs.bootstrap.rm(addr, [options])`
 
 > Remove a peer address from the bootstrap list
 
 ### Parameters
 
 | Name | Type | Description |
 | ---- | ---- | ----------- |
-| addr | [MultiAddr][] | The address of a network peer. If omitted the `all` option must be passed as `true`. |
+| addr | [MultiAddr][] | The address of a network peer |
```
hugomrdias, line 173:
```suggestion
| addr | Array<MultiAddr> | The address of a network peer |
```
This is no longer optional, right ?

docs/core-api/BOOTSTRAP.md
```diff
@@ -113,27 +162,64 @@ console.log(res.Peers)
...
-| all | `boolean` | `false` | If true, remove all peers from the bootstrap list |
 | timeout | `Number` | `undefined` | A timeout in ms |
 | signal | [AbortSignal][] | `undefined` |  Can be used to cancel any long running requests started as a result of this call |
 
-Note: If passing the `all` option, `addr` is an optional parameter (may be `undefined`/`null`) and options may be passed as the first argument. i.e. `ipfs.bootstrap.rm({ all: true })`
+### Returns
+
+| Type | Description |
+| -------- | -------- |
+| `Promise<Object>` | An object that contains an array with all the removed addresses |
```
hugomrdias, line 188:
```suggestion
| `Promise<{ Peers: Array<MultiAddr> }>` | An object that contains an array with all the removed addresses |
```

docs/core-api/DAG.md
```diff
@@ -65,23 +65,23 @@ console.log(cid.toString())
...
 
 > Retrieve an IPLD format node
 
 ### Parameters
 
 | Name | Type | Description |
 | ---- | ---- | ----------- |
-| cid | [CID][] | A DAG node that follows one of the supported IPLD formats |
-| path | `String` | An optional path within the DAG to resolve |
+| cid | [CID][] | A CID that resolves to a node to get |
```
hugomrdias, line 76: why do you right the type like that ? for me this `[CID][]` means an optinal array of cids

docs/core-api/BITSWAP.md
achingbrain: No, it's a Markdown link that's in the footer of the document.

achingbrain: Like this: https://www.markdownguide.org/basic-syntax/#reference-style-links

docs/core-api/BOOTSTRAP.md
achingbrain: Also a reference link.  But quite right, the note about `all` option should be removed.

docs/core-api/DAG.md
achingbrain: Also a reference link, not an array.

docs/core-api/BITSWAP.md
hugomrdias: got it, but the default is the node's peerid right ?

achingbrain: In effect, but there are actually two different methods we call in the bitswap module: https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/core/components/bitswap/wantlist.js#L15-L16

We could split this into two methods at the Core API level too?

hugomrdias: I don't we need to do that, not much to be gained there.
Just need to make it clear that if the user doesn't provide a peer in the options the node's peer id is used

packages/interface-ipfs-core/src/bitswap/utils.js
```diff
@@ -9,7 +9,9 @@ async function waitForWantlistKey (ipfs, key, opts = {}) {
   const end = Date.now() + opts.timeout
 
   while (Date.now() < end) {
-    const list = await ipfs.bitswap.wantlist(opts.peerId)
+    const list = await ipfs.bitswap.wantlist({
+      peer: opts.peerId
```
hugomrdias, line 13: maybe the prop name should be peerId ?

packages/ipfs-http-client/src/files/ls.js
```diff
@@ -7,9 +7,8 @@ const toUrlSearchParams = require('../lib/to-url-search-params')
 
 module.exports = configure(api => {
   return async function * ls (path, options = {}) {
-    if (typeof path !== 'string') {
-      options = path || {}
-      path = '/'
+    if (!path || typeof path !== 'string') {
+      throw new Error('ipfs.files.flush requires a path')
```
hugomrdias, line 11:
```suggestion
      throw new Error('ipfs.files.ls requires a path')
```

docs/core-api/BITSWAP.md
achingbrain: You know, the more I think about it, the more I think it's a separate operation.

achingbrain: I've made it a separate operation.

packages/interface-ipfs-core/src/bitswap/utils.js
achingbrain: It's a separate operation now.

PR #2522, opened 2019-10-08
Migration of Blockstore to use multihash instead of CID as key

AuHau:
Related to https://github.com/ipfs/js-ipfs/issues/2415

Since DHT is not enabled, I have not worked on that part.

#### TODO

 - [ ] Bump `js-ipfs-repo` version upon release
 - [ ] Bump `interface-js-ipfs-core` version upon release
 - [ ] Change DHT to use multihashes instead CIDs 
 - [ ] Fix GC tests that I think never properly worked, but I have idea about what is wrong, seems like `ipfs.block.rm()` actually does not delete the blocks...

#### Blocked by

 - [x] https://github.com/ipfs/js-ipfs-repo-migrations/pull/1
 - [ ] https://github.com/ipfs/js-ipfs-repo-migrations/pull/4
 - [ ] https://github.com/ipfs/interface-js-ipfs-core/pull/540
 - [ ] https://github.com/ipfs/js-ipfs-repo/pull/211
 - [ ] https://github.com/ipfs/js-ipfs-http-client/pull/1128

PR #3124, opened 2020-06-25
feat: store blocks by multihash instead of CID

achingbrain:
Updates the `ipfs-repo` dep to a version that stores blocks by multihash instead of CID to support CIDv1 and CIDv0 access to the same block.

New features:

- Adds a `--multihash` argument to the cli command `ipfs refs local` which prints the base32 encoded multihash of each block

BREAKING CHANGES:

- `ipfs.refs.local` now returns a v1 CID with the raw codec for every block and not the original CID by which it was added to the blockstore


achingbrain:
Supersedes #2522

achingbrain:
@aschmahmann could I get a 👍 or 👎  on the cli change please?  It would be a shame to ship this and then have go-ipfs implement something different later down the line.

The change is:

* `ipfs refs local` used to print out CIDs of blocks in your repo
* Now it prints v1 raw CIDs instead
* There's a new `--multihash` argument to `ipfs refs local` which prints base32 encoded multihashes instead of the v1 raw CIDs

aschmahmann:
@achingbrain I'm having trouble finding which issue describes what we agreed on here. This change seems reasonable to me, but mostly because I'm not sure what people are using `ipfs refs local` for... I recall we use `ipfs refs {cid}` with the preload nodes but that code path is left unchanged.

Any pointers?

achingbrain:
The issue & decision is here: https://github.com/ipfs/specs/issues/242#issuecomment-627533056

Also relevant: https://github.com/ipfs/go-ipfs/issues/6815

aschmahmann:
Neither of those issues seem to explicitly reference the changes to `ipfs refs local`. Although I guess it's probably implied by those issues since there may be orphaned (e.g. unreferenced by a pin) blocks that we just can't figure out the CIDs for.

achingbrain:
> I guess it's probably implied by those issues since there may be orphaned (e.g. unreferenced by a pin) blocks that we just can't figure out the CIDs for

Exactly, it's more of a practical implication of dropping the original CID a block was stored by.

packages/ipfs/src/cli/commands/refs-local.js
```diff
@@ -1,27 +1,40 @@
...
       timeout
     })) {
-      if (ref.err) {
-        print(ref.err, true, true)
+      if (err) {
+        print(err, true, true)
       } else {
-        print(ref.ref)
+        if (multihash) {
+          print(multibase.encode('base32', Buffer.from(ref)).toString().substring(1).toUpperCase())
```
hugomrdias, line 34:
```suggestion
          print(multibase.encoding('base32upper').encode(Buffer.from(ref)))
```
with multibase v1 we can just do this or

```js
multibase.names['base32upper'].encode(Buffer.from(ref))
```

packages/ipfs/test/cli/refs-local.js
```diff
@@ -37,6 +39,23 @@ describe('refs local', () => {
...
+    ipfs.refs.local.withArgs(defaultOptions).returns([{
+      ref
+    }, {
+      err
+    }])
+
+    const out = await cli('refs local --multihash', { ipfs })
+    const lines = out.split('\n')
+
+    expect(lines.includes(multibase.encode('base32', Buffer.from(ref)).toString().substring(1).toUpperCase())).to.be.true()
```
hugomrdias, line 55:
```suggestion
    expect(lines.includes(multibase.encoding('base32upper').encode(Buffer.from(ref))).to.be.true()
```

aschmahmann:
Seems reasonable 👍 from me then. If @hsanjuan or @Stebalien have a second to confirm that'd be great, since they've been more involved with this than I have. However, if they're busy then no complaints from me

packages/ipfs/src/cli/commands/refs-local.js
achingbrain: Neato, I didn't realise.

Stebalien:
@achingbrain that's the planned change for go-ipfs as well.

oed:
So does this change mean that all data already stored in an ipfs node will not be retrievable locally when one updates to v0.48? 
Is there a data migration strategy that can be used here?

achingbrain:
No, there should be no change. A [blockstore migration](https://github.com/ipfs/js-ipfs-repo-migrations/pull/4) takes place when you first start a node running 0.48.0 on a non-migrated repo to convert all the keys so what is currently retrievable should still be retrievable after the upgrade.

PR #3126, opened 2020-06-25
chore: update example with public webrtc servers

vasco-santos:
Per #2779 , this PR adds the public webrtc servers to the browser example.

This allows users to run easily the example, but also provides information for how users should use a signaling server for production.


vasco-santos:
Should we add it to the config browser swarm addresses?

vasco-santos:
> Please leave the test as starting it's own webrtc-star server - our tests should not have dependencies on external infrastructure that maybe down or otherwise unreachable.

Sure, I agree with that! I just want to make sure the tests work first with the setup infrastructure

PR #3128, opened 2020-06-26
chore: browser example with public webrtc servers

vasco-santos:
Replaces #3126 as the examples cache seems impossible to properly clean.

This allows users to run easily the example, but also provides information for how users should use a signaling server for production.


vasco-santos:
Closing this as the problem does not seem related to cache

Issue #2962, opened 2020-03-31
dag.resolve has different return type in http-client


alanshaw:
In `ipfs-http-client` it returns `Promise<{ cid: CID, remPath: string }>` but in `ipfs` it returns `AsyncIterable<{ value: CID|any, remainderPath: string }>`

tarunbatra:
I can take this up. Would we want `dag.resolve` in `ipfs-http-client` to return `AsyncIterable` too? However, not sure why `AsyncIterable` should be returned by `dag.resolve` in `ipfs`.

PS: Might need some help as I'm new to the community.

achingbrain:
It returns an async iterator because we can use it to resolve deeply nested paths inside DAGs - we may end up traversing to other nodes in which case we yield the intermediate nodes along the way.

If I'm honest I don't know how useful this is.

The HTTP API also only returns the last result so although it'd be good for the HTTP API client and the core to be consistent, the API client is limited by what the HTTP API will return, so it might be better to just change the core to return the last result instead of the iterator as the HTTP API client has no way of returning the intermediate nodes.

This is easily done by changing the function signature & taking the logic from the [http api endpoint](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/http/api/resources/dag.js#L327-L354) and moving it into the [core api function](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/core/components/dag/resolve.js) (n.b. it should still return `{ value, remainderPath }`, not `{ Cid, RemPath }`.

There should be interface tests for `dag.resolve` in [this folder](https://github.com/ipfs/js-ipfs/tree/master/packages/interface-ipfs-core/src/dag) but they appear to be missing so some should be added as part of this piece of work.

`ipfs.dag.resolve` should also be added to the [DAG Core API Docs](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/DAG.md).

So, to land this:

- [ ] Move logic from HTTP API into core
- [ ] Add interface tests
- [ ] Add core-api documentation

tarunbatra:
@achingbrain Thanks for the write-up.

Based on how you have explained, using `Promise` does makes sense to me too. We always can create a different method or can parametrize this method to return intermediary nodes, if required in some use case, later.

So I'll start work on the pointers you mentioned. Thanks.

PR #3131, opened 2020-06-29
fix(core): return promise from 'dag.resolve'

tarunbatra:
For #2962

### DIFF
--- a/packages/ipfs/src/core/components/dag/resolve.js
+++ b/packages/ipfs/src/core/components/dag/resolve.js
@@ -1,16 +1,36 @@
 'use strict'
 
+const CID = require('cids')
 const { parseArgs } = require('./utils')
 const { withTimeoutOption } = require('../../utils')
 
 module.exports = ({ ipld, preload }) => {
-  return withTimeoutOption(async function * resolve (cid, path, options) { // eslint-disable-line require-await
+  return withTimeoutOption(async function resolve (cid, path, options) {
     [cid, path, options] = parseArgs(cid, path, options)
 
     if (options.preload !== false) {
       preload(cid)
     }
 
-    yield * ipld.resolve(cid, path, { signal: options.signal })
+    let lastCid = cid
+    let lastRemainderPath = path
+
+    if (path) {
+      for await (const { value, remainderPath } of ipld.resolve(cid, path, {
+        signal: options.signal
+      })) {
+        if (!CID.isCID(value)) {
+          break
+        }
+
+        lastRemainderPath = remainderPath
+        lastCid = value
+      }
+    }
+
+    return {
+      value: lastCid,
+      remainderPath: lastRemainderPath || ''
+    }
   })
 }


tarunbatra:
@achingbrain I'm working on adding tests and documentation. Wanted to validate my approach with you.

Apart from adding tests in [interface-ipfs-core/src/dag/resolve.js](https://github.com/ipfs/js-ipfs/tree/master/packages/interface-ipfs-core/src/dag), would we also want to have tests in [ipfs/test/core/dag.spec.js](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/test/core/dag.spec.js). What is the difference there?

packages/ipfs/src/core/components/dag/resolve.js
```diff
@@ -1,16 +1,45 @@
...
+        }
+      }
+
+      return {
+        value: lastCid,
+        remainderPath: lastRemainderPath || ''
+      }
+
+    } finally {
+      if (release) {
```
achingbrain, line 40: Where is `release` defined?

packages/ipfs/src/core/components/dag/resolve.js
```diff
@@ -1,16 +1,45 @@
 'use strict'
 
+const CID = require('cids')
 const { parseArgs } = require('./utils')
 const { withTimeoutOption } = require('../../utils')
 
 module.exports = ({ ipld, preload }) => {
-  return withTimeoutOption(async function * resolve (cid, path, options) { // eslint-disable-line require-await
+  return withTimeoutOption(async function resolve (cid, path, options) { // eslint-disable-line require-await
```
achingbrain, line 8: This function now includes an `await` so `eslint-disable-line require-await` is not necessary.

packages/ipfs/src/core/components/dag/resolve.js
```diff
@@ -1,16 +1,45 @@
...
 
-    yield * ipld.resolve(cid, path, { signal: options.signal })
+    try {
+      let lastCid = cid
+      let lastRemainderPath = path
+
+      if (path) {
+        for await (const { value, remainderPath } of ipld.resolve(cid, path, {
+          signal,
+          timeout
```
achingbrain, line 23: `ipld.resolve` does not support a `timeout` option.

achingbrain:
> Apart from adding tests in interface-ipfs-core/src/dag/resolve.js, would we also want to have tests in ipfs/test/core/dag.spec.js. What is the difference there?

The tests in `interface-ipfs-core` are the core API tests - that is, everything in [/docs/core-api](https://github.com/ipfs/js-ipfs/tree/master/docs/core-api) should have tests in that package.  They are run against [ipfs core](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs/src/core), against the [http API](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs/src/http) via [ipfs-http-client](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-http-client) and against `go-ipfs` also via `ipfs-http-client`.

The tests in [/packages/ipfs/test](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs/test) cover functionality not in the core-api.

Since the idea here is to document this method in the core api, tests should be in the `interface-ipfs-core` pacakge.

packages/ipfs/src/core/components/dag/resolve.js
tarunbatra: Removed it.

packages/ipfs/src/core/components/dag/resolve.js
tarunbatra: Gotcha! Missed looking up this [doc](https://github.com/ipld/js-ipld#resolvecid-path-options). Removed the `timeout` option.

packages/ipfs/src/core/components/dag/resolve.js
tarunbatra: `release` was supposed to be initiated by `release = gcLock.readLock()` which i picked up from [./put.js](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/core/components/dag/put.js#L45). But now I dont think it makes sense here. So removed this block.

achingbrain: It's in `ipfs.dag.put` because adding blocks and pinning are async operations so gc could run between them which would result in the blocks being removed immediately after they were added.  

We're only reading here, not writing so the lock is unnecessary.

achingbrain:
Fixed by #3152 - I added you as a co-author so you'll appear in the contributors list. Thanks!

tarunbatra:
Thanks @achingbrain. I planned to write tests and documentation but couldn't find time. #3152 looks great. Appreciate your help!

achingbrain:
No worries, sometimes the real world gets in the way - thanks for kicking it off.

PR #3132, opened 2020-06-29
fix: still load dag-pb, dag-cbor and raw when specifying custom formats

### LINKED ISSUE #3129: IPLD formats are overwritten in nodejs
- **Version**: 0.46.0

- **Platform**: nodejs

- **Subsystem**: ipld

#### Description:
When specifying a custom ipld format when creating an ipfs instance the default formats get overwritten. This seems to be happening since the default formats are not defined in the nodejs config for ipld: https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/core/runtime/ipld-nodejs.js (Ipld just uses the passed formats array without merging with it's defaults)

In the browser config it looks like it should work since the passed formats array should be merged with the default one, but I have not confirmed.

Run this in nodejs with a custom `myFormat`.

```js
const ipfs = await Ipfs.create({ ipld: { formats: [myFormat] } })
```

achingbrain:
If we specify a `formats` array as part of the ipld options in in-proc nodes, it replaces the default list of dag-pb, dag-cbor and raw.

This change allows the `loadFormat` function to still resolve those formats even if the user has passed a `format` array that does not contain them.

Fixes #3129

### DIFF
--- a/packages/ipfs-http-client/test/dag.spec.js
+++ b/packages/ipfs-http-client/test/dag.spec.js
@@ -5,7 +5,8 @@
 
 const { Buffer } = require('buffer')
 const { expect } = require('interface-ipfs-core/src/utils/mocha')
-const { DAGNode } = require('ipld-dag-pb')
+const ipldDagPb = require('ipld-dag-pb')
+const { DAGNode } = ipldDagPb
 const CID = require('cids')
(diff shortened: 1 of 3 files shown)


oed:
Thanks @achingbrain :)

PR #3140, opened 2020-07-03
docs(core-api): fix default value column

lidel:
@achingbrain was unsure if its ok to commit such tiny change to `master`, so opened a PR, but let me know if you prefer silent fixes of things like that in the future.


achingbrain:
PRs are fine ([js-ipfs-contributors](https://www.npmjs.com/package/js-ipfs-contributors) will see you are added to the release notes that way too), thanks for the fix.

PR #3148, opened 2020-07-07
feat: turn on delegate nodes by default

achingbrain:
To help find content and peers on the network, turn on delegate nodes by default.

Disables delegate nodes while using the `test` config profile.


achingbrain:
I'd like some thoughts on this - it'll help with peers and DHT queries, but provides will only work for an hour or so (until GC runs on the delegate node).

Does GC clear up DHT records from the datastore?

Do we need some way of perhaps not doing the provide so we can get the `findpeer` and `findprovs` goodness without making content harder to resolve?

vasco-santos:
I was thinking earlier today, if a `go-ipfs` node acting as a delegated router would run the reprovider for the records it received. I am not sure

So, in `js-ipfs` we will need the reprovider with the DHT for what you just described. But, using a third party node DHT, would be the responsibility of this third party node to reprovide? or would it be the peer who asked for it?

Issue #3138, opened 2020-07-03
Implementation bug in normaliseInput


Gozala:
I am pretty sure following lines would throw `TypeError: iterator is not iterable` because `yield *` expression expects iterable and not an iterator.  

https://github.com/ipfs/js-ipfs/blob/8cb8c73037e44894d756b70f344b3282463206f9/packages/ipfs-core-utils/src/files/normalise-input.js#L73-L82

https://github.com/ipfs/js-ipfs/blob/8cb8c73037e44894d756b70f344b3282463206f9/packages/ipfs-core-utils/src/files/normalise-input.js#L114-L122

https://github.com/ipfs/js-ipfs/blob/8cb8c73037e44894d756b70f344b3282463206f9/packages/ipfs-core-utils/src/files/normalise-input.js#L191-L199

It is also worrying that no tests seem to catch that.

Gozala:
Also it appears that parenthesis is in the wrong place here

https://github.com/ipfs/js-ipfs/blob/8cb8c73037e44894d756b70f344b3282463206f9/packages/ipfs-core-utils/src/files/normalise-input.js#L52

achingbrain:
Please can you submit a PR with a failing test, and preferably, a fix.

Gozala:
I'm working on the patch that adds tests and fixes.

Gozala:
Here is the example that illustrates the issue:

```js
class IterableOf {
  constructor(...entries) {
    this.entries = entries
  }
  [Symbol.iterator]() {
    return new IteratorOf(...this.entries)
  }
}

class IteratorOf {
  constructor(...entries) {
    this.position = 0
    this.entries = entries
  }
  next() {
    if (this.position < this.entries.length) {
      return { done: false, value: this.entries[this.position++] }
    } else {
      return { done: true }
    }
  }
}

function * test () {
  const iterable = new IterableOf(1, 2, 3)
  const iterator = iterable[Symbol.iterator]()
  
  yield iterator.next().value
  yield * iterator
}

[...test()] // Uncaught TypeError: iterator is not iterable
```

I should note that browser built-ins (including generators) tend to return iterators that effectively have `[Symbol.iterator]() { return this }` and there for this doesn't appear like a problem. That said iterables are not required to return iterators are also iterables.

achingbrain:
I don't think you can use `yield *` like that.  If I run your example I get:

```
  yield * iterator
  ^

TypeError: undefined is not a function
```

[...]

PR #3151, opened 2020-07-07
feat: non-bufferring multipart body encoder

### LINKED ISSUE #3029: Embracing web native FormData / File where possible
As discussed at https://github.com/ipfs/js-ipfs/issues/2838#issuecomment-599688414 mainstream browsers today do not yet support passing`ReadableStream` as request body. To workaround this limitation currently JS-IPFS encodes stream of content (on ipfs.add / ipfs.write) into multipart/form-data by buffering it:

https://github.com/ipfs/js-ipfs/blob/cb5b9ec420431e43c26d183f76e2bd5e4a8bdf77/packages/ipfs-http-client/src/add.js#L13-L23
https://github.com/ipfs/js-ipfs/blob/cb5b9ec420431e43c26d183f76e2bd5e4a8bdf77/packages/ipfs-http-client/src/lib/multipart-request.js#L58-L64

https://github.com/ipfs/js-ipfs/blob/cb5b9ec420431e43c26d183f76e2bd5e4a8bdf77/packages/ipfs-http-client/src/lib/to-stream.browser.js#L11-L21

This is problematic for webui for instance as some files may not fit into memory or exceed browser quota.

---

Modern browsers do have native `FormData` and `File` / `Blob` primitives that if used carefully can facilitate streaming. It appears that in the past browser loaded all of the FormData / File / Blob content during `fetch` despite not reading it. However that evidently had being fixed and modern browsers (Tested with Firefox, Chrome, Safari) no longer exhibit that problem. 

Following example was used to verify this behavior
https://ipfs.io/ipfs/QmWVrTAeA1FqRSK3owpCfvB69wsz2Dbr2THK6ehn5uEKbp

On my Mac I do not observe visible increase memory when I uploading 5 Gig file (across mentioned browsers). I also observe streaming behavior as as file size change in WebUI is changing during upload:

Screen Shot 2020-05-08 at 3 33 55 PM
Screen Shot 2020-05-08 at 3 36 54 PM
 
---

This investigation suggests that there is an opportunity to avoid buffering overhead in browsers. It also appears that doing it for `ipfs.files.write` should be fairly straight forward since it always writes a single file.

For `ipfs.add` things appear more complicated, as headers need to be added to each multipart part (E.g. unix mode & mtime). Some investigation is required to see how visible that is with native `FormData`. There are also some concerns about supporting directory structures.

Gozala:
> Context: 
> - This PR aims to improve performance of `ipfs.add` in `ipfs-http-client` in browser context by addressing findings from https://github.com/ipfs/js-ipfs/issues/3029.  
> - Alternative approach can be found in https://github.com/ipfs/js-ipfs/pull/3184

### Status

- Implementation is complete, some tests need to be updated.

###### Before

`normaliseInput` used to normalize arbitrary input taken by `ipfs.add` into `AsyncIterable<FileObject>` where `FileObject` is:
```ts
type FileObject = {
  path:string,
  content?: AsyncIterable<ArrayBufferView|ArrayBuffer>,
  mtime?: number | [number, number] | Date | { secs: number, nsecs?: number },
  mode:string|number
}
```

There was (implicit) invariant that if `FileObject` doesn't have `content` it represents a directory.

However representing `content` as `AsyncIterable<ArrayBufferView|ArrayBuffer>` is what lead to [buffering in the browser as `fetch` still does support stream body](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/lib/to-stream.browser.js#L3-L8). 

###### After

This patch changes `normaliseInput` to produce a different output: `AsyncIterable<ExtendedFile|FileStream|Directory>` where

- `Directory` is just like `FileObject` and does not have `content`.
- `ExtendedFile` represents a `FileObject` with known size
    - It is a subclass of the [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File)
      - Polyfill of [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) is used in node
         - Polyfill of [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob) is used in node
    - It is created only from inputs of known sizes like strings, buffers, blobs, byte arrays etc... 
    - It adds `mtime`, `mode` and `path` properties (assumed by ipfs-unixfs-importer).
    - It adds `content` getter which returns `AsyncIterable<Uint8Array>` of it's parts,  which creates compatibility with `FileObject` interface.
- `FileStream` is just like `FileObject` that does have a `content`.
    - It is created only when input is of unknown size like `AsyncIterable<*>` 
    - It is different from `ExtendedFile` because `multipartRequest` can't add it to the `FormData` without buffering it's body, while it can do that with `ExtendedFile`.

#### Multipart Encoder

New `FormDataEncoder` class was added that provides can encode `AsyncIterable<Part>` into `AsyncIterable<BlobPart>` representing body of the multipart request, where `Part` is:

```ts
type Part = {
  name: string,
  content: void|Blob|AsyncIterable<ArrayBufferView|ArrayBuffer>,
  filename?: string,
  headers?: Record<string, string>
}
```

`to-stream` module had being replaced by `to-body` which turns `AsyncIterable<BlobPart>` to readable stream on node and into `Blob` in browser.

With above pieces in place `multipartRequest` now
1. normalized input into `AsyncIterable<ExtendedFile|FileStream|Directory>`
2. Turns it into intermediate representation of `AsyncIterable<Part>` (and ensures that `ExtendedFile` is passed as content instead of passing it's content, to avoid buffering)
3. Turns it into multipart request body encoded as `AsyncIterable<BlobPart>` via `FormDataEncoder`.
4. Turns that into request body via `toBody` (that in node produces readable stream and in browser produces blob).

### Result

1. `ipfs.add` can continue using `normalizeInput` as changes to it should be API (backwards) compatible.
2. ipfs-http-client on node should continue using streams. Only thing that changed there is that some inputs are turned into `Blob`s instead of `AsyncIterator`s but during form data encoding all gets flattened anyway.
3. ipfs-http-client in browser will not buffer as long as input passed in isn't a stream and will fall back to buffering otherwise. E.g. 

   - `ipfs.add([ 'hello', await (await fetch(url)).blob(), { path: '/foo/bar', content: droppedFile } ])` will not incur buffering
   - `ipfs.add([ 'hello', { path: '/foo', content: droppedFile.stream() }, await (await fetch(url)).blob() ])` will only buffer content's of the `droppedFile` and use other pieces as is.

I am not super happy with complexity of all this, nor with the fact that user can accidentally fall of happy path and incur buffering but I do not believe there is a better option without changing an API.

 

attempt to fix #3029


Gozala:
Reminder to myself to include tests discussed in #3138 here

Gozala:
@hugomrdias there one issue that I'm not sure how to resolve. It appears that `electron-renderer` chooses to load `blob.js` over `blob.browser.js` which is causing problems. Is there a way to make it pick up `browser` overrides ?  Otherwise only I other thing I could think of is to do a runtime check.

Gozala:
All tests except the example one (that also fails on master) are passing now. I think this is ready for the review.

Gozala:
> All tests except the example one (that also fails on master) are passing now. I think this is ready for the review.

Not sure what happend, did someone restarted the job and somehow it's fixed ?

achingbrain:
The test was failing due to a temporary infrastructure problem.  All good now.

packages/ipfs-core-utils/src/files/file.node.js
```diff
@@ -0,0 +1,58 @@
...
+   * @param {string} [options.type] -  A DOMString representing the MIME type
+   * of the content that will be put into the file. Defaults to a value of "".
+   * @param {number} [options.lastModified] - A number representing the number
+   * of milliseconds between the Unix time epoch and when the file was last
+   * modified. Defaults to a value of Date.now().
+   */
+  constructor (init, name, options = {}) {
+    super(init, options)
+    /** @private */
+    this._name = name.replace(/\//g, ':')
```
lidel, line 22: What is this about?

packages/ipfs-core-utils/test/files/blob.spec.js
```diff
@@ -0,0 +1,232 @@
...
+      new Blob(['g'])
+    ]
+
+    await assertBlob(new Blob(parts), {
+      size: 7,
+      content: [...parts.slice(0, -1).map(toUint8Array), toUint8Array('g')]
+    })
+  })
+
+  it('Blob silce', async () => {
```
achingbrain, line 106: Typo?  `silce`

packages/ipfs-core-utils/src/files/normalise-input.js
```diff
@@ -1,258 +1,371 @@
...
-        return
-      }
+  // If we got here than we are dealing with async input, which can be either
+  // readable stream or an async iterable (casting former to later)
+  const stream = asReadableStream(input)
+  const asyncIterable = stream
+    ? iterateReadableStream(stream)
+    : asAsyncIterable(input)
+
+  // Async iterable (whech we assume to be homogenous) may represent single file
```
achingbrain, line 88: Typo `whech`

Gozala:
>  * I am worried about added complexity and potential regression in the future.

That worries me as well. I am also not happy with increased complexity. Only other way I can imagine going about this (that would not involve API changes) is to have a `normalise-input.browser.js`. In fact I have tried that approach but it had other major problems:

1. It is where most complexity lies and having two different implementations are very likely to diverge unintentionally.
2. Normalized input is consumed by actual `ipfs.add` (ipfs-unixfs-importer) and the http client.
   - ipfs-unixfs-importer expects  file content to be `AsyncIterable<ArrayBufferView>`. We could push some complexity from here and teach it how to handle blobs, but it seemed to spread complexity instead of reducing it. 
   - This also required diverging `ipfs.add` implementation of http client because in node we'd get normalized files in one form and in browser in the other form.

I think there is opportunity to simplify this approach a bit by using our custom types instead of `Blob` and `File` although that trade-off there would be more custom stuff in browser while Blob and File fit perfectly well.

> Are we able to add tests/benchmarks that safeguard browser-related improvements ?

I was trying to come up with some approach here, e.g.

- We can have an endpoint on echo server that generates as many gigs of data as client asks
- We can have a browser test that does `ipfs.add(await (await fetch('gen-data?size=3gb')).blob())`

However I do not think we can not have a way to tell if browser did any buffering or not. Only thing I could come up with is to generate fragment of data from echo server stop writing until corresponding put occurs on other endpoint. However that is really complex and we need to go through some hoops. There is also no guarantee that browser doesn't read say 2 two chunks at a time.

I think better strategy is to test that when we put in blobs (and alike) what we get on the other end is blobs (not objects with async iterate content). That is a lot easier to test and is free from breaking when browser changes (e.g. how much it fetches before it starts upload).

Gozala:
During our call @hugomrdias asked what were reasons to not use fetch-blob, I could not recall exact reasons so I went back to see. I believe these following reasons were why I've decided against it:

1. It deliberately implements web incompatible API by returning node `Readable` stream instead of web `ReadableStream`.
2. At a time when I looked at it it was doing memcopy, although it looks like that was fixed in https://github.com/node-fetch/fetch-blob/commit/fd4f45e70f295a0fd2d753e41961586c2b56b739#diff-168726dbe96b3ce427e7fedce31bb0bc
3. It did not provided a way to get blob chunks (as they were concatenated).

While I still thing deliberate incompatibility with web API is a mistake, I did try to swap implementation here with `fetch-blob`, but I did encountered test failures, which I have reported https://github.com/node-fetch/fetch-blob/pull/50

Apart from the bug above (which can be fixed) and API incompatibility it also uses weak maps and unnecessary async stuff and node Buffer's instead of Uint8Array's. It also seems to captures tree while this implementation captures flat list of parts  at construction. I'm not sure if all this combined is still a good enough reason to justify our own implementation.  So I think we could either:

1. Go with our implementation, and work with fetch-blob project to improve some of this.
2. Go with fetch-blob instead, and attempt to integrate improvements of this implementation there (obviously we'd need to get mentioned bug fixed to do this).

I'm personally inclined towards option 1.

Gozala:
Added more tests to ensure that result of `normaliseInput` does not use streams / async iterators unless absolutely necessary. Which hopefully addresses some of the @lidel's concern

> I am worried about added complexity and potential regression in the future.
> Are we able to add tests/benchmarks that safeguard browser-related improvements?

There is the caveat, this will not catch all regression e.g. if for some reason `normaliseInput` would e.g. read input blobs and use that to produce output files, but I think such regression is highly unlikely.

Gozala:
Test are failing now due to https://github.com/ipfs/js-ipfs/issues/3169

Gozala:
I had a conversation with @achingbrain earlier today and we have decided:

1. It would be best to factor out DOM `File` and `Blob` poly-fills into separate library.
2. Attempt to converge factored out `Blob` with `fetch-blob`. But that could happen in that factored out library.

I think it might also make sense to factor out introduced `FileStream`. However `mtime` is a bit unusual so it may be too specific to js-ipfs.

Gozala:
Externalized File and Blob implementations.

achingbrain:
I've merged ipfs/js-ipfs/pull/3184 in favour of this.  I hope that it's taken on some of the good ideas from this PR.

It bums me out a little, because you've clearly spent a lot of time and effort on this, but ultimately I think requiring people to use non-standard Blob/FormData/etc implementations to use our HTTP API is a step too far, and taking on the long-term maintenance burden of those custom implementations is not something we should be doing given the available dev capacity.

PR #3153, opened 2020-07-08
feat: enable DHT by Routing.Type config key

achingbrain:
1. Adds [Routing.Type](https://github.com/ipfs/go-ipfs/blob/b86e93045aaa0ac357d939884e1c7996b0fcdbd4/docs/config.md#routing) configuration key with values
  - `'none'` default, DHT disabled
  - `'dht'` DHT start-as-client-auto-upgrade-to-server (not implemented in `js-libp2p-kad-dht` yet)
  - `'dhtserver'` DHT server mode
  - `'dhtclient'` DHT client mode
2. Updates Core API to be compliant with the docs


packages/ipfs/docs/FAQ.md
```diff
@@ -13,7 +15,26 @@
...
+To enable DHT support, before starting your daemon run:
+
+```console
+$ jsipfs config Routing.Type dht
+```
+
+The possible values for `Routing.Type` are:
+
+ -  `'none'` the default, this means the DHT is turned off any you must manually dial other nodes
+ -  `'dht'` start the node in DHT client mode, if it works out it is dialable from outside the network, upgrade to server mode
```
jacobheun, line 29:
```suggestion
 -  `'dht'` start the node in DHT client mode, if it is discovered to be publicly dialable it will automatically switch to server mode
```

achingbrain:
I've updated the minimum required libp2p & kad-dht and will merge when CI is green.

PR #3163, opened 2020-07-10
feat: add `address` info to error object

bluelovers:


jacobheun:
As the error already has a link to the issue I think it would be helpful to instead add steps to correct this programatically in the issue. If you have multiple addresses this could result in multiple edits/starts/errors before it's fixed.

bluelovers:
show the error address in message can help make a auto-fix script

achingbrain:
As implemented this will only print the first failing address - if you only remove a bad address when the daemon fails to start and the user has multiple bad addresses it'll take a few attempts to start the daemon before it succeeds.

The invalid addresses have a recognisable structure that's easy to identify with a regex, it's better to just parse the address list from the config file and remove all that match the structure.

bluelovers:
update for support show multiple invalid address in once

achingbrain:
This change is a shim to help with a very specific migration problem, I don't think this the right thing to do here as going forward this problem does not exist.

If an upgrade script is your only option, the invalid addresses have a recognisable structure that's easy to identify with a regex, it's better to just parse the address list from the config file when upgrading and remove all that match the structure.

PR #3164, opened 2020-07-10
feat: cli `repo`/`config` add `path` for show location

bluelovers:


achingbrain:
I don't think this is the right way to go about this.  We already have `ipfs repo stats` which includes the repo path, adding a dedicated command seems unnecessary.

In future can you please add more context around the changes you wish to make - the problem you have, how your change solves it and please, please add tests.

PR #3180, opened 2020-07-18
chore: update changelog style for record more info

bluelovers:


achingbrain:
Can you give some more context around this change please?

bluelovers:
include Chores, Code Refactoring

achingbrain:
I don't think it's necessary to show all these things in release notes.  The default angular style of 'What's new', 'What's fixed' and 'What's going to break my stuff' is sufficient.  More information is in the commit log if people want to go spelunking for chores/ci/misc tasks.

For docs, perf etc we can use scopes, so `feat(docs): blah blah`, `fix(perf): blah blah`, etc.  It's easier to remember and there's less chance of someone using the wrong category.

PR #3183, opened 2020-07-20
fix(docs): update webrtc instructions for node in faq

achingbrain:
They were out of date and cause confusing errors to be thrown.

### DIFF
--- a/packages/ipfs/docs/FAQ.md
+++ b/packages/ipfs/docs/FAQ.md
@@ -89,6 +89,11 @@ const node = await IPFS.create({
       transport: [WebRTCStar]
     },
     config: {
+      peerDiscovery: {
+        [WebRTCStar.tag]: {
+          enabled: true
+        }
+      },
       transport: {
         WebRTCStar: {
           wrtc
--- a/packages/ipfs/docs/FAQ.md
+++ b/packages/ipfs/docs/FAQ.md
@@ -71,12 +71,10 @@ To add WebRTC support in a IPFS node instance, do:
 
 ```JavaScript
 const wrtc = require('wrtc') // or require('electron-webrtc')()
-const WStar = require('libp2p-webrtc-star')
-const wstar = new WStar({ wrtc })
+const WebRTCStar = require('libp2p-webrtc-star')
 
 const node = await IPFS.create({
   repo: 'your-repo-path',
-  // start: false,
   config: {
     Addresses: {
       Swarm: [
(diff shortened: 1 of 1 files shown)


packages/ipfs/docs/FAQ.md
```diff
@@ -88,8 +86,14 @@ const node = await IPFS.create({
   },
   libp2p: {
     modules: {
-      transport: [wstar],
-      peerDiscovery: [wstar.discovery]
+      transport: [WebRTCStar]
+    },
+    config: {
+      transport: {
+        WebRTCStar: {
```
achingbrain, line 93: It'd be really nice if libp2p could make these keys case-insensitive.

vasco-santos: We will work on improving the config per [libp2p/js-libp2p#576](https://github.com/libp2p/js-libp2p/issues/576). It will probably change a lot the way we config libp2p. But referencing so that we have this in attention

packages/ipfs/docs/FAQ.md
```diff
@@ -88,8 +86,14 @@ const node = await IPFS.create({
   },
   libp2p: {
     modules: {
-      transport: [wstar],
-      peerDiscovery: [wstar.discovery]
```
vasco-santos, line 92: Do we want to remove discovery?

achingbrain: From the old instructions you have to `new` up a WebRTCStar instance to get the discovery property, but now that throws an error about providing an upgrader.

Could you make a suggested change to correct the config here please?

packages/ipfs/docs/FAQ.md
vasco-santos: I just noticed in our config docs that we have this:

```js
const transportKey = WebRTCStar.prototype[Symbol.toStringTag]
```

Should we use it here to avoid confusions on where the naming comes from?

packages/ipfs/docs/FAQ.md
vasco-santos: also PRd libp2p to use the tag in this missing doc: https://github.com/libp2p/js-libp2p/pull/713

achingbrain: Is #3092 ready for review/merging? It's been draft for ages.

packages/ipfs/docs/FAQ.md
achingbrain: Actually I'm not sure if that makes it any clearer.  It would be vastly preferable to be able to just use `[WebRTCStar.tag]` the same as the peer discovery config key.

vasco-santos: We should rething that in the config issue for libp2p. From what I know, the `tag` was added for the discovery side of things. The transports do not include it, except for webrtc-star, as it is both transport and discovery.
Anyway, we will look into it in a future libp2p release

PR #3199, opened 2020-07-28
chore: update unixfs deps

achingbrain:
New version with new dag-pb, dag-cbor and no more node buffers.


achingbrain:
This will be done as part of #2771

PR #3203, opened 2020-08-02
fix: shared node demo

Gozala:
Shared worker example stopped working, and we have not noticed because for some reason CI doesn't seem to run it. There were several problems:

1. I'm guessing ipfs was bumped to 0.48 while things were in the review. And demo was getting older version with older `ipfs.add` API.
2. When I incorporated `add` / `addAll`, I have regressed the example https://github.com/ipfs/js-ipfs/pull/3081/commits/2302b2fcbac64b351d19d201259578ac008d7a39#diff-48144c5d1809a34109f2559846446f22R21.  Have not noticed that, because CI doesn't seem to run it.
3. When I first wrote a example I used `CopyWebpackPlugin` which then I thought was unnecessary and removed. However `index.html` was placed in `dist` from prior runs so after change tests were still passing (and CI never seemed to have run this example).

This pull addresses all of those problems so that example passes tests. I'm still not sure why CI isn't executing this, test but we can figure out in the separate pull.

### DIFF
--- a/examples/browser-sharing-node-across-tabs/package.json
+++ b/examples/browser-sharing-node-across-tabs/package.json
@@ -23,7 +23,7 @@
     "worker-plugin": "4.0.3"
   },
   "dependencies": {
-    "ipfs": "^0.47.0",
+    "ipfs": "^0.48.0",
     "ipfs-message-port-client": "^0.0.1",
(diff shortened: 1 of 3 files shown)


Gozala:
Seems like CI is still has not run tests for this example, I'm not sure why 🤷‍♂️

PR #3206, opened 2020-08-03
fix: correct typo in module name

achingbrain:
The name needs to be correct otherwise the tests don't run as part of CI.


achingbrain:
Runs the tests during CI now:

image

Issue #2553, opened 2019-10-21
node.key.gen('name', { type: 'ed25519', size: 256 }) does not work


Bluebie:
- **Version**: { version: '0.38.0', repo: 7, commit: '' }
- **Platform**: macOS Catalina (nodejs v12.12.0 via Homebrew)
- **Subsystem**: libp2p-keychain?

#### Type:
Bug

#### Severity:
Medium

#### Description:
despite documentation, js-ipfs does not support generating ed25519 keys. libp2p-keychain throws an error that ed25519 keys do not have an `export` function when `ipfs.key.gen` is called

#### Steps to reproduce the error:
```
> const IPFS = require('ipfs'); let ipfs = (new IPFS({ pass: "01234567890123456789" }))
> ipfs.key.gen('ed', { type: 'ed25519', size: 32 })
Promise { <pending> }
> Thrown:
TypeError: keypair.export is not a function
    at ./node_modules/libp2p-keychain/src/keychain.js:215:25
    at ./node_modules/libp2p-crypto/src/keys/ed25519-class.js:97:7
```

achingbrain:
I think this is because the underlying `libp2p-crypto` module can't export `ed25519` keys in the pem format, because the underlying `node-forge` module [doesn't support it](https://github.com/digitalbazaar/forge/issues/667), though they [are accepting pull requests](https://github.com/digitalbazaar/forge/issues/576#issuecomment-383125842).

achingbrain:
More discussion here: https://github.com/libp2p/js-libp2p-crypto/issues/145

alanshaw:
I've updated the JS core interface documentation to clarify the situation https://github.com/ipfs/interface-js-ipfs-core/pull/549

jacobheun:
Fixed in https://github.com/ipfs/js-ipfs/pull/3212

PR #3208, opened 2020-08-04
feat: add support for ed25519 and secp256k1 keys

jacobheun:
This adds support for:
1. Using ed25519 and secp256k1 keys for the ipfs PeerId
2. Using ed25519 and secp256k1 keys with ipns

fixes https://github.com/ipfs/js-ipfs/issues/2553

- [x] Requires https://github.com/libp2p/js-libp2p/pull/725
- [x] Requires https://github.com/libp2p/js-libp2p-crypto/pull/179


jacobheun:
@achingbrain there is currently no way to just specify a key type on initialization, however you can pass a pre generated privateKey as shown in the tests. We could add a `keyType` option to `IPFS.create()` to accomplish this. What do you think about:
```js
await IPFS.create({ init: { keyType: 'ed25519' }})
```

go-ipfs uses `algorithm` though, so we could use that for consistency `ipfs init --algorithm ed25519`

jacobheun:
I went with `algorithm` for now to match Go.

jacobheun:
> Should it throw if you pass bits and algorithm is not 'rsa'?

It could, ideally I think this would be a warning if anything. fwiw, the node crypto library ignores bits for ed keys.

jacobheun:
Closing this in favor of bundling with https://github.com/ipfs/js-ipfs/pull/3212

PR #3211, opened 2020-08-05
feat: backport libp2p noise as fallback for secio (#3074)

jacobheun:
This is a backport of libp2p-noise to IPFS 0.46.

I cut a branch of 0.46.x from the last commit for 0.46. The CI setup isn't currently set to run PRs that don't go into master. I'm not sure what the best approach is for lerna, but the release from the 0.46 branch should work.


achingbrain:
It *should* be ok, as long as we don't backport breaking changes as that would cause a version bump and an attempt to publish over an already published version, but you know, why would we do that.

achingbrain:
CI runs on `release/*` branches so there's a post-merge gate.  Not perfect but we only accept patches to release branches so not terrible.

achingbrain:
This has been published as `ipfs@0.46.1`

PR #3224, opened 2020-08-13
fix: validate ipns records with inline public keys

jacobheun:
The IPNS resolver was unnecessarily (and inaccurately) trying to fetch public keys from the network if they weren't embedded in the record. Embedding has been around for some time so we don't need this anymore.

This also fixes a problem where record validation wouldn't happen for inlined public keys. When PeerId's with inlined keys create ipns records, the public key is not embedded because this is duplicative. (this is the behavior of both Go and JS).

This was caught in interop testing of the latest go-ipfs updates to move to ed25519 keys by default. The failing job for this is https://app.circleci.com/pipelines/github/ipfs/go-ipfs/3329/workflows/33457111-bec0-46a9-80ef-f13fc7844231/jobs/38550.

related https://github.com/ipfs/go-ipfs/pull/7579

### DIFF
--- a/packages/interface-ipfs-core/package.json
+++ b/packages/interface-ipfs-core/package.json
@@ -41,7 +41,7 @@
     "dirty-chai": "^2.0.1",
     "ipfs-unixfs": "^1.0.3",
     "ipfs-unixfs-importer": "^2.0.2",
-    "ipfs-utils": "^2.2.2",
+    "ipfs-utils": "ipfs/js-ipfs-utils#xhr-2",
     "ipld-block": "^0.9.2",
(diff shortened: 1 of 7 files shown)


aschmahmann:
This is essentially the twin of https://github.com/ipfs/go-ipfs/pull/7549

jacobheun:
@achingbrain I'm not sure why the browser tests are failing but I'm seeing the same failures in master from the release yesterday. Any ideas why these might be happening?

jacobheun:
I bumped ipfs-utils to include the reverted progress code which fixes the browser failures.

PR #3237, opened 2020-08-21
chore: libp2p 0.29 integration

vasco-santos:
This PR integrates libp2p@0.29 into `js-ipfs`. The main features included are the usage of Signed Peer Records in Identify and Gossipsub1.1 as described in https://github.com/libp2p/js-libp2p/issues/654

(we might potentially add the UInt8Array refactor here)


achingbrain:
I’ve done this as part of #2771 - just need the libp2p interop PRs merged to finish the libp2p part, then bitswap can be finished the last of the github dep URLs removed.

vasco-santos:
@achingbrain do you mean the UInt8Array stuff?
Anyway, I needed to test gossipsub interop via IPFS and I eventually need to handle a pubsub breaking change here

vasco-santos:
Also FYI we need to sync next week about https://github.com/libp2p/js-libp2p/pull/730 + 0.29, because the 0.29 also has things that need to be changed to UintArray and if we want to get everything in 0.29, we need to make sure that it is also changed

achingbrain:
#2771 has the libp2p 0.29.0 branch as it's dep version and seems happy, so I guess we can call that the integration PR?

vasco-santos:
@achingbrain yes, I will close this PR! I also wanted to check interop tests here with gossipsub1.1 both in go and js, which are cool

PR #3245, opened 2020-08-25
feat: update hapi to v20

achingbrain:
Updates hapi ecosystem modules to v20

BREAKING CHANGES:

- Hapi has dropped support for node < 12


packages/ipfs/src/http/utils/joi.js
```diff
@@ -130,20 +165,25 @@ module.exports = Joi
...
     },
     (joi) => {
       return {
-        name: 'json',
-        base: joi.string(),
-        pre (value, state, options) {
+        type: 'json',
+        base: joi.any(),
+        validate (value, helpers) {
+          if (helpers.schema.$_getFlag('presence') === 'required' && !value) {
```
vasco-santos, line 177: Should we simplify this validation with a function? It seems replicated several times

vasco-santos (commented):
Should we enable CI for node14 per https://github.com/ipfs/js-ipfs/blob/master/.travis.yml#L20 ?

packages/ipfs/src/http/utils/joi.js
achingbrain: Probably. TBH I'm not sure why this is necessary, extending joi is really poorly documented.

achingbrain:
Node 14 isn't working yet, something about streams has changed.  I'll be doing a follow up with a fix.

packages/ipfs/src/http/utils/joi.js
vasco-santos: yeah, totally

achingbrain:
Refs: https://github.com/hapijs/hapi/issues/4149

achingbrain:
I'm going to merge this then do a follow up PR for node 14.

PR #3246, opened 2020-08-25
feat: node 14 support

achingbrain:
Enables testing on node 14.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,7 @@ stages:
 
 node_js:
   - 'lts/*'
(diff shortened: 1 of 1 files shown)


achingbrain:
https://github.com/hapijs/shot/pull/129 fixed a bug with hapi/shot under node 14 which was then reverted in https://github.com/hapijs/shot/pull/130.

I've opened another PR to fix the bug (https://github.com/hapijs/shot/pull/132) but until it or another fix is merged, the tests are broken.

achingbrain:
Currently blocked on a new hapijs/shot release.

Issue #3219, opened 2020-08-10
Return supported protocols in id output


jacobheun:
When running `ipfs.id()` it would be helpful for debugging to return the protocols the peer supports. This is being added in go-ipfs via https://github.com/ipfs/go-ipfs/pull/7409.

_Note_: go-ipfs also has the ability to run `id()` for another peers id, which is also really helpful for debugging. Adding support for that shouldn't be required for adding the protocols list, but it would be nice to also have.

PR #3250, opened 2020-08-27
feat: add protocol list to ipfs id

tk26:
addresses issue https://github.com/ipfs/js-ipfs/issues/3219

### DIFF
--- a/packages/interface-ipfs-core/src/miscellaneous/id.js
+++ b/packages/interface-ipfs-core/src/miscellaneous/id.js
@@ -49,6 +49,19 @@ module.exports = (common, options) => {
       const res = await ipfs.id()
 
       expect(res).to.have.a.property('protocols').that.is.an('array')
+
+      expect(res.protocols).to.have.members([
+        "/floodsub/1.0.0",
+        "/ipfs/bitswap/1.0.0",
+        "/ipfs/bitswap/1.1.0",
+        "/ipfs/bitswap/1.2.0",
+        "/ipfs/id/1.0.0",
+        "/ipfs/id/push/1.0.0",
+        "/ipfs/ping/1.0.0",
+        "/libp2p/circuit/relay/0.1.0",
+        "/meshsub/1.0.0",
+        "/meshsub/1.1.0"
+      ])
     })
   })
 }
--- a/packages/interface-ipfs-core/src/miscellaneous/id.js
+++ b/packages/interface-ipfs-core/src/miscellaneous/id.js
@@ -51,16 +51,16 @@ module.exports = (common, options) => {
       expect(res).to.have.a.property('protocols').that.is.an('array')
 
       expect(res.protocols).to.have.members([
-        "/floodsub/1.0.0",
-        "/ipfs/bitswap/1.0.0",
-        "/ipfs/bitswap/1.1.0",
-        "/ipfs/bitswap/1.2.0",
-        "/ipfs/id/1.0.0",
-        "/ipfs/id/push/1.0.0",
-        "/ipfs/ping/1.0.0",
-        "/libp2p/circuit/relay/0.1.0",
-        "/meshsub/1.0.0",
-        "/meshsub/1.1.0"
+        '/floodsub/1.0.0',
+        '/ipfs/bitswap/1.0.0',
+        '/ipfs/bitswap/1.1.0',
+        '/ipfs/bitswap/1.2.0',
+        '/ipfs/id/1.0.0',
+        '/ipfs/id/push/1.0.0',
+        '/ipfs/ping/1.0.0',
+        '/libp2p/circuit/relay/0.1.0',
+        '/meshsub/1.0.0',
+        '/meshsub/1.1.0'
       ])
     })
   })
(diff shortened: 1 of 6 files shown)


tk26:
I can update the test if the change looks good and makes sense. (was having some issue running the test locally)

achingbrain:
Thanks for opening this, it looks great.

To get this over the line it needs:

1. Exposing in the [HTTP API `/id/` endpoint](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/http/api/resources/id.js#L36-L42)
2. Mapping in the [HTTP API Client](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/id.js) (this may not be necessary as it looks like the protocols will just come through as a string array)
3. Tests added to the [`id` interface test suite](https://github.com/ipfs/js-ipfs/blob/master/packages/interface-ipfs-core/src/miscellaneous/id.js)
4. Those tests skipped in the [HTTP API Client suite](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/test/interface.spec.js#L469-L480) if not implemented in a shipping version of go-ipfs yet

tk26:
@achingbrain Thanks, done! please let me know how it looks.

achingbrain:
The PR looks ok but the code doesn't seem to work - that is the protocols array is always empty whether the node is started or not.

@jacobheun @vasco-santos is the peer store the right place to be fetching supported protocols from for the running node?

packages/interface-ipfs-core/src/miscellaneous/id.js
```diff
@@ -44,5 +44,11 @@ module.exports = (common, options) => {
         expect(Multiaddr.isMultiaddr(ma)).to.be.true()
       }
     })
+
+    it('should have protocols property', async () => {
+      const res = await ipfs.id()
+
+      expect(res).to.have.a.property('protocols').that.is.an('array')
```
achingbrain, line 51: This should assert that there are protocols present, at a minimum I think `/ipfs/id/1.0.0` should be supported

tk26:
> The PR looks ok but the code doesn't seem to work - that is the protocols array is always empty whether the node is started or not.
> 
> @jacobheun @vasco-santos is the peer store the right place to be fetching supported protocols from for the running node?

Thanks for checking. AFAIK peerstore seems to be having it (I referenced https://github.com/ipfs/go-ipfs/blob/master/core/commands/id.go#L154)
Also I had tried to run browser-script-tag example project and the protobook API seems to work fine. 
image

packages/ipfs/src/core/components/id.js
```diff
@@ -9,10 +9,12 @@ module.exports = ({ peerId, libp2p }) => {
   return withTimeoutOption(async function id () { // eslint-disable-line require-await
     const id = peerId.toB58String()
     let addresses = []
+    let protocols = []
 
     if (libp2p) {
       // only available while the node is running
       addresses = libp2p.transportManager.getAddrs()
+      protocols = libp2p.peerStore.protoBook.get(peerId) || []
```
jacobheun, line 17: Self protocols aren't currently stored in the peer store, so they need to be retrieved from the upgrader for now.
```suggestion
      protocols = Array.from(libp2p.upgrader.protocols.keys())
```

packages/interface-ipfs-core/src/miscellaneous/id.js
jacobheun: There should be a handful of protocols, I'd assert them all as they shouldn't change very frequently. There may be differences in Node.js vs browser, depending on modules included, but I think they are currently equivalent.

tk26: Got `["/floodsub/1.0.0", "/ipfs/bitswap/1.0.0", "/ipfs/bitswap/1.1.0", "/ipfs/bitswap/1.2.0", "/ipfs/id/1.0.0", "/ipfs/id/push/1.0.0", "/ipfs/ping/1.0.0", "/libp2p/circuit/relay/0.1.0", "/meshsub/1.0.0"]`, adding those to the assertions. Thanks guys!

packages/ipfs/src/core/components/id.js
tk26: Done! Thanks @jacobheun

achingbrain:
Thanks again for submitting this, it'll go out in the next release.

`go-ipfs` lets you use the `id` command to fetch metadata about other nodes on the network - would you like to have a go at implementing that for `js-ipfs` too?

I think you should be able to use the `protoBook`, `addressBook` etc on the `peerStore` to do this, like how you'd originally implemented this PR.

tk26:
> Thanks again for submitting this, it'll go out in the next release.
> 
> `go-ipfs` lets you use the `id` command to fetch metadata about other nodes on the network - would you like to have a go at implementing that for `js-ipfs` too?
> 
> I think you should be able to use the `protoBook`, `addressBook` etc on the `peerStore` to do this, like how you'd originally implemented this PR.

@achingbrain Thanks for merging this. Sounds interesting, I can take a look at id command for fetching metadata of other nodes.

PR #3251, opened 2020-08-27
fix: use fetch in electron renderer and electron-fetch in main

achingbrain:
Roughly halves the amount of time it takes to run electron tests.

Depends on:

- [x] https://github.com/ipfs/go-ipfs-cmds/pull/201


achingbrain:
Blocked on https://github.com/arantes555/electron-fetch/pull/30

PR #3265, opened 2020-09-03
chore: run electron main tests on linux

achingbrain:
As well as the interop tests.


achingbrain:
Will do this as part of #3251

PR #3275, opened 2020-09-09
fix: disable cors by default

achingbrain:
Brings js-ipfs into line with go-ipfs by not having CORS on by default, instead requiring the user to explicitly configure it.

BREAKING CHANGE:

- CORS origins will need to be [configured manually](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/README.md#cors) before use with ipfs-http-client

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -286,7 +286,7 @@ The RPC API endpoints running on your local node are protected by the [Cross-Ori
 
 When a request is made that sends an `Origin` header, that Origin must be present in the allowed origins configured for the node, otherwise the browser will disallow that request to proceed, unless `mode: 'no-cors'` is set on the request, in which case the response will be opaque.
 
-To allow requests from web browsers, configure the `API.HTTPHeaders.Access-Control-Allow-Origin` setting.  This is an array of URL strings and/or the wildcard `'*'` character.
+To allow requests from web browsers, configure the `API.HTTPHeaders.Access-Control-Allow-Origin` setting.  This is an array of URL strings with safelisted Origins.
 
 ##### Example
 
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -308,7 +308,7 @@ Note that the origin must match exactly so `'http://127.0.0.1:8080'` is treated 
 
 The [Access-Control-Allow-Credentials](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials) header allows client-side JavaScript running in the browser to send and receive credentials with requests - cookies, auth headers or TLS certificates.
 
-For most applications this will not be necessary but if you require this to be set, see the examnple below for how to configure it.
+For most applications this will not be necessary but if you require this to be set, see the example below for how to configure it.
 
 ##### Example
 
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -290,19 +290,21 @@ To allow requests from web browsers, configure the `API.HTTPHeaders.Access-Contr
 
 ##### Example
 
+If you are running a webapp locally that you access via the URL `http://127.0.0.1:3000`, you must add it to the list of allowed origins in order to make API requests from that webapp in the browser:
+
 ```json
 {
   "API": {
     "HTTPHeaders": {
       "Access-Control-Allow-Origin": [
-        "http://127.0.0.1:8080"
+        "http://127.0.0.1:3000"
       ]
     }
   }
 }
 ```
 
-Note that the origin must match exactly so `'http://127.0.0.1:8080'` is treated differently to `'http://127.0.0.1:8080/'`
+Note that the origin must match exactly so `'http://127.0.0.1:3000'` is treated differently to `'http://127.0.0.1:3000/'`
 
 #### `Access-Control-Allow-Credentials`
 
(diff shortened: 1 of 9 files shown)


packages/ipfs/test/http-api/cors.js
```diff
@@ -0,0 +1,123 @@
...
+        id: 'id',
+        publicKey: 'publicKey',
+        addresses: 'addresses',
+        agentVersion: 'agentVersion',
+        protocolVersion: 'protocolVersion'
+      })
+    }
+  })
+
+  describe('should allow configuring CORS', () => {
```
lidel, line 24: This needs additional tests for [preflight `OPTIONS` request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS#Preflighted_requests_in_CORS) with `Access-Control-Request-Method: POST`.

I suspect that right now `OPTIONS` request will fail,  due to `405 Method Not Allowed`, which means CORS is broken even when all headers are in place.

lidel (changes requested):

packages/ipfs/test/http-api/cors.js
achingbrain: I'll add a test.  I didn't initially because Hapi doesn't allow configuring the `Access-Control-Request-Method` header that I can see.

docs/CONFIG.md
```diff
@@ -264,3 +271,53 @@ The "basic" connection manager tries to keep between `LowWater` and `HighWater`
...
+
+HTTP header settings used by the HTTP RPC API server
+
+#### `Access-Control-Allow-Origin`
+
+The RPC API endpoints running on your local node are protected by the [Cross-Origin Resource Sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) mechanism.
+
+When a request is made that sends an `Origin` header, that Origin must be present in the allowed origins configured for the node, otherwise the browser will disallow that request to proceed, unless `mode: 'no-cors'` is set on the request, in which case the response will be opaque.
+
+To allow requests from web browsers, configure the `API.HTTPHeaders.Access-Control-Allow-Origin` setting.  This is an array of URL strings and/or the wildcard `'*'` character.
```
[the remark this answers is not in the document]
achingbrain: Hmm, ok, but obscurity !== security.

packages/ipfs/test/http-api/cors.js
achingbrain: I've added some tests but I'm not 100% sure that go-ipfs is doing the right thing here either: https://github.com/ipfs/go-ipfs/issues/7667

achingbrain:
I've changed the default behaviour of the non-`POST` `405` responses.

The previous implementation was adding extra handlers for `GET`, `DELETE`, etc that just throw `405`s, but it caused Hapi's built in CORS module to treat the presence of a handler for those methods as confirmation that an actual request made to that resource would have a chance of succeeding.

I've moved the `405`-throwing behaviour to an `onRequest` handler, so a preflight OPTIONS request to a resource with an `Access-Control-Request-Method` that isn't `POST` will now `404`.  Sending the actual non-`POST` request will `405` as before.

As an added bonus we now send an `Allow` header along with the `405` response to be RFC7231 compliant.

packages/ipfs/test/http-api/cors.js
achingbrain: Slight aside but for future issue spelunkers Hapi doesn't allow configuring the `Access-Control-Request-Method` header, because it's sent by the client and Hapi's CORS support just [checks to see if a handler is available on the requested resource with that method](https://github.com/hapijs/hapi/blob/eb89f71df7f8ecc48559eeeffbe3414205b45691/lib/cors.js#L103), waving the request through if one exists with CORS config.

lidel:
I want to test this against http://webui.ipfs.io +/- CORS, but we are in the middle of release, so need to wait a bit before I get to it.

lidel (changes requested):
CORS looks fine, but I think there is a regression related to `/webui` on API port:

- :green_heart: https://webui.ipfs.io is unable to use API at `http://127.0.0.1:5002` without CORS safelisting
- :green_heart: https://webui.ipfs.io is able to use API at `http://127.0.0.1:5002` when CORS safelisting is done via `jsipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin '["https://webui.ipfs.io"]'` before starting the daemon 
- :broken_heart:  The WebUI URL printed to the console during `jsipfs daemon` start is broken: `http://127.0.0.1:5002/webui`
 returns `405 Method Not Allowed`

docs/CONFIG.md
```diff
@@ -264,3 +271,53 @@ The "basic" connection manager tries to keep between `LowWater` and `HighWater`
...
+    "HTTPHeaders": {
+      "Access-Control-Allow-Origin": [
+        "http://127.0.0.1:8080"
+      ]
+    }
+  }
+}
+```
+
+Note that the origin must match exactly so `'http://127.0.0.1:8080'` is treated differently to `'http://127.0.0.1:8080/'`
```
lidel, line 305: `127.0.0.1:8080` is bit risky as an example, someone may copy&paste and whitelist entire path-based gateway, giving access to the API to every CID.

Using `:3000` or `example.com:3000` will be safer.

lidel (approved):
WebUI (local or remote) looks fine now.
Small nit below, but other than that lgtm.

PR #3281, opened 2020-09-11
feat: type check & generate defs from jsdoc

Gozala:
This pull request is an attempt to do following:

1. Add missing type info (via jsdoc) and fix existing one so that `tsc --noEmit` can succeed without an error.
2. Fix all the issues that `tsc` pointed out once it got enabled.
3. Add a build script that generates typedefs & typedef maps in `dist` directory & corresponding settings in `package.json` so that when package installed from npm no config will be required to get all the typings.

Unfortunately it is intertwined with some of the eslint work that I have started, in order to get rid of all the disabled warning about jsdoc comments. I was hoping we'd do https://github.com/ipfs/aegir/issues/636, but if aren't doing that we'd need to port some changes into aegir and only after we'll be able to remove those eslint disables (and possibly we'd need to add more of those too).

Note that currently generated typedefs are not a good quality yet, mostly because big parts of code base aren't typed and there is quite a bit dependence on inference which I think could be improved. However I think it would be good to start here and improve by tweaking things in the repo.

@achingbrain I would really like to get this inn as soon as possible because it touches almost every single file in the repo, which would make it difficult to keep it up to date. If you're able to do a preliminary review it would be really great. I also really don't want to block on aegir, I would much rather get typedef generation and validation working and then improved it by migrate some of those pieces to the aegir.

Dependencies:
- [ ] [ESLint with better jsdoc support](https://github.com/ipfs/js-ipfs/pull/3287)
- [ ] [TS typedefs generation](https://github.com/ipfs/aegir/pull/637)


achingbrain:
This is a massive PR, it's going to take some time to go through all the changes here.

Is this really the smallest changeset that accomplishes this task?

Gozala:
> Is this really the smallest changeset that accomplishes this task?

There are few things that I could do to make it somewhat smaller:

1. Remove all the comments that do not contain type info. (Right now big chunk of documentation from .md files had being added as part of jsdoc commens with intent to generate docs later).
2. I could break it a part into smaller patches, but only after landing a last one we would be able to generate docs and ts will succeed in checking / generating. (I'm less eager to do that, because chances are once we land all the pieces things will no longer typecheck as things could have gone out of sync).
3. Detangle eslint related changes from here. Which is something I was going to do anyway. That said it's not going to change size of this pull request much.

Please let me know if you want me to do 1. or the 2. I'm also happy to jump on the call if you want a walkthrough of the changes to make reviewing easier.

packages/ipfs/src/cli/commands/block/put.js
```diff
@@ -2,6 +2,8 @@
 
 const fs = require('fs')
 const multibase = require('multibase')
+/** @type {typeof import('it-concat').default} */
+// @ts-ignore - it-contact has invalid type defs for commonjs
```
Xmader, line 6: Why not make changes to `it-concat` directly? 
https://github.com/alanshaw/it-concat/pull/4

packages/ipfs/src/core/ipns/routing/pubsub-datastore.js
```diff
@@ -10,8 +10,9 @@ const withIs = require('class-is')
...
 const errcode = require('err-code')
 const debug = require('debug')
-const log = debug('ipfs:ipns:pubsub')
-log.error = debug('ipfs:ipns:pubsub:error')
+const log = Object.assign(debug('ipfs:ipns:pubsub'), {
+  error: debug('ipfs:ipns:pubsub:error')
+})
 
 // Pubsub datastore aims to manage the pubsub subscriptions for IPNS
 class IpnsPubsubDatastore {
```
Xmader, line 18:
```suggestion
module.exports = class IpnsPubsubDatastore {
```

Xmader: or maybe we can remove `class-is` dependency completely

packages/ipfs/src/cli/commands/block/put.js
Gozala: Once that propagates, we can remove this. I just don't want to block on the other packages.

packages/ipfs/src/core/ipns/routing/pubsub-datastore.js
Gozala: @Xmader could you please elaborate what are you suggesting here ? I would rather keep this diff as small as possible, and proposed change does not appears strictly necessary, in which case I would much rather defer that to followups.

packages/ipfs/src/core/index.js
```diff
@@ -133,3 +92,100 @@ module.exports = {
...
+ * @property {boolean} [ipnsPubsub] - Enable pub-sub on IPNS. (Default: `false`)
+ * @property {boolean} [sharding] - Enable directory sharding. Directories that have many child objects will be represented by multiple DAG nodes instead of just one. It can improve lookup performance when a directory has several thousand files or more. (Default: `false`)
+ *
+ * @typedef {import('ipfs-repo')} IPFSRepo
+ */
+
+/**
+ * Utility type to write type level conditionals
+ * @template Conditon, Then, Else
+ * @typedef {Conditon extends false
```
Xmader, line 189:
```suggestion
 * @typedef {NonNullable<Conditon> extends false
```

If `"strictNullChecks"` is not set to `true` in `tsconifg.json`,
TypeScript cannot inference type properly unless `init: true, start: true` is explicitly specified

packages/ipfs/src/core/ipns/routing/pubsub-datastore.js
Xmader: `class-is` dependency doesn't have proper TypeScript defs, so this will result an `any` type in the generated `.d.ts` file

Xmader: This can be done in the followups

Gozala: Thanks for the explanation. I think we should just rip out the class-is. No one seems to like it and just tends to get in the way.

packages/ipfs/src/core/index.js
Gozala: This pull request does set it (see https://github.com/ipfs/js-ipfs/pull/3281/files#diff-9a28699e64db6125a0b49074aeb6fdd6R13)

Xmader: If `ipfs` is installed as a dependency, can `.d.ts` files use `tsconifg.json` here?

Gozala: @Xmader that is an interesting question and I get your concern now. I guess we could put back that `NonNullable<Condition>` to avoid that problem. That said, I think we should change `ipfs.create` to return API object that has same interface. To that end I have created #3285

packages/ipfs/src/core/components/start.js
```diff
@@ -343,3 +356,48 @@ function createApi ({
...
+/**
+ * @typedef {Object} CreateAPIConfig
+ * @property {APIManager} apiManager
+ * @property {Bitswap} [bitswap]
+ * @property {Block} [block]
+ * @property {BlockService} blockService
+ * @property {Config} config
+ * @property {StartOptions} constructorOptions
+ * @property {DAG} dag
+ * @property {Files} [files]
```
Xmader, line 369:
```suggestion
 * @property {Bitswap} bitswap
 * @property {Block} block
 * @property {BlockService} blockService
 * @property {Config} config
 * @property {StartOptions} constructorOptions
 * @property {DAG} dag
 * @property {Files} files
```

any reasons to set those properties to be optional here?

Gozala: @Xmader I believe they were optional (as in they were not passed somewhere and in that case defaults were used). I might be mistaken, however I would much rather land this is as and then try to improve than keep making small changes here, because validating those takes quite a bit time that I would rather spend on other issues.

If you do want to help I would suggest maybe creating a separate pull that targets type-gen branch which:

1. Enables type-checking in the CI so we can see if things are still checking.
2. Make these kind of improvements there.

That way we know that those changes could be applied without having to fetch them and validing those locally. I hope this is reasonable.

Gozala:
@jacobheun you've suggested taking some things off of @achingbrain's plate. I incorporated repo overhaul changes from @achingbrain so it should be ready.

achingbrain:
Thanks for tidying this up with all the churn in master.

CI is very red, could you make sure the tests pass please?

Gozala:
Got all the tests passing except examples seem to fail due to

> An error occurred while retrieving a new session: "session not created: This version of ChromeDriver only supports Chrome version 84

I've checked with blank commit ontop of master and same issue seems to happen there https://github.com/ipfs/js-ipfs/runs/1218481366

I'll try to investigate that issue separately tomorrow.

achingbrain:
> An error occurred while retrieving a new session: "session not created: This version of ChromeDriver only supports Chrome version 84

This usually means that the version of Chrome installed on the CI build machine has been silently updated and we need to upgrade the chromedriver dep in `/examples/test-ipfs-example` to the latest version.

achingbrain:
I'm updating chromedriver here, though it's not as straightforward as it has been in the past: https://github.com/ipfs/js-ipfs/pull/3319

packages/ipfs-cli/src/commands/dag/put.js
```diff
@@ -4,6 +4,8 @@ const mh = require('multihashing-async').multihash
 const multibase = require('multibase')
 const dagCBOR = require('ipld-dag-cbor')
 const dagPB = require('ipld-dag-pb')
+/** @type {typeof import('it-concat').default} */
+// @ts-ignore - incorrect typedefs
```
achingbrain, line 8: Does this PR resolve the problem? https://github.com/alanshaw/it-concat/pull/4

packages/ipfs-core-utils/src/cid.js
```diff
@@ -10,23 +10,22 @@ const CID = require('cids')
...
-  options.upgrade = options.upgrade !== false
-
-  if (!CID.isCID(cid)) {
-    cid = new CID(cid)
-  }
+exports.cidToString = (input, options = {}) => {
+  const upgrade = options.upgrade !== false
+  let cid = CID.isCID(input)
+    ? input
+    // @ts-ignore - TS seems to get confused by the type defs in CID repo.
```
achingbrain, line 24: Is there a PR open to fix this?

packages/ipfs-core/src/components/bitswap/stat.js
```diff
@@ -30,7 +30,7 @@ module.exports = ({ bitswap }) => {
   /**
    * @type {Stat<{}>}
    */
-  async function stat (options) { // eslint-disable-line require-await, @typescript-eslint/no-unused-vars
+  async function stat (_options) { // eslint-disable-line require-await
```
achingbrain, line 33: Why the underscore?

packages/ipfs-core/src/components/dht.js
```diff
@@ -69,7 +69,7 @@ module.exports = ({ libp2p, repo }) => {
      * @param {number} [options.numProviders] - maximum number of providers to find
      * @returns {AsyncIterable<{ id: CID, addrs: Multiaddr[] }>}
      */
-    findProvs: withTimeoutOption(async function * (key, options) { // eslint-disable-line require-await
+    vs: withTimeoutOption(async function * (key, options) { // eslint-disable-line require-await
```
achingbrain, line 72: Why did this method name change from `findProvs` to `vs`?

packages/ipfs-core/src/components/files/chmod.js
```diff
@@ -149,9 +150,13 @@ function parseSymbolicMode (input, originalMode, isDirectory) {
...
 function calculateMode (mode, metadata) {
-  if (typeof mode === 'string' || mode instanceof String) {
+  if (mode instanceof String) {
+    mode = mode.toString()
+  }
+
+  if (typeof mode === 'string') {
     if (mode.match(/^\d+$/g)) {
-      mode = parseInt(mode, 8)
+      mode = parseInt(mode.toString(), 8)
```
achingbrain, line 159: This `toString` is unnecessary.

packages/ipfs-core/src/components/name/publish.js
```diff
@@ -78,12 +78,12 @@ module.exports = ({ ipns, dag, peerId, isOnline, keychain }) => {
...
 
-    let pubLifetime
+    let pubLifetime = 0
     try {
-      pubLifetime = parseDuration(lifetime)
+      pubLifetime = parseDuration(lifetime) || 0
 
       // Calculate lifetime with nanoseconds precision
-      pubLifetime = pubLifetime.toFixed(6)
+      pubLifetime = Math.trunc(pubLifetime)
```
achingbrain, line 86: These don't do the same thing?

```javascript
const parseDuration = require('parse-duration').default

const duration = '1.333333333h'
const parsed = parseDuration(duration)

console.info(parsed.toFixed(6))
// "4799999.998800"
console.info(Math.trunc(parsed))
// 4799999
```

packages/ipfs-core/src/components/pin/ls.js
```diff
@@ -19,13 +19,47 @@ function toPin (type, cid, metadata) {
...
+   */
+  async function * ls (options = {}) {
+    /** @type {PinQueryType} */
     let type = PinTypes.all
 
     if (options.type) {
       type = options.type
       if (typeof options.type === 'string') {
+        // @ts-ignore - can't infer toLowerCase()
         type = options.type.toLowerCase()
```
achingbrain, line 63: If you did something like:

```js
`${options.type}`.toLowerCase()
```

maybe you wouldn't need the @ts-ignore comment?

packages/ipfs-core/src/components/pin/pin-manager.js
```diff
@@ -57,7 +66,7 @@ class PinManager {
         yield * this._walkDag(link.Hash, { preload })
       }
     } else if (cid.codec === 'dag-cbor') {
-      for (const [_, childCid] of dagCborLinks(node)) { // eslint-disable-line no-unused-vars
+      for (const [, childCid] of dagCborLinks(node)) { // eslint-disable-line no-unused-vars
```
achingbrain, line 69: Is the eslint comment still necessary with this change?

packages/ipfs-core/src/index.js
```diff
@@ -108,11 +60,19 @@ async function create (options) {
   const startedApi = options.start && initializedApi && await initializedApi.start()
 
   /**
-   * @template T, THEN, ELSE
-   * @typedef {NonNullable<T> extends false
-   * ? THEN : ELSE } IsFalse
+   * create returns object that has different API set based on `options.init`
```
achingbrain, line 63: We've talked about this before - the API returned should be the same.  Would it be possible to fix this bug now that TS will point out what's different?

packages/ipfs-http-server/src/api/resources/object.js
```diff
@@ -505,7 +506,7 @@ exports.patchAppendData = {
       },
       query: {
         cid,
-        cidBase,
+        cidBase: base,
```
achingbrain, line 509: Why was this necessary?  `cidBase` is more descriptive.

achingbrain (changes requested):
Great work here, comments inline.

.travis.yml
```diff
@@ -241,7 +247,7 @@ jobs:
         - npx json -I -f ./lerna.json -e "this.command.bootstrap.nohoist = ['ipfs-css', 'tachyons']"
         - npm run reset
         - npm install
-        - npm run build -- --scope={ipfs,ipfs-http-client,ipfs-message-port-*}
+        - npm run build -- --scope={ipfs-core,ipfs,ipfs-http-client,ipfs-message-port-*}
```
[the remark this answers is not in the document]
Gozala: It did not seem to trigger build on ipfs-core, which generates types that then ipfs relies on to generate it's types.

packages/ipfs-cli/src/commands/daemon.js
```diff
@@ -57,7 +57,7 @@ module.exports = {
     if (argv.initConfig) {
       try {
         const raw = fs.readFileSync(argv.initConfig)
-        config = JSON.parse(raw)
+        config = JSON.parse(raw.toString())
```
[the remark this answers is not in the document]
Gozala: @achingbrain yes but TS can't infer that `encoding` option is passed here because there is nothing that verifies that so it rightfully assumes it could be a buffer. This seemed like a reasonable way to address this. Instead I could do a following, if you prefer

```suggestion
        config = JSON.parse(/** @type {string} */(raw))
```

packages/ipfs-cli/src/commands/dag/put.js
Gozala: It does, but I did not wanted to block this. I would much rather land with comment and then address all the ts-ignores as followups.

packages/ipfs-cli/src/commands/init.js
```diff
@@ -56,7 +56,7 @@ module.exports = {
     if (argv.defaultConfig) {
       try {
         const raw = fs.readFileSync(argv.defaultConfig)
-        config = JSON.parse(raw)
+        config = JSON.parse(raw.toString())
```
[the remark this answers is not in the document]
Gozala: Same in the comment above https://github.com/ipfs/js-ipfs/pull/3281/files#r501817986

packages/ipfs-core-utils/src/cid.js
Gozala: No there is no PR but I opened an issue https://github.com/multiformats/js-cid/issues/128

I would like to get this inn and then go over all the ts-ignores and address them as followups.

packages/ipfs-core/src/components/add-all/index.js
```diff
@@ -58,8 +58,7 @@ module.exports = ({ block, gcLock, preload, pin, options: constructorOptions })
   /**
    * @type {AddAll<{}>}
    */
-  async function * addAll (source, options) {
-    options = options || {}
+  async function * addAll (source, options = {}) {
```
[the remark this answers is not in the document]
Gozala: @achingbrain I can not because if we don't provide the default type checker assigns  it a `undefined|Options` type. And that type remains even after assignment `options = options || {}` (because types do not mutate). So if we don't want to provide default then we have to introduce another variable to deal with that e.g.:

```
const options2 = options || {}
```

Which will then have `Options` type instead of `Options|void`. The reason I did not choose later approach, because it ends up creating a lot more changes then this.

@achingbrain given this information, could you please let me know what should I do here and all other places like this ?

packages/ipfs-core/src/components/bitswap/stat.js
Gozala: TS will complain about unused variables (that's why we had a comment before saying `@typescript-eslint/no-unused-vars`, but it also recognized `_` prefixed bindings as things that may not be used and won't complain.

I like later pattern because it becomes I find it more helpful while reading code, but I could also either:

1. Remove the argument
2. Leave the comment to tell TS to ignore this.

packages/ipfs-core/src/components/block/put.js
```diff
@@ -50,8 +64,8 @@ module.exports = ({ blockService, pin, gcLock, preload }) => {
         const mhtype = options.mhtype || 'sha2-256'
         const format = options.format || 'dag-pb'
 
-        /** @type {CidVersion} */
-        let cidVersion
+        /** @type {CIDVersion} */
+        let cidVersion = 1
```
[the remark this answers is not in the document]
Gozala: I'm afraid answer is bit complicated.  You are right that it is assigned but TS seems to not recognize that here and complains that `cidVersion` maybe undefined.

Use of ternary instead of `if else`  would have worked, but then it end up nesting and ESLint complained about it. So this seemed like a reasonable way to keep everyone happy 🤷‍♂️

packages/ipfs-core/src/components/dht.js
Gozala:
```suggestion
    findProvs: withTimeoutOption(async function * (key, options) { // eslint-disable-line require-await
```

Gozala: Thanks for catching this. Not sure how this happened.

packages/ipfs-core/src/components/name/publish.js
Gozala: Thanks for catching this, I must have overlooked that `6` was passed there. It's just `toFixed()` returns string and not a number that is why I have changed this.

How about ?

```suggestion
      const precision = 1000000
      pubLifetime = Math.trunc(pubLifetime * precision) / precision
```
```

File: packages/ipfs-core/src/components/pin/ls.js (hunk 11 above)
Comment (Gozala) [reply packages/ipfs-core/src/components/pin/ls.js:63]: @achingbrain Problem here is not that TS can't infer a string (it does because there is a typeof check) but the fact that `toLowerCase()` returns value that has a string subtype of `PinQueryType` which is enum of specific strings constants.

I'll refine my comment there to be more clear about it

File: packages/ipfs-core/src/components/pin/ls.js [hunk 20]
```diff
@@ -19,13 +19,47 @@ function toPin (type, cid, metadata) {
...
+   * ```
+   */
+  async function * ls (options = {}) {
+    /** @type {PinQueryType} */
     let type = PinTypes.all
 
     if (options.type) {
       type = options.type
       if (typeof options.type === 'string') {
+        // @ts-ignore - can't infer toLowerCase()
```
Gozala, line 62:
```suggestion
        // @ts-ignore - can't infer that string returned by toLowerCase() is PinQueryType
```

packages/ipfs-core/src/components/pin/pin-manager.js
Gozala:
```suggestion
      for (const [, childCid] of dagCborLinks(node)) {
```

packages/ipfs-core/src/index.js
Gozala: I would like to fix it, but as a followup change because this is large enough in scope that I'm actively trying to cut all the corners I can.

packages/ipfs-http-server/src/api/resources/object.js
Gozala: @achingbrain problem was that `cidToString` on lines: 534, 540, 610, 625, 705, 711, 787, 793 were incorrectly passed `cidBase` option while function expects option named `base`. I could change all those calls to do `{ cid, base: cidBase ... }` but this seemed like smaller change.

Let me know if you want to keep this or change all the calls instead.

packages/ipfs/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
Gozala: I think this comment was made obsolete in one of the changes. Originally it contained some eslint customization, which I put while I was still working on eslint thing that eventually got merged in and was no longer necessary.

packages/ipfs-core/src/components/files/chmod.js
Gozala, line 159:
```suggestion
      mode = parseInt(mode, 8)
```

Gozala: You're right, thanks

Gozala:
@achingbrain Thanks for reviewing this giant patch & catching several issues that I have overlooked. I did address some of the comments and pushed a commit. There are some notes from you that I have responded to and need your response to know what to do. For convenience listing them below:

1. You did not liked that default args like `options={}` were added in bunch of places. That was deliberate change because otherwise type of the options ends up `Options|undefined` and assignment later isn't going to change. We could introduce another variable e.g. `const options2 = options || {}` or rename argument to e.g. `maybeOptions` and then have `const options = maybeOptions || {}`, but default argument seems like a best option.
2. I've added `_` prefixes (e.g. `_options`) to some arguments or variables in bunch of places, which was also deliberate. Otherwise TS will point out that those bindings aren't used (Which ESLint supposed to do as well, but seems to fail). TS recognizes `_` prefixed bindings as ones that may not be used. I could either keep the prefixes or remove those arguments and variables, please let me know what do you want to do ?
3. `cidBase` was renamed to `base` because `{ cidBase, ... }` was incorrectly passed to bunch of calls where `{ base, .... }`  is expected. Let me know if that is ok ? Otherwise we have an option to change call sites to `{ base: cidBase, ... }` instead.
4. There are bunch places where I cut corners in regards to typings in imported libraries. I realize that we should go and fix those libraries, but I think it would be much better to cut corners so this can land and do those improvements incrementally as followups. Otherwise it will require coordinating all those externals and would be nightmare. Let me know if you agree with this.
5. There is also an issue with how I replaced `.toFixed()`, I proposed a solution inline. Please let me know if that looks good and if so I'll push it. Otherwise please let me know what you want to do there.

packages/ipfs-core/src/components/pin/ls.js
Gozala:
```suggestion
        // @ts-ignore - can't infer that string returned by toLowerCase() is PinQueryType
```

packages/ipfs/src/cli/commands/daemon.js
```diff
@@ -57,7 +57,7 @@ module.exports = {
     if (argv.initConfig) {
       try {
         const raw = fs.readFileSync(argv.initConfig)
-        config = JSON.parse(raw)
+        config = JSON.parse(raw.toString())
```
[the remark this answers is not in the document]
Gozala: I think following would work if you'd prefer that:

```
const raw = fs.readFileSync({ ...argv.initConfig, encoding: 'utf8' })
```

packages/ipfs-cli/src/commands/daemon.js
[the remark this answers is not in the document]
Gozala: I think I may have misunderstood your comment, are you suggesting to change a call to `readFileSync` as follows:

```js
const config = JSON.parse(fs.readFileSync({
  ...argv.initConfig,
  encoding: 'utf8'
}))
```

packages/ipfs-core/src/components/bitswap/stat.js
achingbrain: Also, if we start using these jsdoc comments to generate a doc website, won't the argument name show up as `_options`?  Definitely don't want that.

packages/ipfs-cli/src/commands/dag/put.js
achingbrain: It's been merged so this should be ok to remove now?

achingbrain (changes requested):

achingbrain:
I've merged master into this branch

packages/ipfs-cli/src/commands/daemon.js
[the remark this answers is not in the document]
Gozala: Yeah sorry I was silly, makes sense & I'll update.

packages/ipfs-core/src/components/add-all/index.js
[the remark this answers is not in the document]
Gozala: > For consistency each core API method should have this change applied.

Do you want this part of this pull request or a followup ?

packages/ipfs-core/src/components/bitswap/stat.js
Gozala: Can I do `void options` in the source ? That would make TS happy and it's going to be more clear that just `@ts-ignore`

packages/ipfs-core/src/ipns/index.js
```diff
@@ -34,6 +35,7 @@ class IPNS {
 
       // // Add to cache
       const id = peerId.toB58String()
+      // @ts-ignore - parseFloat expects string
```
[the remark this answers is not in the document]
Gozala: I think I made comment about it somewhere else, `parseFloat` is typed as `(input:string) => number`, JS does call `toString()` but TS does expect string not anything that has `toString()` which is almost everything in JS. From TS perspective if you're passing anything but string here it's probably an error.

packages/ipfs-core/src/components/bitswap/stat.js
Gozala: I'll do comments

packages/ipfs-core/src/components/name/publish.js
Gozala, line 86:
```suggestion
      pubLifetime = parseFloat(pubLifetime.toFixed(6))
```

packages/ipfs-http-gateway/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "lint": "aegir lint",
-    "build": "aegir build",
+    "build": "echo 'nothing to build'",
```
Gozala, line 23:
```suggestion
    "build": "aegir build",
```

packages/ipfs-http-server/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "lint": "aegir lint",
-    "build": "aegir build",
+    "build": "echo 'nothing to build'",
```
Gozala, line 23:
```suggestion
    "build": "aegir build",
```

packages/ipfs-cli/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "lint": "aegir lint",
-    "build": "aegir build",
+    "build": "echo 'nothing to build'",
```
Gozala, line 23:
```suggestion
    "build": "aegir build",
```

Gozala:
@achingbrain I have (I believe) addressed all the remaining issues. I did however end up leaving `_options` arguments inn. After our conversation, I have realized that only reason we were going for `@ts-ignore` option instead was so that `_` prefixed names won't show up in docs. Which is why I thought it would be best to address that specific problem in a separate pull request, as we might find a better solution to that problem (one that would not swallow other errors while ignoring specific line).

That said, in case you disagree with this reasoning I have also made `_options` to `@ts-ignore` change in separate pull request #3324 which targets this one, so you it could be merged and squashed here if so desired.

I have also submitted another pull request #3327 (targeting this branch) which adds a separate github action that runs type-checker on pull requests and reports any missmatches inline in the diff view. I think that might be a good compromise for highlighting issues, without necessarily blocking changes. I think it would make sense to land than on sepately.

Gozala:
Trying to figure out why are some tests failing, I can see that here https://github.com/ipfs/js-ipfs/pull/3281/commits/f74e95b6077b92320cb80bcbda17e8e103c657e1 they were passing and after this change they've started failing https://github.com/ipfs/js-ipfs/pull/3281/commits/993b25e4c1d33644d0674b1640885c6bd8cd1e1a, I think it has something to do with this changes, because test fails when comparing two arrays of the same size.

https://github.com/ipfs/js-ipfs/blame/993b25e4c1d33644d0674b1640885c6bd8cd1e1a/packages/ipfs-core/src/components/bootstrap/reset.js#L50

The other regression I think was introduced by https://github.com/ipfs/js-ipfs/pull/3281/commits/58eaaea4c7a9c3a988653a1debdb546b1735515e because if I discard changes in `components/files` things are looking ok on my device. I'm suspecting it's switch from `applyDefaultOptions` to `mergeOptions`, but I get to some sleep before I can investigate these further.

packages/ipfs-http-client/package.json
```diff
@@ -20,13 +20,6 @@
     "go-ipfs": false,
     "ipfs-core-utils/src/files/normalise-input": "ipfs-core-utils/src/files/normalise-input/index.browser.js"
   },
-  "typesVersions": {
```
Gozala: Without this libraries importing http client won’t get any types

achingbrain: I thought it might break something.  Thing is, with this running the typecheck on the ipfs-cli package fails with:

```
ipfs-cli: npm info it worked if it ends with ok
ipfs-cli: npm info using npm@6.14.1
ipfs-cli: npm info using node@v12.16.1
ipfs-cli: npm info lifecycle ipfs-cli@0.0.1~prebuild: ipfs-cli@0.0.1
ipfs-cli: npm info lifecycle ipfs-cli@0.0.1~build: ipfs-cli@0.0.1
ipfs-cli: > ipfs-cli@0.0.1 build /Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli
ipfs-cli: > tsc --emitDeclarationOnly --declarationDir dist
ipfs-cli: src/daemon.js(10,32): error TS2307: Cannot find module 'ipfs-http-client' or its corresponding type declarations.
ipfs-cli: src/utils.js(130,26): error TS2307: Cannot find module 'ipfs-http-client' or its corresponding type declarations.
ipfs-cli: npm info lifecycle ipfs-cli@0.0.1~build: Failed to exec build script
ipfs-cli: npm ERR! code ELIFECYCLE
ipfs-cli: npm ERR! errno 1
ipfs-cli: npm ERR! ipfs-cli@0.0.1 build: `tsc --emitDeclarationOnly --declarationDir dist`
ipfs-cli: npm ERR! Exit status 1
ipfs-cli: npm ERR! 
ipfs-cli: npm ERR! Failed at the ipfs-cli@0.0.1 build script.
ipfs-cli: npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
ipfs-cli: npm timing npm Completed in 4477ms
```

Running `tsc` with `--traceResolution` shows it's prefixing every path with `./dist` when trying to load `ipfs-http-client`:

```
======== Resolving module 'ipfs-http-client' from '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/src/daemon.js'. ========
Explicitly specified module resolution kind: 'NodeJs'.
Loading module 'ipfs-http-client' from 'node_modules' folder, target file type 'TypeScript'.
Directory '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/src/node_modules' does not exist, skipping all lookups in it.
Found 'package.json' at '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/package.json'.
'package.json' has a 'typesVersions' field with version-specific path mappings.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client.ts' does not exist.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client.tsx' does not exist.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client.d.ts' does not exist.
'package.json' does not have a 'typings' field.
'package.json' does not have a 'types' field.
'package.json' has 'main' field 'src/index.js' that references '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/src/index.js'.
'package.json' has a 'typesVersions' entry '*' that matches compiler version '4.0.3', looking for a pattern to match module name 'src/index.js'.
Module name 'src/index.js', matched pattern '*'.
Trying substitution 'dist/*', candidate module location: 'dist/src/index.js'.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/dist/src/index.js' does not exist.
Loading module as file / folder, candidate module location '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/dist/src/index.js', target file type 'TypeScript'.
File name '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/dist/src/index.js' has a '.js' extension - stripping it.
Directory '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/@types' does not exist, skipping all lookups in it.
Directory '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/node_modules' does not exist, skipping all lookups in it.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/node_modules/ipfs-http-client.ts' does not exist.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/node_modules/ipfs-http-client.tsx' does not exist.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/node_modules/ipfs-http-client.d.ts' does not exist.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/node_modules/@types/ipfs-http-client.d.ts' does not exist.
Directory '/Users/alex/Documents/Workspaces/ipfs/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/Documents/Workspaces/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/Documents/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/node_modules' does not exist, skipping all lookups in it.
Directory '/node_modules' does not exist, skipping all lookups in it.
Loading module 'ipfs-http-client' from 'node_modules' folder, target file type 'JavaScript'.
Directory '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/src/node_modules' does not exist, skipping all lookups in it.
Found 'package.json' at '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/package.json'.
'package.json' has a 'typesVersions' field with version-specific path mappings.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client.js' does not exist.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client.jsx' does not exist.
'package.json' has 'main' field 'src/index.js' that references '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/src/index.js'.
'package.json' has a 'typesVersions' entry '*' that matches compiler version '4.0.3', looking for a pattern to match module name 'src/index.js'.
Module name 'src/index.js', matched pattern '*'.
Trying substitution 'dist/*', candidate module location: 'dist/src/index.js'.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/dist/src/index.js' does not exist.
Loading module as file / folder, candidate module location '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/dist/src/index.js', target file type 'JavaScript'.
File name '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/dist/src/index.js' has a '.js' extension - stripping it.
Directory '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/node_modules' does not exist, skipping all lookups in it.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/node_modules/ipfs-http-client.js' does not exist.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/node_modules/ipfs-http-client.jsx' does not exist.
Directory '/Users/alex/Documents/Workspaces/ipfs/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/Documents/Workspaces/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/Documents/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/node_modules' does not exist, skipping all lookups in it.
Directory '/node_modules' does not exist, skipping all lookups in it.
Loading module 'ipfs-http-client' from 'node_modules' folder, target file type 'Json'.
Directory '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/src/node_modules' does not exist, skipping all lookups in it.
Found 'package.json' at '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/package.json'.
'package.json' has a 'typesVersions' field with version-specific path mappings.
'package.json' has 'main' field 'src/index.js' that references '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/src/index.js'.
'package.json' has a 'typesVersions' entry '*' that matches compiler version '4.0.3', looking for a pattern to match module name 'src/index.js'.
Module name 'src/index.js', matched pattern '*'.
Trying substitution 'dist/*', candidate module location: 'dist/src/index.js'.
File '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/dist/src/index.js' does not exist.
Loading module as file / folder, candidate module location '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/ipfs-cli/node_modules/ipfs-http-client/dist/src/index.js', target file type 'Json'.
Directory '/Users/alex/Documents/Workspaces/ipfs/js-ipfs/packages/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/Documents/Workspaces/ipfs/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/Documents/Workspaces/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/Documents/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/alex/node_modules' does not exist, skipping all lookups in it.
Directory '/Users/node_modules' does not exist, skipping all lookups in it.
Directory '/node_modules' does not exist, skipping all lookups in it.
======== Module name 'ipfs-http-client' was not resolved. ========
```

Any ideas how to get it to not do that?

Gozala: It fails to find probably because types for ipfs-http-client had not been generated when you run that. And it should look in `dist` because that’s what typeVersions does. It maps TS search path to the dist folder.

Unfortunately as things stand today ts will not pickup type information from js files if they come from the different package, e.g. ipfs-core will not be able to get type information from ipfs-core-utils even though both have jsdoc comments. It is very unfortunate & I hope TS team will fix it at some point, but as things stand today we need to generate types for them to get picked across package boundaries.

achingbrain: No, the types are there in `/packages/ipfs-http-client/dist`.

PR #3299, opened 2020-09-22
add dag-jose codec

simonovic86:
In situation where `ipfs-http-client` uses `dag-jose` codec as a format, the IPFS instance cannot respond, since it cannot find the codec even if it's registered on bootstrap of the instance:

```js
{
	ipld: {formats: [dagJoseFormat]}
}
```

### DIFF
--- a/packages/ipfs/package.json
+++ b/packages/ipfs/package.json
@@ -109,6 +109,7 @@
     "ipld-git": "^0.6.1",
     "ipld-raw": "^6.0.0",
     "ipld-zcash": "^0.5.0",
+    "dag-jose": "^0.2.0",
     "ipns": "^0.8.0",
     "is-domain-name": "^1.0.1",
     "is-ipfs": "^2.0.0",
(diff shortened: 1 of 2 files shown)


achingbrain:
I'm not massively keen on adding more codecs if we can avoid it, I'd rather remove them!

Can you not [configure your node](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/docs/MODULE.md#optionsipld) to add the `dag-jose` codec in a similar way to the http client?

Also, where does the source code for `dag-jose` live?  The repo/homepage URLs on https://www.npmjs.com/package/dag-jose are broken.

simonovic86:
I agree. I also realized that I would need to use a `multiformats` dependency as well.

On the other side, I'm already passing the format as I'm doing in the `ipfs-http-client` but that format is not being picked up - just `IpldFormats` map is searched.

achingbrain:
It should work, could you share an example how you are doing it?

Alternatively add a test that fails to [this suite](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/test/core/ipld.spec.js).

simonovic86:
I created a separate [ticket](https://github.com/ipfs/js-ipfs/issues/3300) for this. The examples are included there.

simonovic86:
no activity on this PR and the codebase has changed. I'm closing the PR.

PR #3302, opened 2020-09-24
docs: clarify that mtime and mode are optional

lidel:
This should minimize the number of people who pass `mtime` and `mode` to `ipfs.add` when they don't really need them.

Ref. https://discuss.ipfs.io/t/why-has-same-sub-cids-but-not-same-root-cid/8810/6?u=lidel

### DIFF
--- a/docs/core-api/FILES.md
+++ b/docs/core-api/FILES.md
@@ -125,9 +125,9 @@ The regular, top-level API for add, cat, get and ls Files on IPFS
   path?: string
   // The contents of the file (see below for definition)
   content?: FileContent
-  // (Optional) mode to store the entry with (see https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation)
+  // File mode to store the entry with (see https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation)
   mode?: number | string
-  // (Optional) modification time of the entry (see below for definition)
+  // The modification time of the entry (see below for definition)
   mtime?: UnixTime
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/core-api/FILES.md
```diff
@@ -180,7 +180,7 @@ An optional object which may have the following keys:
 | -------- | -------- |
 | `Promise<UnixFSEntry>` | A object describing the added data |
 
-Each yielded object is of the form:
+Each yielded object is of the form (`mtime` and `mode` are optional, so may be missing):
```
achingbrain, line 183: `mode` has a default value so shouldn't be missing.

https://github.com/ipfs/specs/blob/master/UNIXFS.md#metadata

achingbrain (changes requested):
See comment inline

docs/core-api/FILES.md
lidel: @achingbrain Ack, I was unaware of implicit `mode` default. 
This complicates things a bit.. would this work?

```suggestion
Each yielded object is of the form (with optional `mtime` and custom/implicit `mode` [metadata](https://github.com/ipfs/specs/blob/master/UNIXFS.md#metadata)):
```

If not, we could just leave this line as it was before. Let me know.

jacobheun: Perhaps reverting this and just adding a comment in the JS object below would be helpful? Additionally, keeping the optional `?` indicator on the key is a nice way to represent the potential missing response value.

```js
{
  path: '/tmp/myfile.txt',
  cid: CID('QmHash'),
  mode: Number, // implicit if not provided
  mtime?: { secs: Number, nsecs: Number }, // may be missing
  size: 123
}
```

PR #3306, opened 2020-09-29
fix: error invalid version triggered in cli pin add/rm

Xmader:
Error `Invalid version, must be a number equal to 1 or 0` triggered when using CLI pin add/rm. 

The return value of `ipfs.pin.addAll` or `ipfs.pin.rmAll` is just [`AsyncIterable` of `CID`s](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/PIN.md#returns-1)


achingbrain (approved):
Good spot, thanks for this!

PR #3309, opened 2020-09-30
feat: enable custom formats for server/client DAG (put|get) operations

simonovic86:
- add customizable codecs in `ipfs-http-client` for DAG **get**
- add customizable codecs in `ipfs-http-server` for DAG **put**

### DIFF
--- a/packages/ipfs-cli/src/daemon.js
+++ b/packages/ipfs-cli/src/daemon.js
@@ -29,6 +29,7 @@ class Daemon {
 
   /**
    * Starts the IPFS HTTP server
+   * @param {object} [opts] - specify advanced configuration
    * @param {object} [opts.ipld.formats] - IPLD custom formats
    * @return {Promise<HttpApi>}
    */
--- a/packages/ipfs-cli/src/daemon.js
+++ b/packages/ipfs-cli/src/daemon.js
@@ -27,7 +27,12 @@ class Daemon {
     }
   }
 
-  async start () {
+  /**
+   * Starts the IPFS HTTP server
+   * @param {object} [opts.ipld.formats] - IPLD custom formats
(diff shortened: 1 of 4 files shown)


achingbrain:
CI is failing, could you please take a look?

simonovic86:
@achingbrain I've been working on the older branch and didn't merge all the changes. It should be fine now. Tnx!

packages/ipfs-cli/src/daemon.js
```diff
@@ -27,7 +27,13 @@ class Daemon {
...
   }
 
-  async start () {
+  /**
+   * Starts the IPFS HTTP server
+   * @param {object} [opts] - specify advanced configuration
+   * @param {object} [opts.ipld.formats] - IPLD custom formats
+   * @return {Promise<HttpApi>}
+   */
+  async start (opts = {}) {
```
jacobheun, line 36: How are the opts passed into daemon start? If I run `jsipfs daemon` it's not going to pass anything into the start function, so there's no way to set these.

It would be helpful to have a test showing how passing the formats is intended to work, it's not clear to me right now, other than directly using the http server and client.

The updates themselves look good but the usage isn't clear.

achingbrain:
I've tidied up the conflicts and added tests and an example over at #3347 - let's pull this over the line there.

packages/ipfs-cli/src/daemon.js
achingbrain: You can't do this through the CLI as it stands, instead you'd create a wrapper that starts a daemon with the required IPLD options.

I've added an example of how to do this in #3347

simonovic86:
@achingbrain that's awesome! ty!

PR #3319, opened 2020-10-07
chore: update chromedriver

achingbrain:


achingbrain:
Something has broken with streaming media in the latest chrome, the browser-readablestream example fails with:

```
Error: CHUNK_DEMUXER_ERROR_APPEND_FAILED: Append: stream parsing failed. Data size=708 append_window_start=0 append_window_end=inf
```

Works in FireFox. Digging into it.

achingbrain:
Turns out Chrome has got a lot more picky about the kinds of videos it will play.

Issue #3252, opened 2020-08-28
Plans to add ls, object.stat, swarm.connect?


bmann:
We’re planning to use a shared IPFS worker on our webnative apps, so your data is available instantly across all those apps. We’ve added a HTML file and worker JS file to our auth lobby, so you can load that in using an iframe.

Staging version of that HTML file:
https://round-aquamarine-dinosaur.fission.app/ipfs.html

Ideally you won’t need to think about iframes at all, and the webnative sdk will handle that for you.

For this to work we would need a few extra methods on ipfs-message-port-client:

* `ls`, listing directory contents.
* `object.stat`, getting the size of directories and files.
* `swarm.connect`, connecting to peers.

Are you planning on adding these methods?

Also posted to our forum https://talk.fission.codes/t/shared-ipfs-worker/996 cc @icidasset

achingbrain:
Yes, we're planning on fleshing out the API so thanks for mentioning the methods you use, it helps us prioritise accordingly.

We're trying to retire the object API, have you considered using [ipfs.files.stat](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#ipfsfilesstatpath-options) to get the sizes of directories & files instead?

E.g.

```js
const stats = await ipfs.files.stat('/ipfs/Qmfoo')
```

icidasset:
@achingbrain Thanks! We'll try out `ipfs.files.stat`. I didn't know you could use the MFS functions with a `/ipfs/` prefix? Or is that only applicable to `ipfs.files.stat`?

achingbrain:
The docs could definitely be improved here but the IPFS namespace is overlaid on top of the MFS commands so you can use `/ipfs/Qmfoo` style strings as paths where you're not trying to mutate an IPFS path.

Eg:

✅ `ipfs.files.cp('/ipfs/Qmfoo', '/path/in/local/mfs')` - copy a file from the wider IPFS namespace into your MFS
⛔ `ipfs.files.cp('/ipfs/Qmfoo', '/ipfs/Qmbar')` // ipfs is immutable, this makes no sense
⛔ `ipfs.files.rm('/ipfs/Qmfoo')` // ipfs is immutable, this makes no sense
✅ `ipfs.files.stat('/ipfs/Qmfoo')` - get stats on an IPFS path
✅ `ipfs.files.ls('/ipfs/Qmfoo')` - list the contents of an IPFS path
etc

Gozala:
[...]

PR #3322, opened 2020-10-09
fix: implement message-port ipfs.ls

Gozala:
This implements `ipfs.ls` for the `ipfs-message-port-*` libraries as per https://github.com/ipfs/js-ipfs/issues/3252

### DIFF
--- a/packages/ipfs-message-port-server/src/core.js
+++ b/packages/ipfs-message-port-server/src/core.js
@@ -21,6 +21,7 @@ const { decodeCID, encodeCID } = require('ipfs-message-port-protocol/src/cid')
  * @typedef {import('./ipfs').FileObject} FileObject
  * @typedef {import('./ipfs').FileContent} DecodedFileContent
  * @typedef {import('./ipfs').FileInput} DecodedFileInput
+ * @typedef {import('./ipfs').LsEntry} LsEntry
  */
 
 /**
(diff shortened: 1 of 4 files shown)


packages/ipfs-message-port-server/src/ipfs.ts
```diff
@@ -120,13 +127,15 @@ interface LsOptions extends AbortOptions {
...
+export type LsEntry = {
   name: string
+  path: string
   type: FileType
   size: number
+  depth: number
   cid: CID
-  mode: Mode
-  mtime: UnixFSTime
+  mode?: Mode
```
achingbrain, line 137: mode is not optional as it has a [default value](https://github.com/ipfs/specs/blob/master/UNIXFS.md#metadata) for folders and files

packages/ipfs-message-port-server/src/core.js
```diff
@@ -231,6 +230,26 @@ class CoreService {
     const content = this.ipfs.cat(location, { offset, length, timeout, signal })
     return encodeCatResult(content)
   }
+
+  /**
+   * @typedef {Object} LsResult
```
achingbrain, line 235: We should be able to pull the definition for this in from core the same way we're doing in the http client in #3281 right?  Probably in a follow up PR

achingbrain (changes requested):

achingbrain (approved):

packages/ipfs-message-port-server/src/ipfs.ts
Gozala, line 137:
```suggestion
  mode: Mode
```

Gozala:
```suggestion
  mode: Mode
```

PR #3336, opened 2020-10-21
chore: test with ipfsd-ctl branch

achingbrain:


achingbrain:
ipfs/js-ipfsd-ctl#554 was merged so this is no longer necessary

PR #3343, opened 2020-10-23
chore: test aegir global object fix

achingbrain:
Not for merging, just making sure the tests still pass with Karma's webpack output type set to 'umd' instead of 'var'

See: ipfs/aegir#408


achingbrain:
ipfs/aegir#408 has been merged and released so this isn't necessary any more.

Issue #3330, opened 2020-10-16
IPFS.files.ls returns integer not 'directory'/'file' in type attribute


gcohler:
From what I can tell, the documentation and the implementation differ.

Documentation here:  https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#ipfsfileslspath-options.

What I received was a 1 for directory.

achingbrain:
This should return a string in line with the docs - would you like to open a PR that fixes it?

gcohler:
I don't know the code-base well enough.  But I would suggest the following two changes (from my investigation).

1. In the MFS_FILE_TYPES definition, there are two entries with the same number assigned.  That appears to be a problem.

My proposal:  In https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/src/utils.js.  Change line 21 to:
```
  'hamt-sharded-directory': 2
```
2.  In the toOutput function it doesn't translate into a string.

My proposal:  In https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/src/components/files/ls.js:  Change line 18 to:
```
type = Object.entries(MFS_FILE_TYPES).find(x => x[1] == fsEntry.unixfs.type)[0];
```

gcohler:
@achingbrain I submitted a PR.

PR #3345, opened 2020-10-24
Proposed fix for Issue #3330:  files/ls should return string for type (not integer)

gcohler:
I'm not expert enough to be sure this is the complete fix.  But it appears to be about right.

Note that the changes may be breaking changes.


achingbrain:
I've pulled this over the line in #3352 - thanks so much for opening this!

PR #3347, opened 2020-10-26
feat: enable custom formats for dag put and get

achingbrain:
Ensures we can use custom/extra IPLD formats with `ipfs.dag.get` and `ipfs.dag.put` over HTTP as well as directly into core.

Adds an example to show how to do it.

The basic idea is you configure your node with the extra formats and pass that node into the http server, so instead of having the IPLD format logic in the http method endpoint and in core it's just in core.  The http client works the same as it did before.


examples/custom-ipld-formats/README.md
```diff
@@ -0,0 +1,31 @@
+# Custom IPLD formats
+
+This example shows you how to configure and IPFS daemon with the ability to load extra IPLD formats so you can use them in your applications.
```
vasco-santos, line 3:
```suggestion
This example shows you how to configure an IPFS daemon with the ability to load extra IPLD formats so you can use them in your applications.
```

examples/custom-ipld-formats/daemon-node.js
```diff
@@ -0,0 +1,96 @@
...
+const uint8ArrayToString = require('uint8arrays/to-string')
+
+async function main () {
+  // see https://github.com/ipld/interface-ipld-format for the interface definition
+  const format = {
+    codec: codecNumber,
+    defaultHashAlg: multicodec.SHA2_256,
+    util: {
+      serialize (data) {
+        return Buffer.from(JSON.stringify(data))
```
vasco-santos, line 27: Should we use `uint8ArrayFromString` here per [ipld/interface-ipld-format#utilserializeipldnode](https://github.com/ipld/interface-ipld-format#utilserializeipldnode). It also needs to be required

examples/custom-ipld-formats/in-process-node.js
```diff
@@ -0,0 +1,73 @@
...
+  // see https://github.com/ipld/interface-ipld-format for the interface definition
+  const format = {
+    codec: codecNumber,
+    defaultHashAlg: multicodec.SHA2_256,
+    util: {
+      serialize (data) {
+        return Buffer.from(JSON.stringify(data))
+      },
+      deserialize (buf) {
+        return JSON.parse(buf.toString('utf8'))
```
vasco-santos, line 27: use uint8arrays above?

packages/ipfs-http-client/src/lib/ipld-formats.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+module.exports = (options) => {
+  const formats = {
+    [multicodec.DAG_PB]: dagPB,
+    [multicodec.DAG_CBOR]: dagCBOR,
+    [multicodec.RAW]: raw
+  }
+
+  const ipldOptions = (options && options.ipld) || {}
+  const configuredFormats = (ipldOptions && ipldOptions.formats) || []
```
vasco-santos, line 29:
```suggestion
  const configuredFormats = ipldOptions.formats || []
```
We do not need the extra validation per fallback above

packages/ipfs-http-client/src/lib/ipld-formats.js
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Return an object containing supported IPLD Formats
+ *
+ * @param {object} [options] - Options passed to the http client constructor
+ * @param {object} [options.ipld] - IPLD options passed to the http client constructor
+ * @param {Array} [options.ipld.formats] - A list of IPLD Formats to use
+ * @param {Function} [options.ipld.loadFormat] - An async function that can load a format when passed a codec name
+ * @returns {Function}
+ */
+module.exports = (options) => {
```
vasco-santos, line 21:
```suggestion
module.exports = (options = {}) => {
```

We are checking if options exists several times below, can we just set the default value here instead?

vasco-santos (changes requested):
This is great! 👌 Overall this looks good, just some minor things caught in the review

examples/custom-ipld-formats/daemon-node.js
achingbrain: It's not necessary because the daemon only runs under node and node Buffers are Uint8Arrays so the interface contract is fulfilled.

packages/ipfs-http-server/src/api/resources/dag.js
```diff
@@ -198,13 +155,15 @@ exports.put = {
             throw Boom.badRequest('Failed to parse the JSON: ' + err)
           }
         } else {
-          const codec = multicodec[format.toUpperCase().replace(/-/g, '_')]
+          // the node is an uncommon format which the client should have
+          // serialized so deserialize it before continuing
+          const ipldFormat = await request.server.app.ipfs.ipld._getFormat(format)
```
achingbrain, line 160: I've opened https://github.com/ipld/js-ipld/pull/292 to make this `getFormat(..)` instead of `_getFormat(..)`

vasco-santos (approved):
LGTM! 
Pending the `ipld` PR to switch the the `getFormat` function

simonovic86:
@vasco-santos @achingbrain this is great! ty! 🙂

achingbrain:
No problem - just fixing up a few last problems with the typescript defs and this should go out of the door.

PR #3348, opened 2020-10-26
fix: remove buffer export from ipfs-core

### LINKED ISSUE #3312: packages/ipfs-core/src/index.js attempts to export undefined Buffer
- **Version**:
Problem exists from at least 50.1, but is on master as well.

#### Severity:
  High - The main functionality of the application does not work, API breakage, repo format breakage, etc.

#### Description:
Loading code that is browserified will cause js-ipfs to fail to load with the following error
```
[ipfs-provider] ReferenceError: Buffer is not defined
    at Object.669../api-manager (index.js:125)
    at o (_prelude.js:1)
    at _prelude.js:1
    at loadJsIpfsModule (index2.js:8)
    at tryJsIpfs (js-ipfs.js:12)
    at index.js:37
    at index.js:24
    at Object.getIpfs (index.js:52)
    at async HTMLDocument.<anonymous> (index2.js:6)
```

`browserify` the following code and including it in any browser will show the error 

```
'use strict'

const IpfsProvider = require('ipfs-provider')

document.addEventListener('DOMContentLoaded', async () => {
    const { ipfs, provider, apiAddress } = await IpfsProvider.getIpfs({
    loadHttpClientModule: () => require('ipfs-http-client'),
    loadJsIpfsModule: () => require('ipfs'),
    providers: [ 
      IpfsProvider.providers.windowIpfs({permissions: { commands: ['cat'] }}),
      IpfsProvider.providers.jsIpfs(),
      IpfsProvider.providers.httpClient(),
      IpfsProvider.providers.httpClient({apiAddress: "https://gateway.ipfs.io:443" })
    ]
    })
      console.log('IPFS API is provided by: ' + provider + ' with address: ' + apiAddress)
      window.ipfs = ipfs
})
```

Even beyond that, just looking at the current code clearly shows a that `Buffer` is in `module.exports` but is not defined anywhere else: https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/src/index.js

achingbrain:
Everything is Uint8Arrays now so there's no need to export node's
Buffer module any more.

Fixes #3312

BREAKING CHANGE: `Buffer` is no longer exported from core

### DIFF
--- a/packages/ipfs-core/src/index.js
+++ b/packages/ipfs-core/src/index.js
@@ -82,7 +82,6 @@ module.exports = {
   create,
   crypto,
   isIPFS,
-  Buffer,
   CID,
   multiaddr,
   multibase,


vasco-santos (approved):
🎉

Issue #2948, opened 2020-03-25
mfs file type should return same type


bluelovers:
from `stat` is `string`
from `ls` is `number`

this will easy make bug happen when we not check real return value

achingbrain:
The [spec docs say](https://github.com/ipfs/js-ipfs/blob/b18f6e1a791f9c72c9f35ec78c471879bbdc1525/packages/interface-ipfs-core/SPEC/FILES.md#filesls) the `type` returned from `ipfs.files.ls` should be `file` or `directory` so changing this makes sense.

Would you like to create a PR for this that changes it to use strings everywhere?  The HTTP API should remain unchanged but we can translate the types in [core](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/core/components/files/ls.js#L22) & the [client](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/files/ls.js#L46-L50).

You'll most likely have to change the [interface tests](https://github.com/ipfs/js-ipfs/blob/master/packages/interface-ipfs-core/src/files/ls.js) too.  Please make sure HAMT sharded directories are handled correctly.

achingbrain:
Fixed in #3352

PR #3352, opened 2020-10-27
fix: files ls should return string

### LINKED ISSUE #3330: IPFS.files.ls returns integer not 'directory'/'file' in type attribute

achingbrain:
Supersedes #3345 #2939
Fixes #3330 #2948

BREAKING CHANGE: types returned by `ipfs.files.ls` are now strings, in line with the docs but different to previous behaviour

### DIFF
--- a/examples/browser-mfs/filetree.js
+++ b/examples/browser-mfs/filetree.js
@@ -5,8 +5,8 @@ const {
 } = require('./utils')
 
 const FILE_TYPES = {
-  FILE: 0,
-  DIRECTORY: 1
+  FILE: 'file',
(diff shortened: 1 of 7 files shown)


jacobheun (approved):
LGTM, this also brings this inline with the go-ipfs behavior.

vasco-santos (approved):

PR #3353, opened 2020-10-27
chore: ignore typecheck for onFinishCommand

achingbrain:
Adds a ts-ignore for the yargs `onFinishCommand` method as otherwise building fails with:

```
Property 'onFinishCommand' does not exist on type 'Argv<{ silent: boolean; } & { pass: string; } & { migrate: boolean; } & { api: string | undefined; }>'
```


achingbrain:
@Gozala I don't really want to merge this, any idea what's causing this to happen?

To replicate, just run `npm run build` from the root of the repo.

achingbrain:
I can't replicate this now, closing.

Issue #3358, opened 2020-10-27
require('ipfs-core') in ipfs package causes invalid typedef genration.


Gozala:
We had to workaround the fact that following line:

https://github.com/ipfs/js-ipfs/blob/16ecc7485dfbb1f0c827c5f804974bb804f3dafd/packages/ipfs/src/index.js#L4

produces `dist/src/index.d.ts` like

```ts
declare const IPFS: typeof import("ipfs-core/src");
```

Which then TS seems to fail to resolve, by changing require path to `ipfs-core/src` (see  #3356)

We should figure out what is going on there and find a proper fix for that workaround.

Gozala:
I have setup an example TS project that pulls in the ipfs@next to try and figure this problem out. Running `tsc --noEmit --traceResolution` seems to suggest that TS attempts to look for `node_modules/ipfs-core/dist/src` as all the file types bun never as a directory (belowe is the relevant bits and noice removed

```
Loading module 'ipfs-core/src' from 'node_modules' folder, target file type 'TypeScript'.
...
File '/Users/gozala/Projects/ipfs-ts-use/node_modules/ipfs-core/dist/src.ts' does not exist.
File '/Users/gozala/Projects/ipfs-ts-use/node_modules/ipfs-core/dist/src.tsx' does not exist.
File '/Users/gozala/Projects/ipfs-ts-use/node_modules/ipfs-core/dist/src.d.ts' does not exist.
```

But then TS seems to do something strange when trying to resolve as directory 

```
'package.json' has 'main' field 'src/index.js' that references '/Users/gozala/Projects/ipfs-ts-use/node_modules/ipfs-core/dist/src/src/index.js'.
'package.json' has a 'typesVersions' entry '*' that matches compiler version '4.0.3', looking for a pattern to match module name 'src/index.js'.
Module name 'src/index.js', matched pattern '*'.
Trying substitution 'dist/*', candidate module location: 'dist/src/index.js'.
Loading module as file / folder, candidate module location '/Users/gozala/Projects/ipfs-ts-use/node_modules/ipfs-core/dist/src/dist/src/index.js', target file type 'TypeScript'.
File name '/Users/gozala/Projects/ipfs-ts-use/node_modules/ipfs-core/dist/src/dist/src/index.js' has a '.js' extension - stripping it.
```

And for some reason it never tries `.ts`, `.d.ts` extensions there, and just walks up the directory tree

```
[...]

PR #3359, opened 2020-10-27
fix: typedef resolution & add examples that use types

Gozala:
This pull request does following:

1. addresses #3356 in a different way based no findings in #3358. 
2. Adds ts project example that uses ipfs and tests that it type checks.
3. Adds js project example that uses ipfs and runs type checker to ensure types are picked up and inferred.
4. Changes all the `ReturnType<import(...)>`'s to `ReturnType<typeof import(...)>` as former seems to raise errors in stricter TS setup.


examples/ipfs-from-ts/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "ipfs-from-ts",
```
achingbrain, line 2: CI is failing because the name `ipfs-from-ts` is used for both new examples.

Suggest renaming this one (and the directory it's in) to `types-use-ipfs-from-ts`

examples/ipfs-from-ts/package.json
```diff
@@ -0,0 +1,10 @@
...
+  "private": true,
+  "dependencies": {
+    "ipfs": "^0.50.1"
+  },
+  "scripts": {
+    "test": "npx typescript --noEmit"
```
achingbrain, line 8: 1. Can you add `typescript` as a dev dep instead of using npx otherwise we'll get whatever the latest version is at runtime which makes our tests non-deterministic.
2. Don't we need to pass a file to the `typescript` command?

examples/ipfs-from-typed-js/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "ipfs-from-ts",
```
achingbrain, line 2: CI is failing because the name `ipfs-from-ts` is used for both new examples.

Suggest renaming this one (and the directory it's in) to `types-use-ipfs-from-typed-js`

examples/ipfs-from-typed-js/package.json
```diff
@@ -0,0 +1,10 @@
...
+  "private": true,
+  "dependencies": {
+    "ipfs": "^0.50.1"
+  },
+  "scripts": {
+    "test": "npx typescript --noEmit"
```
achingbrain, line 8: 1. Can you add `typescript` as a dev dep instead of using npx otherwise we'll get whatever the latest version is at runtime which makes our tests non-deterministic.
2. Don't we need to pass a file to the `typescript` command?

achingbrain (changes requested):
The examples both need a `README.md` to describe them.

You can crib from the READMEs in the other examples but they need to:

1. Give the name of the example
2. A one-sentence overview of what the example shows
3. Any install instructions
4. How to run the example
5. What output they should expect to see
6. What files they can edit or next steps to take to play with the example

examples/ipfs-from-ts/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "ipfs-from-ts",
+  "private": true,
+  "dependencies": {
+    "ipfs": "^0.50.1"
```
achingbrain, line 5: This needs to be in sync with the version in `/packages/ipfs/package.json` otherwise it may trip that [lerna bug that deletes project files](https://github.com/lerna/lerna/issues/2468#issuecomment-697647195)

examples/ipfs-from-typed-js/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "ipfs-from-ts",
+  "private": true,
+  "dependencies": {
+    "ipfs": "^0.50.1"
```
achingbrain, line 5: This needs to be in sync with the version in `/packages/ipfs/package.json` otherwise it may trip that [lerna bug that deletes project files](https://github.com/lerna/lerna/issues/2468#issuecomment-697647195)

achingbrain (approved):

achingbrain (changes requested):
Some minor typos

PR #3360, opened 2020-10-29
feat: remove all esoteric ipld formats

achingbrain:
We bundle ipld formats for ethereum, zcash, bitcoin and git though they speak beyond the files part of the Interplanetary File System.

Since #3347 we can now configure extra IPLD formats in the http client, in-process and daemon nodes, we no longer have to bundle these formats for them to be supported, instead the user can choose to configure their node with the formats they require.

This makes the behaviour of core the same in node as it is in the browser and also means we don't waste time installing deps our users may not use.

If they do use them, they can configure the node as they see fit.
    
BREAKING CHANGE: only dag-pb, dag-cbor and raw formats are supported out of the box, any others will need to be configured during node startup.


achingbrain:
Cuts the install size for `ipfs-core` from ~160mb to ~135mb

vasco-santos (commented):
Overall this looks good!
What do you think on adding to the `traverse-ipld-graphs` README a small section with the config for adding ipld formats?

```js
const ipfs = await createNode({
    ipld: {
      formats: [
        // Your formats
      ]
    }
  })
```

vasco-santos (approved):
🚢

PR #3363, opened 2020-10-30
fix: cache preloaded CIDs

### LINKED ISSUE #3307: ipfs.cat calls https://node{0..3}.preload.ipfs.io/api/v0/refs?r=true&arg=<hash> on every cat
- **Version**:
```
{version: "0.50.1", repo: 9, commit: "", interface-ipfs-core: "^0.140.0", ipfs-http-client: "^47.0.0"}
```
- **Platform**:
Verified in Firefox 81 and Chrome 85 on Windows, Linux, and Macos

- **Subsystem**:
Unknown

#### Severity:
Medium

#### Description:
If there are many repeat calls of `ipfs.cat` on subpaths of a IPFS hash, then each cat will call refs on the root of the hash. The calls are identical, and they don't seem to end up the browser's indexdb, or at least they aren't referenced there again.

On large trees, you could end up downloading many gigabytes of the same refs exact refs data.

I discovered this while sending over improvements to moshisushi/hlsjs-ipfs-loader

If you look at the network requests made on https://charade.fu.io you will see hundreds of requests to the following 4 URLS.
```
https://node0.preload.ipfs.io/api/v0/refs?r=true&arg=QmbdmJ2JRvEFhWWzHKrAcjjBdkcs46F2N7ggZnrdKKAu4s
https://node1.preload.ipfs.io/api/v0/refs?r=true&arg=QmbdmJ2JRvEFhWWzHKrAcjjBdkcs46F2N7ggZnrdKKAu4s
https://node2.preload.ipfs.io/api/v0/refs?r=true&arg=QmbdmJ2JRvEFhWWzHKrAcjjBdkcs46F2N7ggZnrdKKAu4s
https://node3.preload.ipfs.io/api/v0/refs?r=true&arg=QmbdmJ2JRvEFhWWzHKrAcjjBdkcs46F2N7ggZnrdKKAu4s
```

There are 1360 parts times 5 different bit rates, and each part cat also triggers ipfs-js to call that refs api for the entire tree, not just for that part. The refs here are 2.5MB. As a result the browser is downloading more than 3.5GB of the same 2.5MB from the preload servers.

The movie [Charade (1963) is in the Public Domain](https://en.wikipedia.org/wiki/Charade_(1963_film)), I'm using it as a much longer Big Buck Bunny to enhance the issue. There is no copyright infringement there.

achingbrain:
We 'preload' most CIDs we interact with on the network. In some cases
this can mean preloading the same CID over and over again which is not
necessary.

This PR adds a LRU cache to the preloader with a default size of 1000.
The cache is used to avoid re-preloading the same CID over and over again
until it drops out of the cache.  We use a cache that will evict CIDs
over time to have some sort of upper bound on memory usage.

Fixes #3307

### DIFF
--- a/packages/ipfs-core/src/preload.js
+++ b/packages/ipfs-core/src/preload.js
@@ -6,6 +6,9 @@ const CID = require('cids')
 const shuffle = require('array-shuffle')
 const AbortController = require('native-abort-controller')
 const preload = require('./runtime/preload-nodejs')
+/** @type {typeof import('hashlru').default} */
+// @ts-ignore - hashlru has incorrect typedefs
+const hashlru = require('hashlru')
 
 const log = Object.assign(
   debug('ipfs:preload'),
@@ -14,12 +17,14 @@ const log = Object.assign(
 
 /**
  * @param {Object} [options]
- * @param {boolean} [options.enabled]
- * @param {string[]} [options.addresses]
+ * @param {boolean} [options.enabled] - Whether to preload anything
+ * @param {string[]} [options.addresses] - Which preload servers to use
+ * @param {number} [options.cache] - How many CIDs to cache
  */
 const createPreloader = (options = {}) => {
   options.enabled = Boolean(options.enabled)
   options.addresses = options.addresses || []
+  options.cache = options.cache || 1000
 
   if (!options.enabled || !options.addresses.length) {
     log('preload disabled')
@@ -34,6 +39,9 @@ const createPreloader = (options = {}) => {
   let requests = []
   const apiUris = options.addresses.map(toUri)
 
+  // Avoid preloading the same CID over and over again
+  const cache = hashlru(options.cache)
+
   /**
    * @param {string|CID} path
    * @returns {Promise<void>}
@@ -45,6 +53,14 @@ const createPreloader = (options = {}) => {
       if (typeof path !== 'string') {
         path = new CID(path).toString()
       }
+
+      if (cache.has(path)) {
+        // we've preloaded this recently, don't preload it again
+        return
+      }
+
+      // make sure we don't preload this again any time soon
+      cache.set(path, true)
 
       const fallbackApiUris = shuffle(apiUris)
       let success = false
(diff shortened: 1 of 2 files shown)


packages/ipfs-core/src/preload.js
```diff
@@ -14,12 +17,14 @@ const log = Object.assign(
 
 /**
  * @param {Object} [options]
- * @param {boolean} [options.enabled]
- * @param {string[]} [options.addresses]
+ * @param {boolean} [options.enabled] - Whether to preload anything
+ * @param {string[]} [options.addresses] - Which preload servers to use
+ * @param {number} [options.cache] - How many CIDs to cache
```
vasco-santos, line 22:
```suggestion
 * @param {number} [options.cache = 1000] - How many CIDs to cache
```

packages/ipfs-core/src/preload.js
```diff
@@ -14,12 +17,14 @@ const log = Object.assign(
 
 /**
  * @param {Object} [options]
- * @param {boolean} [options.enabled]
- * @param {string[]} [options.addresses]
+ * @param {boolean} [options.enabled] - Whether to preload anything
+ * @param {string[]} [options.addresses] - Which preload servers to use
```
vasco-santos, line 21:
```suggestion
 * @param {string[]} [options.addresses = []] - Which preload servers to use
```

packages/ipfs-core/src/preload.js
```diff
@@ -14,12 +17,14 @@ const log = Object.assign(
 
 /**
  * @param {Object} [options]
- * @param {boolean} [options.enabled]
- * @param {string[]} [options.addresses]
+ * @param {boolean} [options.enabled] - Whether to preload anything
```
vasco-santos, line 20:
```suggestion
 * @param {boolean} [options.enabled = false] - Whether to preload anything
```

vasco-santos (approved):
LGTM!
Left some suggestions for jsdoc default values

PR #3367, opened 2020-11-04
chore: update-deps

achingbrain:


achingbrain:
@Gozala upgrading to `it-peekable` and friends with the types added in https://github.com/achingbrain/it/pull/9 appears to have broken a bunch of stuff - could you take a look please?

PR #3370, opened 2020-11-05
chore: use dnsaddr for bootstrap nodes

vasco-santos:
This PR adds the new `dnsaddr` multiaddrs for browser bootstrapping peers per https://github.com/libp2p/js-libp2p/pull/772 and https://github.com/libp2p/js-libp2p/pull/782 

cc @mburns

### DIFF
--- a/packages/ipfs-core/.aegir.js
+++ b/packages/ipfs-core/.aegir.js
@@ -15,7 +15,7 @@ let sigServerB
 let ipfsdServer
 
 module.exports = {
-  bundlesize: { maxSize: '518kB' },
+  bundlesize: { maxSize: '520kB' },
   karma: {
     files: [{
       pattern: 'node_modules/interface-ipfs-core/test/fixtures/**/*',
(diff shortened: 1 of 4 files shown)


packages/ipfs-core/.aegir.js
```diff
@@ -15,7 +15,7 @@ let sigServerB
 let ipfsdServer
 
 module.exports = {
-  bundlesize: { maxSize: '518kB' },
+  bundlesize: { maxSize: '520kB' },
```
vasco-santos, line 18: Need to increase due to the addition of the [dns-over-http-resolver](https://github.com/vasco-santos/dns-over-http-resolver) module in multiaddr, which also comes with [receptacle](https://github.com/DylanPiercey/receptacle).

achingbrain (approved):

jacobheun:
@vasco-santos @achingbrain it looks like we missed the update of these for the Node.js config, they should get updated there, that config is using some outdated Peer Ids.

vasco-santos:
yes, I will remove the SoL multiaddrs as well

Issue #2519, opened 2019-10-07
Error handling for streaming API handlers in the browser, has it ever worked?


achingbrain:
A few of our APIs are streaming, in that they produce long-lived HTTP responses containing newline-delimited JSON of arbitrary length.  When these APIs encounter an error mid-stream, they end the response and write HTTP trailers containing error information.

Thing is, no browsers support HTTP trailers so the browser experience is to end the stream (pull stream, readable stream, etc) with no error message.

It seem there's [no plan to add HTTP trailers to the fetch API](https://github.com/whatwg/fetch/issues/34) either, so we need a better way of handling errors.  In fact the whole API needs overhauling to be leaner, more consistent and less tied to golang conventions.  I did a bunch of work on a [RESTful HTTP server for IPFS](https://github.com/ipfs-shipyard/ipfs-http) a while ago, though TBF it could use bringing in line with [ipfsx](https://github.com/ipfs-shipyard/ipfsx).

Errors though, we could:

1. Deprecate the HTTP API and implement RPC over websockets - this would also allow us to do realtime events
  - ✅ Fast (to use)
  - ✅ Simple mapping from existing API
  - ❌ Big job, incredibly disruptive
2. Chunk streaming responses into multipart messages (e.g. instead of each stream entity being delimited by `\n` it would be a full-blown message part with appropriate mime-type headers, etc which would let us know if we are about to parse an error or not)
  - ✅ Uses existing web conventions
  - ❌ Adds overhead for small payloads
3. Return a unique identifier with each response and allow the client to use that id to look up an error for a limited time via an error endpoint
  - ❌ Adds complexity
4. Just write the error into the response body and hope the client notices
  - ❌ Unsuspecting clients likely to encounter significant surprise

My preference would be for 2, though preferably as part of a wider rethink of the API, a la ipfsx.

achingbrain:
cc: @ipfs/repos-javascript

daviddias:
**The crux of this issue is not designing a better API, is upgrading the HTTP API itself**. We need a migration path for go-ipfs to upgrade its HTTP API in a way that it is non breaking for the users.

One of the attempts was to make it spec first -- https://github.com/ipfs/http-api-spec -- but got superseded by autogen docs.

Read https://github.com/ipfs/specs/issues?q=is%3Aissue+is%3Aopen+label%3Ahttp-api for more background.

achingbrain:
[...]

Issue #2943, opened 2020-03-24
ipfs.add + globSource + many files will freeze


bluelovers:
```ts
let ret = await addDirectoryToIPFS(ipfs, `target path`, {
			options: {
				pin:  false,
				//progress: createProgressBar,
			},
			globSourceOptions: {
				hidden: true,
			},
		});

		console.dir(ret.files.length);
		console.dir(ret.root.cid.toString());
```

use v1 will can't see any thing and freeze
use v2 will see some logs, but after many runs will freeze

both can't finish job

test file
https://ipfs.io/ipfs/QmQtUGvQnYv2ZDw5GdRsaMXqjsYGp3h74BwoFHRRSFiWRZ?filename=.cache.7z

v1
```ts
export async function addDirectoryToIPFS(ipfs: IIPFSFileApi, targetDirPath: string, {
	options,
	globSourceOptions,
}: {
	options?: IIPFSFileApiAddOptions,
	globSourceOptions?: IGlobSourceOptions,
} = {})
{
	let files: IIPFSFileApiAddReturnEntry[] = [];

	for await (const file of ipfs.add(globSource(targetDirPath, {
		recursive: true,
		...globSourceOptions,
	}), options))
	{
		console.log(file)
		files.push(file);
	}

	const root = files[files.length - 1]

	return {
		targetDirPath,
		root,
		files,
	}
}
```

v2
```ts
export async function addDirectoryToIPFS(ipfs: IIPFSFileApi, targetDirPath: string, {
	options,
	globSourceOptions,
}: {
	options?: IIPFSFileApiAddOptions,
	globSourceOptions?: IGlobSourceOptions,
} = {})
{

	const stream = globSource(targetDirPath, {
		recursive: true,
		...globSourceOptions,
	})

	let files: IIPFSFileApiAddReturnEntry[] = [];
	let root: IIPFSFileApiAddReturnEntry;

	let i = 0;

	// @ts-ignore
	for await (const file of ipfs.add(stream, options))
	{
		if ((i++ % 100) === 0)
		{
			console.dir(file.path)
			console.log(file.cid.toString());
			//console.dir(root = file)
		}

		root = file;
	}

	return {
		targetDirPath,
		root,
		files: {
			length: i,
		},
	}
}
```

bluelovers:
over http

achingbrain:
Is the daemon running in the background when you use the CLI?

bluelovers:
yes

[...]

Issue #2854, opened 2019-05-01
Live upload progress not showing


MonarthS:
When I Upload any file, it is not showing live uploaded status using progress;

The progress showed after whole file or folder is uploaded.

**CODE:**

```
var ipfsClient = require('ipfs-http-client')
const fs = require('fs');

demoCall = function(data){
    console.log('uploaded : ' + data);
}

ipfs.addFromFs('./soft', { recursive: true, progress: demoCall}, (err, result) => {
    if (err) { throw err }
    console.log(result);
})
```

**OUTPUT:**

![image](https://user-images.githubusercontent.com/40236229/57007579-564d7b00-6c07-11e9-88f9-cae41b8be5ac.png)

achingbrain:
The code behind this module has changed significantly in the last release - are you still seeing this problem?

christopheSeeka:
Here im still having this issue at the moment, here my test code:

```
import { sha256 } from "js-sha256";
import ipfsClient from "ipfs-http-client";
import "regenerator-runtime/runtime";
var buffer = require("buffer");

const ipfs = ipfsClient({
  host: "ipfs.infura.io",
  port: "5001",
  protocol: "https"
});
const addDile = async function(buffer) {
  for await (const result of ipfs.add(buffer, {progress: progress_func})) {
    console.log(result);
  }
}
 let progress_func = function(len) {
   console.log("File progress:", len);
 };
function upload() {
  const reader = new FileReader();
  reader.onloadend = function() {
    const buf = buffer.Buffer(reader.result); // Convert data into buffer
    const files = [
      {
        content: buf
      }
    ];
    console.log(sha256(buf));
    addDile(files);
  };
  const photo = document.getElementById("photo");
  reader.readAsArrayBuffer(photo.files[0]); // Read Provided File
}

document.getElementById("uploadImage").addEventListener("click", upload);
```

I get all the progress logs only at the end. Also for some reaso i can only use ipfs.infura.io, any other gateway from ipfs gateway checker give me a connexion refused or a connexion timeout :/

christopheSeeka:
Made a simple online test here:
https://test-ipfs-ph.herokuapp.com/

You can see in the log after click in upload, i display the hash then there is nothing untill the fille is fully uploaded and all progress log show at the end only.

MonarthS:
> The code behind this module has changed significantly in the last release - are you still seeing this problem?

[...]

Issue #2838, opened 2020-02-19
ipfs.add async Iterable not truly asynchronous. Does not function as intended.


vaultec81:
Hello
I have noticed after doing some experimentation that ipfs-http-client requires the input async iterable to be completed in order for the request to proceed. There is no output until the input async iterable is completed.  Meaning the code functions no different then if an array was specified instead. 
However, if using js-ipfs in process daemon ipfs.add will function as intended. Returning CIDs through the output async iterable as the input async iterable is processed.
In my use case I am reading the output, and continuing to pass input as it is processed.
Example code
```javascript

//Source with no definite end.
for await (var out of ipfs.add(source)) {
    console.log(out.cid.string);
    //Never gets executed
}
```

achingbrain:
The problem here is that in order for `ipfs.add` to be streaming in the browser, browsers need to support streaming request bodies.

The documented way to do that is to pass a [ReadbleStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) as the request body, the problem is that no browser has implemented this - see the `Send ReadableStream in request body` line of the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#Browser_compatibility).

It doesn't help that the MDN docs talk about it like it's something you can use today.

I think we have a few options:

1. Ditch RPC over HTTP and do it over websockets
2. Chunked uploads as per #1540
3. Copy some of the IPFS core logic into the client, e.g. get it to use the [unixfs-importer](https://github.com/ipfs/js-ipfs-unixfs) and orchestrate the various other API calls the core internals make when importing files

My thoughts are not fully-formed on the matter but I think my preference would be for 3 though it'll have an affect on the bundle size.

bluelovers:
maybe some thing like this
https://javascript.info/resume-upload

PR #3371, opened 2020-11-05
feat: add grpc server and client

achingbrain:
Adds a server running a gRPC endpoint over websockets.

So far only supports `ipfs.addAll` but the idea is to implement all streaming methods over websockets instead of HTTP, to give us bidirectional streaming and errors that work in the browser.

Fixes:

#2519
#2838
#2943
#2854

Depends on:

- [ ] https://github.com/ipfs/js-ipfsd-ctl/pull/561


achingbrain:
CI Y U No

achingbrain:
Continued in #3403

PR #3372, opened 2020-11-06
feat: pass file name to add/addAll progress handler

achingbrain:
Since ipfs/js-ipfs-unixfs#87 landed we can now pass the file name to
the progress handler for adding files:

```js
await ipfs.addAll(..., {
  progress: (bytes, fileName) => {
    //...
  }
})
```

This should make showing progress a bit more usable.


vasco-santos (approved):
LGTM!

packages/ipfs-core/src/components/add-all/index.js
```diff
@@ -47,9 +47,9 @@ module.exports = ({ block, gcLock, preload, pin, options: constructorOptions })
       let total = 0
       const prog = opts.progress
 
-      opts.progress = (bytes) => {
+      opts.progress = (bytes, file) => {
```
hugomrdias, line 50:
```suggestion
      opts.progress = (bytes, fileName) => {
```

packages/ipfs-core/src/components/add-all/index.js
```diff
@@ -47,9 +47,9 @@ module.exports = ({ block, gcLock, preload, pin, options: constructorOptions })
...
         total += bytes
-        prog(total)
+        prog(total, file)
```
hugomrdias, line 52:
```suggestion
        prog(total, fileName)
```

packages/ipfs-core/src/components/add-all/index.js
hugomrdias: to match the jsdoc

hugomrdias (changes requested):
Beside the inline suggestions LGTM

hugomrdias (approved):

PR #3374, opened 2020-11-07
feat: implement service worker example

Gozala:
Adds example demonstrating how to use IPFS node in shared worker from service worker.


achingbrain:
It looks like the test for the new `example-browser-service-worker` example is failing?

achingbrain:
There's a webpack error before the test runs:

```
example-browser-service-worker: [webpack-cli] [Error: ENOTDIR: not a directory, open '/home/travis/build/ipfs/js-ipfs/examples/browser-service-worker/dist/main.js'] {
example-browser-service-worker:   errno: -20,
example-browser-service-worker:   code: 'ENOTDIR',
example-browser-service-worker:   syscall: 'open',
example-browser-service-worker:   path: '/home/travis/build/ipfs/js-ipfs/examples/browser-service-worker/dist/main.js'
example-browser-service-worker: }
```

Gozala:
@achingbrain fixed it all passes now, was just a typo in test scripts where I had `&` instead of `&&`.

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+
+  /**
+   * Obtains message port for the SharedWorker operating IPFS node and
+   * creates a client for it.
+   *
+   * @param {Fetch} context
+   */
+  static async activate (context) {
+    // Selects a service worker client that can be used to obtain a message port
+    // from, then send a request to it and once response is obtained create a
```
achingbrain, line 320:
```suggestion
    // from, then sends a request to it and once a response is obtained, creates a
```

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+    // IPFS client and return it.
```
achingbrain, line 321:
```suggestion
    // IPFS client and returns it
```

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+   * Just a wrapper around `activate` that deals with state machine.
+   * 
+   * @param {IPFSService} self
+   * @param {Fetch} context
+   */
+  static async use(self, context) {
+    const { state } = self
+    switch (state.status) {
+      // If "idle" call `activate` and transtion to "pending" state and once
+      // returned promise reloves transition to "ready" state.
```
achingbrain, line 336:
```suggestion
      // returned promise resolves transition to "ready" state.
```

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+          const value = await ready
+          self.state = { status: 'ready', value }
+          return value
+        // If error occurs transition back to `idle` state.
+        } catch (error) {
+          self.state = { status: 'idle' }
+          throw error
+        }
+      }
+      // If "pending" request just wait for completion and return return.
```
achingbrain, line 350:
```suggestion
      // If "pending" request just wait for completion and return
```

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+        return await state.ready
+      }
+      // If "ready" just return the value.
+      case 'ready': {
+        return state.value
+      }
+    }
+  }
+  /**
+   * Just a sugar to for `IPFSService.use(this, event)`
```
achingbrain, line 361:
```suggestion
   * Just sugar for `IPFSService.use(this, event)`
```

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+}
+
+
+/**
+ * Find a window client that can provide a message port for a shared worker.
+ * @param {Fetch} context
+ * @returns {Promise<WindowClient>}
+ */
+const selectClient = async ({ target, clientId  }) => {
+  // Get all the controlled window clients score them and use the best one if
```
achingbrain, line 378:
```suggestion
  // Get all the controlled window clients, score them and use the best one if
```

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+ * @param {string} path
+ */
+const fetchIPFSFile = async (ipfs, path) => {
+  const content = ipfs.cat(path)
+  // We get back `AsyncIterable<Uint8Array>`, but if such file does not exists
+  // we will only know once we start reading. To responding with a stream that
+  // fails we attempt to read first chunk and only then produce a response
+  // instace. This ensures that we either error or return Response that doesn't
+  // immediately fail.
+  const chunks = content[Symbol.asyncIterator]()
```
achingbrain, line 200: Could you go into a bit more detail about the problem this is trying to solve?

Just because the first file chunk was received, there's no guarantee the rest of the chunks are available so how does getting the first chunk before continuing help here?

examples/browser-service-worker/webpack.config.js
```diff
@@ -0,0 +1,51 @@
...
+  output: {
+    path: path.join(__dirname, 'dist'),
+    filename: 'main.js',
+    publicPath: '/'
+  },
+  devServer: {
+    serveIndex: true,
+    index: './index.html',
+    compress: true,
+    port: 3000,
```
achingbrain, line 19: Does this port need aligning with the port in the URLs/screenshots in the README?

achingbrain (changes requested):
There's an awful lot of boilerplate code here, it's not terribly clear what the minimal set of code needed to get a shared worker going is and what is window dressing.

It also uses a lot of technology that may not be absolutely essential and as such are distractions. People look at these examples without a 100% understanding of TypeScript, Webpack, Babel, even JavaScript so anything that can be removed should be removed.

E.g. People will largely copy/paste the example and use it to start a new project. Ask yourself what tech choices have you then made for them? Maybe don't make those choices. For example if we can get away with just using a script tag and linking to `node_modules/some-module/dist/index.min.js` instead of using a bundler, do that instead - it'll make the example more concise (good for the reader) and more maintainable since there will be fewer dependencies to be constantly upgrading (good for us).

Could you please restructure the example to show clearly what is boilerplate/unnecessary and what is not.

examples/browser-service-worker/test.js
```diff
@@ -0,0 +1,33 @@
...
+    browser
+      .url(process.env.IPFS_EXAMPLE_TEST_URL)
+      .waitForElementVisible('body')
+
+
+    browser.expect.element('body').text.to.contain('Load content by adding IPFS path to the URL')
+    browser.expect.element('a').text.to.contain('/ipfs/bafy')
+
+    // Sometimes service worker activation takes a bit so we give it a moment
+    browser.pause(1000)
```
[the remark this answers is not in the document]
Gozala: Good point, I have changed it to inserting meta element in the document so we can wait for that instead.

examples/browser-service-worker/src/main.js
```diff
@@ -0,0 +1,81 @@
...
+      }, [worker.port])
+    }
+  }
+}
+
+/**
+ * Creates a shared worker instance that exposes JS-IPFS node over MessagePort.
+ * @returns {SharedWorker}
+ */
+const useIPFSWorker = () => new SharedWorker(
```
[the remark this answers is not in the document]
Gozala: I don't main changing the name, however if anything SharedWorker API is misleading when you instantiate one it does not actually create one, but rather creates a connection to one. This is even more true here because we don't create new IPFS worker instead we use the one that is already there.

examples/browser-service-worker/src/service.js
Gozala: I'm not sure how to make this comment any more clear, could you please suggest something ?

So if you look at `fetchIPFSContent` function you can see that it tries to use `fetchIPFSFile` first and if it fails due to `dag node is a directory` it falls back to `fetchIPFSDirectory` (that renders diretory index) or if failed with `no link named` it returns 404.

Without pulling first chunk these errors happens inside the stream, meaning request will succeed with 200, but than it's body will fail and nothing is served.

That is why I mentioned other day that APIs that just return `AsyncIterable`s aren't very ergonomic and can easily lead to mistakes, because they tangle two different class of errors:

1. "incorrect requests"
2. "io error satisfying requests"

I think they are fundamentally different because they require very different recovery logic.

Gozala:
@achingbrain as per our discussion I have moved some pieces into utils and added a general overview section into readme that attempts to provide guidence in terms of how to navigate the code base.

examples/browser-service-worker/src/service.js
achingbrain: If you want to know if a CID resolves to a file or a directory before catting it, you can use `ipfs.files.stat` and make a decision based on that.

It'll also tell you if the file exists (by which I mean blocks are present in the local repo, I'm assuming that's what the comment means too, please correct if not), sort of, in that if the CID cannot be resolved it'll try to pull it from the network and will eventually time out.

My point is that if a file consists of 100 chunks, just because the first chunk arrives there's no guarantee that the 68th chunk will arrive so the stream can still fail, though from your clarification above I don't think that's the problem you are trying to solve.

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+      const [,protocol] = url.pathname.split('/')
+      switch (protocol) {
+        // If requests are for `/ipfs/...` or `/ipns/` URLs we respond with
+        // a content viewer which is page which are pages containing iframe
+        // for actual content like <iframe src=`/view${url.pathname}` />
+        // and a script that we can talk to to obtain message port for IPFS node.
+        // This might be confusing but this wrapper page is what allows this
+        // service worker to obtain a mesasge port, otherwise there may not even
+        // be a page that will start a shared worker, nor a way to get a message
+        // port if on is present.
```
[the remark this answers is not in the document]
achingbrain: `if one is present`?

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,389 @@
...
+  })
+}
+
+/**
+ * @param {IPFS} ipfs 
+ * @param {string} path
+ */
+const fetchIPFSDirectory = async (ipfs, path) => {
+  try {
+    // Collect diretory entries. If happens to contain `index.html` create a
```
achingbrain, line 242:
```suggestion
    // Collect directory entries. If happens to contain `index.html` create a
```

examples/browser-service-worker/src/service.js
achingbrain: @Gozala I think this is the last thing to be resolved before it can be merged. Unless there's a good reason to leave it as it is can you please refactor to use `ipfs.files.stat` to find out what type of DAG is being accessed instead of using exceptions for flow control?

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,530 @@
...
+    // response containing that file otherwise respond with a diretory listings.
+    const entries = []
+    for await (const entry of ipfs.ls(path)) {
```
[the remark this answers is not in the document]
Gozala: Not sure, but gateway seems to render index for dirs so this just attempts to replicate same experience instead of just doing 5xx

examples/browser-service-worker/src/service.js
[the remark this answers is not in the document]
Gozala: yeah I think it was a typo

examples/browser-service-worker/src/service.js
```diff
@@ -0,0 +1,389 @@
...
+      const [,protocol] = url.pathname.split('/')
+      switch (protocol) {
+        // If requests are for `/ipfs/...` or `/ipns/` URLs we respond with
+        // a content viewer which is a page containing an iframe
+        // for actual content like <iframe src=`/view${url.pathname}` />
+        // and a script that we can talk to in order to obtain a connection to the shared IPFS node via a MessagePort.
+        // This might be confusing but this wrapper page is what allows this
+        // service worker to obtain a MessagePort, otherwise there may not even
+        // be a page that will start a shared worker, nor a way to get a message
+        // port if on is present.
```
Gozala, line 42:
```suggestion
        // port for it.
```

examples/browser-service-worker/src/service.js
Gozala: So motivation here was to demonstrate some of the nitty gritty details of the API. I suspect that any serious usage of SW will try to avoid additional roundtrips for every single request (which in common cases would be redundant). And I feel like it is easy to overlook problem so showing how to address that was deliberate.

If you do think it is out of scope for this example, I can go ahead and change it as suggested.

examples/browser-service-worker/src/service.js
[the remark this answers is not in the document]
achingbrain: The gateway is a bit more low level in that it renders the links under a given DAG node, which if it's a shard will be might look like: `A0`, `A1`, A3somefile.txt`, but it'll be capped at the max children a node will have (e.g. 174 by default).

This code buffers the entire contents of the directory - if it's a shard `ipfs.ls` will transparently convert it's contents into the same output format as a regular directory, except potentially much, much larger.

examples/browser-service-worker/src/service.js
achingbrain: I think maybe this is a bit more advanced than is necessary for an example and having it be concise and easy to digest would be more valuable.

It's a fair point though, how about adding a 'further reading' or 'what next' section to the readme with the sort of problems you're likely to encounter and how to resolve them?

Gozala:
@achingbrain I have addressed remaining notes

- instead of error base dir rendering fallback callback now code uses `ipfs.file.stat` to choose between dir & file renderring.
   - There are still some error handling e.g. to redirect non file / dir entries to https://explore.ipld.io for the nodes
- dir renderring no longer buffers but rather streams response.
  - Both chrome and firefox seem to do incremental rendering of entries 
  - I have added limit to the number of entries, if limit is reached note about pagination user exercise is renderred

achingbrain (approved):

PR #3384, opened 2020-11-12
fix: ensure correct progress is reported

achingbrain:
Fixes a bug where due to parallel file imports, the reported size of
ingested bytes was getting mixed up.

We now tally imported bytes against the reported file name instead of
assuming a sequential import, taking care to not let the tally grow
until memory is exhausted.


vasco-santos (approved):
LGTM! 👌

PR #3386, opened 2020-11-12
feat: allow passing the id of a network peer to ipfs.id

achingbrain:
Adds parity with go-IPFS and allows looking up known peer IDs using
the `ipfs.id` command.

Depends on:

- [x] https://github.com/libp2p/js-libp2p/issues/799
- [x] https://github.com/libp2p/js-libp2p/pull/943


achingbrain:
Needs js-libp2p to store the `agentVersion` and `protocolVersion` for remote peers somewhere.

hugomrdias (approved):

packages/ipfs-core/src/components/id.js
```diff
@@ -14,44 +16,71 @@ module.exports = ({ peerId, libp2p }) => {
...
+    let publicKey = id.pubKey
     let addresses = []
     let protocols = []
+    let agentVersion = `js-ipfs/${pkgversion}`
+    let protocolVersion = '9000'
+
+    if (options.peerId) {
+      if (PeerId.isPeerId(options.peerId)) {
+        id = options.peerId
+      } else {
```
Gozala, line 40: NIT: do we really need to support different representations ? Seems like user of the API could do the coercion instead.

Gozala (approved):
Left some comments, but feel free to disregard those.

packages/ipfs-core/src/components/id.js
achingbrain: No, we don't.  Also, this should be a string and not a PeerId, otherwise we pull `peer-id` and all the associated crypto junk into `ipfs-core-types` as it'll form part of the API types and so into the http client and other places that don't need them.

packages/ipfs-core/src/components/id.js
```diff
@@ -14,44 +16,71 @@ module.exports = ({ peerId, libp2p }) => {
...
+        id = PeerId.createFromB58String(options.peerId.toString())
+      }
+
+      if (!libp2p) {
```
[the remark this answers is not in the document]
achingbrain: The peer store should work offline, so we should be able to query it for peers we knew about last time we were online.

The implementation here may need a little more thought.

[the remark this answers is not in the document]
Gozala: maybe in the short term we should just return empty arrays instead of errors then

docs/core-api/MISCELLANEOUS.md
```diff
@@ -47,6 +47,7 @@ An optional object which may have the following keys:
 | ---- | ---- | ------- | ----------- |
 | timeout | `Number` | `undefined` | A timeout in ms |
 | signal | [AbortSignal][] | `undefined` |  Can be used to cancel any long running requests started as a result of this call |
+| peerId | `String|PeerId` | `undefined` | Look up the identity for this peer instead of the current node |
```
achingbrain, line 50:
```suggestion
| peerId | `string` | `undefined` | Look up the identity for this peer instead of the current node |
```

PR #3387, opened 2020-11-12
chore: replace SoL arnet from node config

vasco-santos:
Replaces `SoL-arnet` bootstrap nodes from Node.js config to the new `dnsaddr` per https://github.com/ipfs/js-ipfs/pull/3370#issuecomment-726083859 and https://github.com/libp2p/js-libp2p-bootstrap/pull/98

These new addresses resolve into DNS TXT records with tcp and wss based addresses for each peer (with `ipv4` and `ipv6`.


jacobheun (approved):
bootstrap node changes lgtm

achingbrain (approved):

PR #3388, opened 2020-11-12
test: add tests for different types of connection config

achingbrain:
We claim to support URLs, MultiAddrs, strings and individual URL components
so add tests for them.

Also, add code to actually support the above permutations..

### DIFF
--- a/packages/ipfs-http-client/src/lib/core.js
+++ b/packages/ipfs-http-client/src/lib/core.js
@@ -9,19 +9,17 @@ const merge = require('merge-options')
 const toUrlString = require('ipfs-core-utils/src/to-url-string')
 
 /**
- * @param {ClientOptions|URL|Multiaddr|string} options
+ * @param {ClientOptions|URL|Multiaddr|string} [options]
  * @returns {ClientOptions}
  */
 const normalizeOptions = (options = {}) => {
   let url
-  let opts
+  let opts = {}
 
   if (typeof options === 'string' || Multiaddr.isMultiaddr(options)) {
     url = new URL(toUrlString(options))
-    opts = {}
   } else if (options instanceof URL) {
     url = options
-    opts = {}
   } else if (typeof options.url === 'string' || Multiaddr.isMultiaddr(options.url)) {
     url = new URL(toUrlString(options.url))
     opts = options
--- a/packages/ipfs-http-client/src/lib/core.js
+++ b/packages/ipfs-http-client/src/lib/core.js
@@ -1,58 +1,59 @@
 'use strict'
 /* eslint-env browser */
 const Multiaddr = require('multiaddr')
-const toUri = require('multiaddr-to-uri')
 const { isBrowser, isWebWorker } = require('ipfs-utils/src/env')
-const { URL } = require('iso-url')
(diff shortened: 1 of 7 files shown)


packages/ipfs-http-client/src/lib/core.js
```diff
@@ -1,58 +1,59 @@
...
  */
-const normalizeInput = (options = {}) => {
-  if (isMultiaddr(options)) {
-    options = { url: toUri(options) }
-  } else if (typeof options === 'string') {
-    options = { url: options }
-  }
+const normalizeOptions = (options = {}) => {
+  let url
+  let opts
```
vasco-santos, line 17:
```suggestion
  let opts = {}
```

packages/ipfs-http-client/src/lib/core.js
```diff
@@ -1,58 +1,59 @@
...
+
+  if (typeof options === 'string' || Multiaddr.isMultiaddr(options)) {
+    url = new URL(toUrlString(options))
+    opts = {}
```
vasco-santos, line 21:
```suggestion
```

packages/ipfs-http-client/src/lib/core.js
```diff
@@ -1,58 +1,59 @@
...
+  } else if (options instanceof URL) {
+    url = options
+    opts = {}
```
vasco-santos, line 24:
```suggestion
```

packages/ipfs-http-client/src/lib/core.js
```diff
@@ -1,58 +1,59 @@
...
-    Multiaddr(input) // eslint-disable-line no-new
-    return true
-  } catch (e) {
-    return false
-  }
-}
-
-/**
- * @param {any} options
+ * @param {ClientOptions|URL|Multiaddr|string} options
```
vasco-santos, line 12:
```suggestion
 * @param {ClientOptions|URL|Multiaddr|string} [options]
```

packages/ipfs-http-client/src/lib/core.js
```diff
@@ -1,58 +1,59 @@
...
+  if (opts.apiPath) {
+    url.pathname = opts.apiPath
+  } else if (url.pathname === '/' || url.pathname === undefined) {
+    url.pathname = 'api/v0'
+  }
+
+  return {
+    ...opts,
+    host: url.host,
+    protocol: url.protocol.replace(':', ''),
```
vasco-santos, line 52: Do we need this with the previous replace?

vasco-santos (approved):
This looks good! Left a few suggestion

packages/ipfs-http-client/src/lib/core.js
achingbrain: Yes because URL objects add a `:` on the end of `.protocol`:

```js
new URL('http://example.com').protocol // 'http:'

new URL('http:://example.com') // throws

const u = new URL('http://example.com')
u.protocol = 'https::::::'
u.protocol // 'https:'

u.protocol = 'http'
u.protocol // 'http:'
```

vasco-santos (approved):

PR #3407, opened 2020-11-18
chore: move withTimeoutOption to core-utils

achingbrain:
Pulls non-grpc changes out of #3403 to ease the continued merging of
master into that branch.

- Moves withTimeoutOption into core-utils
- Moves TimeoutError into core-utils
- Adds missing ts project links
- Adds more add-all tests to interface suite
- Ignores unpassable tests for non-grpc or core implementations
- Normalises mode and mtime in normalise-input function
- Dedupes mtime normalisation between core and http client


packages/ipfs-core-utils/src/with-timeout-option.js
```diff
@@ -0,0 +1,106 @@
...
+
+const TimeoutController = require('timeout-abort-controller')
+const { anySignal } = require('any-signal')
+const parseDuration = require('parse-duration').default
+const { TimeoutError } = require('./errors')
+
+/**
+ * @template {any[]} ARGS
+ * @template {Promise<any> | AsyncIterable<any>} R - The return type of `fn`
+ * @param {Fn<ARGS, R>} fn
```
Gozala, line 12: @achingbrain this seems to have introduced a regression, because `Fn` typedef ended up in `index.js` which is not imported here. Which for some reason TS still seems to kind of pick up, but it no longer behaves correctly & causing returned function to be inferred as any.

Gozala: I think the whole `Fn` type is kind of redundant and we should probably change type annotation for `withTimeoutOptions` to  something like this instead:

```js
/**
 * @template {any[]} Args
 * @template {Promise<any> | AsyncIterable<any>} R - The return type of `fn`
 * @param {(...args:Args) => R} fn
 * @param {number} [optionsArgIndex]
 * @returns {(...args:Args) => R}
 */
```

Issue #3400, opened 2020-11-15
In-browser incorrect ipfs.swarm.connect on refresh


hunterInt:
Version:
0.50.2

Platform:
Linux ubuntu 5.4.0-52-generic #57-Ubuntu SMP Thu Oct 15 10:57:00 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

Severity:
Medium - A non-essential functionality does not work, performance issues, etc.

#### Description:
When I first open my webpage and attempt to connect to another IPFS Node that I own I can successfully connect:

![1st](https://user-images.githubusercontent.com/14793417/99198271-d5a49000-2754-11eb-9d4c-730744d3b815.png)

However, when I refresh the **same** page I get this error:

![2nd](https://user-images.githubusercontent.com/14793417/99198280-e3f2ac00-2754-11eb-9653-f6aae04e6bbe.png)

You'll notice that both the successful and unsuccessful connections are attempting to connect to the *same* peerID (`QmeYrvjcWzJTGRv8aJVfManvQfsBaN5fPwhXmMCzaLL`) but with the incorrect multiAddr the second time (`/ip4/127.0.0.1/tcp/4002/p2p/QmeYrvjcWzJTGRv8aJVfManvQfsBaN5fPwhXmMCzaLL`). 

I'm assuming this is because JS-IPFS is attempting to connect to the external node via its peerID through the JS-IPFS node within the browser. Which obviously won't work since the browser is not listening to `127.0.0.1`. 

How can I prevent JS-IPFS from attempting to connect to a local (`127.0.0.1`) when I already gave it a domain and multiAddr to connect to (`/dns4/example.com/tcp/4005/wss/p2p/QmeYrvjcWzJTGRv8aJVfManvQfsBaN5fPwhXmMCzaLL`)? 

#### Steps to reproduce the error:
```
ipfsNode = await IPFS.create();

const serverMultiAddr = `/dns4/example.com/tcp/4005/wss/p2p/QmeYrvjcWzJTGRv8aJVfManvQfsBaN5fPwhXmMCzaLL`;
console.log('serverMultiAddr', serverMultiAddr);

const connected = await ipfsNode.swarm.connect(serverMultiAddr);
console.log('connected', connected);
```

achingbrain:
libp2p stores known connection details in it's peer store and will try to reconnect when the node starts up (on page load after refreshing the page in this case).

[...]

PR #3409, opened 2020-11-20
feat: announce addresses via config

vasco-santos:
This PR adds the ability to add announces addresses via the config file and create options.

This is really useful for browser contexts, so that other peers can announce their public addresses, like dns addresses.

Ref: https://github.com/ipfs/js-ipfs/issues/3400


vasco-santos:
@achingbrain CI is failing with unrelated problems to this PR as far as I can tell

achingbrain:
#3411 will have fixed the CI issues, please merge master into this branch and it should go green.

achingbrain (approved):
Just waiting for green CI

vasco-santos:
It is 🟢

PR #3417, opened 2020-11-23
fix: do not write to prefix outside of output directory

achingbrain:
Adds a check to prevent overwriting files not under the output
directory, which you can override with a `--force` option.


jacobheun (approved):
pending CI changes LGTM

PR #3421, opened 2020-11-25
fix: make port transferables unique

### LINKED ISSUE #3402: ipfs-message-port-* should ensure no transferables are duplicated
Forking https://github.com/ipfs/js-ipfs/issues/3252#issuecomment-728471990 into separate issue.

Looks like some of our API calls return values that refer to same `ArrayBuffer`, which in turn causes one of the encoder functions https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-message-port-protocol/src to add it to the transfer list twice and error on `postMessage`.

icidasset:
Fixes #3402

The `transfer` list, ie. the second argument to `postMessage`, should always contain unique values, multiple of the same pointers cause errors. This PR fixes that, see linked issue above for more info. My solution was to just make everything in the list unique by converting the array to a set and back. Let me know if this works, or if another solution is preferred.

### DIFF
--- a/.github/workflows/bundlesize.yml
+++ b/.github/workflows/bundlesize.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         run: npm install
       - name: Bundlesize ${{ matrix.project }}
-        uses: ipfs/aegir/actions/bundle-size@v28.0.0
+        uses: ipfs/aegir/actions/bundle-size@v29.0.0
         with:
           project: ${{ matrix.project }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/packages/ipfs-message-port-client/README.md
+++ b/packages/ipfs-message-port-client/README.md
@@ -102,7 +102,7 @@ values instead of copying.
 
 > **Note:** Transferring data will empty it on the sender side which can lead to
 > errors if that data is used again later. To avoid these errors transfer option
-> was added so user can explicitily give up reference when it is safe to do so.
+> was added so user can explicitly give up reference when it is safe to do so.
 
 ```js
 /**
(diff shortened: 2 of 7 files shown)


achingbrain:
This looks great, thanks for submitting it. Please can you add a test to ensure regressions won't occur in future?

icidasset:
@achingbrain I added a test for `ipfs-message-port-server`.
Can confirm that the test fails if I remove the code changes.

When the test is ran in the browser environment, it does complain about `worker_threads` not being available, even though I'm detecting the correct environment. Possibly webpack that tries to resolve the require before the code is actually being executed? 🤷 Anyhow, it does seem to actually run the tests.

achingbrain (approved):

Gozala (approved):
Thanks @icidasset for identifying this issue and also fixing it! It looks good. There would be few nice to haves, but I would much rather land this now and address them in followups. Feel free to take a stab if you like, otherwise I'll do it when I'll get around to it.

1. Implement solution does more or less following `[...new Set(...transfer)]` to eliminate duplicates. But ideally we would just make `transfer` itself into a `Set` so we don't have to eliminated duplicates.
2. `postMessage` actually accepts Sets (or any iterable) for transfer list, so conversion back to array is unnecessary.
3. I think it would probably be better to amend [`ipfs-message-port-protocol` tests](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-message-port-protocol/test) to exercise moving [nodes e.g with duplicates CIDs](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-message-port-protocol/test/dag.spec.js).

achingbrain:
> I would much rather land this now and address them in followups.

@Gozala please can you open issues for the outstanding pieces of work so they don't get missed

Gozala:
> @Gozala please can you open issues for the outstanding pieces of work so they don't get missed

https://github.com/ipfs/js-ipfs/issues/3402#issuecomment-736701901

PR #3426, opened 2020-11-26
feat: enable upnp nat hole punching

achingbrain:
On by default, set [Swarm.DisableNatPortMap](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#swarmdisablenatportmap) to `true` to disable.

Depends on:

- [x] libp2p/js-libp2p#810


gammazero (approved):
Assuming `bundlesize: { maxSize: '527kB' },` is correct, then
LGTM

achingbrain:
> Assuming bundlesize: { maxSize: '527kB' }, is correct

Sadly it is. About 10% of that is from `ipfs-core`, the rest is dependencies, mostly crypto libraries.

You can see the breakdown by running:

```console
$ cd packages/ipfs-core
$ npm run build:js -- -b
```

PR #3427, opened 2020-11-27
chore: update libp2p to 0.30

vasco-santos:
Integrates `libp2p@0.30` release.

This PR changes needed types accordingly.

Needs:
- [x] libp2p@0.30
- [x] #3445 
- [x] #3442
- [x] https://github.com/ipfs/interop/pull/203
- [x] https://github.com/libp2p/js-libp2p-floodsub/pull/112
- [x] https://github.com/ChainSafe/js-libp2p-gossipsub/pull/146

A note should be included in the next release notes regarding the `websockets` [update](https://github.com/libp2p/js-libp2p-websockets/releases/tag/v0.15.0) to anticipate possible problems for local environment tests of IPFS users


packages/ipfs-core/src/components/index.js
```diff
@@ -70,6 +70,8 @@ class IPFS {
     })
     const dns = createDNSAPI()
     const isOnline = createIsOnlineAPI({ network })
+    // @ts-ignore This type check fails as options.
+    // libp2p can be a function, while IPNS router config expects libp2p config
```
vasco-santos, line 74: This is not an easy fix. I recommend we do it individually

packages/ipfs-core/src/components/key/export.js
```diff
@@ -23,11 +23,10 @@ module.exports = ({ keychain }) => {
    * ```
    * @param {string} name - The name of the key to export
    * @param {string} password - Password to set on the PEM output
-   * @param {import('.').AbortOptions} options
    * @returns {Promise<string>} - The string representation of the key
    */
-  const exportKey = (name, password, options) =>
-    keychain.exportKey(name, password, options)
+  const exportKey = (name, password) =>
+    keychain.exportKey(name, password)
```
vasco-santos, line 29: We do not support Aborts/options in keychain.

packages/ipfs-core/src/components/pubsub.js
```diff
@@ -41,6 +41,7 @@ module.exports = ({ network, config }) => {
    */
   async function subscribe (topic, handler, options) {
     const { libp2p } = await network.use(options)
+    // @ts-ignore Libp2p Pubsub is deprecating the handler, using the EventEmitter
```
vasco-santos, line 44: At some point we might get https://github.com/libp2p/js-libp2p/blob/0.30.x/src/pubsub-adapter.js this out. I can add in a follow up PR the handlers behaviour for IPFS

packages/ipfs-core/src/components/pubsub.js
```diff
@@ -111,7 +113,7 @@ module.exports = ({ network, config }) => {
    */
   async function ls (options) {
     const { libp2p } = await network.use(options)
-    return libp2p.pubsub.getTopics(options)
+    return libp2p.pubsub.getTopics()
```
vasco-santos, line 116: Pubsub has no options for getTopics + getSubscribers

packages/ipfs-core/src/components/storage.js
```diff
@@ -43,6 +43,8 @@ class Storage {
 
     const { peerId, keychain, isNew } = await loadRepo(repo, options)
 
+    // TODO: throw error?
```
vasco-santos, line 46: Keychain is not available in init IIRC. That is why it might exist or not

packages/ipfs-core/package.json
```diff
@@ -88,19 +88,19 @@
     "it-first": "^1.0.4",
     "it-last": "^1.0.4",
     "it-pipe": "^1.1.0",
-    "libp2p": "^0.29.3",
+    "libp2p": "libp2p/js-libp2p#fix/types-with-dist",
```
hugomrdias, line 91: the prepare script doesnt work ?

packages/ipfs-core/src/components/storage.js
```diff
@@ -52,7 +54,7 @@ module.exports = Storage
  *
  * @param {Repo} repo
  * @param {RepoOptions & InitOptions} options
- * @returns {Promise<{peerId: PeerId, keychain:Keychain, isNew:boolean }>}
+ * @returns {Promise<{peerId: PeerId, keychain: Keychain | undefined, isNew:boolean }>}
```
hugomrdias, line 57:
```suggestion
 * @returns {Promise<{peerId: PeerId, keychain?: Keychain, isNew:boolean }>}
```

packages/ipfs-core/src/components/storage.js
```diff
@@ -96,7 +98,7 @@ const openRepo = async (repo) => {
 /**
  * @param {Repo} repo
  * @param {RepoOptions & InitOptions} options
- * @returns {Promise<{peerId: PeerId, keychain:Keychain}>}
+ * @returns {Promise<{peerId: PeerId, keychain: Keychain | undefined}>}
```
hugomrdias, line 101:
```suggestion
 * @returns {Promise<{peerId: PeerId, keychain?: Keychain}>}
```

packages/ipfs-core/src/components/storage.js
```diff
@@ -197,7 +199,7 @@ const peerIdToIdentity = (peerId) => ({
  *
  * @param {Repo} repo
  * @param {ConfigureOptions} options
- * @returns {Promise<{peerId: PeerId, keychain:Keychain}>}
+ * @returns {Promise<{peerId: PeerId, keychain: Keychain | undefined}>}
```
hugomrdias, line 202:
```suggestion
 * @returns {Promise<{peerId: PeerId, keychain?: Keychain}>}
```

hugomrdias (changes requested):

packages/ipfs-core/package.json
vasco-santos: Was going to add it now. But there are some CID type check failing now.
So I did not try it out yet

vasco-santos: Added it now, but type check will fail :(

hugomrdias: CID errors should be fixed in the latest version

vasco-santos: it seems to be working in CI. But my local master branch is failing with it now

vasco-santos:
We seem to have a build problem now and libp2p's node_modules are replicated in the bundle.

This PR:

image

#master

image

cc @hugomrdias

packages/ipfs-core/src/components/storage.js
Gozala: So there was / is this code:

https://github.com/ipfs/js-ipfs/blob/2bafc884da6f1153f8ab2eea4f2418b372ca80e0/packages/ipfs-core/src/components/storage.js#L137-L156

And this code path

https://github.com/ipfs/js-ipfs/blob/2bafc884da6f1153f8ab2eea4f2418b372ca80e0/packages/ipfs-core/src/components/storage.js#L214-L230

And I'm still not sure why libp2p is created in two different ways to be honest. It would be good to get these cleaned up (in some followup). I'd also happily pair on this as lack of uncertainty here really bothered me.

Gozala, line 46: Seems like throw is not a great option here, if we keychain does not exists we should probably create it ? Otherwise I don't see how user is supposed to use this without getting an error.

packages/ipfs-core/src/components/swarm/disconnect.js
```diff
@@ -16,7 +16,7 @@ module.exports = ({ network }) => {
    */
   async function disconnect (addr, options) {
     const { libp2p } = await network.use(options)
-    return libp2p.hangUp(addr, options)
+    return libp2p.hangUp(addr)
```
Gozala, line 19: Just to do the same thing you did with `dial`

```suggestion
    await libp2p.hangUp(addr)
```

Gozala (approved):
Thanks for the pull @vasco-santos, looks good to me.

packages/ipfs-core/src/components/storage.js
vasco-santos: I agree with you on this, it also confused me. We should improve this on a followup

vasco-santos: As far as I understand, Keychain will always exist at this point as this is only used after start. The problem is that the types are not sure about it.

packages/interface-ipfs-core/src/utils/ipfs-options-websockets-filter-all.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const WS = require('libp2p-websockets')
+const filters = require('libp2p-websockets/src/filters')
+const transportKey = WS.prototype[Symbol.toStringTag]
+
+module.exports = () => ({
```
hugomrdias, line 7: this could be just an object since it doesn't take any input, it would reduce LoC in the others files

hugomrdias (changes requested):

packages/interface-ipfs-core/src/utils/ipfs-options-websockets-filter-all.js
vasco-santos: The main reason I created a function here is to enable the options to be customized from the caller without modifying the actual content thanks to the new reference. We had issues in the past in libp2p tests, where its base configuration used in tests was modified in some tests, which caused other tests to have different configurations than expected required.
We currently only use it directly and I can turn that into an object. What do you think?

hugomrdias: sounds good maybe freeze it to be safe if it makes sense.

hugomrdias (approved):

achingbrain (approved):

PR #3428, opened 2020-11-27
chore: test with downgraded electron

achingbrain:
Testing, not for merging

### DIFF
--- a/packages/interface-ipfs-core/package.json
+++ b/packages/interface-ipfs-core/package.json
@@ -30,7 +30,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "aegir": "^28.2.0",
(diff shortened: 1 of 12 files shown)


hugomrdias:
maybe we are not setting timeout correctly https://travis-ci.com/github/ipfs/js-ipfs/jobs/449951659#L344

achingbrain:
I'm not sure, it's breaking other builds that test on electron too: https://travis-ci.com/github/ipfs/js-ipfsd-ctl/builds/204955250

Previously when I've seen this sort of thing, it's been when using the `node-fetch` module on the electron main process, over a few requests the whole process sort of grinds to a halt. Using `electron-fetch` solves the problem.

Issue #3413, opened 2020-11-21
Discussion: Improving typeings


Gozala:
### Constraints

Here is the set of constraints following proposal attemts to satisfy (please call
out if you find something is missing or doesn't belong)

- Provide a way to ensure that `ipfs-core`  and `ipfs-http-client` implement
  same base API.
  - They could intentionally diverge a bit (e.g. http-client takes extra HTTPOptions)
    and so could `ipfs-core` but there should be common subset between two.
	- Compatibility should be ensured type checker as opposed to discipline or
    peer review
- Should require no type annotation repetitions.
- Types should be simple and stright forward, meaning avoid
  [magic types](https://www.typescriptlang.org/docs/handbook/utility-types.html)
  that are hard to understand.
- Cross package type dependecies should not depend on internal file structure
  (as much as possible)

### Proposal

I propose to satisify all of the listed constraints by doing following

1. Create a designated directory for reusable type definitions (e.g.
    `src/interface` or whatever wins bikeshed contest and have component specific
    interface definitions for `ipfs.pin`, `ipfs.files`, `ipfs.dag`, etc..

    > Note: This still creates file structure but it is agnostic of implementation
    details. Alternative could be to use [TS namespaces](https://www.typescriptlang.org/docs/handbook/namespaces.html),
    but they are left over from pre-esm days in TS and seems like non futureproof
    approach, which is why using naming convetion seems like a better option
    `ipfs.pin` -> `interface/pin`, `ipfs.files` -> `interface/files`, etc...

    Below is illustration of what `interface/pin.ts` could look like:

    ```ts
    import CID from "cids"
    import { AbortOptions } from "./utils"

    export interface API {
      addAll (source: ToPins, options?: AddOptions & AbortOptions): AsyncIterable<CID>
      add(source: ToPin, options?: AddOptions & AbortOptions): Promise<CID>
      rmAll(source: ToPins, options?: RemoveOptions & AbortOptions): AsyncIterable<CID>
      rm(source: ToPin, options?: RemoveOptions & AbortOptions): Promise<CID>
      ls(options?: ListOptions & AbortOptions): AsyncIterable<PinEntry>
    }

    export type ToPin =
      | CID
      | string
      | ToPinWithPath
      | ToPinWithCID

    export type ToPinWithPath = {
      path: string
      recursive?: boolean
      metadata?: any
      cid?: undefined
    }

    export type ToPinWithCID = {
      cid?: CID
      recursive?: boolean
[...]

Issue #3053, opened 2020-05-28
feature request: include reference to IPFS inside node instance


tabcat:
Hi, I'd like to be able to access classes like CID from just an 'instance' of ipfs made from the .create method on the IPFS factory object.
My use case is that I have a package that exports a class that takes an instance of ipfs as a parameter. I want to use the CID class included with the IPFS package instead of having to require it. Adding a reference to the IPFS factory object from the ipfs node instance object is one way to enable this.

achingbrain:
> I want to use the CID class included with the IPFS package instead of having to require it.

This doesn't sound like a great idea, an instance should not have a reference to its factory.

Could you expand a little more on why you would do that?  Why not just require it?

```js
const {
  CID
} = require('ipfs')
```

tabcat:
Pardon my ignorance on this but it seems like Class instances in javascript all have references to their factories via `.constructor`. Is it different since IPFS is not a Class or is it a good rule? Explicitly making a reference to the factory for the ipfs instance does sound useless, except for being able to access those bundled Classes.

> Could you expand a little more on why you would do that? It's bundled with IPFS as a top-level export, why not just require it?

I believe I just wished to inject CID with the IPFS instance. OrbitDB and my package that uses OrbitDB take an ipfs instance as a parameter to keep things simple for users. I was probably specifically looking to eliminate an older version of whats happening [here](https://github.com/cypsela/sailplane-node/blob/76e33c11018bacc4fc69649799c6e72abf272868/src/SharedFS.js#L90), which is actually not that bad because I have to create those preceding CID instances anyway.

achingbrain:
> it seems like Class instances in javascript all have references to their factories via .constructor

They have a reference to their constructor, but a factory function might not be the constructor:

```js
class Foo {
  constructor () {

  }
}

function createFoo () {
  return new Foo()
}
```

IPFS is a class, but you get instances of it via the `create` factory method because it needs to do some async work before the node is ready to use (set up the repo, open ports, etc) which cannot be done in a constructor since they cannot be marked async.

> OrbitDB and my package that uses OrbitDB take an ipfs instance as a parameter
> I was probably specifically looking to eliminate an older version of whats happening here

In this project I would declare `ipfs` as a peer dependency and just require `CID` from it as per my previous comment above.

This also says to the user 'my project is compatible with this version of ipfs` and the user will get a warning if they don't declare a dep on ipfs at the right version.

PR #3439, opened 2020-12-07
Export IPFS instance type

Xmader:
solves https://github.com/ipfs/js-ipfs/issues/2945#issuecomment-731784171

## Usage 

```ts

import IPFS from 'ipfs'

function doSomething (ipfs: IPFS) {
    // ...
}

// ...
doSomething(await IPFS.create())
```

### DIFF
--- a/packages/ipfs-core/src/index.js
+++ b/packages/ipfs-core/src/index.js
@@ -13,6 +13,10 @@ const multihash = multihashing.multihash
 const CID = require('cids')
 const IPFS = require('./components')
 
+/**
+ * @typedef {import('./components')} default
+ */
+
 module.exports = {
   create: IPFS.create,
   crypto,
--- a/packages/ipfs-core/src/index.js
+++ b/packages/ipfs-core/src/index.js
@@ -11,14 +11,14 @@ const multicodec = require('multicodec')
 const multihashing = require('multihashing-async')
 const multihash = multihashing.multihash
 const CID = require('cids')
-const IPFS = require('./components')
+const { create } = require('./components')
 
 /**
  * @typedef {import('./components')} default
  */
 
 module.exports = {
-  create: IPFS.create,
+  create,
   crypto,
   isIPFS,
   CID,
--- a/packages/ipfs-core/src/index.js
+++ b/packages/ipfs-core/src/index.js
@@ -14,6 +14,11 @@ const CID = require('cids')
 const { create } = require('./components')
 
 /**
+ * Export IPFS instance type
+ *
+ * This will result in `export default import("./components")`
+ * in the generated `d.ts` file
+ *
  * @typedef {import('./components')} default
  */
 
--- a/packages/ipfs/src/index.js
+++ b/packages/ipfs/src/index.js
@@ -3,7 +3,12 @@
 const IPFS = require('ipfs-core')
 
 /**
- * @typedef {import('ipfs-core/src/components')} IPFS default
+ * Export IPFS instance type
+ *
+ * This will overlap onto the default export
+ * in the generated `d.ts` file
+ *
+ * @typedef {import('ipfs-core').default} IPFS
  */
 
 module.exports = IPFS
--- a/packages/ipfs/src/index.js
+++ b/packages/ipfs/src/index.js
@@ -1,14 +1,12 @@
 'use strict'
 
-const IPFS = require('ipfs-core')
+module.exports = require('ipfs-core')
 
 /**
  * Export IPFS instance type
  *
- * This will overlap onto the default export
+ * This will result in `export default …`
  * in the generated `d.ts` file
- *
- * @typedef {import('ipfs-core').default} IPFS
  */
-
-module.exports = IPFS
+// @ts-ignore duplicate identifier
+/** @typedef {import('ipfs-core').default} default */
--- a/packages/ipfs/src/index.js
+++ b/packages/ipfs/src/index.js
@@ -2,4 +2,8 @@
 
 const IPFS = require('ipfs-core')
 
+/**
+ * @typedef {import('ipfs-core/src/components')} IPFS
+ */
+
 module.exports = IPFS
--- a/packages/ipfs/src/index.js
+++ b/packages/ipfs/src/index.js
@@ -1,12 +1,14 @@
 'use strict'
 
-module.exports = require('ipfs-core')
+const IPFS = require('ipfs-core')
 
 /**
  * Export IPFS instance type
  *
- * This will result in `export default …`
+ * This will overlap onto the default export
  * in the generated `d.ts` file
+ *
+ * @typedef {import('ipfs-core').default} IPFS
  */
-// @ts-ignore duplicate identifier
-/** @typedef {import('ipfs-core/src/components')} default */
+
+module.exports = IPFS
--- a/packages/ipfs/src/index.js
+++ b/packages/ipfs/src/index.js
@@ -9,4 +9,4 @@ module.exports = require('ipfs-core')
  * in the generated `d.ts` file
  */
 // @ts-ignore duplicate identifier
-/** @typedef {import('ipfs-core').default} default */
+/** @typedef {import('ipfs-core/src/components')} default */
--- a/packages/ipfs/src/index.js
+++ b/packages/ipfs/src/index.js
@@ -3,7 +3,7 @@
 const IPFS = require('ipfs-core')
 
 /**
- * @typedef {import('ipfs-core/src/components')} IPFS
+ * @typedef {import('ipfs-core/src/components')} IPFS default
  */
 
 module.exports = IPFS
--- a/packages/ipfs/src/index.d.ts
+++ b/packages/ipfs/src/index.d.ts
@@ -1,0 +1,3 @@
+type IPFS = import('ipfs-core/src/components')
+declare const IPFS: typeof import('ipfs-core')
+export = IPFS
--- a/packages/ipfs/src/index.d.ts
+++ b/packages/ipfs/src/index.d.ts
@@ -1,3 +1,0 @@
-type IPFS = import('ipfs-core/src/components')
-declare const IPFS: typeof import('ipfs-core')
-export = IPFS


achingbrain:
@Gozala could you take a look at this please?  I was under the impression #3359 resolved this issue.

Gozala (approved):

Gozala:
> @Gozala could you take a look at this please? I was under the impression #3359 resolved this issue.

I think I have overlooked the fact that `IPFS` class does not make all the way to `module.exports` but instead just it’s `create` does:

https://github.com/ipfs/js-ipfs/blob/c3c46073d31445411bf1f379f05c562eb628c860/packages/ipfs-core/src/index.js#L16-L29

@Xmader would you mind instead changing above lines so we end up exporting `IPFS` class and avoid having to introduce a type here ? I’m assuming @achingbrain is onboard with that idea, otherwise this is probably a way to go.

Gozala (changes requested):
@Xmader would you mind instead changing `ipfs-core/src/index` so that `IPFS` ends up been a class and we don’t have to introduce a type here too ? I’m assuming @achingbrain is onboard with this.

Xmader:
> @Xmader would you mind instead changing `ipfs-core/src/index` so that `IPFS` ends up been a class and we don’t have to introduce a type here too ? I’m assuming @achingbrain is onboard with this.

There is no need to introduce a type in `ipfs-core/src/index`.

Xmader:
> @Xmader would you mind instead changing `ipfs-core/src/index` so that `IPFS` ends up been a class and we don’t have to introduce a type here too ? I’m assuming @achingbrain is onboard with this.

I tried changing https://github.com/ipfs/js-ipfs/blob/c3c46073d31445411bf1f379f05c562eb628c860/packages/ipfs-core/src/index.js#L16-L29 to 
```js
module.exports = class extends IPFS {
  static create = IPFS.create
  static crypto = crypto
  static isIPFS = isIPFS
  static CID = CID
  static multiaddr = multiaddr
  static multibase = multibase
  static multihash = multihash
  static multihashing = multihashing
  static multicodec = multicodec
  static PeerId = PeerId
  static globSource = globSource
  static urlSource = urlSource
}
``` 
, but does not work.

achingbrain:
I'm not sure exporting the IPFS class as the default export from `ipfs` is a great idea because we export `CID`, `crypto` etc as properties of the default export, so they'd become static fields of the `IPFS` class which seems like the wrong place to put them.

We could export the class as another property on the default:

```js
const { IPFS } = require('ipfs-core')

const ipfs = new IPFS(...)
```

..but we don't do that right now by design because the node needs to do some async work before it's ready to be used.  We found so much copy pasting of:

```js
const IPFS = require('ipfs')

const node = new IPFS(opts)

node.on('error', (e) => {
  // .. error handling
})

node.on('ready', () => {
  // ... real works starts here
})
```

..that we introduced the `create` factory function which is much simpler to use, can be wrapped in try/catch, etc:

```js
const { create } = require('ipfs')

const node = await create(opts)
```

..so we'd be intentionally exporting something that isn't terribly useful and we don't want people to actually use just to satisfy the TypeScript compiler, which seems like a high price to pay conceptually and in user confusion.  If they're faced with a class and a factory method, which should they use?  Why do they have to make this decision?  Is one better than the other, etc.

Is there a way to just cause the return type of the factory method to end up in the generated typedef file with a JSDoc comment?

Xmader:
> Is there a way to just cause the return type of the factory method to end up in the generated typedef file with a JSDoc comment?

```js
/**
 * @typedef {ReturnType<typeof import('ipfs-core')['create']> extends Promise<infer U> ? U : never} IPFS
 */
```
which ends up to be
```js
/**
 * @typedef {import('ipfs-core/src/components')} IPFS
 */
```
(this pull request)

Gozala:
> ..so we'd be intentionally exporting something that isn't terribly useful and we don't want people to actually use just to satisfy the TypeScript compiler, which seems like a high price to pay conceptually and in user confusion. If they're faced with a class and a factory method, which should they use? Why do they have to make this decision? Is one better than the other, etc.

Not trying to argue, just presenting different point of view on this: There is nothing inherently wrong with doing `new IPFS({ ... })` however that would require passing **initialized** repo and **started** network component, that users could do by doing all that nitty gritty on their own. In most cases however is is probably more convenient to use `IPFS.create` instead that will take care of that.
 

> Is there a way to just cause the return type of the factory method to end up in the generated typedef file with a JSDoc comment?

Problem is that `IPFS` class (or a type) isn't exposed by `ipfs` module so that TS user can't write a function annotation like `doSomethingWithIPFS(ipfs: IPFS, ....)` because they can't import `IPFS` here.

I thought just making `require('ipfs')` be that `IPFS` class was a better approach than what this pr does, because you don't wind up with something like:

```ts
import IPFS from 'ipfs`

const startApp (ipfs:IPFS) => {
}

const async main = () => {
   const ipfs = await IPFS.create(...)
   startApp(ipfs)
}
```

And instead do:

```ts
import IPFS from 'ipfs`
import { IPFS as IPFSType } from 'ipfs'

const startApp (ipfs:IPFSType) => {
}

const async main = () => {
   const ipfs = await IPFS.create(...)
   startApp(ipfs)
}
```

Gozala (approved):
Given that @achingbrain doesn't seem to like idea of exposing `IPFS` class, I think this is the best alternative we have. Lets do this for now and we can always improve this in the future.

Xmader:
> And instead do:
> 
> ```ts
> import IPFS from 'ipfs'
> import { IPFS as IPFSType } from 'ipfs'
> ```

People can do 
```ts
import IPFS from '.'
import type IPFSType from '.'
```
with this pr

Gozala:
> > And instead do:
> > ```ts
> > import IPFS from 'ipfs'
> > import { IPFS as IPFSType } from 'ipfs'
> > ```
> 
> People can do
> 
> ```ts
> import IPFS from '.'
> import type IPFSType from '.'
> ```
> 
> with this pr

Cool, that is better! But I do still think that it's kind of confusing & not sure that is even possible to express in JSDoc syntax. So personally I would prefer just exposing `IPFS` class and assume that users that choose to use `new IPFS` know what they are doing. 

I also recognize that following is not ideal:

>  we export CID, crypto etc as properties of the default export, so they'd become static fields of the IPFS class which seems like the wrong place to put them.

But than again I don't personally find it to be so concerning and we even had a request to do #3053

It is also worth considering that if and when we switch to ESM we will no longer need to stick them into class statics as IPFS could be a default export and CID, crypto etc named ones.

Gozala:
@Xmader, @achingbrain other alternative worth considering might be to stick `IPFS` class into module exports along with `CID` and others. That way it is possible to just do following:

```ts
import { IPFS } from "ipfs"

const startApp (ipfs:IPFS) => {
}

const async main = () => {
   const ipfs = await IPFS.create(...)
   startApp(ipfs)
}
```

But it's also possible to do what we do today:

```js
const IPFS = require('ipfs')

const main = async () => {
  const ipfs = await IPFS.create()
}
```

And I think that would also align with potential ESM in the future.

Xmader:
> not sure that is even possible to express in JSDoc syntax. 

It can. 

```js
const IPFS = require('ipfs')

/**
 * @typedef {import('ipfs')} IPFS
 */
```

achingbrain:
If I have `test.ts` like this (nb. importing `ipfs-core` and not `ipfs` since unless you need the daemon, the HTTP API and the CLI components you should be using `ipfs-core` and not `ipfs`):

```ts
// test.ts
import { create, IPFS } from "ipfs-core"

const startApp = async (ipfs:IPFS): Promise<void> => {
  await ipfs.id()
  // ... more code here
}

const main = async (): Promise<void> => {
   const ipfs = await create({})
   await startApp(ipfs)
}
```



Adding the typedef to `/packages/ipfs-core/src/index.js` like:

```js
// ... more code here
const CID = require('cids')
const IPFS = require('./components')

/**
 * @typedef {import('./components')} IPFS     <-- added typedef
 */

module.exports = {
  create: IPFS.create,
// ... more code here
```

..results in a `.d.ts` like:

```ts
export type IPFS = import("./components");
export const crypto: typeof import("libp2p-crypto");
export const CID: any;
// ... more code here
export declare const create: typeof import("./components").create;
//# sourceMappingURL=index.d.ts.map
```

..`test.ts` then compiles happily.

But we should support requiring `ipfs` too for those who missed [the blog post](https://blog.ipfs.io/2020-10-29-js-ipfs-0-50/).

The `ipfs` module should just re-export all of the types from `ipfs-core`, including `IPFS`, though I'm having problems getting it to do that.  It'll export everything from `ipfs-core` /except/ the IPFS type, which I can only export by renaming it (otherwise it complains about a duplicate identifier "IPFS"):

```js
// index.js
const IPFSCore = require('ipfs-core')

/**
 * @typedef {import('ipfs-core/src').IPFS} RenamedIPFS
 */

module.exports = IPFSCore

// index.d.ts
export = IPFSCore;
declare const IPFSCore: typeof import("ipfs-core/src");
declare namespace IPFSCore {
    export { RenamedIPFS };
}
type RenamedIPFS = import("ipfs-core/src/components");
```

If I remove the name it gets called 'exports' which is less than ideal:

```js
// index.js
const IPFSCore = require('ipfs-core')

/**
 * @typedef {import('ipfs-core/src').IPFS}
 */

module.exports = IPFSCore

// index.d.ts
export = IPFSCore;
declare const IPFSCore: typeof import("ipfs-core/src");
declare namespace IPFSCore {
    export { exports };
}
type exports = import("ipfs-core/src/components");
```

If I go in and manually edit `.d.ts` file to change the name, it seems to work and I can compile `test.ts` with the `import` changed to from `ipfs-core` to `ipfs`:

```ts
export = IPFSCore;
declare const IPFSCore: typeof import("ipfs-core/src");
declare namespace IPFSCore {
    export { IPFS };
}
type IPFS = import("ipfs-core/src/components");
```

Any ideas how to get `ipfs` to generate this `.d.ts` file or otherwise to just export all of the types from `ipfs-core` @Gozala @Xmader ?

Xmader:
@achingbrain 

Add `// @ts-ignore` in `/packages/ipfs-core/src/index.js`

```js
// @ts-ignore duplicate identifier
/** @typedef {import('ipfs-core/src/components')} IPFS */
```

The `.d.ts` file can be generated perfectly.

```ts
export = IPFS;
declare const IPFS: typeof import("ipfs-core/src");
declare namespace IPFS {
    export { IPFS };
}
type IPFS = import("ipfs-core/src/components");
//# sourceMappingURL=index.d.ts.map
```

Xmader:
By the way, people can [replace `ipfs` with `ipfs-core`](https://blog.ipfs.io/2020-10-29-js-ipfs-0-50/#ipfs-core-a-new-lighter-ipfs-for-application-authors) using [`npm i ipfs@npm:ipfs-core`](https://github.com/npm/rfcs/blob/latest/implemented/0001-package-aliases.md), so that it doesn't need to update the require name

Xmader:
Is that OK to export `IPFS` type under `export default` as well, in `ipfs-core`?

```js
// packages/ipfs-core/src/index.js

/**
 * @typedef {import('./components')} IPFS
 * @typedef {IPFS} default
 */

```

generated `d.ts` file

```ts
export type IPFS = import("./components");
type _default = import("./components");
export default _default;
export const crypto: typeof import("libp2p-crypto");
export const isIPFS: any;
export const CID: typeof import("cids");
// ... more code here
export declare const create: typeof import("./components").create;
//# sourceMappingURL=index.d.ts.map
```

achingbrain:
Interesting.  How about this:

```js
// packages/ipfs-core/src/index.js

// ... other code
const IPFS = require('./components')

/**
 * @typedef {import('ipfs-core/src/components')} default
 */

module.exports = {
  create: IPFS.create,
// ... other code
```

Generates:

```ts
type _default = import("./components");
export default _default;
export const crypto: typeof import("libp2p-crypto");
export const isIPFS: any;
export const CID: typeof import("cids");
export const multiaddr: typeof import("multiaddr");
export const multibase: typeof import("multibase/src/");
export const multihash: any;
export const multihashing: any;
export const multicodec: any;
export const PeerId: typeof import("peer-id");
export const globSource: any;
export const urlSource: any;
export declare const create: typeof import("./components").create;
//# sourceMappingURL=index.d.ts.map
```

```js
// packages/ipfs/src/index.js

const IPFS = require('ipfs-core')

/** @typedef {import('ipfs-core/src/components')} IPFS default */

module.exports = IPFS
```

Generates:

```ts
export = IPFS;
declare const IPFS: typeof import("ipfs-core/src");
declare namespace IPFS {
    export { IPFS };
}
/**
 * default
 */
type IPFS = import("./components");
//# sourceMappingURL=index.d.ts.map
```

Then:

```ts
import type IPFS from "ipfs"  // <--- can be "ipfs" or "ipfs-core"
import { create } from "ipfs" // <--- can be "ipfs" or "ipfs-core"

const startApp = async (ipfs:IPFS): Promise<void> => {
  await ipfs.id()
}

const main = async (): Promise<void> => {
   const ipfs = await create({})
   await startApp(ipfs)
}
```

No `@ts-ignore` required.

If having `IPFS` as a separate type import is preferable and @Gozala is happy with this approach we can get this merged ASAP.

Xmader:
I don't know which approach is more straightforward, and can get people less confused
 `export default`, export as a separate type, or both?

I personally prefer just `export default`.

achingbrain:
My opinion's not that strong, I like not requiring `@ts-ignore` to make it work but I don't know if having a separate `import type` will violate some sort of convention.  Any thoughts @Gozala or is this good to go?

packages/ipfs-core/src/index.js
```diff
@@ -13,6 +13,10 @@ const multihash = multihashing.multihash
 const CID = require('cids')
 const IPFS = require('./components')
```
Gozala, line 14: Seems like it might be simpler to just do this ?

```suggestion
const { create } = require('./components')

/**
 * @typedef {import(‘./components’)} IPFS
 */
```

packages/ipfs/src/index.js
```diff
@@ -2,4 +2,8 @@
 
 const IPFS = require('ipfs-core')
```
Gozala, line 3: I think it would be simpler to do this instead

```suggestion
/**
 * @typedef {import(‘ipfs-core/src/components’)} IPFS
 */
 
 module.exports = require('ipfs-core')
```

Gozala (commented):
I do not want to block this, but the `default` export as a means to export type makes me a little uncomfortable because:

- I’m not exactly sure how all this is going to work when we have actual ESM modules.
- Overlapping namespace of types and implementations is just confusing & TS with all it’s configs and issues makes me worried it’s going to be source of pain.

I think it would be a lot simpler to just export `IPFS` named type from both files and avoid naming conflicts by giving a different name to local bindings (as suggested in the PR)

packages/ipfs/src/index.js
```diff
@@ -2,4 +2,8 @@
...
 
+/**
+ * @typedef {import('ipfs-core/src/components')} IPFS default
```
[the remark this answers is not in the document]
Xmader: `IPFS` is the exported type, `default` is the description saying it will be the default export

Xmader:
> I’m not exactly sure how all this is going to work when we have actual ESM modules.

It can still work 

`packages/ipfs-core/src/index.js`

```js
// ...
const multihash = multihashing.multihash
const CID = require('cids')

/**
 * @typedef {import('./components')} default
 */

export { create } from './components'
export {
  crypto,
  isIPFS,
  CID,
  multiaddr,
  multibase,
  multihash,
  multihashing,
  multicodec,
  PeerId,
  globSource,
  urlSource
}
```

Generates: (not much differ from what we have now)
```ts
export { create } from "./components";
type _default = import("./components");
export default _default;
export const crypto: typeof import("libp2p-crypto");
export const isIPFS: any;
export const CID: typeof import("cids");
export const multiaddr: typeof import("multiaddr");
export const multibase: typeof import('multibase');
export const multihash: any;
export const multihashing: any;
export const multicodec: any;
export const PeerId: typeof import("peer-id");
export const globSource: any;
export const urlSource: any;
//# sourceMappingURL=index.d.ts.map
```

`packages/ipfs/src/index.js`

```js
/**
 * @typedef {import('ipfs-core').default} default
 */
export * from 'ipfs-core'
```

Generates:
```ts
export * from "ipfs-core/src";
type _default = import("ipfs-core/src/components");
export default _default;
//# sourceMappingURL=index.d.ts.map
```

packages/ipfs/src/index.js
Xmader: It does not work because `@typedef {import('ipfs-core/src/components')} IPFS` will not overlap onto the exports

Generated `d.ts` file:

```ts
declare const _exports: typeof import('ipfs-core/src')
export = _exports;
export type IPFS = import('ipfs-core/src/components');
```

We can't use 
```ts
import type IPFS from "ipfs"
```

packages/ipfs-core/src/index.js
Xmader: Yes

packages/ipfs/src/index.js
Gozala: @Xmader I realize that, but point I was trying to make is, having export type `IPFS` is just simpler and in the nutshell will give us:

```ts
import { IPFS, create } from "ipfs"
```

Furthermore I expect that as we adopt approach proposed in #3413 we might end up with `IPFS` interface that in the future could be exported in place of `IPFS` type.

That said, it's an @achingbrain call

Gozala:
@Xmader I end up creating another pull request https://github.com/ipfs/js-ipfs/pull/3447 to illustrate what I was suggesting. Please do not take it as a subversion to your efforts, your contributions and time is really valuable, it just felt that having two different options to compare pros & cons side by side seemed more effective way to get to a consensus.

achingbrain:
[Like I said](https://github.com/ipfs/js-ipfs/pull/3439#issuecomment-744629690), my opinion on:

```ts
import type IPFS from "ipfs"
import { create } from "ipfs"
```

vs

```ts
import { IPFS, create } from "ipfs"
```

is just not that strong.  I think less verbosity is better in general but I'm happy to be talked around.

> I’m not exactly sure how all this is going to work when we have actual ESM modules.

I don't think we should worry about that right now as that will almost certainly need to be released as a breaking change anyway.

We can go with #3447 as it [doesn't make CID etc look like class fields of IPFS](https://github.com/ipfs/js-ipfs/pull/3439#issuecomment-741847305).

@Xmader I second @Gozala - thanks so much for taking the time to look in to this and to make so many changes to the PR, your contributions are very much appreciated.

PR #3440, opened 2020-12-10
Fix ipfs.ls() for a single file object

ikreymer:
This fixes a typo where `ipfs.ls()` returns instead of yielding from a generator.
The result is that ipfs.ls() does not work when the path is a file. I think this should fix it, with `ls` returning a generator with a single file result.

(Note: I wanted to add a test but could not find where this command is being tested).

### DIFF
--- a/packages/interface-ipfs-core/src/ls.js
+++ b/packages/interface-ipfs-core/src/ls.js
@@ -220,11 +220,12 @@ module.exports = (common, options) => {
     })
 
     it('should ls single file', async () => {
+      const dir = randomName('DIR')
       const file = randomName('F0')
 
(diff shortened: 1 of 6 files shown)


achingbrain:
Thanks for submitting this, great catch & the fix looks good.

The tests for this part of the API live here: https://github.com/ipfs/js-ipfs/blob/master/packages/interface-ipfs-core/src/ls.js - could you please add a test for this and I'll get a patch release out.

achingbrain (approved):

achingbrain:
Thanks for adding the tests, I added a couple more and fixed up the breakage when invoking the api over http.

PR #3442, opened 2020-12-11
fix: regressions introduced by new releases of CID & multicodec

Gozala:
Few regressions had been introduced by

https://github.com/multiformats/js-cid/commit/0e11f035c9230e7f6d79c159ace9b80de88cb5eb#diff-25a6634263c1b1f6fc4697a04e2b9904ea4b042a89af59dc93ec1f5d44848a26

https://github.com/multiformats/js-multicodec/commit/25cf33fdc392da21d776f6f4e85b678101724364

This changes fix two remaining problems:

1. Change type of codec from string to code name.
2. No longer import removed JSON file and use exported table instead.


vasco-santos (approved):
LGTM!
Thanks for looking into this @Gozala

vasco-santos (commented):
It seems that (another release?) broke anything else. I am now getting:

```
../ipfs-core-utils/src/cid.js:32:34 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type '"base58btc" | "identity" | "\0" | "0" | "7" | "9" | "f" | "F" | "v" | "V" | "t" | "T" | "b" | "B" | "c" | "C" | "h" | "k" | "K" | "z" | "Z" | "m" | "M" | "u" | "U" | "base2" | "base8" | "base10" | ... 18 more ... | undefined'.
  Type 'string' is not assignable to type '"base58btc" | "identity" | "\0" | "0" | "7" | "9" | "f" | "F" | "v" | "V" | "t" | "T" | "b" | "B" | "c" | "C" | "h" | "k" | "K" | "z" | "Z" | "m" | "M" | "u" | "U" | "base2" | "base8" | "base10" | ... 18 more ... | undefined'.

32   return cid.toBaseEncodedString(options.base)
                                    ~~~~~~~~~~~~


Found 1 error.
```

Same in the examples job in travis

hugomrdias (approved):

Gozala:
> It seems that (another release?) broke anything else. I am now getting:
> 
> ```
> ../ipfs-core-utils/src/cid.js:32:34 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type '"base58btc" | "identity" | "\0" | "0" | "7" | "9" | "f" | "F" | "v" | "V" | "t" | "T" | "b" | "B" | "c" | "C" | "h" | "k" | "K" | "z" | "Z" | "m" | "M" | "u" | "U" | "base2" | "base8" | "base10" | ... 18 more ... | undefined'.
>   Type 'string' is not assignable to type '"base58btc" | "identity" | "\0" | "0" | "7" | "9" | "f" | "F" | "v" | "V" | "t" | "T" | "b" | "B" | "c" | "C" | "h" | "k" | "K" | "z" | "Z" | "m" | "M" | "u" | "U" | "base2" | "base8" | "base10" | ... 18 more ... | undefined'.
> 
> 32   return cid.toBaseEncodedString(options.base)
>                                     ~~~~~~~~~~~~
> 
> 
> Found 1 error.
> ```
> 
> Same in the examples job in travis
> 
> EDIT:
> 
> We can make `@param {import('cids').BaseNameOrCode} [options.base]` in this case. Several other issues appear from this though

I was failing to reproduce any of that, but now I see it that it's because you've locked the CID version  #3445. I'll revert that change here and try to figure out remaining issues.

Gozala:
@vasco-santos fixed all the remaining issues

vasco-santos (commented):
This looks good to me. However you need to add back the bundle size increase that was in the reverted commit

vasco-santos:
Also the example-custom-ipld-formats needs an update on the base table import

achingbrain (approved):
LGTM, just waiting on green CI

vasco-santos (approved):

examples/custom-ipld-formats/daemon-node.js
```diff
@@ -5,7 +5,7 @@
 const codecName = 'dag-test'
 const codecNumber = 392091
 
-const baseTable = require('multicodec/src/base-table.json')
+const { baseTable } = require('multicodec/src/base-table')
```
achingbrain, line 8: Not sure this is working as expected.  CI says:

```
example-custom-ipld-formats: Error: `put` requires a format
example-custom-ipld-formats:     at IPLDResolver.put (/home/travis/build/ipfs/js-ipfs/node_modules/ipld/src/index.js:180:13)
example-custom-ipld-formats:     at put (/home/travis/build/ipfs/js-ipfs/node_modules/ipfs-core/src/components/dag/put.js:75:30)
example-custom-ipld-formats:     at Object.put (/home/travis/build/ipfs/js-ipfs/node_modules/ipfs-core/src/utils.js:231:46)
example-custom-ipld-formats:     at main (/home/travis/build/ipfs/js-ipfs/examples/custom-ipld-formats/in-process-node.js:49:30)
```

vasco-santos: I can try to see what is happening here in a couple of hours

vasco-santos: So, an Object.freeze is exported from base-table. This means that the object is not changed by the following change

Gozala:
It appears to be some problem with generated types, probably `file.cid` is `any` here which causes the problem in the check.

Gozala:
> It appears to be some problem with generated types, probably `file.cid` is `any` here which causes the problem in the check. I'm looking into this

Problem appears to be due to fact that ts generates`dist/src/components/add-all/index.d.ts` file with a fragment:

```
export type CID = import("cids/src/");
```

Which seems to fail to resolve, because https://github.com/microsoft/TypeScript/issues/41281

which aren't caught by mapping we use 😭

https://github.com/multiformats/js-cid/blob/master/package.json#L10-L13

examples/types-use-ipfs-from-typed-js/src/main.js
```diff
@@ -17,7 +17,6 @@ async function main () {
 
   console.log('Added file:', file.path, file.cid.toString())
   try {
-    // @ts-expect-error CID has no toUpperCase method
```
Gozala, line 20: I think this defeats the point of the point of this example to a degree. It was there to ensure that TS is able to pick up types properly. Unfortunately it got broken, but we should fix not remove the check.

I think it would be better to temporarily disable this example test instead of just removing this line.

achingbrain: The `@ts-expect-error` was only added [yesterday](https://github.com/ipfs/js-ipfs/pull/3447) so the purpose of this this example has only ever been defeated.

The it's never run in CI because the module name was not picked up by the [scope filter](https://github.com/ipfs/js-ipfs/blob/master/.travis.yml#L243).

Please can you submit a fix in a follow up PR.  This one cannot wait any longer.

PR #3445, opened 2020-12-14
chore: fixed cid and multicodec versions

vasco-santos:
This PR sets fixed versions for cid and multicodec prior to their types PR being merged. These updates lead to a few regressions that should be fixed for updating these deps

Context:
- https://github.com/ipfs/js-ipfs/pull/3442
- Some follow up breaking changes will probably land https://github.com/multiformats/js-multicodec/pull/73


vasco-santos:
@achingbrain bundle size has increased a bit. Should I increase it in the PR?

achingbrain:
> Should I increase it in the PR?

Yes please, otherwise it'll break master.

achingbrain (approved):

PR #3447, opened 2020-12-15
fix: export IPFS type

Gozala:
This is an attempt to illustrate a different approach for exporting IPFS type than one proposed in #3439.

Here instead of using `default export` for type `IPFS` type is just exported with it's name so that user in ts can write `import { IPFS, create } from "ipfs"` and equivalent in js as well.

Pull also updates two examples we have in place to both illustrate & test this.

### DIFF
--- a/examples/types-use-ipfs-from-typed-js/src/main.js
+++ b/examples/types-use-ipfs-from-typed-js/src/main.js
@@ -1,11 +1,11 @@
 const { create } = require('ipfs')
 /**
- * @typedef {import('ipfs').IPFS} IPFS 
+ * @typedef {import('ipfs').IPFS} IPFS
  * @typedef {import('cids')} CID
  */
 
 async function main () {
-  const node = await IPFS.create()
+  const node = await create()
   const version = await node.version()
 
   console.log('Version:', version.version)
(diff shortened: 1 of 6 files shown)


examples/types-use-ipfs-from-typed-js/src/main.js
```diff
@@ -1,4 +1,8 @@
-const IPFS = require('ipfs')
+const { create } = require('ipfs')
+/**
+ * @typedef {import('ipfs').IPFS} IPFS 
+ * @typedef {import('cids')} CID
+ */
 
 async function main () {
   const node = await IPFS.create()
```
achingbrain, line 8: Is IPFS defined now?  The require was deleted.

achingbrain: Ah, I think the test for this module isn't being run in CI because it's name in the `package.json` does not start with `example-` - see https://github.com/ipfs/js-ipfs/blob/master/.travis.yml#L243

achingbrain: I've updated the PR so that the tests now run in CI

PR #3449, opened 2020-12-16
chore: refactor common types

Gozala:
This pull request targets #3442 and addresses following issue https://github.com/ipfs/js-ipfs/pull/3442#issuecomment-745564672 by refactoring some of the common types used by root APIs as per #3413


achingbrain:
Please could you rebase on top of master?

achingbrain:
Also, if we could not have `ipfs-interface` and `interface-ipfs-core` it would help everyone's sanity.  The `interface` term is [used](https://github.com/libp2p/interface-stream-muxer) [in](https://github.com/libp2p/interface-peer-discovery) [loads](https://github.com/ipfs/interface-datastore) [of](https://github.com/libp2p/interface-content-routing) [other](https://github.com/libp2p/interface-connection) [places](https://github.com/libp2p/interface-transport) so let's call it something different.

How about calling it `types-ipfs` or `ipfs-types`?

`ipfs-types` is taken on npm already, @bluelovers would you consider donating the name so we can have first-class type support with regular releases done in sync with the main implementation etc?

achingbrain (approved):
LGTM

PR #3456, opened 2020-12-17
fix(types): correct factory options types

AuHau:
This extends the range of allowed `ipfsClient`'s options that is possible to pass to `ipfsClient`. Relates to https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/lib/core.js#L19


achingbrain:
I think this needs a bit more work, the type check gh action is failing?

PR #3471, opened 2021-01-06
IGNORE: try ci

Gozala:
Just trying to see if errors on CI are caused by changes in #3468


achingbrain:
You can open these sorts of PRs as drafts, CI will still run.

achingbrain:
I don't think this PR is doing what you need it to do - because of the `$RUN_SINCE` variable, CI only runs tests etc in packages that have changed in a given commit set, and packages that depend on changed packages (as assuming a test suite is deterministic, there's no need to test code that hasn't changed).

There are no changes in this PR which is why the tests don't run in CI (examples aside).

Issue #3464, opened 2020-12-29
HTTP daemon performance is low


ukstv:
- **Version**:
ipfs-http-server v0.1.4

- **Platform**:
  - Darwin feather 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64
  - Node.js v14.15.1

- **Subsystem**:
ipfs-http-daemon

#### Severity:
Medium: Performance Issues

#### Description:
JS-IPFS HTTP daemon refuses any new HTTP connection over certain threshold, which on my machine is 20. I believe, Node.js is a bit more performant than this.

As far as I understand, js-ipfs http daemon is built on top of [HAPI](https://hapi.dev), which apparently is [one of the slowest HTTP frameworks for Nodejs](https://www.techempower.com/benchmarks/#section=data-r19&hw=cl&test=fortune&l=zik0sf-1p&a=2). I do believe, this commands the inability to handle more than 20 simultaneous connections. As an experiment, I got `dag.put` HTTP endpoint implemented on top of Fastify, with same IPFS instance configuration. Handles 1000 simultaneously initiated connections just fine.

I do understand, that one could put something like HAProxy in front, and that JS-IPFS maybe has never been oriented towards performance. Though, it feels like you guys should be aware of the issue.

#### Steps to reproduce the error:
Adjust `TIMES` variable in https://gist.github.com/ukstv/25f77d94113f32c0b2d200f8f1e0c3a1 to 20, and run it against local js-ipfs instance. Works all right, reports no refused connections. If you set it back to 1000, then a huge part of the connections would be refused.

jacobheun:
We're looking into this.

May be related to https://github.com/ipfs/js-ipfs/issues/3469

achingbrain:
It's important to set a baseline for performance expectations.  In [this gist](https://gist.github.com/achingbrain/0c856d9d2ce62c33b778b1aa4e83d233) I set up an HTTP server and use a client to make requests to it, using only node core - no HTTP framework and no HTTP client abstractions.

Running on Mac OS X 10.15.6 I see the following:

```console
$ node max-concurrent-requests.js
testing 100 concurrent requests
max in flight 100
testing 111 concurrent requests
max in flight 111
testing 123 concurrent requests
max in flight 123
testing 136 concurrent requests
request 125 failed
max in flight 0
[...]

PR #3474, opened 2021-01-12
feat: allow passing a http.Agent to ipfs-http-client in node

achingbrain:
Right now no `http.Agent` is used for requests made using the http client in node, which means each request opens a new connection which can end up hitting process resource limits which means connections get dropped.

The change here sets a default `http.Agent` with a `keepAlive: true` and `maxSockets` of 6 which is consistent with [browsers](https://tools.ietf.org/html/rfc2616#section-8.1.4) and [native apps](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1407597-httpmaximumconnectionsperhost?language=objc).

The user can override the agent passed to the `ipfs-http-client` constructor to restore the previous functionality:

```js
const http = require('http')
const createClient = require('ipfs-http-client')

const client = createClient({
  url: 'http://127.0.0.1:5002',
  agent: new http.Agent({
    keepAlive: false,
    maxSockets: Infinity
  })
})
```

Refs: #3464

### DIFF
--- a/packages/ipfs-http-client/test/node/agent.js
+++ b/packages/ipfs-http-client/test/node/agent.js
@@ -1,0 +1,119 @@
+/* eslint-env mocha */
+'use strict'
+
+const { isNode } = require('ipfs-utils/src/env')
+const { expect } = require('aegir/utils/chai')
+const ipfsClient = require('../../src')
+const delay = require('delay')
+
+const PORT = 29832
+
+function startServer (handler) {
+  return new Promise((resolve) => {
+    // spin up a test http server to inspect the requests made by the library
+    const server = require('http').createServer((req, res) => {
+      req.on('data', () => {})
+      req.on('end', async () => {
+        const out = await handler(req)
+
+        res.writeHead(200)
+        res.write(JSON.stringify(out))
+        res.end()
+      })
+    })
+
+    server.listen(PORT, () => {
+      resolve({
+        close: () => server.close()
+      })
+    })
+  })
+}
+
+describe('agent', function () {
+  // do not test in browser
+  if (!isNode) {
+    return
+  }
+
+  let ipfs
+  let agent
+
+  before(() => {
+    const { Agent } = require('http')
+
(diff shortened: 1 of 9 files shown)


packages/ipfs-http-client/test/node/agent.js
```diff
@@ -0,0 +1,119 @@
...
+      })
+    })
+  })
+}
+
+describe('agent', function () {
+  // do not test in browser
+  if (!isNode) {
+    return
+  }
```
hugomrdias, line 37: this isnt needed, only the .spec.js files run in all envs

hugomrdias (changes requested):

packages/ipfs-http-client/test/node/agent.js
achingbrain: Have removed, I think it just came from the headers test.

Gozala (approved):
Looks good to me. I would however suggest making `keepAlive` and `makSockets` (or better yet `maxConnections`) as options directly instead of tying it all to the `Agent`, so it is more immune to changes in the implementation (e.g. switching to library that does not allow providing `Agent`).

achingbrain:
> I would however suggest making keepAlive and makSockets (or better yet maxConnections) as options directly instead of tying it all to the Agent

It's all a tradeoff. If we specify all agent options as part of our API, we need to ensure compatibility with whatever options are available in a given version of node or otherwise handle the incompatibility.  If we start renaming options, we break compatibility with node and users have to learn our API instead of using the one that is familiar to them.

I think it's better to just treat the Agent as an opaque datatype and pass it through to the underlying request library.  This way people can also use experimental Agent implementations should they wish to, and we are none the wiser.

> switching to library that does not allow providing Agent

I'm not sure we'd ever do this, there would be no way to control connection pooling in node, since it's all done by Agents.  Unless I suppose the library implemented it's own HTTP client over raw sockets but the chances of this seem small.

PR #3479, opened 2021-01-14
Re-enabling orbit-db and ipfs-log tests

### LINKED ISSUE #3202: Running OrbitDB tests in js-ipfs CI
This is an idea that's been discussed between me and a few of the `js-ipfs` devs, as well as @autonome.

**Context**
[OrbitDB](https://orbitdb.org) is a popular library that builds on top of IPFS. It is a distributed database system that uses the IPFS DAG to store its operations log in a data structure called a CRDT. It interfaces with IPFS via either an in-process js-ipfs node or via the IPFS HTTP API via js-ipfs-http-client. Right now, the OrbitDB devs must pay "catch up" when a new js-ipfs release comes out, and as such we are not able to immediately use the latest and greatest features of IPFS.

**Proposal**
What I'd like to propose is to write a GitHub action that runs the battery of [OrbitDB tests](https://github.com/orbitdb/orbit-db/tree/master/test) on the pull requests this repository. Ideally, it would only respond to changes in code to either `packages/ipfs` and `packages/ipfs-http-client`.  

I don't expect that these checks would necessarily block any merges to code, since there may be changes required in OrbitDB's code itself, but it will be extremely useful to learn about such things far sooner, so that the OrbitDB devs can move much faster. In a perfect world, we could release OrbitDB versions in lock-step with IPFS versions.

aphelionz:
The goal of this PR is to re-enable the automated tests that run against [orbit-db](https://github.com/orbitdb/orbit-db) and [ipfs-log](https://github.com/orbitdb/orbit-db) when Travis CI wants to run them.

I have created [this branch of `orbit-db-test-utils`](https://github.com/orbitdb/orbit-db-test-utils/compare/bleeding-edge) and published it to npm as @next. If all goes according to plan then it will automatically pull in the latest versions of `go-ipfs`, `js-ipfs`, `ipfs-ctl`, and `ipfs-http-client`. Then it should run the OrbitDB tests.

Opening this now to get feedback, since I can't get the `--deps` argument to work locally :sweat_smile: but I'm nearly certain this approach should work. Let me know what I'm missing and if this approach will get us back to testing OrbitDB automatically

Thank you!

Fixes #3202

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -347,14 +347,14 @@ jobs:
       if: branch = master AND type = push AND fork = false
       name: external - orbit-db
       script:
-        - npm run test:external -- -- -- https://github.com/orbitdb/orbit-db.git --deps=ipfs@next
+        - npm run test:external -- -- -- https://github.com/orbitdb/orbit-db.git --deps=orbit-db-test-utils@next
 
     - stage: test-external
       # only run on changes to master
       if: branch = master AND type = push AND fork = false
       name: external - ipfs-log
       script:
-        - npm run test:external -- -- -- https://github.com/orbitdb/ipfs-log.git --deps=ipfs@next
+        - npm run test:external -- -- -- https://github.com/orbitdb/ipfs-log.git --deps=orbit-db-test-utils@next
 
     - stage: test-external
       # only run on changes to master


achingbrain:
I'd really like to get the automated testing running again so thanks for opening this.

Overriding transitive dependencies to test doesn't seem like the right way to do it and it's something we've [explicitly not done in the past](https://github.com/ipfs/aegir/pull/455) since we're not really guaranteeing that we won't break a library just by changing the `ipfs` version.

What if `orbit-db-test-utils` declared `ipfs` as a peer dependency, then `orbit-db` and `ipfs-log` would need to declare `ipfs` as a dev-dep, the version of which could then be overridden in the tests.  

It gives `orbit-db` and `ipfs-log` control over which version of `ipfs` they are tested against and lets us say useful things about which version of `ipfs` works with which dependant modules.

We follow a similar pattern with the [ipfs interop tests](https://github.com/ipfs/interop/blob/master/package.json#L88-L92).

aphelionz:
I may just be being dense here (a distinct possibility) but I'm having trouble seeing the difference between for example:

1. `orbit-db` pulling in `orbit-db-test-utils@next`, which will includes `ipfs@next` as per the PR, and
2. `orbit-db` pulling in `ipfs@next`

It would seem they both achieve the same thing, and it's less of a difference in both of our codebases. Can you help me understand a case where this would break?

Note the `orbit-db-test-utils@next` branch isn't likely to change very much if at all, and will always point to the `next` tag of the IPFS libs

achingbrain:
A green CI build tells me IPFS works with module A at version X, that's simple to reason about and when someone reports a problem, we can say, please use module A at version X - the tests pass you should be good.

If module B depends on module C that depends on IPFS, then I can test module C at version Y with IPFS and say that it works, but I can't say for certain that IPFS works with module B at any version because module C has it's own release schedule so unless module B locks all of it's dependencies and the dependencies of it's dependencies to specific versions, installing module B at a given version does not guarantee which version of module C you'll get and as such, you cannot guarantee you'll get a version that we've tested with a specific version of IPFS.

> Note the orbit-db-test-utils@next branch isn't likely to change very much if at all, and will always point to the next tag of the IPFS libs

This is true until it's not :stuck_out_tongue_winking_eye: - CI cannot guarantee this and so we can't rely on it.

aphelionz:
Ok, I've started this process here: https://github.com/orbitdb/orbit-db-test-utils/pull/31

Fingers crossed :crossed_fingers:

achingbrain:
@aphelionz I see that PR was merged - it doesn't look like there's been a release of the orbit-db-test-utils module since then, does something need to happen there to get this moving?

aphelionz:
Thanks for the nudge! I'll carry forth the nudge on my end.

lidel:
@aphelionz  any update on this?

lidel:
Closing due to inactivity and main branch moving away a bit.
 Feel free to fill a new one

PR #3481, opened 2021-01-14
chore: Replace transfer arrays with sets in ipfs-message-port-*

### LINKED ISSUE #3480: Arraybuffer at index 0 is already detached
- **Version**:
ipfs-message-port-server v0.4.3
js-ipfs v0.52.3

- **Platform**:
Darwin Kernel Version 20.1.0
Firefox 84.0.2
Opera 73

- **Subsystem**:
https://github.com/ipfs/js-ipfs/blob/6b207d77cd66bc5326f260221e53914192807861/packages/ipfs-message-port-server/src/server.js#L208

#### Severity:
Critical

#### Description:
Screenshot 2021-01-14 at 19 18 24

We've implemented a filesystem on top of IPFS, which is a mix of DAG PB and DAG CBOR nodes. IPFS is loaded using `ipfs-message-port-client` through an iframe, which then loads `ipfs-message-port-server` and `js-ipfs`.

I make a bunch of concurrent changes to the file system, resulting in several `dag.put` calls. And it's somewhere here I suspect the error originates from.

Hard to reproduce, feels like a concurrency error 🤔 

You can try it at https://nightly.diffuse.sh/
→ Login with Fission
→ Add some content (add music, mark some stuff as favourite, make a playlist)
→ Export
→ Import
→ ... wait a minute or so ...

#### Moving forward
**Ok so, I'm working on a PR to fix this, but I don't know how to proceed exactly.** How does the ArrayBuffer become detached exactly? Does it become detached if you put in the transfer list of a MessagePort? Should we make a copy of the ArrayBuffer, or should we ignore it? 🤔

icidasset:
1. ~~Fixes #3480~~
As described in the issue, I wasn't sure how to tackle this, but I decided to go the "remove already detached buffers" route. This seems to work, I tested this with my app (see issue for more info). Thoughts?

2. Instead of passing of an Array to `postMessage`, we can just pass the Set iterable.
The only downside of this seems to be that Typescript doesn't like this 🤔 
Typescript expects an Array, not an iterable like it should.

3. ~~I just discovered there's a thing called `SharedArrayBuffer`, maybe we should use that instead? 🤔~~
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer

### DIFF
--- a/packages/ipfs-message-port-protocol/src/core.js
+++ b/packages/ipfs-message-port-protocol/src/core.js
@@ -2,6 +2,7 @@
 
 /* eslint-env browser */
 const { encodeError, decodeError } = require('./error')
+const { ensureUniqueBuffers } = require('./buffer')
 
 /**
  * @template T
@@ -100,8 +101,6 @@ exports.decodeIterable = decodeIterable
  */
 const encodeIterable = (iterable, encode, transfer) => {
   const { port1: port, port2: remote } = new MessageChannel()
-  /** @type {Transferable[]} */
-  const itemTransfer = []
   /** @type {Iterator<I>|AsyncIterator<I>} */
   const iterator = toIterator(iterable)
 
@@ -114,14 +113,18 @@ const encodeIterable = (iterable, encode, transfer) => {
             port.postMessage({ type: 'next', done: true })
             port.close()
           } else {
-            itemTransfer.length = 0
+            /** @type {Transferable[]} */
+            const itemTransfer = []
+            const encodedValue = encode(value, itemTransfer)
+
             port.postMessage(
               {
                 type: 'next',
                 done: false,
-                value: encode(value, itemTransfer)
+                value: encodedValue
               },
-              itemTransfer
+              // @ts-expect-error - Typescript expects the transfer list to be an Array, but it can actually be any iterable.
+              ensureUniqueBuffers(itemTransfer)
             )
           }
         } catch (error) {
@@ -197,7 +200,8 @@ const decodeCallback = ({ port }) => {
    * @returns {void}
    */
   const callback = (args, transfer = []) => {
-    port.postMessage(args, [...new Set(transfer)])
+    // @ts-expect-error - Typescript expects the transfer list to be an Array, but it can actually be any iterable.
+    port.postMessage(args, new Set(transfer))
   }
 
   return callback
--- a/packages/ipfs-message-port-server/src/server.js
+++ b/packages/ipfs-message-port-server/src/server.js
@@ -3,6 +3,7 @@
 /* eslint-env browser */
 
 const { encodeError } = require('ipfs-message-port-protocol/src/error')
+const { ensureUniqueBuffers } = require('ipfs-message-port-protocol/src/buffer')
 
 /**
  * @typedef {import('ipfs-message-port-protocol/src/data').EncodedError} EncodedError
@@ -212,11 +213,14 @@ exports.Server = class Server {
     if (!query.signal.aborted) {
       try {
         const value = await query.result
-        const transfer = [...new Set(value.transfer || [])]
+        const transfer = ensureUniqueBuffers(value.transfer || [])
+
+        // Don't need the transfer value in the result
         delete value.transfer
 
         port.postMessage(
           { type: 'result', id, result: { ok: true, value } },
+          // @ts-expect-error - Typescript expects the transfer list to be an Array, but it can actually be any iterable.
           transfer
         )
       } catch (error) {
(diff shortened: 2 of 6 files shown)


Gozala:
Thanks @icidasset for tackling the problem. I'll respond to some of your comments inline below:



> * Fixes #3480
>   As described in the issue, I wasn't sure how to tackle this, but I decided to go the "remove already detached buffers" route. This seems to work, I tested this with my app (see issue for more info). Thoughts?
> 

I would really like this fixed, but I can't shake the feeling that proposed changes

1. Work around the issue without actually fixing it.
2. Actual issue is still there it's just fission happens to not detect it yet.
3. Dropped buffer would lead on the receiver end getting `Uint8Array` copy with size `0` as opposed to expected bytes. And that `Uint8Array` is most likely was not intended to be `0` size.

I really think we should at least figure out what causes this issue. I'll provide some specific suggestions on this in the #3480  

> * Instead of passing of an Array to `postMessage`, we can just pass the Set iterable.
>   The only downside of this seems to be that Typescript doesn't like this 🤔
>   Typescript expects an Array, not an iterable like it should.
> 

`@ts-ignore` is unfortunate, but at this point we're used to typescript standard lib not been perfect.

>
> * I just discovered there's a thing called `SharedArrayBuffer`, maybe we should use that instead? 🤔
>   https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer

`SharedArrayBuffer`s had been disabled in response to [Spectre vulnerability](https://en.wikipedia.org/wiki/Spectre_(security_vulnerability)), I think chrome enabled them back but it's unclear when others will do it.
But even if we had them available, they aren't necessarily a good fit here, they create a shared memory and require synchronized access. Transferring buffers seems like a lot better fit when one thread writes a chunk of memory and then passes it onto the other one to read/write into it.

packages/ipfs-message-port-server/src/server.js
```diff
@@ -212,11 +213,14 @@ exports.Server = class Server {
     if (!query.signal.aborted) {
       try {
         const value = await query.result
-        const transfer = [...new Set(value.transfer || [])]
+        const transfer = ensureNonDetachedUniqueBuffers(value.transfer || [])
```
Gozala, line 216: I'll provide more input in #3480 but I do think this solves actual problem, it just pushes it further down the stack.

packages/ipfs-message-port-server/src/server.js
```diff
@@ -212,11 +213,14 @@ exports.Server = class Server {
...
+
+        // Don't need the transfer value in the result
         delete value.transfer
 
         port.postMessage(
           { type: 'result', id, result: { ok: true, value } },
+          // @ts-ignore - Typescript expects the transfer list to be an Array, but it can actually be any iterable.
```
Gozala, line 223: I think it's better to use `@ts-expect-error` so that when ts fixes it we can remove those comments.

```suggestion
          // @ts-expect-error - Typescript expects the transfer list to be an Array, but it can actually be any iterable.
```

packages/ipfs-message-port-protocol/src/core.js
```diff
@@ -197,7 +200,8 @@ const decodeCallback = ({ port }) => {
    * @returns {void}
    */
   const callback = (args, transfer = []) => {
-    port.postMessage(args, [...new Set(transfer)])
+    // @ts-ignore - Typescript expects the transfer list to be an Array, but it can actually be any iterable.
```
Gozala, line 203:
```suggestion
    // @ts-expect-error - Typescript expects the transfer list to be an Array, but it can actually be any iterable.
```

Gozala (changes requested):
@icidasset I'm happy to take changes that switch from arrays to sets, however we need to figure out proper fix for #3480 as dropping such array buffers do not really fix the problem they'll just push it further down the stack.

icidasset:
@Gozala While we discuss the mysterious buffer transfer issues in #3480, I repurposed this PR to just replace the arrays with sets. I also applied the changes you asked for.

icidasset:
@Gozala Do we still want this tiny PR that sends a `Set` over postMessage instead of `[...set]`?

Gozala:
> @Gozala Do we still want this tiny PR that sends a `Set` over postMessage instead of `[...set]`?

@icidasset thanks for starting this effort, I have taken on your work to create #3573 which removes transferable arrays everywhere and just makes them sets. I would push changes here, but github won't allow because PR is from your fork. So I'll close this in favor of the other one.

PR #3488, opened 2021-01-18
Enable text streaming support for React Native (iOS and Android) on pubsub.subscribe

acostalima:
React Native's Networking API accepts several arguments, two of which that:
- Drive the [native response type](https://github.com/facebook/react-native/blob/v0.63.4/Libraries/Network/NativeNetworkingIOS.js#L23) (one of `blob`, `base64` or `text`).
- Drive whether we want to subscribe to [incremental data events](https://github.com/facebook/react-native/blob/v0.63.4/Libraries/Network/NativeNetworkingIOS.js#L24).

The reception of incremental data is only [supported](https://github.com/facebook/react-native/blob/v0.63.4/Libraries/Network/RCTNetworking.mm#L544-L547) when the native response type is set to `'text'`. As such, [@react-native-community/fetch](https://github.com/react-native-community/fetch), the implementation of the fetch API that provides text streaming support for React Native, accepts a custom option to drive the arguments mentioned above.

In the [demo app](https://github.com/ipfs-shipyard/react-native-ipfs-demo), this option is currently included with a [patch](https://github.com/ipfs-shipyard/react-native-ipfs-demo/blob/master/patches/ipfs/ipfs-http-client%2B48.2.0.patch).

### DIFF
--- a/packages/ipfs-http-client/src/pubsub/subscribe.js
+++ b/packages/ipfs-http-client/src/pubsub/subscribe.js
@@ -34,7 +34,11 @@ module.exports = configure((api, options) => {
           arg: topic,
           ...options
         }),
-        headers: options.headers
+        headers: options.headers,
+        // Enables text streaming support for React Native when using @react-native-community/fetch
+        reactNative: {
+          textStreaming: true
+        }
       })
         .catch((err) => {
           // Initial subscribe fail, ensure we clean up


achingbrain:
Pubsub subscriptions in the HTTP client are long-lived requests that for all intents and purposes may never end.

If I receive an arbitrary number of pubsub messages, and we get incremental data updates to process those messages, can responseText be reset or will it grow over time and eventually cause the device to run out of memory/the process or app to be killed?

acostalima:
> Pubsub subscriptions in the HTTP client are long-lived requests that for all intents and purposes may never end.
> 
> If I receive an arbitrary number of pubsub messages, and we get incremental data updates to process those messages, can responseText be reset or will it grow over time and eventually cause the device to run out of memory/the process or app to be killed?

In case of the fetch implementation that ships with React Native I'd say that would be the case, yes, because it is implemented on top of `XMLHttpRequest` and response strings are buffered in memory until the request ends. The alternative implementation I mentioned does not use `XMLHttpRequest`. Instead, it's built directly on top of React Native's Networking API which, AFAIK, does not hold on into previous responses.

BigLep:
@acostalima : is this still being worked on?  If not, we'll plan on closing this PR.  Thanks!

acostalima:
Hi @BigLep, on a second thought, I'll move this piece of logic to https://github.com/ipfs/js-ipfs/pull/3517 instead, which makes more sense in the scope of the work done over there as a whole. Feel free to close this. Thanks!

PR #3495, opened 2021-01-21
fix: issue with isolateModules flag

### LINKED ISSUE #3494: ipfs-core-types causes problems when isolateModules flag is set
See following issue for the context https://github.com/ipfs-shipyard/ipfs-webui/pull/1655#issuecomment-763846124

Gozala:
Fixes #3494

It appears that `react-scripts` swaps `isolateModules` which [causes problems in webui](https://github.com/ipfs-shipyard/ipfs-webui/pull/1655#issuecomment-763846124) because according to TS:

> Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.  TS1205

Changes here address that problem as follows:

- `export type {...}` is used as per TS error.
- `isolateModules` flag is set to `true` so that such issues can be caught locally.
- Setting flags seems to have triggered problems in `ipfs-client`, so I had to add some type annotations to fix that.
- Added `references` in `ipfs-client/tsconfig.json`, otherwise it does not pick up changes in the dependencies unless they are manually rebuild.

### DIFF
--- a/packages/ipfs-client/package.json
+++ b/packages/ipfs-client/package.json
@@ -22,6 +22,9 @@
       ]
     }
   },
+  "eslintConfig": {
+    "extends": "ipfs"
+  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/ipfs/js-ipfs.git"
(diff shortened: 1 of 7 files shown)


hugomrdias (approved):

achingbrain (approved):
LGTM

PR #3497, opened 2021-01-21
feat: Add pin.add to ipfs-message-port

icidasset:
Adds `pin.add` to `ipfs-message-port-client` and `ipfs-message-port-server`.

### DIFF
--- a/packages/ipfs-message-port-client/package.json
+++ b/packages/ipfs-message-port-client/package.json
@@ -42,14 +42,14 @@
   "dependencies": {
     "browser-readablestream-to-it": "^1.0.1",
     "ipfs-core-types": "^0.3.0",
-    "ipfs-message-port-protocol": "^0.6.0"
+    "ipfs-message-port-protocol": "file:../ipfs-message-port-protocol"
   },
   "devDependencies": {
     "aegir": "^30.3.0",
     "interface-ipfs-core": "^0.144.1",
     "ipfs": "^0.54.2",
     "ipfs-core": "^0.5.2",
-    "ipfs-message-port-server": "^0.6.1",
+    "ipfs-message-port-server": "file:../ipfs-message-port-server",
     "rimraf": "^3.0.2"
   },
   "engines": {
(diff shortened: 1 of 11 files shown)


achingbrain:
Looks good as a first pass, thanks for submitting this.

Can you please add the `ipfs.pin.add` interface test suite to [/packages/ipfs-message-port-client/test/interface-message-port-client.js](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-message-port-client/test/interface-message-port-client.js) to ensure this gets tested in CI etc.

achingbrain:
By way of an example, see how the pinning tests are [run for testing the http client against go-IPFS](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/test/interface-http-go.js#L580-L587) - nb. you'll need to add skips for the parts of the pinning API that haven't been implemented in this PR.

icidasset:
@achingbrain Done. I did have to skip all the tests because `pin.ls` is not implemented. Might look into that later if I have some time and if it's not too difficult.

Gozala:
> @achingbrain Done. I did have to skip all the tests because `pin.ls` is not implemented. Might look into that later if I have some time and if it's not too difficult.

@icidasset This looks good, but I think we would need to add `pin.ls` (even if just limited version and only for testing) otherwise things may not work or break and we won't even know that.

P.S.: Bunch of APIs here were actually implemented just to enable testing  not because they were intentionally included.

icidasset:
@Gozala I've added `pin.ls` and `pin.rmAll`. Most of the tests are passing, but there's a few that are still failing that I can't seem to get fixed. Could you take a look?

There's one metadata issue with `pin.ls` that's weird, I should be passing the metadata along from with result. And another issue with `normaliseInput` and `pin.rmAll`, for some reason it's not able to decode some CIDs.

Gozala:
> @Gozala I've added `pin.ls` and `pin.rmAll`. Most of the tests are passing, but there's a few that are still failing that I can't seem to get fixed. Could you take a look?
> 
> There's one metadata issue with `pin.ls` that's weird, I should be passing the metadata along from with result. And another issue with `normaliseInput` and `pin.rmAll`, for some reason it's not able to decode some CIDs.

Thanks for championing this effort! I've resolved remaining issues in #3575

Issue #3502, opened 2021-01-23
dht.findProvs: TypeError: key.toBaseEncodedString is not a function


tymmesyde:
- **Version**: 0.53.2

- **Platform**: Windows 10

- **Subsystem**:

#### Description:
Unable to get providers from dht with a string cid.

#### Steps to reproduce the error:
Code:
```js
const providers = await all(this.ipfs.dht.findProvs(`Qm...`));
console.log(providers);
```
Error:
`TypeError: key.toBaseEncodedString is not a function`

achingbrain:
The `cid` argument to `ipfs.dht.findProvs` should be of the [CID](https://www.npmjs.com/package/cids) type and not a string: https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/DHT.md#ipfsdhtfindprovscid-options

jacobheun:
Closing as the resolution has been suggested.

TheDiscordian:
Requesting re-open, for 2 reasons.

1. The linked doc uses a string in the example (`ipfs.dht.findProvs('QmdPAhQRxrDKqkGPvQzBvjYe3kU8kiEEAd2J6ETEamKAD9')`)
2. Not using a string produces the exact same error.

```js
const IPFS = require('ipfs-core');
const CID = require('cids');
 
async function run() {
	const node = await IPFS.create({
		repo: 'ok' + Math.random(), // random so we get a new peerid every time, useful for testing
		libp2p: {
			config: {
				dht: {
					enabled: true
				}
			}
		}
	});

	const { cid } = await node.add('Hello world');
	
	console.log(cid);
	// CID(QmNRCQWfgze6AbBCaT1rkrkV5tJ2aP4oTNPb5JZcXYywve)
	console.log(CID.isCID(cid));
	// true

	const providers = node.dht.findProvs(cid);
	
	for await (const provider of providers) {
		console.log(provider.id.toString());
	}
}
 
run();
```

```
[user@TheDesktop DHT]$ node index.js
generating 2048-bit (rsa only) RSA keypair...
to get started, enter:

        jsipfs cat /ipfs/QmfGBRT6BbWJd7yUc2uYdaUZJBbnEFvTqehPFoSMQ6wgdr/readme

Swarm listening on /ip4/127.0.0.1/tcp/4002/p2p/QmVG4ZKiXgw1cU8VJ4fAj5P5aF45RAsE2tnskmghXNs2J7
[...]

PR #3503, opened 2021-01-23
fix(core): convert cid to CID type

tymmesyde:
see the issue for details

Refs #3502


achingbrain:
The `cid` argument to `ipfs.dht.findProvs` should be of the [CID](https://www.npmjs.com/package/cids) type and not a string: https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/DHT.md#ipfsdhtfindprovscid-options - if anything we should be throwing an error here demanding the correct argument type.

Issue #3511, opened 2021-01-27
Making js-ipfs supports webpack5


FledgeXu:
- **Version**:
"ipfs": "0.53.2",
- **Platform**:
Darwin FledgeXus-MBP 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64

#### Severity:
Hight: Making JS-ipfs supports webpack5

#### Description:
[Webpack5 removes Node.js core modules polyfills](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-nodejs-polyfills-removed) and it makes that js-ipfs can not just work in browser. How will the IPFS support webpack5?

achingbrain:
We've removed node globals from all the repos in the ipfs/ipld/multiformats orgs.  There are still a few dependencies of IPFS that do things like reference `process` or `Buffer` without `require`ing them first, those will take longer to fix.

If you would like to help you can PR those repos to remove the use of node globals.  I did [cbor](https://github.com/hildjj/node-cbor) earlier today.  I think some of the level repos need PRs, as does node-forge and some of the crypto libraries.

Otherwise the only solution is to stick to webpack 4 for now, or use a different bundler.

Gozala:
We actually have one example that uses webpack5,
https://github.com/ipfs/js-ipfs/blob/master/examples/browser-service-worker/webpack.config.js

PR #3512, opened 2021-01-28
fix: updates webpack example to use v5

achingbrain:
Takes the config from the service worker example and uses it to patch up the webpack example to use v5

fixes: #3511


Hrushi20:
How long will this commit take? I am not able to use ipfs-http-client with cra due to webpack v5

PR #3516, opened 2021-01-30
Fixed http-client dht/put

rob-deutsch:
The http-client dht/put seems to refuse any possible key I can dream up to send it. It looks like its because its not operating on the same principle as dht/get.

### DIFF
--- a/packages/ipfs-core/src/components/dht.js
+++ b/packages/ipfs-core/src/components/dht.js
@@ -33,7 +33,7 @@ module.exports = ({ network, repo }) => {
      * form, this will write that value to the DHT with
      * that key.
      *
-     * @param {Uint8Array} key
+     * @param {Uint8Array|string} key
      * @param {Uint8Array} value
      * @param {AbortOptions} [options]
      * @returns {AsyncIterable<QueryEvent>}
--- a/packages/ipfs-http-client/src/dht/put.js
+++ b/packages/ipfs-http-client/src/dht/put.js
@@ -5,6 +5,7 @@ const multiaddr = require('multiaddr')
 const toCamel = require('../lib/object-to-camel')
 const configure = require('../lib/configure')
 const toUrlSearchParams = require('../lib/to-url-search-params')
+const uint8ArrayToString = require('uint8arrays/to-string')
 const multipartRequest = require('../lib/multipart-request')
 const { anySignal } = require('any-signal')
 const AbortController = require('native-abort-controller')
@@ -21,7 +22,7 @@ module.exports = configure(api => {
       timeout: options.timeout,
       signal,
       searchParams: toUrlSearchParams({
-        arg: key,
+        arg: key instanceof Uint8Array ? uint8ArrayToString(key) : key,
         ...options
       }),
       ...(


achingbrain:
Thanks for submitting this, sorry for the delay in reviewing.

The problem this addresses is that passing a [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) as a value to a [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) results in the `Uint8Array` being stringified and not encoded as a string.

E.g.

```js
Uint8Array.from([0, 1, 2]).toString()
// "0,1,2"
```

So:

```js
new URLSearchParams({ data: Uint8Array.from([0, 1, 2]) }).toString()
// "data=0%2C1%2C2"
```

This is then interpreted as a byte array encoded as an ascii string by the HTTP API server:

```js
decodeURIComponent('0%2C1%2C2')
// "0,1,2"
Uint8Array.from("0,1,2")
// Uint8Array(5) [0, 0, 1, 0, 2]
```

This PR will fix the issue, but only where byte values can be represented as a string by JavaScript - note that for byte values 128-254 this is not the case:

```js
const bytes = Uint8Array.from([254])
// Uint8Array [254]
const str = new TextDecoder().decode(bytes)
// "�"
new TextEncoder().encode(str)
// Uint8Array(3) [239, 191, 189]
```

Ascii/windows-1252 encoding doesn't quite do what you think it'd do either:

```js
const bytes = Uint8Array.from([254])
// Uint8Array [254]
const str = new TextDecoder('ascii').decode(bytes)
// "þ"
new TextEncoder().encode(str)
// Uint8Array(3) [195, 190]
```

`encodeURIComponent` doesn't fare much better:

```js
encodeURIComponent("þ")
// "%C3%BE"
```

This is what would end up being sent and I don't think go-IPFS will interpret `%C3%BE` correctly - certainly if you send it as [pubsub message data on the query string](https://github.com/ipfs/go-ipfs/issues/7939) it's sent to subscribers as `w74=` which decoded is `Uint8Array(2) [195, 190]` and not `Uint8Array [254]`.

For it to be interpreted as `Uint8Array [254]` we'd need to send `%FE` which we can't do using built-in JS functions.  We'd have to write our own string encoder/decoder which then means everyone using the HTTP API from JS would need to do the same thing, or use this client library.

Really for this to be properly fixed we need to be able to [encode bytes sent in the query string](https://github.com/multiformats/multibase).

This is simple to do from the JS side but will need some coordination from go-IPFS.

achingbrain:
Writeup for encoding bytes sent in the query string in go-IPFS - https://github.com/ipfs/go-ipfs/issues/7958

lidel:
Needs https://github.com/ipfs/go-ipfs/issues/7958 first.

achingbrain:

`ipfs-http-client` should be swapped out for `kubo-rpc-client` - it's a drop in replacement and where future bug fixes and features will land.

If this is still a problem, a new PR should be opened against `kubo-rpc-client`.

PR #3528, opened 2021-02-04
fix: update to new aegir

hugomrdias:
- update aegir
- fix some linting
- enable multiple test runs at the same time (all servers now try to find free ports)

todo:
- [x] ipfs-message-port-client

needs:
https://github.com/ipfs/interop/pull/223


.github/workflows/bundlesize.yml
```diff
@@ -7,33 +7,20 @@ on:
...
-        uses: actions/checkout@v2
-        with:
-          path: node_modules
-          key: ${{ matrix.node-version }}-${{ runner.OS }}-build-${{ hashFiles('package.json') }}
-      - name: Install dependencies
-        run: npm install
+      - run: npm install
       - name: Bundlesize ${{ matrix.project }}
-        uses: ipfs/aegir/actions/bundle-size@v29.0.0
+        uses: ipfs/aegir/actions/bundle-size@master
```
achingbrain, line 23: This should go back to being a version number, no?

packages/interface-ipfs-core/src/ping/ping.js
```diff
@@ -49,11 +49,16 @@ module.exports = (common, options) => {
       expect(pongs.length).to.equal(count)
     })
 
-    it('should fail when pinging a peer that is not available', () => {
+    it('should fail when pinging a peer that is not available', async () => {
       const notAvailablePeerId = 'QmUmaEnH1uMmvckMZbh3yShaasvELPW4ZLPWnB4entMTEn'
       const count = 2
 
-      return expect(all(ipfsA.ping(notAvailablePeerId, { count }))).to.eventually.be.rejected()
```
achingbrain, line 56: Why was this necessary?

packages/ipfs-http-client/package.json
```diff
@@ -45,15 +45,14 @@
     "test:chrome": "aegir test -t browser -t webworker -- --browsers ChromeHeadless",
     "test:firefox": "aegir test -t browser -t webworker -- --browsers FirefoxHeadless",
     "lint": "aegir lint",
-    "prepare": "aegir build --no-bundle",
```
achingbrain, line 48: Why was this removed?

packages/ipfs-message-port-client/package.json
```diff
@@ -31,24 +31,23 @@
...
-    "test:interface:message-port-client": "npm run prepare && aegir test -t browser -f ./test/interface-message-port-client.js",
+    "test:interface:message-port-client": "aegir test -t browser --bail -f ./test/interface-message-port-client.js",
     "lint": "aegir lint",
-    "prepare": "aegir build --no-bundle && aegir build -- --config ./test/util/webpack.config.js",
-    "prepublishOnly": "aegir build && aegir build -- --config ./test/util/webpack.config.js",
+    "prepublishOnly": "aegir build",
     "coverage": "npx nyc -r html npm run test:node -- --bail",
     "clean": "rimraf ./dist",
-    "dep-check": "aegir dep-check -i ipfs-core -i rimraf -i ipfs-core-types"
+    "dep-check": "aegir dep-check -i ipfs-core -i rimraf -i ipfs-core-types -i esbuild"
```
achingbrain, line 39: `esbuild` isn't a dep here?

achingbrain (changes requested):

packages/interface-ipfs-core/src/ping/ping.js
hugomrdias: i was trying to catch an uncaught rejection that happens sometimes, because the new runner fails when these happen now.
this seems to help a bit but 100%

packages/ipfs-message-port-client/package.json
hugomrdias: dep-check doesnt look into .aegir.js

achingbrain:
The [fs](https://www.npmjs.com/package/fs) module prints out:

```
I'm `fs` modules
```

in the browser console, needs to be overridden to `false` or something.

hugomrdias:
> The [fs](https://www.npmjs.com/package/fs) module prints out:
> 
> ```
> I'm `fs` modules
> ```
> 
> in the browser console, needs to be overridden to `false` or something.

https://github.com/ipfs-shipyard/js-mock-ipfs-pinning-service > https://github.com/isa-group/oas-tools/blob/master/package.json#L26 > https://github.com/npm/security-holder > open node_modules/fs/index.js 🤷🏻‍♂️

packages/ipfs-http-client/package.json
hugomrdias: we dont need it here, the prepare for types is just to help with PR with gh url and mono-repo doesnt allow for gh url dep install. Plus it slows down everything

achingbrain: > mono-repo doesnt allow for gh url dep install

Good news!  Now you can do:

```console
$ npm install https://gitpkg.now.sh/ipfs/js-ipfs/packages/ipfs-core?some-branch-identifier
```

https://gitpkg.now.sh

achingbrain (approved):

PR #3534, opened 2021-02-05
fix: use correct version IPFS version number

vmx:
The version number that `ipfs.version()` outputs should be the one
of the release.

Prior to this change it was the version of the `ipfs-core` package.

This change is powered by the hackweek :)

### DIFF
--- a/packages/ipfs-core/src/components/version.js
+++ b/packages/ipfs-core/src/components/version.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const pkg = require('../../package.json')
+const pkg = require('../../../ipfs/package.json')
 const withTimeoutOption = require('ipfs-core-utils/src/with-timeout-option')
 
 /**


packages/ipfs-core/src/components/version.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const pkg = require('../../package.json')
+const pkg = require('../../../ipfs/package.json')
```
achingbrain, line 3: This will fail at runtime as it traverses outside of the module folder.  E.g. if you only have `ipfs-core` installed (e.g. your app is running an in-process IPFS node and you don't need a CLI/HTTP API server/daemon, etc).

achingbrain (changes requested):

packages/ipfs-core/src/components/version.js
achingbrain: A better solution may be to have the build write a version file out at build time with the version number, git hash, etc, then we can try to load it here.  It would also mean that the whole `package.json` doesn't get included in the browser bundle which would be a nice side benefit.

lidel:
Seems this needs more robust solution, let's continue in https://github.com/ipfs/js-ipfs/issues/3622

Issue #2796, opened 2020-02-26
Web UI hangs indefinitely when opening it in a browser after starting the js ipfs daemon


pcowgill:
After using `js-ipfs` to start a daemon, it says the web UI will be available here `http://127.0.0.1:5002/webui`, but when you go to that page in a browser it never finishes loading.

pcowgill:
I considered using IPFS Desktop as an alternative, but that's an upcoming feature for `js-ipfs` https://github.com/ipfs-shipyard/ipfs-desktop/issues/601

In the meantime, is `cat`ing files from the terminal my best bet?

achingbrain:
Does this still happen for you? I can't replicate it.

pcowgill:
@achingbrain Yep. Trying it in Chrome with the latest version of the `js-ipfs` master branch.

```
Initializing IPFS daemon...
js-ipfs version: 0.41.2
System version: x64/darwin
Node.js version: 12.14.1
Swarm listening on /ip4/127.0.0.1/tcp/4002/p2p/QmRGuGpzcEoQRFFJsK8Eh7uLeKuFM9C2xFv3R3cMmVzVQi
Swarm listening on /ip4/192.168.7.42/tcp/4002/p2p/QmRGuGpzcEoQRFFJsK8Eh7uLeKuFM9C2xFv3R3cMmVzVQi
Swarm listening on /ip4/169.254.98.207/tcp/4002/p2p/QmRGuGpzcEoQRFFJsK8Eh7uLeKuFM9C2xFv3R3cMmVzVQi
Swarm listening on /ip4/127.0.0.1/tcp/4003/ws/p2p/QmRGuGpzcEoQRFFJsK8Eh7uLeKuFM9C2xFv3R3cMmVzVQi
API listening on /ip4/127.0.0.1/tcp/5002/http
Gateway (read only) listening on /ip4/127.0.0.1/tcp/9090/http
Web UI available at http://127.0.0.1:5002/webui
Daemon is ready
```


<img width="354" alt="Screenshot 2020-03-11 11 02 05" src="https://user-images.githubusercontent.com/2731712/76437626-d6944f00-6387-11ea-9372-cca5331312b2.png">

pcowgill:
@hugomrdias Can you reproduce this?

hugomrdias:
no works here

pcowgill:
Maybe it's a Mac issue? Does @alanshaw have a Mac to try it on?

[...]

PR #3545, opened 2021-02-09
fix: fail to open the webui page in the browser

LehaoLin:
I met the same issue as #2796, then I added this piece of code and the problem was solved.
Environment:
macOS 10.15.3
js-ipfs version: 0.4.2
System version: x64/darwin
Node.js version: 15.8.0

### DIFF
--- a/packages/ipfs-core/src/runtime/config-nodejs.js
+++ b/packages/ipfs-core/src/runtime/config-nodejs.js
@@ -52,5 +52,48 @@ module.exports = () => ({
   },
   Routing: {
     Type: 'none'
+  },
+  "Mounts": {
+    "IPFS": "/ipfs",
+    "IPNS": "/ipns",
+    "FuseAllowOther": false
+  },
+  "Gateway": {
+    "HTTPHeaders": {
+      "Access-Control-Allow-Headers": [
+        "X-Requested-With",
+        "Range",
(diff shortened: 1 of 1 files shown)


achingbrain:
Thanks for opening this PR.  Most of this config is not used by js-IPFS so it's not relevant, it also opens up the HTTP API to any origin when running the HTTP API client in the browser so can't be merged as-is.

Which part do you think solved the problem?

LehaoLin:
@achingbrain 
I think the `Mounts` part and the `Gateway` part mainly solved the problem, which are missed in the `.jsipfs/config`.
I can't open the webui without these two parts.
After I add them, I can open it successfully.

achingbrain:
This can't be the case because `Mounts` a go-IPFS specific config key (it's not used by js-IPFS) and the `Gateway` key only affects the gateway, not the WebUI.  The headers you have added are related to CORS so should not stop you loading the page, though will loosen the security of any page you load through the gateway.

If you can add a test that fails against master and passes with this patch then that would help narrow the problem down but I don't think this can be merged as it stands as it's not clear how it solves the problem.

lidel:
I believe CORS handling  of API port was refactored recently, I'm closing this as no longer relevant, cc @achingbrain to configm.

PR #3550, opened 2021-02-15
chore: upgrade deps with new typedefs

achingbrain:
Upgrades to new version with types

- Uses default aegir ts config
- Fixes all ts errors
- Fully types core-api in ipfs-core-types package
- Makes ipfs-core implement types from ipfs-core-types package
- Removes duplicate types, ipfs-core-types as single source of type truth
- Reduces use of external APIs by internal components in ipfs-core

Depends on:

- [x] https://github.com/ipfs/js-ipfs-unixfs/pull/114
- [x] https://github.com/ipfs-shipyard/js-hamt-sharding/pull/3
- [x] https://github.com/ipld/js-ipld-dag-pb/pull/189
- [x] https://github.com/ipfs/js-ipfs-unixfs/pull/122
- [x] https://github.com/ipfs/js-datastore-pubsub/pull/60
- [x] https://github.com/ipfs/js-ipfs-unixfs/pull/133
- [x] https://github.com/ipfs/js-ipfs-repo-migrations/pull/82
- [x] https://github.com/ipfs/js-ipfsd-ctl/pull/616
- [x] https://github.com/ipfs/js-ipfs-utils/pull/111
- [x] https://github.com/alanshaw/it-concat/pull/6
- [x] https://github.com/ipfs/interop/pull/257
- [x] https://github.com/rvagg/cborg/issues/8
- [x] https://github.com/rvagg/cborg/issues/7
- [x] https://github.com/rvagg/cborg/issues/9
- [ ] https://github.com/browserify/browser-resolve/pull/101 or some variation of https://github.com/multiformats/js-multiformats/pull/66
- [x] https://github.com/rvagg/cborg/pull/10
- [x] https://github.com/rvagg/cborg/pull/12

To do:

- [x] Entype core-api in `ipfs-core-types` package
- [x] Make `ipfs-core` implement types from `ipfs-core-types` package
- [x] Make `ipfs-grpc-*` implement types from `ipfs-core-types` package
- [x] Make `ipfs-message-port-*` implement types from `ipfs-core-types` package
- [x] Make `ipfs-http-client` implement types from `ipfs-core-types` package
- [x] Make `ipfs-http-server` implement types from `ipfs-core-types` package
- [x] Make `ipfs-http-gateway` implement types from `ipfs-core-types` package
- [x] Make `ipfs-daemon` implement types from `ipfs-core-types` package
- [x] Make `ipfs-cli` implement types from `ipfs-core-types` package
- [x] Make `ipfs` implement types from `ipfs-core-types` package
- [x] Resolve main/exports esm/cjs aegir/ipjs imports
- [x] Fix tests


packages/ipfs-core/src/components/files/utils/add-link.js
```diff
@@ -8,6 +8,7 @@ const CID = require('cids')
 const log = require('debug')('ipfs:mfs:core:utils:add-link')
 const UnixFS = require('ipfs-unixfs')
 const DirSharded = require('ipfs-unixfs-importer/src/dir-sharded')
+const defaultImporterOptions = require('ipfs-unixfs-importer/src/options')
```
hugomrdias, line 11: shouldn't this be exported by the entrypoint to avoid these deep requires ?

packages/ipfs-core/src/components/files/utils/hamt-utils.js
```diff
@@ -102,6 +117,7 @@ const generatePath = async (context, fileName, rootNode) => {
       prefix: toPrefix(currentBucket._posAtParent)
     })
 
+    // @ts-ignore
```
hugomrdias, line 120: can we get a reason in the comment please ?

hugomrdias (changes requested):

packages/ipfs-core/src/components/files/chmod.js
```diff
@@ -210,13 +210,19 @@ module.exports = (context) => {
           dagBuilder: async function * (source, block, opts) {
             for await (const entry of source) {
               yield async function () {
-                const cid = await persist(entry.content.serialize(), block, opts)
+                /** @type {DAGNode} */
+                // @ts-ignore
```
rvagg, line 214: I think you get to remove this ignore now

packages/ipfs-core/src/components/files/stat.js
```diff
@@ -77,77 +77,78 @@ const statters = {
...
+  file: (file) => {
     /** @type {Stat} */
-    const output = {
+    const stat = {
       cid: file.cid,
       type: 'file',
-      size: size,
-      cumulativeSize: cumulativeSize,
-      blocks: blocks,
+      size: file.unixfs.fileSize(),
```
rvagg, line 88: do we have a guarantee now that there is a `file.unixfs`?

packages/ipfs-core/src/components/files/stat.js
```diff
@@ -77,77 +77,78 @@ const statters = {
...
+      cumulativeSize: file.node.size,
+      blocks: file.unixfs.blockSizes.length,
       local: undefined,
       sizeLocal: undefined,
-      withLocality: false
+      withLocality: false,
+      mode: file.unixfs.mode
     }
 
-    if (file.unixfs) {
```
rvagg, line 100: the things I think are missing from this deleted block now are - `stat.type` setting (it's always `'file'` now), `stat.size` set to `0` for `isDirectory()`, should these be added back?

rvagg: ... or maybe this is now safely disambiguating the file vs directory case before we even get here ... I guess that makes them unnecessary here!

packages/ipfs-core/src/components/files/utils/hamt-utils.js
```diff
@@ -62,13 +76,14 @@ const addLinksToHamtBucket = async (links, bucket, rootBucket) => {
...
-          hashFn: DirSharded.hashFn
+          hash: rootBucket._options.hash,
+          bits: rootBucket._options.bits
         }, bucket, pos))
 
         return Promise.resolve()
       }
 
-      return (rootBucket || bucket).put(link.Name.substring(2), {
+      return rootBucket.put(link.Name.substring(2), {
```
rvagg, line 86: I'm not seeing the guarantee in the changes here that there _will_ be a `rootBucket` at this point. It's added in `recreateHamtLevel` if there's not a `parentBucket`, but are we sure that there's always a `rootBucket` if there's a `parentBucket`? A quick `git grep` suggests that `recreateHamtLevel` either has only the first arg or all of the args, but I don't know whether those args always have a non-null value.

packages/ipfs-core/src/components/files/utils/remove-link.js
```diff
@@ -127,7 +127,7 @@ const updateShard = async (context, positions, child, options) => {
 
   log(`Updating shard ${prefix} with name ${newName}`)
 
-  const size = DAGNode.isDAGNode(result.node) ? result.node.size : result.node.Tsize
+  const size = result.node instanceof DAGNode ? result.node.size : result.node.Tsize
```
rvagg, line 130: I always get told off by browser people for using `instanceof`, isn't this something we're supposed to avoid by duck typing at a minimum?

packages/ipfs-core/src/components/files/utils/add-link.js
achingbrain: That's happening here: https://github.com/ipfs/js-ipfs-unixfs/pull/122

achingbrain: This has been done

packages/ipfs-core/src/components/files/utils/hamt-utils.js
achingbrain: Done

packages/ipfs-core-types/src/files/index.d.ts
```diff
@@ -0,0 +1,355 @@
...
+import { ToMTime } from './files'
+import CID, { CIDVersion } from 'cids'
+import { CodecName } from 'multicodec'
+import { HashName } from 'multihashes'
+import { Mtime } from 'ipfs-unixfs'
+import type { AddProgressFn } from '../root'
+
+export interface API<OptionExtension = {}> {
+  chmod: (path: string, mode: number | string, options?: ChmodOptions & OptionExtension) => Promise<void>
+  cp: (from: IPFSPath | IPFSPath[], to: string, options?: CpOptions & OptionExtension) => Promise<void>
```
hugomrdias, line 11: whats a `IPFSPath` ? `ipfs://<cid>`, `<cid>/filename.png` ?

hugomrdias (changes requested):
partial review 
some comments inline, im a little worried that our inputs are forcing the users to pull in `CID`, `PeerId`, `multiaddr`, `ipld-block` etc etc  
i know we talked about this a million times and i understand why its like this but i dunno if its the right way.
So if we force the input to be the instance of those classes we should do it everywhere, and put an @see 'url to package' in the js docs. 
Or we can accept the low level version of those like string for CID or UInt8array for Block etc.

packages/ipfs-core-types/src/basic.d.ts
```diff
@@ -0,0 +1,136 @@
+import CID from 'cids'
+import { AwaitIterable } from './basic'
```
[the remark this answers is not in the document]
achingbrain: Oops, looks like it.

packages/ipfs-core-types/src/basic.d.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/**
+ * File content in arbitrary (supported) representation. It is used in input
+ * positions and is usually normalized to `Blob` in browser contexts and
+ * `AsyncIterable<Uint8Array>` in node.
+ */
+export type ToContent =
+  | string
+  | InstanceType<typeof String>
```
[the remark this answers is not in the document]
achingbrain: If we want to support `'string'` and `new String('string')` then no as they are different.

Who does that though?

packages/ipfs-core-types/src/files/index.d.ts
achingbrain: Variously we have:

`Qmfoo`
`/ipfs/Qmfoo`
`/ipfs/Qmfoo/some/deep/file.txt`
`/ipns/Qmfoo` (!)

Here it could also be:

`/path/in/my/mfs.txt`

I'm starting to be of the opinion that these things should just be `CID`s or MFS paths.  If you have `/ipfs/Qmfoo/some/deep/file.txt` you should `ipfs.dag.resolve` it first and get a `CID`.

packages/ipfs-cli/src/commands/get.js
```diff
@@ -2,6 +2,7 @@
 
 const fs = require('fs')
 const path = require('path')
+// @ts-ignore no types
 const toIterable = require('stream-to-it')
```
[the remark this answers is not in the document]
achingbrain: Node readable streams are async iterators, but we use this module to turn a writable into a sink for use with it-pipe.

packages/ipfs-cli/src/commands/ls.js
```diff
@@ -79,8 +100,8 @@ module.exports = {
     }
 
     for await (const link of ipfs.ls(key, { recursive, timeout })) {
-      const mode = formatMode(link.mode, link.type === 'dir')
-      const mtime = formatMtime(link.mtime)
+      const mode = link.mode != null ? formatMode(link.mode, link.type === 'dir') : ''
+      const mtime = link.mtime != null ? formatMtime(link.mtime) : '-'
```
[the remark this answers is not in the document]
achingbrain: Yes, see the test - https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-cli/test/ls.js#L134-L135

packages/ipfs-cli/src/commands/add.js
```diff
@@ -161,6 +165,37 @@ module.exports = {
...
+   * @param {boolean} argv.progress
+   * @param {string[]} argv.file
+   * @param {number} argv.mtime
+   * @param {number} argv.mtimeNsecs
+   * @param {boolean} argv.recursive
+   * @param {boolean} argv.hidden
+   * @param {boolean} argv.preserveMode
+   * @param {boolean} argv.preserveMtime
+   * @param {number} argv.mode
+   * @param {import('multibase/src/types').BaseName} argv.cidBase
```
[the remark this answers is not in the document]
achingbrain: Done

packages/ipfs-cli/src/commands/object/patch/set-data.js
```diff
@@ -1,17 +1,23 @@
 'use strict'
 
 const fs = require('fs')
-const concat = require('it-concat')
+const all = require('it-all')
+const uint8arrayConcat = require('uint8arrays/concat')
```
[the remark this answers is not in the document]
achingbrain: I mis-read the types for it-concat, have reverted.

packages/ipfs-core-types/src/block/index.d.ts
```diff
@@ -0,0 +1,80 @@
+import { AbortOptions, PreloadOptions } from '../basic'
+import CID, { CIDVersion } from 'cids'
+import Block from 'ipld-block'
+import { CodecName } from 'multicodec'
+import { HashName } from 'multihashes'
+
+export interface API<OptionExtension = {}> {
+  get: (cid: CID | string | Uint8Array, options?: AbortOptions & PreloadOptions & OptionExtension) => Promise<Block>
```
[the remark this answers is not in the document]
achingbrain: I was trying to only break APIs where typing was impossible, e.g. `ipfs.files.rm(...paths, options)` had to change to `ipfs.files.rm(string | string[], opts) as rest params have to be last.

Mostly this documents the current state of the API whether we like it or not but I think we should definitely look at reducing the our input types and the general API surface area as part of a simplification project.

packages/ipfs-core-types/src/block/index.d.ts
```diff
@@ -0,0 +1,80 @@
...
+  put: (block: Block | Uint8Array, options?: PutOptions & OptionExtension) => Promise<Block>
```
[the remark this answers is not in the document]
achingbrain: See https://github.com/ipfs/js-ipfs/pull/3550#discussion_r598943203

packages/ipfs-core-types/src/basic.d.ts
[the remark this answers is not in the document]
hugomrdias: lol right, this is just a nit so feel free to ignore

hugomrdias (approved):

PR #3556, opened 2021-02-17
feat: upgrade to the new multiformats

vmx:


oed:
Since this changes the underlaying nodejs version supported which has consequences for various build tooling, will the major version be bumped when this is released?

achingbrain:
The supported node versions are [Current and Active LTS](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#supported-versions) which at the time of writing are `14.x` and `16.x`.

CI is [still run](https://github.com/ipfs/js-ipfs/blob/master/.travis.yml#L22-L24) on these versions, the change you may be referring to here is switching what version the GH actions run on, which just do typechecking and report bundle size changes.

Where modules are `0.x.x` the minor will be bumped, where they are `x.x.x` the major will be bumped, but this is because the constructor args have changed to accept [IPLD codecs](https://www.npmjs.com/package/multiformats#ipld-codecs-multicodec) in place of [ipld-format](https://www.npmjs.com/package/interface-ipld-format)s, and the block API now accepts/returns `Uint8Array`s in place of the `{ cid, data }` tuple.

PR #3565, opened 2021-02-19
chore: use npm7 workspaces instead of lerna bootstrap

achingbrain:
It's much faster!  npm@7 is required so people will need to upgrade
if they are running node 14.  Node 15 comes with npm@7 so all good
there.

```
npm7

real	2m30.996s
user	8m11.495s
sys	0m48.380sit

lerna

real	4m19.727s
user	4m36.371s
sys	0m51.310s
```


package.json
```diff
@@ -26,13 +25,13 @@
     "clean": "lerna run clean",
     "lint": "lerna run lint",
     "dep-check": "lerna run dep-check",
-    "configure-examples": "run-s configure-examples:* release:pre:reinstall release:pre:bundle",
-    "configure-examples:add-examples": "json -I -f ./lerna.json -e \"this.packages.push('examples/*'); this.packages = [...new Set(this.packages)]\"",
-    "configure-examples:add-hoisted-modules": "json -I -f ./lerna.json -e \"this.command.bootstrap.nohoist = ['ipfs-css', 'tachyons']; this.command.bootstrap.nohoist = [...new Set(this.command.bootstrap.nohoist)]\"",
+    "configure-examples": "run-s configure-examples:* release:pre:reinstall build",
+    "configure-examples:add-examples-lerna": "npx json -I -f ./lerna.json -e \"this.packages.push('examples/*'); this.packages = [...new Set(this.packages)]\"",
+    "configure-examples:add-examples-npm": "npx json -I -f ./package.json -e \"this.workspaces.push('examples/*'); this.workspaces = [...new Set(this.workspaces)]\"",
```
hugomrdias, line 30: if we define workspaces in the package.json we still need it in lerna.json ?

hugomrdias (changes requested):

package.json
achingbrain: For the time being yes - when [v4.0.0](https://github.com/lerna/lerna/issues/2703#issuecomment-777089520) is stable we can probably get rid of it.

achingbrain: Actually, I type that and lerna 4 is now tagged `latest` instead of `next` so maybe it's ready

achingbrain: Though there's this: https://github.com/lerna/lerna/issues/2841

achingbrain:
So close!  Needs https://github.com/ipfs-examples/js-ipfs-examples/pull/29

PR #3573, opened 2021-02-26
fix: transfer set

### LINKED ISSUE #3402: ipfs-message-port-* should ensure no transferables are duplicated

Gozala:
This pull requests supersedes https://github.com/ipfs/js-ipfs/pull/3481 that passes transferables as sets instead of arrays (in order to avoid duplicates in them). However https://github.com/ipfs/js-ipfs/pull/3481 still kept using arrays during encoding and only afterwards it turned it into a set to eliminate duplicates.

This change instead updates type of transfer to set to eliminate intermediary arrays in favor of sets.

Fixes #3402


Gozala:
Example seems to be failing because, webpack-cli get hoisted and instead of using local webpack@5 it uses hoisted webpack@4 and fails 😭

I'll try to figure out how to resolve this and push a fix, that said I think it's ready to be reviewed.

Gozala:
I've spend couple of hours today trying to get the last test passing, but I could not work out the hoisting problem. I also tried to just update all webpack deps to webpack@5 but that broke aegir build which  started to getting webpack@5 instead webpack@4 it expects.

lidel:
@Gozala  is this waiting for TS work to land first, or is it blocked on something else?

Gozala:
> @Gozala is this waiting for TS work to land first, or is it blocked on something else?

This PR just needs to be updated to merge all the changes in the tree. However due to bunch of issues with duplicate deps in the tree (caused by TS Work) it's best to wait for that to settle instead of duplicating effort.

Gozala:
@achingbrain do you by a chance know why code in message port client changed from class syntax prototype base one ?

bmann:
Ping! Looking to get this merged, any updates?

icidasset:
Anything we can do to help get this merged?

achingbrain (approved):

icidasset:
Thank you! ☺️

achingbrain:
@icidasset there’s an rc available with this fix in, could you please try the ‘next’ version of the message port deps and let me know if it resolves the issue & I’ll do a full release.

icidasset:
@achingbrain It works, thanks! 🙌 🎉 

---

One thing I've noticed is
https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-message-port-protocol/src/cid.js#L22

That CID encoding & decoding doesn't account for the old [cids](https://www.npmjs.com/package/cids), which don't have the `cid.multihash.bytes` property. So it'd need to be `cid.multihash.buffer` instead.

Any thoughts on that? It's an issue for us because we wouldn't have any backwards compatibility.

achingbrain:
Interesting, we've fully migrated to the new CID implementation from the multiformats module, are you unable to do the same?

icidasset:
We are, but, in our architecture there will be older clients.
Meaning, if the ipfs node on our shared worker uses the new protocol, older clients won't work anymore.
We have some solutions in mind, but I was just wondering what your thoughts were on backwards compatibility.

achingbrain:
You can convert CID instances from the old version to the new with:

```js
import { CID } from 'multiformats/cid'

function catFile (maybeOldCid) {
  const cid = CID.asCID(maybeOldCid)

  return ipfs.cat(cid)
}
```

There `maybeOldCid` could be from either the `multiformats` or the `cids` module.

If you're exposing an API to your clients, you could do the conversion there?

icidasset:
Oh, that's a good idea. I could potentially "override" the `IPFSService` or `Server` from the ipfs-message-protocol-server package and swap out the CID, before it gets decoded 🤔 I'll try that, thanks again!

Issue #3391, opened 2020-11-12
Incidental Complexity: Dependency injection is getting in the way


Gozala:
@hugomrdias @achingbrain we have discussed this a bit on the call, but I don't think I did good job explaining exactly what I meant and comments like this https://github.com/ipfs/js-ipfs/pull/3365/files/fa7c2c13c43bab922052104fe8aff863d9725e9b?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.ts#r520398194 prompted me to do actual write up.

If I captured what @achingbrain told js-ipfs in the past used to use more object oriented approach in a sense that there was a class containing all the APIs where each function would just access things it needed on the instance and do it's job. Then at some point things were refactored to the current style.

I don't know the full history so I'm guessing that problem with OOP style was fairly common that is so many things initialized at various times making it really difficult to see what depends on what and what is the order in which things need to be initialized.

Guessing again it was improved by moving to the dependency injection style used today where individual functions are assembled by passing set of components they depend upon, which makes it clear what depends upon what and makes order in which initialization should happen a lot more clear.

However my argument is that:

1. Current approach introduces complexity when component A depends on component B.

   > Which is not theoretical, there were bunch of TODO comments that https://github.com/ipfs/js-ipfs/pull/3365/ fixes by breaking components further apart.

2. All this dependency injection just tends to introduce indirection, in practice every single function just needs to call other function, but it does not happens to exists until component creates it and threads it through.

What I would like to suggest is radically simplify this by decoupling (mutable) state from functions. Here is concrete example to put all this into perspective:

- `ipfs.add` is created from `addAll` 
   https://github.com/ipfs/js-ipfs/blob/583503bd35b23bc02f2eca2413a18d33cddccdcd/packages/ipfs-core/src/components/add.js#L5-L18

- Which is created by `init`
   https://github.com/ipfs/js-ipfs/blob/583503bd35b23bc02f2eca2413a18d33cddccdcd/packages/ipfs-core/src/components/init.js#L150

- Which in turns requires other components
   https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/src/components/add-all/index.js#L18-L27

- I want go into all of them but `pin` is interesting because, `addAll` doesn't really need whole pin API just an `pin.add` function:
[...]

PR #3588, opened 2021-03-10
feature: implement remote pinning API

yusefnapora:
This is an implementation of the `ipfs.pin.remote` API as well as I can gather from reading the tests in `interface-ipfs-core`.

I haven't figured out how to actually run the tests yet, so we'll see 😄

Leaving as a draft for now, since I probably missed some stuff and am out of steam for the day. Also it's depending on an unpublished npm package (the `js-ipfs-pinning-service-client` I wrote earlier) so we should sort that out.

cc @Gozala :)

### DIFF
--- a/packages/ipfs-core/package.json
+++ b/packages/ipfs-core/package.json
@@ -89,6 +89,7 @@
     "it-first": "^1.0.4",
     "it-last": "^1.0.4",
     "it-pipe": "^1.1.0",
+    "js-ipfs-pinning-service-client": "github:yusefnapora/js-ipfs-pinning-service-client#master",
     "libp2p": "^0.30.7",
     "libp2p-bootstrap": "^0.12.1",
     "libp2p-crypto": "^0.19.0",
--- a/packages/ipfs-core/src/components/index.js
+++ b/packages/ipfs-core/src/components/index.js
@@ -29,6 +29,7 @@ const createIDAPI = require('./id')
 const createConfigAPI = require('./config')
 const DagAPI = require('./dag')
 const PinManagerAPI = require('./pin/pin-manager')
+const PinRemoteAPI = require('./pin/remote')
 const createPreloadAPI = require('../preload')
 const createMfsPreloadAPI = require('../mfs-preload')
 const createFilesAPI = require('./files')
@@ -57,6 +58,8 @@ class IPFS {
   constructor ({ print, storage, options }) {
     const { peerId, repo, keychain } = storage
     const network = Service.create(Network)
+    const swarm = new SwarmAPI({ network })
+    const config = createConfigAPI({ repo })
 
     const preload = createPreloadAPI(options.preload)
 
@@ -86,7 +89,8 @@ class IPFS {
     })
     const resolve = createResolveAPI({ ipld, name })
     const pinManager = new PinManagerAPI({ repo, dagReader })
-    const pin = new PinAPI({ gcLock, pinManager, dagReader })
+    const pinRemote = new PinRemoteAPI({ swarm, config, peerId })
+    const pin = new PinAPI({ gcLock, pinManager, dagReader, pinRemote })
     const block = new BlockAPI({ blockService, preload, gcLock, pinManager, pin })
     const dag = new DagAPI({ ipld, preload, gcLock, pin, dagReader })
     const refs = Object.assign(createRefsAPI({ ipld, resolve, preload }), {
@@ -155,7 +159,7 @@ class IPFS {
     this.version = createVersionAPI({ repo })
     this.bitswap = new BitswapAPI({ network })
     this.bootstrap = new BootstrapAPI({ repo })
-    this.config = createConfigAPI({ repo })
+    this.config = config
     this.ping = createPingAPI({ network })
 
     this.add = add
@@ -170,7 +174,7 @@ class IPFS {
     this.object = new ObjectAPI({ ipld, preload, gcLock, dag })
     this.repo = new RepoAPI({ gcLock, pin, repo, refs })
     this.stats = new StatsAPI({ repo, network })
-    this.swarm = new SwarmAPI({ network })
+    this.swarm = swarm
 
     // For the backwards compatibility
     Object.defineProperty(this, 'libp2p', {
--- a/packages/ipfs-core/src/components/pin/index.js
+++ b/packages/ipfs-core/src/components/pin/index.js
@@ -5,7 +5,6 @@ const createAddAll = require('./add-all')
 const createLs = require('./ls')
 const createRm = require('./rm')
 const createRmAll = require('./rm-all')
-const RemotePinAPI = require('./remote')
 
 class PinAPI {
   /**
@@ -13,8 +12,9 @@ class PinAPI {
    * @param {GCLock} config.gcLock
    * @param {DagReader} config.dagReader
    * @param {PinManager} config.pinManager
+   * @param {PinRemoteAPI} config.pinRemote
    */
-  constructor ({ gcLock, dagReader, pinManager }) {
+  constructor ({ gcLock, dagReader, pinManager, pinRemote }) {
     const addAll = createAddAll({ gcLock, dagReader, pinManager })
     this.addAll = addAll
     this.add = createAdd({ addAll })
@@ -22,7 +22,7 @@ class PinAPI {
     this.rmAll = rmAll
     this.rm = createRm({ rmAll })
     this.ls = createLs({ dagReader, pinManager })
-    this.remote = new RemotePinAPI()
+    this.remote = pinRemote
   }
 }
 module.exports = PinAPI
@@ -34,4 +34,5 @@ module.exports = PinAPI
  * @typedef {import('..').PinManager} PinManager
  * @typedef {import('..').AbortOptions} AbortOptions
  * @typedef {import('..').CID} CID
+ * @typedef {import('./remote')} PinRemoteAPI
  */
--- a/packages/ipfs-core/src/components/pin/index.js
+++ b/packages/ipfs-core/src/components/pin/index.js
@@ -5,6 +5,7 @@ const createAddAll = require('./add-all')
 const createLs = require('./ls')
 const createRm = require('./rm')
 const createRmAll = require('./rm-all')
+const RemotePinAPI = require('./remote')
 
 class PinAPI {
   /**
@@ -21,6 +22,7 @@ class PinAPI {
     this.rmAll = rmAll
     this.rm = createRm({ rmAll })
     this.ls = createLs({ dagReader, pinManager })
+    this.remote = new RemotePinAPI()
   }
 }
 module.exports = PinAPI
--- a/packages/ipfs-core/src/components/pin/remote/index.js
+++ b/packages/ipfs-core/src/components/pin/remote/index.js
@@ -1,120 +1,169 @@
 'use strict'
 
+const multiaddr = require('multiaddr')
+const PinningClient = require('js-ipfs-pinning-service-client')
+
 /**
- * @typedef {import('cids')} CID
- *
- * @typedef {'queued'|'pinning'|'pinned'|'failed'} PinStatus
- * 
- * @typedef {object} RemotePin
- * @property {string} [cid]
- * @property {PinStatus} [status]
- * @property {?string} [name]
- * @property {?object} [meta]
+ * PinRemoteAPI provides an API for pinning content to remote services.
  */
+class PinRemoteAPI {
 
-class RemotePinAPI {
-
-  constructor() {
-    this.service = new RemotePinServiceAPI()
+  /**
+   * @param {Object} opts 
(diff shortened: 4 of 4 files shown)


yusefnapora:
Hey @Gozala & @alanshaw, I think this is basically ready to review, but there are a couple things that I'd like to sort out before merging. 

First is the `js-ipfs-pinning-service-client` package, which isn't published to NPM and is just living in a github repo under my personal username. Would this belong in js-ipfs as one of the subpackages? I can pull it into this PR if so. Otherwise I guess it should at least get moved to the ipfs github org or similar and published to npm.

Second, I didn't implement saving / loading pinning service configs to the main IPFS config, although the `config` service instance does get passed through to the right spot, so it should't be too hard to wire up. The cli commands (`js-ipfs pin remote add`, etc) are also unimplemented, but maybe that could wait for a separate PR.

Anyway, things seem to have reached "works on my machine" status, so it would be great if someone else could check it out and kick the tires a bit.

lidel (commented):
> quick drive-by comments :)

Perhaps as a part of this PR you could add a simple browser example under  `examples/browser-remote-pinning` with:
- input for CID (or a file picker, so user can `ipfs.add` a file and fill CID field that way)
- inputs for pinning service  Endpoint and Key
- "pin!" button

It would make testing against different services easier + will be a useful artifact for the community that wants to add support for remote pinning to their dapp.

> First is the js-ipfs-pinning-service-client package, which isn't published to NPM and is just living in a github repo under my personal username. Would this belong in js-ipfs as one of the subpackages? 

I don't think it should be part of js-ipfs, it will be useful for people who don't run IPFS node at all (https://github.com/ipfs/pinning-services-api-spec/issues/43)
We already have go counterpart at https://github.com/ipfs/go-pinning-service-http-client,  so perhaps  move it to https://github.com/ipfs/ and then publish to NPM, so I can add it to https://github.com/ipfs/pinning-services-api-spec#client-libraries :)

yusefnapora:
@lidel, that's a good idea about the example. I haven't tested this in browsers yet, so that will be good proof that it works :) I'll try to add the example in tomorrow.

I agree that the pinning HTTP client is useful outside of js-ipfs, so it makes sense to put it into the ipfs github org. I'd like to add some tests to the client first, but that shouldn't take long. Then someone could review it and pull it into the ipfs org and do the npm thing. I'll send a flare up tomorrow or Wednesday for review :)

packages/ipfs-core/src/components/pin/remote.js
```diff
@@ -0,0 +1,441 @@
...
+   * @param {Array<string>} options.cid - CID(s) to remove from remote pinning service
+   * @param {?Array<PinStatus>} options.status - only remove pins that have one of the specified statuses (queued, pinning, pinned, failed). Default: pinned
+   * @param {?number} options.timeout - request timeout (seconds)
+   * @returns {Promise<void>}
+   */
+  async rm (options) {
+    const { service, ...rmOpts } = options
+    if (!service) {
+      throw new Error('service name must be passed')
+    }
```
Gozala, line 85: Might be worth consolidating handling lack of `service` option under `this.service.serviceNamed` instead.

packages/ipfs-core/src/components/pin/remote.js
```diff
@@ -0,0 +1,441 @@
...
+    if (!options.endpoint) {
+      throw new Error('option "endpoint" is required')
+    }
+
+    if (!options.key) {
+      throw new Error('option "key" is required')
+    }
+
+    const svcOpts = Object.assign({ swarm: this.swarm, peerId: this.peerId }, options)
+    this._services[name] = new RemotePinningService(name, svcOpts)
```
Gozala, line 146: We would want to persist these (I think in) config so registered services are available across sessions.

packages/ipfs-core/src/components/pin/remote.js
```diff
@@ -0,0 +1,441 @@
+'use strict'
+
+const multiaddr = require('multiaddr')
+const CID = require('cids')
+const PinningClient = require('js-ipfs-pinning-service-client')
+const withTimeoutOption = require('ipfs-core-utils/src/with-timeout-option')
+const log = require('debug')('ipfs:components:pin:remote')
+
+/**
+ * PinRemoteAPI provides an API for pinning content to remote services.
```
Gozala, line 10:
```suggestion
 * PinRemoteAPI provides an API for pinning content to remote services.
 *
 * @typedef {import('ipfs-core-types/src/pin/remote').API} API
 * @implements {API}
```

packages/ipfs-core/src/components/pin/remote.js
```diff
@@ -0,0 +1,441 @@
...
+      throw new Error('service name must be passed')
+    }
+    const svc = this.service.serviceNamed(service)
+    return svc.rmAll(rmOpts)
+  }
+}
+
+/**
+ * RemotePinServiceAPI provides methods to add, remove, and list the configured
+ * remote pinning services that are used by the remote pinning api.
```
Gozala, line 112:
```suggestion
 * remote pinning services that are used by the remote pinning api.
 *
 * @typedef {import('ipfs-core-types/src/pin/remote/service').API} ServiceAPI
 * @implements {ServiceAPI}
```

packages/ipfs-core/src/components/pin/remote.js
```diff
@@ -0,0 +1,441 @@
...
+    delete this._services[name]
+  }
+
+  /**
+   * Returns a RemotePinningService object for the given service name. Throws if no service has been configured with the given name.
+   *
+   * @param {string} name
+   * @returns {RemotePinningService}
+   */
+  serviceNamed (name) {
```
Gozala, line 199: I think we would want to make it a static method instead to avoid polluting a namespace.

Gozala (changes requested):
Thanks @yusefnapora again for driving this effort, I am very excited about this making it's way to js-ipfs.

I did a first pass and left some notes. Overall it looks good but there are few things we'd need to figure out before we can move this forward

1. Separate generated pinning service client. I am not sure if having it in separate package provides enough value to justify added overhead of coordinating changes.
2. Generated client pulls in different http library that would increase bundle size in browser and probably increase a maintenance burden. I think we would want to reuse existing http code used in rest of the code base instead (and that may imply not generating, but writing client code). However ultimately it's @achingbrain call.
3. Currently things are not persisted across sessions, which I think we should add before landing this.
4. We already have interface and type definitions in place (see https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core-types/src/pin/remote.ts) and it would be good to utilize them instead of duplicate them. Specifically `@implements {API}` allows us to ensure that same API is implanted across HTTP client and js-ipfs. If we need to update types that's fine, but we should try to reuse and comply to them unless there's a good reason not to.
5. I think we would prefer to not introduce another code style. I hate been this person as I've personally been pushing to revisiting current dependency injection style used to plain functions wrapped in class based sugar approach outlined here https://github.com/ipfs/js-ipfs/issues/3391. For the reference remote pinning service API on IPFS client follows it https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/pin/remote/index.js

I realize there's a lot here, but I'm available to help with moving certain pieces.

yusefnapora:
@Gozala thanks for all the review feedback 👍

I'm fine with refactoring to fit the current style; I was kind of confused by it when I first started hacking, but I think I'm getting it now 😄

I also kind of hate the generated client for the same reason... plus I just don't really like the code it generates and would be happy to ditch it. The HTTP API itself is pretty simple, so it's not hard to write a client using any HTTP request library. I'll plan on just doing that, and I can just leave the js client on my github org for now. We can always move it to shipyard or something if people start using it in other projects.

I'll plan to hack on that and implement the config persistence, and I'll ping you next week sometime for another review pass.

packages/ipfs-core/src/components/pin/remote/client.js
```diff
@@ -0,0 +1,283 @@
...
+    for (const s of status) {
+      searchParams.append('status', s)
+    }
+
+    let resp = await api.get('/pins', { searchParams })
+    let body = await resp.json()
+    const total = body.count
+    let yielded = 0
+    while (true) {
+      if (body.results.length < 1) {
```
Gozala, line 164: This seems redundunt if it's empty next loop will be noon and next `if` clause will return and if doesn't it's probably better to fail.

packages/ipfs-core/src/components/pin/remote/client.js
```diff
@@ -0,0 +1,283 @@
...
+  async function awaitPinCompletion (pinResponse) {
+    const pollIntervalMs = 100
+
+    let { status, requestid } = pinResponse
+    while (status !== 'pinned') {
+      if (status === 'failed') {
+        throw new Error('pin failed: ' + JSON.stringify(pinResponse.info))
+      }
+
+      await delay(pollIntervalMs)
```
Gozala, line 95: I wonder if we should make `pollInterval` part of the optional `add` options so it could be configured if necessary to longer / shorter interval.

packages/ipfs-core/src/components/pin/remote/rmAll.js
```diff
@@ -0,0 +1,26 @@
...
+module.exports = ({ serviceRegistry }) => {
+  /**
+   * Remove all pins that match the given criteria from a remote pinning service.
+   *
+   * @param {Query & AbortOptions} options
+   * @returns {Promise<void>}
+   */
+  async function rmAll (options) {
+    const { service } = options
+    const svc = serviceRegistry.serviceNamed(service)
```
Gozala, line 14: Unless I'm missing something this will return `null` unload config was loaded which may not be the case here.

packages/ipfs-core/src/components/pin/remote/ls.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = ({ serviceRegistry }) => {
+  /**
+   * List objects that are pinned by a remote service.
+   *
+   * @param {Query & AbortOptions} options
+   * @returns {AsyncGenerator<Pin>}
+   */
+  async function * ls (options) {
+    const { service } = options
+    const svc = serviceRegistry.serviceNamed(service)
```
Gozala, line 14: Unless I'm missing something this will return `null` unload config was loaded which may not be the case here.

packages/ipfs-core/src/components/pin/remote/rm.js
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * Remove a single pin from a remote pinning service.
+   * Fails if multiple pins match the specified query. Use rmAll to remove all pins that match.
+   *
+   * @param {Query & AbortOptions} options
+   * @returns {Promise<void>}
+   */
+  async function rm (options) {
+    const { service } = options
+    const svc = serviceRegistry.serviceNamed(service)
```
Gozala, line 15: Unless I'm missing something this will return `null` unload config was loaded which may not be the case here.

Gozala (changes requested):
Thanks @yusefnapora for all the work that went into it. I think it's mostly there, I made bunch of minor comments to improve things here and there. There is a thing however that I think need to be addressed

1. Unless I've missed something, it appears to me that the way things are setup `ipfs.pin.remote.*` calls will fail in a next sessions, because config isn't going to be loaded.

And a  before thing that I think would be really nice to have, but doesn't need to block this:

`pin/remote/client.js` seems unnecessary and only complicates things at this point. I think you end up with it because client used to be a separate lib, now that it's not the case it would make more sense to just have that logic in `add/rm/ls/*` and just pull endpoint/key instead.

packages/ipfs-core/src/components/config.js
```diff
@@ -524,4 +525,12 @@ module.exports.profiles = profiles
...
  * @typedef {import('ipfs-core-types/src/basic').ToJSON} ToJSON
  * @typedef {import('.').AbortOptions} AbortOptions
+ *
+ * @typedef {Object} PinningConfig
+ * @property {Object<string, RemotePinningServiceConfig>} [RemoteServices]
+ *
+ * @typedef {Object} RemotePinningServiceConfig
+ * @property {Object} API
+ * @property {string} API.Endpoint
+ * @property {string} API.Key
```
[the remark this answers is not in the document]
lidel: The smarter thing to do is to move all secrets to separate files outside the JSON config.
The only secret right now is `Identity.PrivKey` this PR adds second type of secret under `Pinning.RemoteServices[*].API.Key`.
I may be less work to move secrets outside, than to sanitize the output of `ipfs config` and make sure  `ipfs config replace` does not forget the secrets.

lidel (changes requested):

BigLep:
@yusefnapora : what's the plan for this PR?  Will you be working land it?  My goal is to understand the next steps.  Thanks!

lidel:
Closing until https://github.com/protocol/web3-dev-team/pull/58  is picked up by one of project teams

brickpop:
Any chance that the PR is ever considered? Pinning data on a remote node is very important for us
Thank you

PR #3599, opened 2021-03-25
fix: add bignumber.js dependency to core types

vmx:
The Bitswap types need bignumber.js. Without this dependency you could get an
error like

    ../../node_modules/ipfs-core-types/src/bitswap.ts:1:29 - error TS2307: Cannot find module 'bignumber.js' or its corresponding type declarations.

    1 import type BigInteger from 'bignumber.js'
                                  ~~~~~~~~~~~~~~

### DIFF
--- a/packages/ipfs-core-types/package.json
+++ b/packages/ipfs-core-types/package.json
@@ -28,6 +28,7 @@
   ],
   "license": "(Apache-2.0 OR MIT)",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


achingbrain:
This should no longer be required, now that BigInt is supported everywhere we support there's no need to use bignumber.js in the core api any more.

achingbrain:
Closing as we use BigInt in the core api.

PR #3606, opened 2021-04-01
chore: update it-to-buffer dep

achingbrain:
Pulls in latest version to drop Buffer dependency.


achingbrain:
Examples are broken due to https://github.com/Gozala/web-encoding/issues/19

PR #3608, opened 2021-04-01
chore: fix types

achingbrain:
Not sure why the compiler doesn't pick up things like this.


achingbrain:
I'm going to merge this branch to unblock #3556 - the outstanding problems will be resolved by in-range dependency updates so there doesn't seem any point in delaying this.

PR #3633, opened 2021-04-22
fix: interface core test swarm peers with webrtc does not dial the correct address

vasco-santos:
Error in CI: https://travis-ci.com/github/ipfs/js-ipfs/jobs/500525900

In the past webrtc did not proper support multiple listeners. This basically mean that if trying to listen on multiple addresses the webrtc sdp offers could end up with wrong signal server address throwing random errors in CI.

https://github.com/libp2p/js-libp2p-webrtc-star/pull/330 fixed this and now a libp2p node allows multiple listeners. When a peer with multiple webrtc-star listeners attempts to dial another peer via a star signalling server, the signal server of the target peers is identified among the listeners and used for starting the dial.

In the test changed in this PR (note that ports might change):

```
nodeA: ['/ip4/127.0.0.1/tcp/53466/ws/p2p-webrtcstar/p2p/QmPqSo3Kqf9XtWG6g3WKVrx2PZ2q3BoFZssVjixaweABNW']
nodeB: [
          '/ip4/127.0.0.1/tcp/53466/ws/p2p-webrtc-star/p2p/QmNLojvkVdRTD2bmy7Yzpc7vUBv6bUDEpamrgQeRwPQAg1',
          '/ip4/127.0.0.1/tcp/53467/ws/p2p-webrtc-star/p2p/QmNLojvkVdRTD2bmy7Yzpc7vUBv6bUDEpamrgQeRwPQAg1'
]
```

and the test did:

```
await nodeA.swarm.connect(nodeB.peerId.addresses[isBrowser ? 1 : 0])
```

We were using nodeA to dial nodeB with an address of a signalServer it does not use, which obviously failed. The solution here is to either have nodeB dial nodeA address (has same signalServer), or we need to do in the test a match function to understand what address of nodeB we should use (with the same signalServer).

The error in CI is poor and webrtc-star should validate we actually have the listener for the signalServer and throw a proper error if not instead of the Cannot read property 'listener' of undefined. This will be fixed shortly in a new PR

### DIFF
--- a/packages/interface-ipfs-core/src/swarm/peers.js
+++ b/packages/interface-ipfs-core/src/swarm/peers.js
@@ -126,10 +126,7 @@ module.exports = (common, options) => {
         }
       })).api
 
-      // TODO: the webrtc-star transport only keeps the last listened on address around
-      // so the browser has to use 1 as the array index
-      // await nodeA.swarm.connect(nodeB.peerId.addresses[0])
-      await nodeA.swarm.connect(nodeB.peerId.addresses[isBrowser ? 1 : 0])
+      await nodeB.swarm.connect(nodeA.peerId.addresses[0])
 
       await delay(1000)
       const peersA = await nodeA.swarm.peers()
--- a/packages/ipfs-core/package.json
+++ b/packages/ipfs-core/package.json
@@ -105,7 +105,7 @@
     "libp2p-noise": "^2.0.5",
     "libp2p-record": "^0.10.3",
     "libp2p-tcp": "^0.15.4",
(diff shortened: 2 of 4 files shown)


achingbrain (approved):
LGTM.

Could you please update the libp2p-webrtc dep version wherever it appears in package.json files to make sure we always get the right patch version with the fix, then this is good to go.

vasco-santos:
> Could you please update the libp2p-webrtc dep version wherever it appears in package.json files to make sure we always get the right patch version with the fix, then this is good to go.

done

achingbrain (approved):

PR #3634, opened 2021-04-23
fix: do not republish self key twice

vasco-santos:
IPNS Republish was republishing self private key twice, first as part of the root IPNS Record published when creating the IPFS Repo and then as part of keychain keys.

This was specially problematic when doing a flow such as:

```sh
node packages/ipfs/src/cli.js init
DEBUG=ipfs:i* node packages/ipfs/src/cli.js daemon --pass 12345678901234567890
```

As we provided a keychain password for the daemon, this password was going to be used for creating the keychain dek. However, self was created as part of the PeerId in the init command (without password).

### DIFF
--- a/packages/ipfs-core/test/name.spec.js
+++ b/packages/ipfs-core/test/name.spec.js
@@ -50,6 +50,26 @@ describe('name', function () {
       expect(republisher._republishEntries.calledTwice).to.equal(true)
     })
 
+    it('should not republish self key twice', async function () {
+      const mockKeychain = {
+        listKeys: () => Promise.resolve([{ name: 'self' }])
+      }
+      republisher = new IpnsRepublisher(sinon.stub(), sinon.stub(), sinon.stub(), mockKeychain, {
+        initialBroadcastInterval: 500,
+        broadcastInterval: 1000,
+        pass: 'pass'
+      })
+      republisher._republishEntry = sinon.stub()
+
+      await republisher.start()
+
+      expect(republisher._republishEntry.calledOnce).to.equal(false)
+
+      // Initial republish should happen after ~500ms
+      await delay(750)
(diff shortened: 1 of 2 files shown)


packages/ipfs-core/test/name.spec.js
```diff
@@ -50,6 +50,26 @@ describe('name', function () {
...
+        listKeys: () => Promise.resolve([{ name: 'self' }])
+      }
+      republisher = new IpnsRepublisher(sinon.stub(), sinon.stub(), sinon.stub(), mockKeychain, {
+        initialBroadcastInterval: 500,
+        broadcastInterval: 1000,
+        pass: 'pass'
+      })
+      republisher._republishEntry = sinon.stub()
+
+      await republisher.start()
```
achingbrain, line 64: Does this need stopping?

packages/ipfs-core/test/name.spec.js
```diff
@@ -50,6 +50,26 @@ describe('name', function () {
...
+
+      expect(republisher._republishEntry.calledOnce).to.equal(false)
+
+      // Initial republish should happen after ~500ms
+      await delay(750)
```
achingbrain, line 69: Can we configure a smaller delay to make the tests faster?

packages/ipfs-core/test/name.spec.js
vasco-santos: It is stopped in the afterEach (line 29), like remaining tests

packages/ipfs-core/test/name.spec.js
vasco-santos: changed to 200 with initial broadcast of 100

vasco-santos: also decreased the other test timers

achingbrain (approved):

PR #3642, opened 2021-04-27
chore: update ipld-dag-cbor

achingbrain:
Pulls in the latest borc for the latest buffer


vasco-santos (approved):
🎉 LGTM, other than needing to update bundle size

achingbrain:
The bundle size increase was caused by a new version of `ipld-dag-cbor` meaning the version depended on by `ipld` got out of sync and suddenly there were two of them.  I've updated `ipld` so we're back to one version.

PR #3650, opened 2021-04-28
feat: add multibaseName option for addAll

chrisdukakis:
Requires https://github.com/multiformats/js-cid/pull/144 & https://github.com/ipfs/js-ipfs-unixfs/pull/146

### DIFF
--- a/packages/ipfs-core-types/src/root.d.ts
+++ b/packages/ipfs-core-types/src/root.d.ts
@@ -189,7 +189,7 @@ export interface AddOptions extends AbortOptions {
   /**
    * The multibase name to use. (defaults to 'base58btc' for v0 & 'base32' for v1)
    */
-  multibaseName?: string
+  multibaseName?: BaseName
 
   /**
    * Multihash hashing algorithm to use. (Defaults to 'sha2-256')
--- a/packages/ipfs-core-types/src/root.d.ts
+++ b/packages/ipfs-core-types/src/root.d.ts
@@ -187,6 +187,11 @@ export interface AddOptions extends AbortOptions {
   cidVersion?: CIDVersion
 
   /**
+   * The multibase name to use. (defaults to 'base58btc' for v0 & 'base32' for v1)
+   */
+  multibaseName?: string
+
+  /**
    * Multihash hashing algorithm to use. (Defaults to 'sha2-256')
    */
   hashAlg?: string


packages/ipfs-core-types/src/root.d.ts
```diff
@@ -186,6 +186,11 @@ export interface AddOptions extends AbortOptions {
    */
   cidVersion?: CIDVersion
 
+  /**
+   * The multibase name to use. (defaults to 'base58btc' for v0 & 'base32' for v1)
+   */
+  multibaseName?: string
```
achingbrain, line 192: Please use a `BaseName` type from [multibase](https://www.npmjs.com/package/multibase) instead of a string.

achingbrain (changes requested):
Could you please add tests to `packages/interface-ipfs-core/src/add.js` and `packages/interface-ipfs-core/src/add-all.js` to prevent future regressions.

chrisdukakis:
Hi, I notice that cli has a [cidBase option](https://github.com/ipfs/js-ipfs/blob/28ad9ad6e50abb89a366ecd6b5301e848f0e9962/packages/ipfs-cli/src/commands/add.js#L321) for formatting output. Should I pass it to `addAll` options?

hugomrdias:
we should either use the same name in the cli and api or just rely on cid version at least in this api

BigLep:
2021-05-24 next step: @hugomrdias and @achingbrain are going to discuss.

lidel:
Additional comment from triage: given that JS APIs return CID, I think https://github.com/multiformats/js-cid/pull/144  is enough.
One can choose multibase via `cid.toString(base=base2)` already, we need to think twice if we want to create a precedent of additng this parameter to  every command that returns CID.

chrisdukakis:
Hi @lidel, idea is to create self-describing cid, not to convert. So methods _creating_ cids may use `multibaseName`, while others which return existing cids don't have to.

BigLep:
2021-06-14: @lidel is going to respond here on how we should proceed?

lidel:
Thank you @chrisdukakis  for submitting this, it helped us to clarify our policy regarding CIDs.
We marinated on this during weekly triage and the decision is to close this pecific  PR without merging.

Rationale:
- `add` operations are  not the only ones that return a CID – if we accept this, then we would either have incosistent APIs, or had to complicate every API by adding the multibase parameter
- Creating  vs converting is a fuzzy topic: multibase matters only when CIDv1 is serialized to text, and one could use CID without ever doing that, so introducing a parameter that controls something  indirect like that, at different abstraciton layer, is somehow wasteful
-  CID object provides the ability to customize base encoding when serializing to text already (`CID.toString(base)`),  and `toV1().toString(base)` could be used to ensure arbitrary representation as CIDv1 without the need for changing all IPFS commands

PR #3652, opened 2021-04-28
chore: update libp2p to final release version

achingbrain:


vasco-santos (approved):
still need kad dht + gossipsub, but I can add them as follow up

PR #3654, opened 2021-04-29
chore: fix docker images

achingbrain:
Install node-pre-gyp before IPFS and switch to official node image as the size difference is now quite small.

`--unsafe-perm` seems unnecessary too.


achingbrain:
typescript wat

achingbrain:
A new version of protobufjs has been released that has new types, causing the breakage.

PR #3656, opened 2021-04-29
fix: ignore the ts error caused by the recent protobufjs type change

achingbrain:
The json generated by protobufjs is now not compatible with the types on protobufjs's `fromJSON` method.


achingbrain:
https://github.com/ipfs/js-ipfs/pull/3655 is a better fix, but isn't possible until protobufjs supports streaming gRPC methods - see https://github.com/protobufjs/protobuf.js/pull/1115

PR #3657, opened 2021-04-29
fix: update types after feedback from ceramic

### LINKED ISSUE #3640: [Types] IPFS.create() options uses Libp2pConfig instead of Libp2pOptions
Ref: https://github.com/ceramicnetwork/js-ceramic/pull/1194#discussion_r606629404

https://github.com/ipfs/js-ipfs/blob/63b11e5ac9d0b36d89ec23966e4523721878a7c0/packages/ipfs-core/src/types.d.ts#L8

Sooo .. it looks like what's going on here is that ipfs-core is pulling in `Libp2pConfig` when it should be pulling in `Libp2pOptions` because `Libp2pConfig` is used for the `config` property inside `Libp2pOptions`. So in https://github.com/ceramicnetwork/js-ceramic/pull/1194 I was having to pull up the elements in the `libp2p.config` object into `libp2p` to satisfy this condition. I think this should just be:

```js
import type { Libp2pOptions } from 'libp2p'
```

achingbrain:
Some of the types need a little correcting.

Fixes #3640

### DIFF
--- a/packages/ipfs-core/src/types.d.ts
+++ b/packages/ipfs-core/src/types.d.ts
@@ -5,7 +5,7 @@ import type { ProfileNames } from 'ipfs-core-types/src/config/profiles'
 import type IPLD from 'ipld'
 import type { Options as IPLDOptions } from 'ipld'
 import type Libp2p from 'libp2p'
-import type { Libp2pConfig as Libp2pOptions } from 'libp2p'
+import type { Libp2pOptions } from 'libp2p'
 import type IPFSRepo from 'ipfs-repo'
 import type { ProgressCallback as MigrationProgressCallback } from 'ipfs-repo-migrations'
 import type { Datastore } from 'interface-datastore'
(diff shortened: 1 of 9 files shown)


packages/ipfs-core/src/types.d.ts
```diff
@@ -117,7 +117,7 @@ export interface Options {
  * in Node.js, [`libp2p-browser.js`](../src/core/runtime/libp2p-browser.js) in
  * browsers.
    */
-  libp2p?: Libp2pOptions | Libp2pFactoryFn
+  libp2p?: Partial<Libp2pOptions> | Libp2pFactoryFn
```
rvagg, line 120: TIL about `Partial<>`.
What's the reason for doing this here, is `Libp2pOptions` too constrained, or is this what `getLibp2pOptions()` is doing, filling out defaults for the required?

achingbrain: The libp2p option type has a bunch of required fields, so make it a partial so we can accept an incomplete config object from the user.  Otherwise we’d have to duplicate the type to mark the fields optional.

It’s not perfect as it’s non-recursive so if that’s an issue maybe we can relax the required fields in libp2p itself in the future.

PR #3660, opened 2021-04-30
fix: only use public api in http api server

### LINKED ISSUE #3639: [Types] ipfs-http-server HttpApi requires JSIPFS type
Ref: https://github.com/ceramicnetwork/js-ceramic/pull/1194

https://github.com/ipfs/js-ipfs/commit/a418a521574c878d7aabd0ad2fd8d516908a3756 introduced this beast to the `HttpApi` constructor: `@typedef {IPFS & { ipld: IPLD, libp2p: libp2p }} JSIPFS`. How do we expect users to get one of these things. The two places we use `HttpApi` in the test don't seem to have types coverage. Is a `@ts-ignore` just how it has to be for now or is there an alternative that wraps up this little beasty from an existing `ipfs` instance?

achingbrain:
We needed to reach into ipld in order to get at the formats so we can deserialize the data.

Instead, put the data as a block and then read it out using ipfs.dag.get.

Slightly inefficient but better to use the public API.

Fixes #3639

### DIFF
--- a/packages/ipfs-http-server/package.json
+++ b/packages/ipfs-http-server/package.json
@@ -43,6 +43,7 @@
     "ipfs-core-utils": "^0.7.2",
     "ipfs-http-gateway": "^0.3.2",
     "ipfs-unixfs": "^4.0.3",
+    "ipld-block": "^0.11.1",
     "ipld-dag-pb": "^0.22.1",
     "it-all": "^1.0.4",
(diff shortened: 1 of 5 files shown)


vasco-santos (approved):
LGTM! yeah, this is cleaner

rvagg:
👌 noice

PR #3662, opened 2021-04-30
fix: fix types

achingbrain:
Fix type errors exposed by integrating with powergate-client.


vasco-santos (approved):
LGTM! So bad that aegir ts checker does not get this

achingbrain:
inorite.  Their tsc config did catch it so it must just be one of the options.

achingbrain:
I think it's because we have `skipLibCheck: true` in our aegir config.  Turning that off reveals the errors this PR fixes but also errors in the types of modules outside our control 😭

PR #3669, opened 2021-05-04
fix: mark ipld options as partial

achingbrain:
Otherwise the user has to specify a blockstore etc.

We make the options object whole before instantiating ipld.


vasco-santos (changes requested):
This apparently broke the types on other ipfs modules

vasco-santos (approved):

PR #3672, opened 2021-05-04
docs: add missing repo API docs link

alanshaw:


achingbrain (approved):
The CI error would be solved by rebasing but it seems unnecessary as this is just a doc change.

PR #3673, opened 2021-05-06
fix: loosen input type for swarm.connect and swarm.disconnect

### LINKED ISSUE #3638: [Types] SwarmAPI.connect requires a Multiaddr but libp2p.dial allows PeerId|Multiaddr|string
Ref: https://github.com/ceramicnetwork/js-ceramic/pull/1194

Similar to https://github.com/ipfs/js-ipfs/issues/3637

https://github.com/ipfs/js-ipfs/blob/63b11e5ac9d0b36d89ec23966e4523721878a7c0/packages/ipfs-core-types/src/swarm/index.d.ts#L16

But this passes straight through to `libp2p.dial()` which is more permissive:

https://github.com/libp2p/js-libp2p/blob/5372f7af2f8f1653384504af15efea8ba776534b/src/index.js#L454-L465

Should we be similarly lax here?

achingbrain:
Accept peer ids as strings or multiaddrs as Multiaddrs.

Also removes accepting peerIDs as CIDs for consistency as I don't think we actually return peerIDs as CIDs from the api anywhere (from `ipfs.swarm.peers`, for example), only ever as strings.

Fixes #3638

### DIFF
--- a/packages/ipfs-core-types/src/bitswap/index.d.ts
+++ b/packages/ipfs-core-types/src/bitswap/index.d.ts
@@ -24,7 +24,7 @@ export interface API<OptionExtension = {}> {
    * // [ CID('QmHash') ]
    * ```
    */
-  wantlistForPeer: (peerId: CID | string, options?: AbortOptions & OptionExtension) => Promise<CID[]>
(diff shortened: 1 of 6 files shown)


vasco-santos (approved):
👌🏼  LGTM!

PR #3674, opened 2021-05-06
fix: reject requests when cors origin list is empty

achingbrain:
If CORS origin list is empty, Hapi throws an error as it considers
that to be invalid configuration.

We want to reject requests that send and origin or a referer when
no allowed origins have been configured, so when these headers are
sent, reject the request if no allowed origins are present in the
config.

### DIFF
--- a/packages/ipfs-http-server/src/index.js
+++ b/packages/ipfs-http-server/src/index.js
@@ -67,7 +67,13 @@ function isAllowedOrigin (str, allowedOrigins = []) {
     return false
   }
 
-  const { origin } = new URL(str)
+  let origin
+
+  try {
+    origin = (new URL(str)).origin
+  } catch {
+    return false
+  }
 
   for (const allowedOrigin of allowedOrigins) {
     if (allowedOrigin === '*') {
--- a/packages/ipfs-http-server/src/index.js
+++ b/packages/ipfs-http-server/src/index.js
@@ -59,16 +59,15 @@ async function serverCreator (serverAddrs, createServer, ipfs, cors) {
 }
 
 /**
- * @param {string} str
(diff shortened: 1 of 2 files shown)


packages/ipfs-http-server/src/index.js
```diff
@@ -58,6 +58,31 @@ async function serverCreator (serverAddrs, createServer, ipfs, cors) {
   return servers
 }
 
+/**
+ * @param {string} str
```
vasco-santos, line 62:
```suggestion
 * @param {string} [str]
```
it might be empty

packages/ipfs-http-server/src/index.js
```diff
@@ -58,6 +58,31 @@ async function serverCreator (serverAddrs, createServer, ipfs, cors) {
...
+ * @param {string[]} allowedOrigins
```
vasco-santos, line 63:
```suggestion
 * @param {string[]} [allowedOrigins]
```
not required

vasco-santos (changes requested):

lidel (approved):
I did not run this, but should do the trick of aligning with what we do in  go-ipfs.
Small nits below.

vasco-santos (approved):

packages/ipfs-http-server/test/cors.js
```diff
@@ -149,5 +163,53 @@ describe('cors', () => {
 
       expect(res).to.have.property('statusCode', 404)
     })
+
+    it('rejects requests when cors origin list is empty and origin is sent', async () => {
```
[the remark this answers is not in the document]
achingbrain: Done!

PR #3675, opened 2021-05-06
fix: only accept cid for ipfs.dag.get

### LINKED ISSUE #3637: [Types] DAGAPI get requires CID but should also be `string|Uint8Array`?
Ref: https://github.com/ceramicnetwork/js-ceramic/pull/1194#discussion_r606629195

https://github.com/ipfs/js-ipfs/blob/63b11e5ac9d0b36d89ec23966e4523721878a7c0/packages/ipfs-core-types/src/dag/index.d.ts#L51

->

https://github.com/ipfs/js-ipfs/blob/63b11e5ac9d0b36d89ec23966e4523721878a7c0/packages/ipfs-core/src/components/dag/get.js#L14-L21

But `toCidAndPath()` still allows `string` and `Uint8Array`:

https://github.com/ipfs/js-ipfs/blob/63b11e5ac9d0b36d89ec23966e4523721878a7c0/packages/ipfs-core-utils/src/to-cid-and-path.js#L8-L12

achingbrain:
If you have a path within the DAG to resolve, pass it as `path` in the options object.

Makes the API conform to the documentation.

Fixes #3637

### DIFF
--- a/examples/traverse-ipld-graphs/eth.js
+++ b/examples/traverse-ipld-graphs/eth.js
@@ -34,14 +34,14 @@ async function main () {
     await ipfs.block.put(new Block(data, cid))
   }
 
-  const block302516 = 'z43AaGEywSDX5PUJcrn5GfZmb6FjisJyR7uahhWPk456f7k7LDA'
-  const block302517 = 'z43AaGF42R2DXsU65bNnHRCypLPr9sg6D7CUws5raiqATVaB1jj'
+  const block302516 = new CID('z43AaGEywSDX5PUJcrn5GfZmb6FjisJyR7uahhWPk456f7k7LDA')
+  const block302517 = new CID('z43AaGF42R2DXsU65bNnHRCypLPr9sg6D7CUws5raiqATVaB1jj')
   let res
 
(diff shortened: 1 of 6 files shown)


rvagg:
Travis is failing because `examples/traverse-ipld-graphs/eth.js` is passing string CIDs, I guess we don't have type checking extending down into there too?

achingbrain:
> I guess we don't have type checking extending down into there too?

This is intentional - the idea is to show a single idea as simply as possible and to not make any other technical decisions for the user, e.g. just plain js as far as possible, as they'll likely copy/paste example code into their own app so we don't want them to then have to remove extra scaffolding or framework code etc.

CSDUMMI:
> Travis is failing because `examples/traverse-ipld-graphs/eth.js` is passing string CIDs, I guess we don't have type checking extending down into there too?
@rvagg the issue still exists in the release. #3689

CSDUMMI:
I do not see, why we need to restrict the types.
**This is a breaking change and not intuitive!**

Because most js-IPFS functions (see Files API) 
accept both CID and String.

PR #3680, opened 2021-05-10
chore: fix flaky test

achingbrain:
Because the requests are initiated with `Promise.all`, there's no
guarantee which order they arrive at the server in so when we check
the responses, just make sure the length and contents are correct
instead of the order.


vasco-santos (approved):
LGTM

Issue #3591, opened 2021-03-15
JS-IPFS misrepresents key IDs when Ed25519 keys are generated


nadimkobeissi:
- **Version**: 
```javascript
{
  version: '0.5.4',
  repo: 10,
  commit: '511147bedd51be3151de44a90fefe9425bfbcd50',
  'interface-ipfs-core': '^0.144.2'
}
```

- **Platform**: ```Darwin Nadims-Mac.home 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64 x86_64```

- **Subsystem**: Probably `libp2p-crypto`

#### Description:
JS-IPFS does not correctly encode key IDs in the case of Ed25519 keys:

```javascript
let keyList = await ipfs.key.list()
console.log(keyList)
```

Gives us:
```javascript
[
  {
    name: 'exampleEd25519Key',
    id: '22,81,109,83,110,56,76,85,115,104,119,77,113,74,86,56,99,122,56,90,76,87,87,89,72,113,110,85,80,99,121,76,109,66,119,107,103,101,120,85,121,121,84,69,88,74,54'
  },
  {
    name: 'self',
    id: 'QmWC64HjhUmNjCNksZZzPVhmffx1LZoGobkFx2xFini4p6'
  }
]
```

The key ID for the `self` key, an RSA key, appears to be encoded correctly as a Qm multihash. However the key ID for the example Ed25519 key appears to be encoded as an odd string of bytes represented in decimal format. It looks to me like this encoding is incorrect or corrupted somehow.

I'm not sure what the procedure is to turn this bad encoding into a real ID, although, the following does seem to produce something resembling a QmHash:

```javascript
let idWeirdEncoding = ed25519ExampleKey.id.split(`,`)
let id = []
idWeirdEncoding.forEach((v, i) => {
    id[i] = String.fromCharCode(v)
})
```

I don't know if I can expect the resulting value from the above to be reliable. Either way the output does not seem correct and clarification on this issue is appreciated.

TheDiscordian:
The bytes listed decode to `QmSn8LUshwMqJV8cz8ZLWWYHqnUPcyLmBwkgexUyyTEXJ6`.

I believe if these are encoded using `b58mh`, this key should begin with `12D` for a `Ed25519` key. Strange.

nadimkobeissi:
[...]

PR #3693, opened 2021-05-18
feat: ed25519 keys by default

lidel:
This PR is wip to switch from RSA to ed25519 to match what we already do in go-ipfs.

When merged, this should close https://github.com/ipfs/js-ipfs/issues/3591

## TODO

- [x] clarify helptext that bits is only applied to RSA keys
- [x] `ipfs init` should default to ed25519
- [x] `ipfs key gen` should default to ed25519
- [x] tests should pass
- [x] why  `Ed25519` (current js-ipfs) vs `ed25519` (go-ipfs)? a typo or on purpose?
  - FYSA `./packages/ipfs-cli/src/commands/init.js` has ` @param {'RSA' | 'Ed25519' | 'secp256k1'} argv.algorithm`


achingbrain:
> why Ed25519 (current js-ipfs) vs ed25519 (go-ipfs)? a typo or on purpose?

It's the [KeyType declared by libp2p-crypto](https://github.com/libp2p/js-libp2p-crypto/blob/master/src/index.d.ts#L4), though we should accept case-insensitive input.

achingbrain (approved):

PR #3698, opened 2021-05-25
fix: remove optional chaining from code that will be transpiled

achingbrain:
Optional chaining is supported in node and in browsers but tools that
transpile code to run in older browsers doesn't understand the syntax.

Specifically the version of acorn that at least vue uses via a dependency
on webpack 4 does not support this.

It'll be fixed by them upgrading to webpack 5 but the timescale on that
is unclear, which is why we're using a prerelease version in our examples.

The short term fix is to just remove optional chaining from where we
are using it in code that will be transpiled for browser use.

### DIFF
--- a/packages/ipfs-core/src/components/add-all/index.js
+++ b/packages/ipfs-core/src/components/add-all/index.js
@@ -189,7 +189,7 @@ function pinFile (pin, opts) {
     for await (const file of source) {
       // Pin a file if it is the root dir of a recursive add or the single file
       // of a direct add.
-      const isRootDir = !file.path?.includes('/')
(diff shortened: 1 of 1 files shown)


lidel (approved):
:+1: for short term fix, but this type of thing often comes back as new PRs get merged.

Should we enforce https://eslint.org/docs/rules/no-unsafe-optional-chaining check at ci/eslint/aegir level for now? (cc @hugomrdias)

achingbrain:
I don't think that rule would do it, we'd want to disallow optional chaining altogether rather than just when the thing we think the property of is optional is undefined itself.

PR #3700, opened 2021-05-25
chore: allow bundlesize gh action to fail

achingbrain:
Sometimes it's because there's a duplicate dep version that needs to
be fixed.

Usually it's because some transitive dep increases it's bundle size
by a kilobyte or two and has nothing to do with the current PR, so all
we do is increase the `bundlesizeMax` value to make it pass.

Let's just make the check informational instead.


lidel (commented):
Sounds sensible. 
Should we upstream this to https://github.com/ipfs/aegir/blob/master/md/github-actions.md? (cc @hugomrdias)?

PR #3705, opened 2021-06-01
fix: move wrtc to optional deps

achingbrain:
npm7 has started installing peer deps by default, which means if you
are running a future tech node.js version you may end up compiling
native addons that may fail.  These addons should be opt-in so move
them to optional deps which means if they fail to install, the whole
installation does not fail.

Also adds `@mapbox/node-pre-gyp` as a dependency so we can attempt to 
compile modules that have native addons where no prebuilt binary is
available instead of requiring users to install it globally which is
required due to a misconfiguration in the `webrtc` module.

### DIFF
--- a/.github/workflows/bundlesize.yml
+++ b/.github/workflows/bundlesize.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


achingbrain:
We may be able to remove the `@mapbox/node-pre-gyp` dep if https://github.com/node-webrtc/node-webrtc/pull/694 ever gets merged.

PR #3706, opened 2021-06-03
fix: add onError to pubsub.subscribe types

achingbrain:
This was missed in #3468.  Also runs the type checker on the pubsub http client tests to ensure typing is correct.


achingbrain:
Types are failing - they've been broken by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52776#pullrequestreview-675605684 but will be fixed by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53591

PR #3737, opened 2021-07-05
docs: improve README of browser-create-react-app

kvutien:
I found a typo in the last line `egt` in place of `get`.

Then I rephrased some sentences. They puzzled me when I read them now from a beginner's point of view (OMG, it's SOO easy to start using jargon prematurely and confuse the reader)


examples/browser-create-react-app/README.md
```diff
@@ -1,15 +1,15 @@
...
+This is a minimal demonstration of how to use `js-ipfs` in a `create-react-app` generated app.
 
-It boots up a js-ipfs instance (an IPFS node) via a custom React hook in `./src/hooks/use-ipfs-factory.js`, which is called from `./src/App.js`. Once the IPFS node is set up, `App.js` displays its ident and its version number.
+It boots up a `js-ipfs` instance (an IPFS node) via a custom React hook in `./src/hooks/use-ipfs-factory.js`, which is called from `./src/App.js`. Once the IPFS node is set up, `./src/App.js` displays the ident of this node and the version number of `js-ipfs` used to spawn it.
 
-> _Remember that a Peer ID of an IPFS node is [the multihash of the public key of this node](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#peer-ids), and the public-private key pair of a node is generated by typing `ipfs init`._
+> _Side note: what we call ident here, or "id", is the Peer ID of the IPFS node and is [the multihash of the public key of this node](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#peer-ids). The public-private key pair of a node is generated by typing `ipfs init`._
```
achingbrain, line 7: This is a bit weird, I don't think we refer to a peer id as an "ident" anywhere else.

```suggestion
> _Side note: The PeerId of the IPFS node is [the multihash of the public key of this node](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#peer-ids). The public-private key pair of a node is generated by typing `ipfs init`._
```

achingbrain (approved):
This is great, thanks for clearing it up.  I've made some minor edits but let's get this in.

kvutien:
Great. Very kind of you.

> This is a bit weird, I don't think we refer to a peer id as an "ident" anywhere else.
...

Ah? OK. 

Your are right. It was not "ident", but `id` is used in `App.js` lines 7-10. I put a question as comments here, BTW. There was something that puzzled me when I tried to understand the code. I don't have the answer and maybe some other learner may have the same question. You may want to change the comments into your explanation?

PR #3746, opened 2021-07-16
Update package.json

bluelovers:


achingbrain:
`ipfs-unixfs` uses the new multiformats module which has a new implementation of the CID class.

It's not compatible with the codebase in master.

It will be upgraded by https://github.com/ipfs/js-ipfs/pull/3556

PR #3749, opened 2021-07-18
fix(ipfs-core-types): top `OptionExtension`

bluelovers:


achingbrain:
Thanks for opening this and the other PRs.

You haven't added a description so it's unclear why you wish to make the change so it's hard to review.  It's causing CI failures so it cannot be merged without further work.

Can you please for this and all future PRs, add a description that clearly states:

1. The problem you perceive
2. How your solution addresses the problem
3. Any potential problems with your solution or tradeoffs - breaking changes, etc

achingbrain:
Fixed by #3763

PR #3750, opened 2021-07-20
fix: support @web-std/file in normalize input

vasco-santos:
Related to https://github.com/web3-storage/ipfs-car/issues/62

### DIFF
--- a/packages/ipfs-core-utils/src/files/normalise-input/utils.js
+++ b/packages/ipfs-core-utils/src/files/normalise-input/utils.js
@@ -1,7 +1,5 @@
 'use strict'
 
-const { Blob } = globalThis
-
 /**
  * @param {any} obj
  * @returns {obj is ArrayBufferView|ArrayBuffer}
@@ -15,7 +13,7 @@ function isBytes (obj) {
  * @returns {obj is Blob}
  */
 function isBlob (obj) {
-  return typeof Blob !== 'undefined' && obj instanceof Blob
+  return obj.constructor && (obj.constructor.name === 'Blob' || obj.constructor.name === 'File')
 }
 
 /**
--- a/packages/ipfs-core-utils/src/files/normalise-input/utils.js
+++ b/packages/ipfs-core-utils/src/files/normalise-input/utils.js
@@ -10,10 +10,12 @@ function isBytes (obj) {
 
 /**
  * @param {any} obj
- * @returns {obj is Blob}
+ * @returns {obj is globalThis.Blob}
  */
 function isBlob (obj) {
-  return obj.constructor && (obj.constructor.name === 'Blob' || obj.constructor.name === 'File')
+  return obj.constructor &&
+    (obj.constructor.name === 'Blob' || obj.constructor.name === 'File') &&
+    typeof obj.stream === 'function'
 }
 
 /**
--- a/packages/ipfs-core-utils/test/files/normalise-input.spec.js
+++ b/packages/ipfs-core-utils/test/files/normalise-input.spec.js
@@ -6,7 +6,7 @@ const { expect } = require('aegir/utils/chai')
 const blobToIt = require('blob-to-it')
 const uint8ArrayFromString = require('uint8arrays/from-string')
 const all = require('it-all')
-const { File } = require('@web-std/file')
+const { File, Blob: WsBlob } = require('@web-std/file')
 const { Blob, ReadableStream } = globalThis
 const { isBrowser, isWebWorker, isElectronRenderer } = require('ipfs-utils/src/env')
 
@@ -39,8 +39,8 @@ async function verifyNormalisation (input) {
     try {
       expect(content).to.be.an.instanceOf(Blob)
     } catch (err) {
-      // Fallback to instance of File
-      expect(content).to.be.an.instanceOf(File)
+      // Fallback to instance of WsBlob
+      expect(content).to.be.an.instanceOf(WsBlob)
     }
     content = blobToIt(content)
   }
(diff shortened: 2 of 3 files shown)


packages/ipfs-core-utils/src/files/normalise-input/utils.js
```diff
@@ -15,7 +13,7 @@ function isBytes (obj) {
  * @returns {obj is Blob}
```
achingbrain, line 13: Because the `const { Blob } = ...` line was removed, this is now the global `Blob` so TS will be unhappy if someone explicitly passes `@web-std/file` objects in your code.

Might just need to remove the `@returns` annotation so tsc will derive a boolean return type.

packages/ipfs-core-utils/test/files/normalise-input.spec.js
```diff
@@ -35,7 +36,12 @@ async function verifyNormalisation (input) {
   let content = input[0].content
 
   if (isBrowser || isWebWorker || isElectronRenderer) {
-    expect(content).to.be.an.instanceOf(Blob)
+    try {
```
achingbrain, line 39: `File` extends `Blob` so under what conditions would we need to fall back here?

achingbrain (changes requested):

packages/ipfs-core-utils/src/files/normalise-input/utils.js
```diff
@@ -15,7 +13,7 @@ function isBytes (obj) {
...
  */
 function isBlob (obj) {
-  return typeof Blob !== 'undefined' && obj instanceof Blob
+  return obj.constructor && (obj.constructor.name === 'Blob' || obj.constructor.name === 'File')
```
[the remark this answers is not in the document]
vasco-santos: this sounds a good idea

packages/ipfs-core-utils/test/files/normalise-input.spec.js
vasco-santos: `instanceOf` is tricky. 
File https://github.com/web-std/file/blob/default/src/lib.js#L6 extends from Blob as you mention, and Blob is https://github.com/web-std/io/blob/main/blob/src/lib.js (so the same as here).

But, somehow this will not work with instanceOf Blob, and works with `@web-std/blob`. So, if you create a Blob and a @web-std/blob, their instanceOf are not compatible...

packages/ipfs-core-utils/src/files/normalise-input/utils.js
vasco-santos: return the `@returns` causes typescript to also not be happy, I added globalThis to the type

packages/ipfs-core-utils/test/files/normalise-input.spec.js
achingbrain: This sounds like `@web-std/blob` needs to return the native implementation if it's available?

achingbrain: Also of note here is that node 16 has a [Blob](https://nodejs.org/api/buffer.html#buffer_class_blob) implementation.

vasco-santos: > This sounds like @web-std/blob needs to return the native implementation if it's available?

It does: https://github.com/web-std/io/blob/main/blob/src/lib.js

but instanceOf returns false 🤷🏼‍♂️

achingbrain: > but instanceOf returns false 🤷🏼‍♂️

Then it can't be, otherwise instanceOf would return true?

vasco-santos: I changed it now to Blob to be:

```js
const { Blob } = require('@web-std/blob')
```

instead of 

```js
const { Blob } = globalThis
```

It works this way, but then other tests fail (which probably use the native Blob 🤷🏼‍♂️

vasco-santos: reverting this, as this is even more problematic

vasco-santos: So, when creating a https://github.com/web-std/io/tree/main/file in the browser, it will not be an instance of `Blob`, but it will be of https://github.com/web-std/io/tree/main/blob. This is really weird, given that ws Blob is really [globalThis.Blob](https://github.com/web-std/io/blob/main/blob/src/lib.js#L3).

This [codepen](https://codepen.io/vascosantos/pen/bGWrqMx?editors=0011) shows that a browser File should be instance of both Blob and ws Blob.

My only assumption for this to not be the case in the tests here, is that somehow playwright has a different globalThis scope in the test that makes them being different... @hugomrdias is this reasonable? do you have any clues?

vasco-santos: Now `@web-std/file` uses the Node.js version with Eletron renderer 😡

packages/ipfs-core-utils/src/files/normalise-input/utils.js
```diff
@@ -12,10 +10,12 @@ function isBytes (obj) {
 
 /**
  * @param {any} obj
- * @returns {obj is Blob}
+ * @returns {obj is globalThis.Blob}
  */
 function isBlob (obj) {
-  return typeof Blob !== 'undefined' && obj instanceof Blob
+  return obj.constructor &&
```
vasco-santos, line 16: @achingbrain what are your feelings on using here `@web-std/blob` ? That would work instead of "supporting browser only"

achingbrain (approved):

PR #3752, opened 2021-07-20
feat(ipfs-core-types): update ipfs.config.get

bluelovers:


lidel:
@bluelovers mind ensuring CI pass?

achingbrain:
@bluelovers similar to https://github.com/ipfs/js-ipfs/pull/3749#issuecomment-882458857 this PR has been opened without context or description of what you are trying to fix.

Please can you explain:

1. The problem you perceive
2. How your solution addresses the problem
3. Any potential problems with your solution or tradeoffs - breaking changes, etc

bluelovers:
> @bluelovers similar to [#3749 (comment)](https://github.com/ipfs/js-ipfs/pull/3749#issuecomment-882458857) this PR has been opened without context or description of what you are trying to fix.
> 
> Please can you explain:
> 
> 1. The problem you perceive
> 2. How your solution addresses the problem
> 3. Any potential problems with your solution or tradeoffs - breaking changes, etc

```ts
export async function ipfsApiAddresses(ipfs): Promise<string>
{
	return (ipfs as IPFS).config.get<string>('Addresses.API')
}
```

vs

```ts
export async function ipfsApiAddresses(ipfs): Promise<string>
{
	return (ipfs as IPFS).config.get('Addresses.API') as any
}
```

vs

```ts
export async function ipfsApiAddresses(ipfs): Promise<string>
{
	return (ipfs as IPFS).config.get('Addresses.API') // <=== error
}
```

achingbrain:
The config object is [fully typed](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core-types/src/config/index.d.ts#L32-L44) so it's better to make use of that.  Could you not do something like this instead?

```ts
export async function ipfsApiAddresses(ipfs: IPFS): Promise<string>
{
  const config = await ipfs.config.getAll()
  const value = config.Addresses?.API

  if (!value) {
    // handle missing value
  }

  return value
}
```

You don't need to introduce the generic type to cast the returned config value this way, and you'll get a compile time error if the shape of the config object ever changes.

If you use `ipfs.config.get(string)` it'll only break at run time.

bluelovers:
@achingbrain 

another type for auto return type of give path

```ts
type IsAUnion<T, Y = true, N = false, U = T> = U extends any
  ? ([T] extends [U] ? N : Y)
  : never;

type IsASingleStringLiteral<
  T extends string,
  Y = true,
  N = false
  > = string extends T ? N : [T] extends [never] ? N : IsAUnion<T, N, Y>;

type StringKeyOf<T> = Extract<keyof T, string>

type GetValueFromConfigByPath<T extends string, C = Config, V = object | string> =
T extends DotPathOfConfig<infer K1, infer S, C>
  ? K1 extends StringKeyOf<C> ?
    S extends StringKeyOf<C[K1]>
      ? C[K1][S] : GetValueFromConfigByPath<S, C[K1], V>
    : K1
  : T extends StringKeyOf<C> ? C[T] : IsASingleStringLiteral<T> extends true ? never : V
  ;

type DotPathOfConfig<K1 extends string, S extends string, C> = `${IsASingleStringLiteral<K1, K1>}.${IsASingleStringLiteral<S, S>}`
```

```ts
export interface API<OptionExtension = {}> {
  get: <T extends string>(key: T, options?: AbortOptions & OptionExtension) => Promise<GetValueFromConfigByPath<T, Config>>
}
```

==============================

> demo

```ts
export let rr = (async () => {

	let ca: API;

	let r1 = await ca.get('Addresses');
	let r2 = await ca.get('Addresses.API');
	let r3 = await ca.get('Datastore');
	let r4 = await ca.get('Datastore.Spec4');
	let r5 = await ca.get('Datastore.Spec.mounts');
	let r6 = await ca.get('Datastore.Spec.type');
	let r7 = await ca.get(`Datastore.Spec.${null as 'mounts' | 'type'}`);

	// @ts-expect-error
	r2 = 1;

	return {
		r1,
		r2,
		r3,
		r4,
		r5,
		r6,
		r7,
	}
})()
```

> output

```ts
export declare let rr: Promise<{
    r1: import("../index").AddressConfig;
    r2: string;
    r3: import("../index").DatastoreConfig;
    r4: never;
    r5: import("../index").DatastoreMountPoint[];
    r6: string;
    r7: string | import("../index").DatastoreMountPoint[];
}>;
```

bluelovers:
image

lidel:
@bluelovers thank you for the suggestion, but we decided to close this.

As @achingbrain noted, adds too much complexity for a feature that does not bring much value in JS context – querying subsets of config make sense in go-ipfs CLI but in JS we just want full config and operate on JSON object instead.

PR #3753, opened 2021-07-20
fix(ipfs-core-types): wrong extension

bluelovers:


lidel:
@bluelovers mind adding description why this PR is needed and make sure CI pass?

bluelovers:
it should be .d.ts not , .ts , and ci fail not by this pr

achingbrain (approved):

PR #3754, opened 2021-07-20
fix(ipfs-core-types): better typescript for IPFSEntry

bluelovers:
```ts
for await (const file of ipfs.get(cid))
			{
				if (file.path === cid && !file.content)
				{
					continue;
				}
```

# before

image

# after

image

### DIFF
--- a/packages/ipfs-core-types/src/root.d.ts
+++ b/packages/ipfs-core-types/src/root.d.ts
@@ -166,6 +166,7 @@ export interface Directory {
    * Directory path
    */
   path: string
+  content?: undefined
   mode?: number
   mtime?: Mtime
   size: number


achingbrain:
A directory doesn't have content in the same way a file does and it seems redundant to add an optional field that will never have a value.

The intention here is that you would use the `type` field on the `IPFSEntry` to disambiguate between `File` and `Directory` types before accessing their properties, eg:

```js
for await (const file of ipfs.get(cid)) {
  if (file.type === 'file') {
    // do something with file
  } else if (file.type === 'dir') {
    // do something with directory
  }
}
```

bluelovers:
> A directory doesn't have content

that is why set to `undefined` for `Directory`

this make a better type 

this change didn't break anything, but will make it more useful for typescript

achingbrain:
Thanks for bringing this up but looking into it more the current implementation of `ipfs.get` doesn't really align with the intention behind the (go) API at all so it's not surprising it's a bit confusing.

The idea for `ipfs.get` is that it lets you export a tarball containing files/whole directories but as it's currently implemented the js version is sort of like a recursive `ipfs.ls` with added the added `.content` property that may or may not be there.

Worse, in order to iterate over a directory using `ipfs.get` via the http client as currently implemented, you have to consume the content of a file before you can iterate to the next one, since it's parsing the output from the tarball behind the scenes which is a fairly significant foot gun.

Instead `ipfs.ls` should be used to list directory contents, then `ipfs.cat` to get the file contents, and `ipfs.get` should be used to get tarballs.

#3785 brings this more into line with the go-IPFS API.

Please can you try with the latest RC?

```console
$ npm install ipfs-core@next
or
$ npm install ipfs-http-client@next
```

You should be able to do something like:

```js
for await (const file of ipfs.ls(cid)) {
  if (file.type === 'file') {
    // do something with file, e.g. `for await (const buf of ipfs.cat(file.cid))`
  } else if (file.type === 'dir') {
    // do something with directory
  }
}
```

achingbrain:
This was shipped in `ipfs-http-client@52.0.0` and `ipfs-core@0.10.0`, please try the code snippet above with them.

PR #3755, opened 2021-07-20
fix: await pubsub unsubscribe to fix flaky test

achingbrain:
We manage pubsub subscriptions locally in the http client, when we unsubscribe from a given topic, we abort the long-lived connection we're holding open but do not wait for it to actually close.

In our tests we unsubscribe then check the subscription list - if we do this before the subscription is torn down it can report that we are still subscribed to the topic we just unsubscribed from.

The fix here is to wait for the connection to be torn down before returning from the unsubscribe call, by which point we should no longer have the topic in our subs list.

### DIFF
--- a/packages/ipfs-http-client/package.json
+++ b/packages/ipfs-http-client/package.json
@@ -64,6 +64,7 @@
     "multiformats": "^9.4.1",
     "nanoid": "^3.1.12",
     "native-abort-controller": "^1.0.3",
+    "p-defer": "^3.0.0",
     "parse-duration": "^1.0.0",
     "stream-to-it": "^0.2.2",
     "uint8arrays": "^2.1.6"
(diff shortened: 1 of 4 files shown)


achingbrain:
Fixes this sort of error:

```
ipfs:        .pubsub.unsubscribe
ipfs:          should subscribe and unsubscribe 5 times:
ipfs:       AssertionError: expected [ Array(1) ] to deeply equal []
ipfs:       + expected - actual
ipfs:       -[
ipfs:       -  "pubsub-tests-fAnzZgkOvVVMyDEl9ewm8"
ipfs:       -]
ipfs:       +[]
ipfs:       
ipfs:       at file:///home/travis/build/ipfs/js-ipfs/node_modules/chai-as-promised/lib/chai-as-promised.js:302:22
```

achingbrain:
Bleugh, didn't fix it, needs more investigation.

achingbrain:
Superseded by #3761

PR #3757, opened 2021-07-21
feat(ipfs-core-types): add `DAGLinkRawLike`

bluelovers:
https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/OBJECT.md#example-6

https://github.com/ipfs/js-ipfs/blob/91a84e4ecb793f450e97d9cadaa6782ac9017aee/packages/ipfs-http-client/src/object/patch/add-link.js#L21-L27

https://github.com/ipfs/js-ipfs/blob/91a84e4ecb793f450e97d9cadaa6782ac9017aee/packages/ipfs-http-client/src/object/patch/rm-link.js#L21-L25

### DIFF
--- a/packages/ipfs-core-types/src/object/patch/index.d.ts
+++ b/packages/ipfs-core-types/src/object/patch/index.d.ts
@@ -2,9 +2,23 @@ import type CID from 'cids';
 import type { AbortOptions } from '../../utils'
 import type { DAGLink } from 'ipld-dag-pb'
 
+export type DAGLinkRawLike = ({
+  Name: string
+  Hash: CID | string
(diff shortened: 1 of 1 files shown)


achingbrain:
`ipld-dag-pb` has been removed from the `ipfs` stack, it now uses `@ipld/dag-pb` which exports the [PBNode](https://github.com/ipld/js-dag-pb/blob/master/src/interface.ts#L14-L17).

The long term goal here is to tighten up the input types and not loosen them - having alternative field names and/or multiple casings of those names moves us in the opposite direction so I'm sorry but this cannot be merged.

PR #3760, opened 2021-07-27
feat: add `error.response` for better error handling

bluelovers:
image

really bad and useless

### DIFF
--- a/packages/ipfs-http-client/src/lib/core.js
+++ b/packages/ipfs-http-client/src/lib/core.js
@@ -102,11 +102,15 @@ const errorHandler = async (response) => {
   // This is what go-ipfs returns where there's a timeout
   if (msg && msg.includes('context deadline exceeded')) {
     error = new HTTP.TimeoutError('Request timed out')
+    // @ts-ignore
+    error.response = response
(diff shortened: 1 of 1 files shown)


achingbrain:
This is the wrong place to make this change, it should be added to the [TimeoutError constructor](https://github.com/ipfs/js-ipfs-utils/blob/master/src/http/error.js#L4) instead, then typescript would be happy and you wouldn't need to add the `@ts-ignore`.

What extra information are you getting from the response that makes it necessary to add it to the error?

bluelovers:
> This is the wrong place to make this change, it should be added to the [TimeoutError constructor](https://github.com/ipfs/js-ipfs-utils/blob/master/src/http/error.js#L4) instead, then typescript would be happy and you wouldn't need to add the `@ts-ignore`.
> 

this also didn't have typescript support

image


> What extra information are you getting from the response that makes it necessary to add it to the error?

what api i call make this timeout error and what really happen and let me can found where are code is

lidel:
Superseded by https://github.com/ipfs/js-ipfs-utils/pull/135

Issue #3776, opened 2021-07-29
Multiformats import error in typescript


NaikAayush:
- **Version**:

- **Platform**:
Linux fedora 5.13.4-200.fc34.x86_64 #1 SMP Tue Jul 20 20:27:29 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

- **Subsystem**:
ipfs-core-types

#### Severity:
Critical - System crash, application panic.

#### Description:
- Used ipfs-http-client in Angular
- Typechecking while compiling throws an error
- There is a spelling error in the ipfs-core-types files, where "multiformats" is spelt as "multiformts"

#### Steps to reproduce the error:

![image](https://user-images.githubusercontent.com/57558584/127545197-d23d2fbb-d9a6-470f-9934-5f9801e94588.png)

rvagg:
https://github.com/ipfs/js-ipfs/pull/3778

PR #3778, opened 2021-07-29
fix: typo in 'multiformats' type defs

rvagg:
Fixes: https://github.com/ipfs/js-ipfs/issues/3776

It'd be nice to have some kind of tests that caught this, but that might mean adding typescript examples that use this which is probably a fair bit of work. @achingbrain thoughts?

### DIFF
--- a/packages/ipfs-core-types/src/bitswap/index.d.ts
+++ b/packages/ipfs-core-types/src/bitswap/index.d.ts
@@ -1,4 +1,4 @@
-import type { CID } from 'multiformts/cid'
+import type { CID } from 'multiformats/cid'
 import type { AbortOptions } from '../utils'
 
 export interface API<OptionExtension = {}> {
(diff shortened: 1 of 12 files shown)


rvagg:
whoa boy, that's uncovered a bunch of type failures in Actions, mostly where we're allowing strings in JS but our types are strict about `CID`

achingbrain:
We really need to do something.  There is an [example](https://github.com/ipfs/js-ipfs/tree/master/examples/types-use-ipfs-from-ts) for using ipfs from ts but it doesn't catch this stuff, not 100% sure why not.

Adding types to the interface tests would probably help here but it's a big job.

Ref: https://github.com/ipfs/aegir/issues/849

rvagg:
k, so the change here is to go with the strict CID requirement and remove the option to provide strings in these places. SGTM although it'll be interesting to see user feedback as more people start adopting the latest and find their flexibility removed.

PR #3785, opened 2021-08-04
feat: make ipfs.get output tarballs

achingbrain:
`ipfs get` on the cli writes files/folders to disk, also tarballs and gzipped tarballs. Also gzipped individual files.

In the API it's very similar to `ipfs.ls`.  Make it more like the cli so there's a clear reason to use one or the other.

Also adds types to `interface-ipfs-core` and the `ipfs-core` tests to better catch broken types in `ipfs-core-types`.


achingbrain:
Depends on:

- [ ] https://github.com/alanshaw/it-tar/pull/5
- [ ] Probably a change to ipfsd-ctl

PR #3821, opened 2021-08-17
chore: move examples to external repo

oliveriosousa:
- Removes all the examples from this repository
- Change CI to execute the test examples from external repo (https://github.com/ipfs-examples/js-ipfs-examples.git)

**UPDATE:**

- Update `.md` files and add links to point to the new examples org/repos
- Remove scripts related with example tests
- Close the existing PRs against the current examples folder

CC @hugomrdias @achingbrain


achingbrain:
Could you please also:

* update the various `.md` files and/or add links to point to the new examples org/repos.
* Remove the [release:post:update-example-dependencies](https://github.com/ipfs/js-ipfs/blob/96b39099efb051b7a76f0afc2ff9429997c73971/package.json#L44) step from the build and supporting files
* Remove the various [configure-examples](https://github.com/ipfs/js-ipfs/blob/master/package.json#L29-L31) npm scripts
* Ensure the examples are actually tested in CI

achingbrain:
Also, it'd be nice to test the examples individually as it's the longest build step by far so good to break it into chunks.

achingbrain:
It'd be great to get this merged.  I think the last thing to do is to test the examples individually.

By which I mean instead of:

```yaml
    - stage: test
      name: examples
      script:
        # Travis lets scripts continue even if previous steps fail so need to use &&: https://github.com/travis-ci/travis-ci/issues/1066
        - npm run configure-examples &&
          npm run test -- $RUN_SINCE --scope=example* --concurrency=1
```

Do:

```yaml
    - stage: test
      name: examples - ipfs-http-client upload file
      script:
        - npm run test:external -- -- -- https://github.com/ipfs-examples/js-ipfs-http-client-upload-file.git

    - stage: test
      name: examples - ipfs-http-client bundle webpack
      script:
        - npm run test:external -- -- -- https://github.com/ipfs-examples/js-ipfs-http-client-bundle-webpack.git

    - stage: test
      name: examples - ipfs browser webpack
      script:
        - npm run test:external -- -- -- https://github.com/ipfs-examples/js-ipfs-browser-webpack.git

// etc
```

..so that all the example tests run in parallel.

oliveriosousa:
Hi @achingbrain,

Thanks for the explanation but it raises some concerns.
That way every time a new example is added to ipfs-example we must add a new entry in the travis file. If we leave as is, it is not needed to make any changes to the repo from now on.

Can we make this improvement of having tests running in parallel done in the other repo?

achingbrain:
> That way every time a new example is added to ipfs-example we must add a new entry in the travis file. If we leave as is, it is not needed to make any changes to the repo from now on.

This is true, but are you expecting a large amount of churn in the number of examples to be added over time?  If not, adding new entries may not be a huge burden.

Furthermore adding examples to the the `ipfs-example` org could end up changing the work the js-ipfs build needs to do between runs which makes it less deterministic.

> Can we make this improvement of having tests running in parallel done in the other repo?

This will still run in a single travis runner though, right?  Historically they've been too weak to do things like run multiple browser based tests simultaneously.

achingbrain (approved):

Issue #3914, opened 2021-10-07
http-client dag.put failed with invalid byte


jtsmedley:
- **Version**: ipfs-http-client 53.0.1, go-ipfs v0.10.0

- **Platform**: Windows

- **Subsystem**:

rvagg:
@achingbrain the DAG API is now out of sync between go-ipfs and js-ipfs-http-client and errors like this are going to start popping up. This particular error is going to be about the codec defaults, and what's more, the arguments have changed. `format` and `input-enc` are now replaced with `store-codec` and `input-codec` and both the defaults and behaviour are different, in subtle (but arguably much more clear, now) ways.

The default `input-enc` here is `raw` and `format` is `dag-cbor` and there's pre-processing done here before it's sent off to the server. So I think what's happening for this error is the JS object `{hello: "world"}` is encoded with dag-cbor and sent as bytes to the server. But the default `input-codec` is `dag-json` on the server and its receiving cbor bytes, so it's getting `0x5a` when it really wants a valid JSON value character.

Defaults and options need to be updated here to match 0.10. I just don't know how you deal with the version mismatch if someone's using 0.9, but maybe you just don't?

https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#v0100-2021-09-30

PR #3917, opened 2021-10-11
feat: update DAG API to match go-ipfs@0.10 changes

rvagg:
_Current status:_

* Focuses mainly on `dag.put` since that's where the bulk of the changes are
  - Option names have changed
  - Behaviour has now changed, strictly "input codec" and "store codec" where "codec" is that IPLD kind, no longer the special JSON form that was accepted for "input-enc". But in the case of the HTTP and Core APIs you give it the JS object as the node to store so "input codec" doesn't make as much sense - but I've made it so that if you give it "input codec" and a Uint8Array then it'll do a decode for you, or send it upstream to do the decode. So we have equivalent behaviour that you can use if there's an optimisation that makes sense for your use-case (perhaps you have the encoded bytes and want to pass that on and have it stored as an alternative codec than the one it's encoded with?).
* CLI has `--output-codec` for `dag get` which matches go-ipfs, roughly:
  - no newline character
  - uses the codec you requested, defaulting to dag-json
  - if you want to print a Bytes node and request an output codec of "raw" then `--data-enc` will still work, but otherwise that argument isn't used like it was before (and I'm pretty sure it's js-ipfs specific)
* Added an equivalent of first half of sharness t0053 from go-ipfs that pushes the DAG API around a bit
* Added dag-json as a first-class codec

Outstanding:

* https://github.com/ipfs-examples/js-ipfs-examples/pull/77
* CIDv0 was removed from go-ipfs 0.10 DAG API, you only have the option to get CIDv1. I think we should follow suit here.
* It was raised in https://github.com/ipfs-examples/js-ipfs-custom-ipld-formats/pull/1 but I think that DAG GET API _shouldn't_ be using the BLOCK API (neither should DAG PUT), it should be a pass-through to the server. If it's more efficient to use the BLOCK API then the user should use that directly. There are a couple of reasons:
  - The DAG API on the server should be responsible for codecs, go-ipfs has a coded plugin architecture for this, if the client doesn't have the codec but the server does, you should be able to get it from the server anyway
  - Using the BLOCK API means you need the entire block even if you only want a tiny bit of it - if you want the entire block, use the BLOCK API directly
  - The DAG API is responsible for navigating through the DAG, using the BLOCK API as a substitute on the client means you need all of the blocks from the server and then navigate to the final node to get what you want - if you're 5 blocks deep and they're all 1Mb and all you want is a single 1Kb string in the middle of the final node then you've had to download 5Mb and decode them all, in sequence, just to get the final node. The server is better placed to do this and hand you what you want.
  - It should be a matter of documentation - use the DAG API if you want to traverse into nodes, or want a convenient way to get a decoded JS object for a single block. If you want to grab whole blocks and you have the codecs then use the BLOCK API. If you have a use-case where dealing with raw-bytes, and possibly doing codec work on the client side, then use the BLOCK API. If you have a custom codec on the server and not on the client, then use the DAG API.

---

This is a Draft PR because there's more to do. The initial, most important thing is solving #3914 so the DAG PUT API works against 0.10, which this does. I think it'll end up being a non-breaking change for users unless they're using `format` option or expecting to get a v0 CID from a dag-pb (which go-ipfs doesn't do anymore).

As a fun side effect, this introduces a possible optimisation for users in some situations - if they have the already-encoded bytes of a block then they can use DAG PUT to send that in and bypass the encode/decode cycles on the client side and the server will store it as is if they provide `inputCodec`. Although the BLOCK API is probably more appropriate for that use-case.

To do:

 * [ ] Server PUT & GET changes
 * [ ] CLI PUT and GET changes
 * [ ] Consider what to do with the changed dag-pb & unixfs output and pathing semantics - at a minimum we might want to follow suit and ditch the custom dag-pb object shape output and use pure dag-pb output (yay)

---

feat!: match go-ipfs@0.10 dag put options and semantics

Fixes: https://github.com/ipfs/js-ipfs/issues/3914
Ref: https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#v0100-2021-09-30

--format and --input-enc have been replaced with --input-codec and
--store-codec and mean something a little different. You now supply
raw input and instruct the server which --input-codec that data is
in which it will decode, then re-encode with --store-codec before
storing it and providing you with the CID.

We accept plain JavaScript objects to encode with --store-codec via
the API here, defaulting to dag-cbor, and send that to the server as
encoded bytes using that codec, to be stored using that codec.

If you supply an --input-codec then we assume you're supplying raw,
encoded bytes using that codec and we pass that directly on to the
server to handle.


rvagg:
I don't understand the lint error on CI, I can't repro locally

oed:
Any update on this? It's a blocker for us as well.

BigLep:
@rvagg : are you able to get the build to pass?  @achingbrain will take on the code review afterwards.

rvagg:
Pushed update which sorts out a couple more issues that were exposed after the aegir / dependency-check fix cleaned up CI. But now CI is not running, hopefully just a temporary blip with GH.

rvagg:
Now playing wac-a-mole with the OBJECT API and CIDv0 / CIDv1 returns to get the suite to pass unfortunately, there seems to be some inconsistency in how that's been updated in go-ipfs that's tricky to narrow down.

rvagg:
So current status is that there's an awkward game going on with the OBJECT API, as I said above, where the API itself hasn't actually changed in go-ipfs, but some of it internally defers to the DAG API which now defaults to CIDv1. So there's some cases where it'll give you CIDv1 and some where you'll get v0, but it's not consistent and mostly comes back as v0. The HTTP client here has defaults for CID version in a few places, and a `version` parameter which isn't respected by go-ipfs but is by js-ipfs. Additionally, the js-ipfs server doesn't defer to the DAG API in exactly the same places the go-ipfs server does, so it's not a simple matter of fixing just those paths. The trick is finding the right combination of defaults to make the tests see the same thing against a js-ipfs and go-ipfs server. And it's messy, cause there's defaults in multiple places (client, server, core)--I change one thing and something else breaks.

I didn't come here to do the OBJECT API, but I guess this PR is now a general 0.10 compat PR.

rvagg:
OK, I've got to green in CI, but that's from disabling two examples in a960d28d4 `js-ipfs-custom-ipld-formats` & `js-ipfs-traverse-ipld-graphs`. I did contribute a fix for the first over in the js-ipfs-examples repo, I haven't had a chance to look at the latter and what it's problem is (not obvious from the failure output).

oed:
What's the next step here? Anything we can do to help move this forward?

ukstv:
@rvagg As annoying as it is, please, could you post some guidance on how to continue the work or contribute somehow. These last 5 meters to peak of Everest are undoubtedly the hardest ones, and there are some hands and brains willing to help here.

rvagg:
this is waiting on @achingbrain's bandwidth to review now

.github/workflows/test.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
         node: [16]
+      fail-fast: false
```
achingbrain, line 74: AFAIK gh orgs only have a certain number of workers available for actions - this may end up hogging resources and preventing other builds from running when we know a given CI run will fail.

lidel: If we truly care about not using too many workers then fail-fast does not buy us much.
We should limit the number of concurrent jobs with [`concurrency`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) – did a poc in https://github.com/ipfs/js-ipfs/pull/3917/commits/50cbc1f766a2871c0338da31708030dcd671875b

rvagg: sorry, this wasn't meant to be here for the final version - I turned this off because of the examples failures, which are currently commented out and need to be addressed before merge but I'm not sure how to do that dance when they depend on eachother being fixed

achingbrain: > If we truly care about not using too many workers

We should be optimistic that builds will complete successfully, so they should consume as many resources as necessary to do that expediently (within reason).  If they fail, they should not stand in the way of other builds completing, so should fail fast.

Not sure what reasonable is yet but so far so good.

>  Bit confused why CI stopped passing without any package.json changes

Right now it's failing because the types in `iso-random-stream` are broken - needs https://github.com/hugomrdias/iso-random-stream/pull/91

> I'm not sure how to do that dance when they depend on each other being fixed

We can check out branches of the example tests with fixes applied which should get the build passing.

e.g. in `.github/workflows/test.yml`

```
- name: ipfs browser video streaming
  repo: https://github.com/ipfs-examples/js-ipfs-browser-video-streaming.git
  deps: ipfs-core@$PWD/packages/ipfs-core/dist
```

would become:

```
- name: ipfs browser video streaming
  repo: https://github.com/ipfs-examples/js-ipfs-browser-video-streaming.git#fix/this-is-a-branch-with-a-fix
  deps: ipfs-core@$PWD/packages/ipfs-core/dist
```

If that's not practical, I think it's ok to merge with the example tests disabled, then fix the examples and re-enable them in a follow up PR.  As long as there isn't a massive lag between the two.

hugomrdias:
Released the fix for iso-random-stream

Hugo Dias

.github/workflows/test.yml
```diff
@@ -67,6 +67,7 @@ jobs:
     name: Unit tests ${{ matrix.project }} node ${{ matrix.node }} ${{ matrix.os }}
     needs: build
     runs-on: ${{ matrix.os }}
+    concurrency: js-ipfs-test-${{ github.ref }}
```
achingbrain, line 70: This is going to make CI take aaaaaages.

packages/ipfs-core/src/components/index.js
```diff
@@ -241,7 +242,7 @@ const addEmptyDir = async (ipfs) => {
 
   const cid = await ipfs.block.put(buf, {
     mhtype: 'sha2-256',
-    format: 'dag-pb'
+    storeCodec: 'dag-pb'
```
achingbrain, line 245: Is this done on purpose?  There's been no change to `ipfs.block.put` in `ipfs-core` to support a `storeCodec` options.

packages/ipfs-core/src/components/index.js
```diff
@@ -5,6 +5,7 @@ import errCode from 'err-code'
 import { UnixFS } from 'ipfs-unixfs'
 import * as dagPB from '@ipld/dag-pb'
 import * as dagCBOR from '@ipld/dag-cbor'
+import * as dagJSON from '@ipld/dag-json'
```
achingbrain, line 8: Does go-ipfs support dag-json by default now?

packages/ipfs-core/src/components/index.js
achingbrain: An aside, but it's Very Tedious that the type checker doesn't catch this.

achingbrain: Aha, we can make it catch it.  In `packages/ipfs-core/src/componenets/block/put.js`, change:

```js
  /**
   * @type {import('ipfs-core-types/src/block').API["put"]}
   */
  async function put (block, options = {}) {
```

to:

```js
  /**
   * @type {import('ipfs-core-types/src/block').API<{}>["put"]}
   */
  async function put (block, options = {}) {
```

then we get:

```
src/components/index.js:245:5 - error TS2345: Argument of type '{ mhtype: string; storeFormat: string; }' is not assignable to parameter of type 'PutOptions'.
  Object literal may only specify known properties, and 'storeFormat' does not exist in type 'PutOptions'.

245     storeFormat: 'dag-pb'
        ~~~~~~~~~~~~~~~~~~~~~
```

JSDoc doesn't pick up the default option extension in `/packages/ipfs-core-types/src/componenets/block/index.ts`.  That's annoying

achingbrain: https://github.com/ipfs/js-ipfs/pull/3965

packages/ipfs-core/src/components/index.js
achingbrain: Ah looks like it does.  One more for the kitchen sink.

```console
$ ipfs dag put --help
...
  --input-codec  string - Codec that the input object is encoded in. Default: dag-json.
```

packages/ipfs-core/src/components/index.js
achingbrain: Merged into this branch, all good!

achingbrain (approved):

PR #3918, opened 2021-10-12
Fixes typo in pull request 3607

dancingfrog:
See comment on https://github.com/ipfs/js-ipfs/pull/3607#issuecomment-941005529

### DIFF
--- a/packages/ipfs-http-gateway/src/index.js
+++ b/packages/ipfs-http-gateway/src/index.js
@@ -72,7 +72,7 @@ export class HttpGateway {
     // @ts-ignore TODO: move config typedefs to repo
     const config = await ipfs.config.getAll()
     const addresses = config.Addresses || { Swarm: [], Gateway: [] }
-    const gatewayAddrs = addresses?.Gateway || []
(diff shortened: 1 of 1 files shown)


dancingfrog:
After spending more time going through `index.js` I have concluded that this file is definitely not TypeScript... none of the variables have type declarations, which will always be an implicit error (`Parameter 'XXX' implicitly has an 'any' type.`). The content of `index.s` is certainly intended to be interpreted as JavaScript, so the use of the `?.` syntax is an error in this context.

achingbrain:
This is not a TypeScript project, you are quite right.  The [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator is available in JavaScript, but it is only supported by node 14+.

Are you using an older version?

dancingfrog:
Yup, v12.19.0

I guess it would help you out of Node version were a parameter of the issue
report template. So, basically js-ipfs supports ECMAScript 2020 at
*minimum*?

I think that should be boldly stated in some early and prominent section of
the documentation.

Peace & Love,
John ❤

PR #3922, opened 2021-10-19
fix(pubsub): multibase in pubsub http rpc

lidel:
This PR  aims to restore interop with go-ipfs by applying the same changes as in https://github.com/ipfs/go-ipfs/pull/8183 

TLDR is that we clean up and unify the API. 

## BREAKING CHANGE

We had to make breaking changes to `pubsub` commands sent over HTTP RPC  to fix data corruption caused by topic names and payload bytes that included `\n`.
More details in https://github.com/ipfs/go-ipfs/issues/7939 and https://github.com/ipfs/go-ipfs/pull/8183 

## TODO

- [x] ipfs-http-client (JS lib used by people to talk to `/api/v0` exposed by js-ipfs or go-ipfs)
- [x] ipfs-http-server (`/api/v0` exposed by js-ipfs)
- [x] <del>js-ipfs cli (file instead of "file per stin line"</del> descoped. Independent of the programmatic JS API.
- [x] update `pubsub` tests in https://github.com/ipfs/interop/pull/387 to test v1 and v2 circuits  to the degree possible


lidel:
@achingbrain i believe we want to merge this when we ship go-ipfs 0.11.0-rc1, so people can test the new wire format with `@next` version of http-client.
fysa the updated  interop tests that use code from this PR can be previewed in https://github.com/ipfs/interop/pull/387

lidel:
Interop from `ipfs/interop#feat/pubsub-require-multibase` branch (https://github.com/ipfs/interop/pull/387) is passing.


@achingbrain my understanding is that this is ready, but:
- We need to merge https://github.com/ipfs/interop/pull/387, release `ipfs/interop` and switch this PR to release version. 
- Some tests using  `dag` commands fail because of changes in go-ipfs  0.10.0 (and this PR pulls them in because we switched to 0.11.0-rc1). 
  - iiuc we need to land https://github.com/ipfs/js-ipfs/pull/3917 first, then rebase this PR to make CI green.

packages/ipfs-cli/test/pubsub.spec.js
```diff
@@ -101,19 +101,19 @@ describe('pubsub', () => {
     }
 
     it('should publish message', async () => {
-      const subName = 'sub-name'
-      const data = 'data'
+      const subName = 'sub-name-1'
+      const data = 'data\r\nfirst\nZażółć gęślą jaźń😇'
```
achingbrain, line 105: Yellow goose self?

lidel: Sounds better in Polish ;) (it is a valid sentence that includes all polish diacritics, good for testing useful utf8)

lidel:
@achingbrain updated to recent changes from master
and fixed some tests caused by update to go-ipfs 0.11 (eg.  sharding fix in https://github.com/ipfs/js-ipfs/pull/3922/commits/e9603751c8a6573f18bbc4cfa47273a3bdbb14bc).  

This seems to narrow down the remaining CI issues to:
- [ ] a bug in [ Interface tests test:interface:http-go browser](https://github.com/ipfs/js-ipfs/runs/4448326052?check_suite_focus=true#step:6:948):
   ```
   ipfs:        .files.rm
   ipfs:          should not remove not found file/dir, expect error:
   ipfs:      AssertionError: expected promise to be rejected but it was fulfilled with undefined
  ```
- [ ] a bug in [Test example ipfs browser ipns publish ](https://github.com/ipfs/js-ipfs/runs/4448323323?check_suite_focus=true#step:6:351):
   ```
   1) tests/test.js:62:3 › http ipns publish: › should connect to IPFS node and publish a file via ipns 
   ```
- [ ] a bug in [Unit tests node 16 windows-latest ](https://github.com/ipfs/js-ipfs/runs/4448326360?check_suite_focus=true#step:6:16) 
   ```
   ipfs-http-client: 'aegir' is not recognized as an internal or external command,
   ``` 

Lmk if I should poke at those or are those already fixed elsewhere.

oed:
Is this the last outstanding issue before `js-ipfs-http-client` can be used with go-ipfs 0.11?

lidel:
@oed  correct. When this PR gets merged, you will be able to use `ipfs-http-client@next` (built from the main branch automatically) until the real release ships.

packages/ipfs-http-client/src/files/rm.js
```diff
@@ -20,7 +21,15 @@ export const createRm = configure(api => {
...
-    await res.text()
+    const body = await res.text()
+    // we don't expect text body to be ever present
+    // (if so, it means an error such as https://github.com/ipfs/go-ipfs/issues/8606)
+    if (body !== '') {
+      /** @type {Error} */
+      const error = new HTTP.HTTPError(res)
+      error.message = body
+      throw error
+    }
```
lidel, line 32: @achingbrain would this work around be acceptable?

achingbrain: :groan:

lidel:
Ok, so the only failing thing in CI were examples from https://github.com/ipfs-examples/js-ipfs-examples that use pubsub and ipns.

`PWDEBUG=console npm test` confirmed this is caused by http wire format change for pubsub over http rpc api: 
>  

I disabled those two example tests in (TODO: will push shortly). 

My understanding is that:
1. We need to merge this PR and release a new  ipfs-http-client  compatible with go-ipfs 0.11
2. Then, switch https://github.com/ipfs-examples/js-ipfs-examples to go-ipfs 0.11 and new ipfs-http-client that works with new pubsub multibase wire format
3. Finally, re-enable pubsub examples in the CI of js-ipfs 

@achingbrain  lmk if this approach is sensible. 
If so, there is nothing more to do here – this PR is ready for your final review.

achingbrain (approved):

PR #3943, opened 2021-11-19
docs: add CORS to getting started

### LINKED ISSUE #3597: HTTP client: CORS setup is two links deep from README
One of the first things you need to do to use the HTTP client is to setup your IPFS node with CORS. At the moment I have to click into "Read the docs" and _then_ "How to configure CORS for use with the http client" to access this information.

IMHO this is very important information that should not be hidden deep within the documentation, it should be one of the first things I read in the "Getting Started" section.

hacdias:
Fixes #3597.

I assumed that the request is to have a link near the top for the CORS configuration. If the idea is to rework where those docs are placed and also make it easier to find through the docs directory, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ We've come a long way, but this project is still in Alpha, lots of development i
 ## Getting started
 
 * Read the [docs](https://github.com/ipfs/js-ipfs/tree/master/docs)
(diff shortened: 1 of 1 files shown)


achingbrain (approved):

achingbrain:
Thanks!

PR #3950, opened 2021-11-24
fix(dag): replace custom dag walk with multiformats/traversal

rvagg:
Ref: https://github.com/multiformats/js-multiformats/pull/127

/cc @patrickwoodhead


packages/ipfs-core/src/components/dag/export.js
```diff
@@ -80,52 +81,30 @@ export function createExport ({ repo, preload, codecs }) {
...
+  return async (cid) => {
+    const bytes = await repo.blocks.get(cid, options)
+    const codec = await codecs.getCodec(cid.code)
+
+    let block
+    if (codec) {
+      block = createUnsafe({ bytes, cid, codec })
+    } else if (NO_LINKS_CODECS.includes(cid.code)) {
+      // an opaque block with that will yield no links(), which is all walk() cares about
+      block = createUnsafe({ bytes, cid, value: null })
```
rvagg, line 101: This is a bit of a hack; maybe OK for now.

I'd like to do something like go-ipld-prime's [`SkipMe`](https://pkg.go.dev/github.com/ipld/go-ipld-prime/traversal#SkipMe), but the equivalent isn't very JS—it's returned as an error from the load function and the traversal engine checks if the error is a `SkipMe` and then ignores that branch. I don't really want to be throwing errors that are not errors.

We _could_ return `null` in the case of a `SkipMe`, that would be more JSish and leave us with a return type of `Promise<Block<T>|null>` and give us the ability to throw proper errors still.

rvagg:
test failures unrelated

BigLep:
@rvagg : I know it's been a couple of months on this.  Is this still needed?  If so, can you get CI green and Alex will take a look.

Issue #3953, opened 2021-11-25
TypeError: isOnline returns the type 'Promise '  instead of type 'boolean'


jovijovi:
- **Version**:

ipfs@0.60.2

- **Platform**:

Linux 5.4.0-90-generic x86_64 x86_64 x86_64 GNU/Linux

- **Subsystem**:

ipfs-core-types@0.8.4

#### Severity:

Medium

#### Description:

```javascript
isOnline: () => boolean;
```

The function `isOnline` defined in `packages/ipfs-core-types/src/root.ts` returns `boolean`. But if you call `isOnline`, you will get a `Promise` return instead of a `boolean`.

#### Steps to reproduce the error:

```javascript

import {create} from 'ipfs-http-client'; // version 54.0.2

const client = create();

function Ping(): boolean {
	const rsp = client.isOnline();  // rsp is type 'Promise'
	console.log(rsp);
	return rsp;
}

```

SgtPooki:
@achingbrain I am still seeing the incorrect type in ipfs-core-types@0.10.2. Can we release this update?

achingbrain:
`ipfs-core-types@0.10.3` is out - please try that

BigLep:
Closing since the unblocking release went out over a month ago.

luna-fire-dev:
This issue re-appears in the latest version (ipfs-core-types@npm:0.11.1)

https://github.com/ipfs/js-ipfs/blob/7a7e091c5d7110542ca7ab6eca1c0c9abb19e54b/packages/ipfs-core-types/src/root.ts#L139

PR #3954, opened 2021-11-25
fix: isOnline returns the type 'Promise' (#3953)

jovijovi:
The function `isOnline` defined in `packages/ipfs-core-types/src/root.ts` returns the type `Promise<boolean>` .

### DIFF
--- a/packages/ipfs-core-types/src/root.ts
+++ b/packages/ipfs-core-types/src/root.ts
@@ -135,7 +135,7 @@ export interface API<OptionExtension = {}> {
    * Returns true if this IPFS node is online - that is, it's listening on network addresses
    * for incoming connections
    */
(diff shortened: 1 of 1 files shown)


achingbrain:
Thanks for submitting this.  It looks like the build is failing, likely because you updated the signature for `isOnline` to return a promise but not every implementing function returns one - they just need to be marked `async`.

jovijovi:
> Thanks for submitting this. It looks like the build is failing, likely because you updated the signature for `isOnline` to return a promise but not every implementing function returns one - they just need to be marked `async`.

If mark `async`, code analyzer will report `Redundant 'await' for a non-promise type`.
Although it doesn't look very graceful, it seems that there is no good way.
I'll close this PR, and think of another way.
Thx bro~

PR #3947, opened 2021-11-22
feat: dht client

achingbrain:
* Enables `libp2p-kad-dht` in client mode
* Updates types with new DHT events

Depends on:

 - [x] https://github.com/ipfs/js-ipfs-bitswap/pull/384
 - [x] https://github.com/libp2p/js-libp2p/pull/1009

BREAKING CHANGE: The DHT API has been refactored to return async iterators of query events

PR #3955, opened 2021-11-29
Switch from IPFS_REUSEPORT to LIBP2P_TCP_REUSEPORT

aschmahmann:


aschmahmann:
@achingbrain any idea why this PR is failing, is it just flakiness/pre-existing failures or something new here? The `LIBP2P_TCP_REUSEPORT` env var has been around in go-libp2p for a few years and they're now dropping support  for IPFS_REUSEPORT (https://github.com/libp2p/go-tcp-transport/pull/104)

achingbrain:
Looks like a regression was caused by https://github.com/libp2p/js-libp2p-crypto/pull/202

achingbrain:
Actually on closer inspection this is caused by multiple versions of `libp2p-crypto` in the dep tree.

#3947 brings it all into line, though it needs releases of:

- [ ] https://github.com/ChainSafe/js-libp2p-gossipsub/pull/173
- [ ] https://github.com/ChainSafe/js-libp2p-noise/pull/116

achingbrain (approved):

PR #3959, opened 2021-12-01
feat: Add support for dag-jose

oed:
Support for dag-jose was recently merged into go-ipfs https://github.com/ipfs/go-ipfs/pull/8569. This PR adds support for dag-jose in the `js-ipfs-http-client` so that the new codec in go-ipfs can be used from javascript environments.

A few notes:
- This is not super urgent as it's easy to plug in additional ipld codecs in javascript. More of a nice to have to be symmetric with go-ipfs
- Maybe the `ipfs-core` package should also support dag-jose?

### DIFF
--- a/packages/ipfs-http-client/package.json
+++ b/packages/ipfs-http-client/package.json
@@ -58,6 +58,7 @@
     "@ipld/dag-pb": "^2.1.3",
     "abort-controller": "^3.0.0",
     "any-signal": "^2.1.2",
+    "dag-jose": "^1.0.0",
     "debug": "^4.1.1",
     "err-code": "^3.0.1",
     "ipfs-core-types": "^0.8.4",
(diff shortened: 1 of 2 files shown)


achingbrain:
> it's easy to plug in additional ipld codecs in javascript

This is very much intentional - there's no need to add extra codecs here which increase the bundle size, install times, etc.

The idea is that if you need a special codec or there's a new release of one or whatever you can configure it yourself without needing to PR this project or wait for a release.

achingbrain:
..I say that, but of course `go-ipfs` is shipping with support for `dag-jose` by default so `js-ipfs` will need to do the same thing.

To merge this PR it needs:

1. support for `dag-jose` added to `ipfs-core` - https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/src/components/index.js#L290
2. The version of go-ipfs used in tests upgraded to one that has `dag-jose` support - https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/package.json#L87
3. tests added to the `ipfs.dag.get`/`ipfs.dag.put` suites in `interface-ipfs-core` that exercise the `dag-jose` support - https://github.com/ipfs/js-ipfs/tree/master/packages/interface-ipfs-core/src/dag
	- see the npm scripts for running interface tests - https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/package.json#L63-L66

oed:
Thanks @achingbrain will prioritize this!

oed:
Replaced by #3989, closing.

PR #3962, opened 2021-12-03
chore: re-enable next js test

achingbrain:
The fix to enhanced-resolve should have made it into a release now.


rvagg (approved):
🎉 
Aside: what's with these create-node private key flaky test failures happening regularly now? Do we understand the cause for those and have a path to addressing them?

achingbrain:
> what's with these create-node private key flaky test failures happening regularly now?

It's being caused by there being multiple versions of `libp2p-crypto` in the dep tree so `instanceof` is failing:

```
AssertionError: expected { Object (_key, _publicKey) } to be an instance of Ed25519PrivateKey
```

> Do we ... have a path to addressing them

I've been upgrading all of our deps, it should be largely fixed now, this build was just in the wrong place at the wrong time.

.github/workflows/test.yml
```diff
@@ -424,6 +424,7 @@ jobs:
       - name: Install Dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
+          npm install -g @mapbox/node-pre-gyp
```
achingbrain, line 427:
```suggestion
```

lidel:
Seem that this needs more work:
- bump next.js  after they update their version of webpack

PR #3972, opened 2021-12-12
chore: fix monitoring

achingbrain:
Allow GET requests for prom stats and return correct response type


achingbrain:
CI is broken due to `@noble/secp256k1` having bad `package.json` config.  Fix in https://github.com/paulmillr/noble-secp256k1/pull/37

PR #3973, opened 2021-12-13
fix: use peer store for id

achingbrain:
Instead of using all the various books, just get the peer from the libp2p peer store, trying to fetch the public key from the dht if it's missing.


achingbrain:
CI is broken due to `@noble/secp256k1` having bad `package.json` config.  Fix in https://github.com/paulmillr/noble-secp256k1/pull/37

PR #3980, opened 2021-12-15
fix: always close writer so iterator throws on error

alanshaw:
When `traverseWrite` throws (abort signal for example), the writer is never closed and the iterator never ends.


achingbrain (approved):

rvagg (commented):
belated 👌, good catch

PR #3984, opened 2021-12-16
docs: correct the ipfs.stats.bw API options object

SgtPooki:
This update reflects exactly what is displayed on https://docs.ipfs.io/reference/http/api/#api-v0-stats-bw

Related to #3985


lidel:
Those two params are not in HTTP RPC API, but they ARE part of JS CORE API, so need to stay :)
General takeaway here we need to figure out a way for automated docs generation, so we no longer maintain these JS API docs by hand. (We already have that for HTTP RPC)

PR #4013, opened 2022-01-12
fix: dag.import description in core docs

alanshaw:


rvagg (approved):
you could `s/file/format` while you're in there too if you feel inclined, trying to banish "file" from "CAR" is a bit of an exercise since they most commonly map to actual files, but they're supposed to be more of a transport or container than just a file format.

achingbrain (approved):

PR #4015, opened 2022-01-15
fix: remove abort-controller deps

achingbrain:
`AbortController` and `AbortSignal` are global in all supported runtimes so we don't need the polyfill any more.


achingbrain:
Nb. will need bitswap and libp2p releases before it's completely gone from the stack.

packages/interface-ipfs-core/package.json
```diff
@@ -50,7 +50,7 @@
     "clean": "rimraf ./dist",
     "build": "aegir build && copyfiles ./test/fixtures/**/*, ./dist/cjs && copyfiles ./test/fixtures/**/*, ./dist/esm",
     "lint": "aegir ts -p check && aegir lint",
-    "dep-check": "aegir dep-check -i abort-controller -i ipfs-core-types -i rimraf -i copyfiles"
+    "dep-check": "aegir dep-check -i -i ipfs-core-types -i rimraf -i copyfiles"
```
clarkjoao, line 53: remove double '-i'

"dep-check": "aegir dep-check -i ipfs-core-types -i rimraf -i copyfiles"

achingbrain:
Test failures are unrelated.

PR #4061, opened 2022-03-03
fix: BWOptions.interval accepts number|string

### LINKED ISSUE #3985: ipfs-core-types@0.9.0 - Out of date types for BWOptions
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `ipfs-core-types@0.9.0` for the project I'm working on.

A number is no longer valid for the `interval` property of `BWOptions`. In order to successfully call the `ipfs.stats.bw` endpoint with TypeScript, you must either `//@ts-ignore` or avoid passing `interval`.

See https://docs.ipfs.io/reference/http/api/#api-v0-stats-bw
 
Here is the diff that solved my problem:

```diff
diff --git a/node_modules/ipfs-core-types/src/stats/index.ts b/node_modules/ipfs-core-types/src/stats/index.ts
index 43e5f24..30dbf86 100644
--- a/node_modules/ipfs-core-types/src/stats/index.ts
+++ b/node_modules/ipfs-core-types/src/stats/index.ts
@@ -16,7 +16,7 @@ export interface BWOptions extends AbortOptions {
   peer?: string
   proto?: string
   poll?: boolean
-  interval?: number
+  interval?: string
 }
 
 export interface BWResult {
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

SgtPooki:
fix #3985


SgtPooki:
@lidel these tests seem to be failing unrelated to my changes. However, I couldn't get `npm install` to succeed locally; I assumed due to M1 issues 😬

achingbrain:
I have an M1, things install fine for me - what's the error message you are seeing on install?

The build is failing because during the esbuild step it's hitting [this line](https://github.com/ipfs/aegir/blob/master/utils/resolve.js#L3) and failing to resolve the node 'path' module.

This is likely because something in the dep tree had it previously and now does not which means it only worked previously by coincidence.  Oops.

The quick fix is either to add the [path](https://www.npmjs.com/package/path) polyfill to aegir or to add `"path": false` to the [browser overrides](https://github.com/ipfs/aegir/blob/master/package.json#L22-L25), but the better fix might be to work out why the `resolve` module is getting referenced during the browser build in the first place since it should only really be loaded from node.

BigLep:
@SgtPooki : can you please respond with next steps so we can close this or get it across the line?

achingbrain (approved):

PR #4076, opened 2022-04-04
fix: exclude fs from bundle

achingbrain:
We used to have 'fs' in our dep tree via ora via mock-ipfs-pinning-service but that seems to have changed.

We only want to import 'fs' during node tests but esbuild tries to resolve it, so add the dummy fs package as a dev dep to the affected packages.


achingbrain:
Needs: https://github.com/ipfs-shipyard/js-mock-ipfs-pinning-service/pull/21

PR #4085, opened 2022-04-20
fix: update car dependency for CARv2 read support

rvagg:
4.x adds basic CARv2 read support, so you could `ipfs dag import` a CARv2 with this; also with https://explore.ipld.io/ getting CAR support, this would allow importing and inspection of a CARv2 there.

The breakage was in the return type definition of `CarReader#header()` which I don't believe is touched in js-ipfs, which only cares about `getRoots()` in the header which remains the same.

### DIFF
--- a/packages/interface-ipfs-core/package.json
+++ b/packages/interface-ipfs-core/package.json
@@ -60,7 +60,7 @@
   ],
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


achingbrain (approved):

achingbrain:
Thanks!

PR #4087, opened 2022-04-21
Improper Input validation

bhaskarvilles:
Setting setHeader header to "*" in setHeader might be too permissive. This allows malicious code on other domains to communicate with the application, which is a security risk.


achingbrain:
The code you have updated is only run during testing, it's not published as part of IPFS.  Are you seeing something different?

bhaskarvilles:
@achingbrain I don't see any workflows running in actions and it has to be approved by maintainers, this is my first contribution here. i've tested it locally on my computer and i'ts working.

bhaskarvilles:
> The code you have updated is only run during testing, it's not published as part of IPFS.  Are you seeing something different?

Any update ??

achingbrain:
The code you have changed is only executed while running unit tests, so while you are right it’s very permissive, it’s unlikely to be exploited.

BigLep:
Closing since this is just test code and this isn't the focus of the test.

PR #4124, opened 2022-06-03
docs: update block api documentation

### LINKED ISSUE #3913: `ipfs.block` API is out of sync with the implementation
Here is what I've noticed:
* `ipfs.block.get`
  * docs say it returns a `Block`, but it returns a `Uint8Array`
  * docs are missing the `preload` option
* `ipfs.block.put`
  * docs say it can also take a `Block` as a parameter, but it can only be a `Uint8Array`
  * docs say it can take a `cid` as an option, but the implementation ignores such a parameter
  * docs say it returns a `Block`, but it's just a `Uint8Array`

hacdias:
Closes #3913.

I have some inline notes and questions.

### DIFF
--- a/docs/core-api/BLOCK.md
+++ b/docs/core-api/BLOCK.md
@@ -39,6 +39,7 @@ An optional object which may have the following keys:
 | ---- | ---- | ------- | ----------- |
 | timeout | `Number` | `undefined` | A timeout in ms |
 | signal | [AbortSignal][] | `undefined` |  Can be used to cancel any long running requests started as a result of this call |
+| preload | `boolean` | `false` |  Whether to preload all blocks created during this operation |
 
 ### Returns
 
@@ -63,7 +64,7 @@ A great source of [examples][] can be found in the tests for this API.
 
 | Name | Type | Description |
 | ---- | ---- | ----------- |
-| block | A `Uint8Array` or [Block][] instance | The block or data to store |
+| block | `Uint8Array` | The block of data to store |
 
 ### Options
 
@@ -73,19 +74,18 @@ An optional object which may have the following keys:
 | ---- | ---- | ------- | ----------- |
 | format | `String` | `'dag-pb'` | The codec to use to create the CID |
 | mhtype | `String` | `sha2-256` | The hashing algorithm to use to create the CID |
-| mhlen | `Number` | | |
+| mhlen | `Number` | `undefined` | The hash length (only relevant for `go-ipfs`) |
 | version | `Number` | `0` |  The version to use to create the CID |
 | pin | `boolean` | `false` |  If true, pin added blocks recursively |
 | timeout | `Number` | `undefined` | A timeout in ms |
 | signal | [AbortSignal][] | `undefined` | Can be used to cancel any long running requests started as a result of this call |
-
-**Note:** If you pass a [`Block`][block] instance as the block parameter, you don't need to pass options, as the block instance will carry the CID value as a property.
+| preload | `boolean` | `false` |  Whether to preload all blocks created during this operation |
 
 ### Returns
 
 | Type | Description |
 | -------- | -------- |
-| `Promise<Block>` | A [Block][block] type object, containing both the data and the hash of the block |
+| `Promise<CID>` | A [CID][CID] type object containing the hash of the block |
 
 ### Example
 
@@ -95,21 +95,6 @@ const buf = new TextEncoder().encode('a serialized object')
 const decoder = new TextDecoder()
 
 const block = await ipfs.block.put(buf)
-
-console.log(decoder.decode(block.data))
-// Logs:
-// a serialized object
-console.log(block.cid.toString())
-// Logs:
-// the CID of the object
-
-// With custom format and hashtype through CID
-import { CID } from 'multiformats/cid'
-import * as dagPB from '@ipld/dag-pb'
-const buf = new TextEncoder().encode('another serialized object')
-const cid = CID.createV1(dagPB.code, multihash)
-
-const block = await ipfs.block.put(blob, cid)
 
 console.log(decoder.decode(block.data))
 // Logs:
@@ -191,6 +176,7 @@ An optional object which may have the following keys:
 | ---- | ---- | ------- | ----------- |
 | timeout | `Number` | `undefined` | A timeout in ms |
 | signal | [AbortSignal][] | `undefined` | Can be used to cancel any long running requests started as a result of this call |
+| preload | `boolean` | `false` |  Whether to preload all blocks created during this operation |
 
 ### Returns
 
--- a/packages/ipfs-core-types/src/block/index.ts
+++ b/packages/ipfs-core-types/src/block/index.ts
@@ -32,19 +32,6 @@ export interface API<OptionExtension = {}> {
    * console.log(block.cid.toString())
    * // Logs:
    * // the CID of the object
-   *
-   * // With custom format and hashtype through CID
-   * import { CID } from 'multiformats/cid'
-   * const another = encoder.encode('another serialized object')
-   * const cid = CID.createV1(dagPB.code, multihash)
-   * const block = await ipfs.block.put(another, cid)
-   * console.log(decoder.decode(block.data))
-   *
-   * // Logs:
-   * // a serialized object
-   * console.log(block.cid.toString())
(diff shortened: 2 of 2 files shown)


docs/core-api/BLOCK.md
```diff
@@ -39,6 +39,7 @@ An optional object which may have the following keys:
 | ---- | ---- | ------- | ----------- |
 | timeout | `Number` | `undefined` | A timeout in ms |
 | signal | [AbortSignal][] | `undefined` |  Can be used to cancel any long running requests started as a result of this call |
+| preload | `boolean` | `false` |  Whether to preload all blocks created during this operation |
```
hacdias, line 42: https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-core-types/src/block/index.ts#L14

docs/core-api/BLOCK.md
```diff
@@ -63,7 +64,7 @@ A great source of [examples][] can be found in the tests for this API.
 
 | Name | Type | Description |
 | ---- | ---- | ----------- |
-| block | A `Uint8Array` or [Block][] instance | The block or data to store |
+| block | `Uint8Array` | The block of data to store |
```
hacdias, line 67: https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-core-types/src/block/index.ts#L14

docs/core-api/BLOCK.md
```diff
@@ -73,19 +74,18 @@ An optional object which may have the following keys:
 | ---- | ---- | ------- | ----------- |
 | format | `String` | `'dag-pb'` | The codec to use to create the CID |
 | mhtype | `String` | `sha2-256` | The hashing algorithm to use to create the CID |
-| mhlen | `Number` | | |
+| mhlen | `Number` | `undefined` | The hash length (only relevant for `go-ipfs`) |
```
hacdias, line 77: https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core-types/src/block/index.ts does not include this option. However, the `go-ipfs` CLI seems to support this option and found this mentions in this repository:

https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-cli/src/commands/block/put.js#L34-L37

https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-http-server/src/api/resources/block.js#L111

- Is this option relevant? Should it be removed?
- Is it supported by `js-ipfs` or only `go-ipfs`?

docs/core-api/BLOCK.md
```diff
@@ -73,19 +74,18 @@ An optional object which may have the following keys:
...
 | version | `Number` | `0` |  The version to use to create the CID |
 | pin | `boolean` | `false` |  If true, pin added blocks recursively |
 | timeout | `Number` | `undefined` | A timeout in ms |
 | signal | [AbortSignal][] | `undefined` | Can be used to cancel any long running requests started as a result of this call |
-
-**Note:** If you pass a [`Block`][block] instance as the block parameter, you don't need to pass options, as the block instance will carry the CID value as a property.
+| preload | `boolean` | `false` |  Whether to preload all blocks created during this operation |
```
hacdias, line 82: https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-core-types/src/block/index.ts#L84

docs/core-api/BLOCK.md
```diff
@@ -73,19 +74,18 @@ An optional object which may have the following keys:
...
 
 ### Returns
 
 | Type | Description |
 | -------- | -------- |
-| `Promise<Block>` | A [Block][block] type object, containing both the data and the hash of the block |
+| `Promise<CID>` | A [CID][CID] type object containing the hash of the block |
```
hacdias, line 88: https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-core-types/src/block/index.ts#L50

docs/core-api/BLOCK.md
```diff
@@ -96,21 +96,6 @@ const decoder = new TextDecoder()
...
-// Logs:
-// the CID of the object
-
-// With custom format and hashtype through CID
-import { CID } from 'multiformats/cid'
-import * as dagPB from '@ipld/dag-pb'
-const buf = new TextEncoder().encode('another serialized object')
-const cid = CID.createV1(dagPB.code, multihash)
-
-const block = await ipfs.block.put(blob, cid)
```
hacdias, line 112: Couldn't find evidence of this being possible. Both the TypeScript typings and the implementation do not support a `CID` as parameter.

https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-core-types/src/block/index.ts#L50

https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-core/src/components/block/put.js#L20

docs/core-api/BLOCK.md
```diff
@@ -191,6 +176,7 @@ An optional object which may have the following keys:
 | ---- | ---- | ------- | ----------- |
 | timeout | `Number` | `undefined` | A timeout in ms |
 | signal | [AbortSignal][] | `undefined` | Can be used to cancel any long running requests started as a result of this call |
+| preload | `boolean` | `false` |  Whether to preload all blocks created during this operation |
```
hacdias, line 179: https://github.com/ipfs/js-ipfs/blob/548cfb215e8afa8dca64d3c38fa13feecbe577e2/packages/ipfs-core-types/src/block/index.ts#L81

packages/ipfs-core-types/src/block/index.ts
```diff
@@ -32,19 +32,6 @@ export interface API<OptionExtension = {}> {
...
-   * const another = encoder.encode('another serialized object')
-   * const cid = CID.createV1(dagPB.code, multihash)
-   * const block = await ipfs.block.put(another, cid)
-   * console.log(decoder.decode(block.data))
-   *
-   * // Logs:
-   * // a serialized object
-   * console.log(block.cid.toString())
-   * // Logs:
-   * // the CID of the object
```
hacdias, line 47: Same as https://github.com/ipfs/js-ipfs/pull/4124#discussion_r888768519

docs/core-api/BLOCK.md
achingbrain: It's not supported by `js-ipfs` and I'm not clear on what it even does for `go-ipfs`.  It [defaults to `-1`](https://docs.ipfs.io/reference/http/api/#api-v0-block-put) which means.. 🤷 

If you could find out the what and the why maybe we could even remove it entirely.

docs/core-api/BLOCK.md
achingbrain: `CID` used to be supported as a parameter, but it's not safe as you could store the wrong block for a CID so it was removed.

achingbrain (changes requested):
LGTM, have responded inline to the queries, we can merge once they are resolved.

docs/core-api/BLOCK.md
hacdias: Well, it seems to be literally the [multihash length](https://github.com/ipfs/go-ipfs/search?q=mhlen), but doesn't seem to be very useful looking at the very limited search results. I tried figuring out if this was useful when opening this PR, but couldn't find anything that explained why it may be needed.

lidel: I had to refactor relevant go-ipfs code around this recently `mhlen`  gets passed deep into go-multihash code responsible for digert (`Sum`) calculation: https://github.com/multiformats/go-multihash/blob/75ae3688857d036ea15947be9df68d3172b19470/sum.go


tldr: 
- `-1` is the implicit default which means "set  length based on hashing function output"
  - same applies to any value `<=0`
- setting custom `mhlen`  allows for truncating  longer digests produces by alternative hash functions 
- setting it to longer than the default output of specific hash function will produce error

achingbrain (approved):

PR #4151, opened 2022-06-29
deps!: update to libp2p@0.38.x

achingbrain:
Upgrades all deps to latest versions including libp2p


kenshyx:
Hey, it would have been nice to explain why all the contributors were removed from package.json

PR #4172, opened 2022-07-30
add secp256k1 key type

inverted-capital:
In the [docs](https://github.com/inverted-capital/js-ipfs/blob/master/docs/MODULE.md#optionsinit) we are promised the secp256k1 key type being available, but it is not.  Is there some error with doing this ?

### DIFF
--- a/packages/ipfs-core-types/src/root.ts
+++ b/packages/ipfs-core-types/src/root.ts
@@ -260,7 +260,7 @@ export interface CatOptions extends AbortOptions, PreloadOptions {
 export interface GetOptions extends AbortOptions, PreloadOptions {
   archive?: boolean
   compress?: boolean
-  compressionLevel?: -1 | 0 | 1 | 2 | 3 | 4 | 5 | 6| 7 | 8| 9
+  compressionLevel?: -1 | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
 }
 
 export interface ListOptions extends AbortOptions, PreloadOptions {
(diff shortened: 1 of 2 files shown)


lidel:
It happens to be listed there because IPFS nodes reuse libp2p keystore, and  libp2p's [specs for peerid](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types) list these key types:

```
        RSA = 0;
	Ed25519 = 1;
	Secp256k1 = 2;
	ECDSA = 3;
```

afaik no IPFS implementation uses anything other than `RSA` and `Ed25519`, so allowing it here may open interoperability issues, especially in the context of [IPNS specificiation](https://github.com/ipfs/specs/blob/main/IPNS.md) which uses the same keystore, but does not explicitly list which key types should be supported by minimal  IPNS implementation. 


I'm marking this is blocked until we clarify the specs, at least IPNS one.

----

@inverted-capital mind elaborating why do you want to use secp256k1? is there a practical need (if so, it is useful for us to understand it, so we can adjust specs, if it is generic enough) or did you fill  it only for completeness?

inverted-capital:
Thanks for the review @lidel - we are using the private key to provide other forms of digital signing, and the scp256k1 key fills application specific needs for us.  We would like to reuse the IPFS key management systems, and we would like only one secret key in our application, so it is convenient if they are the same.  We think some of our preference for this algorithm overlap with what is useful to IPFS, namely:
1. audited and dependency free JS implementation in [`@noble/secp256k1`](https://www.npmjs.com/package/@noble/secp256k1)
2. Very fast signing performance - less than 0.5ms in our benchmarks
3. Short key and signature lengths making it easier on the eye when reading data that is signed in this fashion, or when reading the public keys.

So this pull request is not based on some feature of IPFS that we need, but our more general preference and trust the algorithm and implementation.

PR #4174, opened 2022-08-04
docs: switch links to .tech tld

lidel:
> Part of https://github.com/protocol/bifrost-infra/issues/2018

This PR updates links to use `.tech` TLD, avoiding redirect and potential issues with DNS sinkholing of `ipfs.io` 

(js-ipfs repo and README files published on NPM are how people discover a lot of our websites, including docs and forums)


docs/MIGRATION-TO-ASYNC-AWAIT.md
```diff
@@ -864,7 +864,7 @@ e.g.
 const IpfsHttpClient = require('ipfs-http-client')
 const ipfs = IpfsHttpClient()
 
-const files = await ipfs.addFromURL('https://ipfs.io/images/ipfs-logo.svg')
+const files = await ipfs.addFromURL('https://example.com/images/ipfs-logo.svg')
```
BigLep, line 867: I assume here and other places we are ok to use a URL for a resource that doesn't exist.  If so then the PR seems good to me?  Everything else checked out.

packages/ipfs-core-config/README.md
```diff
@@ -1,8 +1,7 @@
 # ipfs-core-utils <!-- omit in toc -->
 
-[![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](http://protocol.ai)
-[![](https://img.shields.io/badge/project-IPFS-blue.svg?style=flat-square)](http://ipfs.io/)
-[![](https://img.shields.io/badge/freenode-%23ipfs-blue.svg?style=flat-square)](http://webchat.freenode.net/?channels=%23ipfs)
+[![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](https://protocol.ai)
```
achingbrain, line 3: Could you PR https://github.com/ipfs/aegir/blob/master/src/check-project/readme/header.js#L12-L18 with these changes as well please?

achingbrain (approved):

achingbrain:
Just needs conflicts sorting out

docs/MIGRATION-TO-ASYNC-AWAIT.md
achingbrain: IANA [reserves this domain](https://www.iana.org/domains/reserved) specifically for use in documentation, but it might be nicer if the user could copy/paste this sort of example and have it work.  What do you think @lidel?

lidel:
@achingbrain  resolved&rebased

packages/interface-ipfs-core/src/miscellaneous/dns.js
```diff
@@ -28,9 +28,9 @@ export function testDns (factory, options) {
 
     after(() => factory.clean())
 
-    it('should non-recursively resolve ipfs.io', async function () {
+    it('should non-recursively resolve ipfs.tech', async function () {
       try {
-        const res = await ipfs.dns('ipfs.io', { recursive: false })
+        const res = await ipfs.dns('ipfs.tech', { recursive: false })
```
achingbrain, line 33: This seems to have broken CI:

```
ipfs:   1 failing
ipfs:        .dns
ipfs:          should non-recursively resolve ipfs.tech:
ipfs:      AssertionError: expected '/ipfs/QmdcBdTSBLhS1DVVRXfFYHhGpSQaght…' to match /\/ipns\/.+$/
ipfs:       at Context.<anonymous> (file:///home/runner/work/js-ipfs/js-ipfs/packages/interface-ipfs-core/src/miscellaneous/dns.js:36:24)
ipfs:       at runMicrotasks (<anonymous>)
ipfs:       at processTicksAndRejections (node:internal/process/task_queues:96:5)
ipfs: Command failed with exit code 1: mocha test/interface-http-go.js --ui bdd --require source-map-support/register --timeout=60000 --bail
```

achingbrain (changes requested):

packages/interface-ipfs-core/src/miscellaneous/dns.js
lidel: That is because on `ipfs.tech` we no longer have a DNSLink that points to other DNSLink: we use CNAME pointing at Fleek instead, and get final `/ipfs/cid` directly, which saves us one DNS lookup.

@achingbrain  I've updated the tests to accept both `/ipfs/` and `/ipns/`: ad0ca96e33684b58cbe425dc8128a3da05d2bf57

docs/MIGRATION-TO-ASYNC-AWAIT.md
lidel: If we ever reorganize `ipfs.io` and change path of logo, the example will no longer work (happened the last time ipfs.io was revamped).

It is better to use `example.com` than "real" websites: 
communicates well it is not a "live" example.

packages/interface-ipfs-core/src/miscellaneous/dns.js
achingbrain: Is there another handy DNSLink we can use to test recursive lookups?

lidel: No, nothing stable enough to include in CI/tests.

I think at some point Kub oswitched all resolve methods to be recursive by default?

Also, we've deprecated `ipfs dns ipfs.tech` in Kubo, 
and now simply point people at `ipfs resolve /ipns/ipfs.tech` (recursive by default).

achingbrain:
Closing as js-ipfs is deprecated and this is no longer necessary.

PR #4178, opened 2022-08-05
deps: upgrade @multiformats/multiaddr to 10.4.0

lidel:
Shipping fix from https://github.com/multiformats/js-multiaddr/pull/253 is required to fix https://github.com/libp2p/go-libp2p-relay-daemon/issues/18 and unblock @thediscordian to update https://docs.ipfs.tech/how-to/create-simple-chat-app/ tutorial 

Filling this PR to see if it is an easy bump that can be shipped as a patch, or more work needs to happen.

## TODO

- [ ] re-run after CID is fixed in  `master`


lidel:
@achingbrain CI fails due to an unrelated reason (same in master: https://github.com/ipfs/js-ipfs/commits/master)

PR #4187, opened 2022-08-15
Support for TypeScript + ESM

pleerock:
We need typings inside "export" to support TypeScript with ESM enabled. See [TypeScript docs](https://www.typescriptlang.org/docs/handbook/esm-node.html).

### DIFF
--- a/packages/ipfs-http-client/package.json
+++ b/packages/ipfs-http-client/package.json
@@ -17,7 +17,8 @@
   ],
   "exports": {
     ".": {
-      "import": "./src/index.js"
+      "import": "./src/index.js",
+      "types": "./dist/src/index.d.ts"
     }
   },
   "eslintConfig": {


packages/ipfs-http-client/package.json
```diff
@@ -17,7 +17,8 @@
   ],
   "exports": {
     ".": {
-      "import": "./src/index.js"
+      "import": "./src/index.js",
+      "types": "./dist/src/index.d.ts"
```
achingbrain, line 21: The ts docs say the `"types"` key must appear first in the list - see the code example, right at the bottom of https://www.typescriptlang.org/docs/handbook/esm-node.html

Also, can you please add the `"types"` key to every exports map entry in every package in this repo rather than just one of them?

achingbrain (changes requested):

achingbrain:
This should be fixed in the latest RC - could you please try with:

```console
$ npm i ipfs-http-client@next
```

BigLep:
@pleerock : can you confirm if this is resolved now?  Thanks.

pleerock:
Yes, `next` works because it contains the changes I proposed in this PR. Thank you.

PR #4191, opened 2022-08-24
do not store node private key in plaintext

inverted-capital:
Considerable effort has been put in to integrating a password based key derivation scheme in js-ipfs, and yet the primary encryption keys of the node are not stored there.  The private key for the node can be read in plaintext from the config slice of the repo.

### DIFF
--- a/packages/ipfs-core/test/config.spec.js
+++ b/packages/ipfs-core/test/config.spec.js
@@ -6,6 +6,9 @@ import { Multiaddr } from '@multiformats/multiaddr'
 import { isBrowser, isWebWorker } from 'ipfs-utils/src/env.js'
 import createNode from './utils/create-node.js'
 import createConfig from 'ipfs-core-config/config'
+import { fromString as uint8ArrayFromString } from 'uint8arrays/from-string'
+import { unmarshalPrivateKey } from '@libp2p/crypto/keys'
+import { peerIdFromKeys } from '@libp2p/peer-id'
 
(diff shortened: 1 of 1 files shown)


BigLep:
@inverted-capital: You're right.  Can you please open an issue in https://github.com/ipfs/js-ipfs-repo repo on how we can better store keys?  There are various usability things we need to decide so the implementation can be picked up.

inverted-capital:
@BigLep thanks for writing back - I do not believe the correct repo for a fix is the https://github.com/ipfs/js-ipfs-repo repo because the keys are stored in config as part of the js-ipfs way of doing things.  js-ipfs-repo is providing the config slice as a place to store an arbitrary POJO, and it is offering the keychain slice as a way of storing password based key derived encrypted data - it forces no format or usage restrictions beyond that.

I think the fix is to modify the way js-ipfs provisions itself to store the 'self' key in the keychain slice of the repo, encrypted with a blank password, and then if the user or programmer wants to set a password, this triggers a key rotation which is already built in to the repo implementation.

Shall I open an issue in js-ipfs that proposes this fix ?

lidel:
Triage notes:
- this is a sensible feature, but it should be opt-in via js-ipfs config 
- a proper place to continue is:
  - create issue in https://github.com/ipfs/js-ipfs describing proposed design (config, API changes in relevant libraries)
  - get approval from maintainer
  - open PR 
  
 I am closing this one @inverted-capital, feel free to link js-ipfs issue when you create it.

PR #4207, opened 2022-09-12
fix!: require IPNS V2 signatures

lidel:
This PR  makes js-ipfs ignore V1 signatures when validating IPNS records, and always require V2.

See https://github.com/ipfs/js-ipns/pull/180 for low level details (needs that to be released before merging this)

Closes https://github.com/ipfs/js-ipfs/issues/4197

BREAKING  CHANGE: IPNS Records that do not have V2 but have V1 signature will no longer pass validation, even if V1 is correct. V2 is mandatory to pass validation. See "Record validation" in https://github.com/ipfs/specs/pull/319 for details.


lidel:
@achingbrain switched this PR to js-ipns v3.0.0, ready for your final review. We are planning to release https://github.com/ipfs/kubo/pull/9265 in Kubo 0.16 RC  this/next week, so would be good to ship JS version in the next JS-IPFS release too.

achingbrain (approved):

Issue #4233, opened 2022-10-14
feat: re-export component APIs


DougAnderson444:
https://github.com/ipfs/js-ipfs/blob/dfc43d4e9be67fdf25553677f469379d966ff806/packages/ipfs-core/src/components/index.js#L34

Could we re-export the component APIs (such as `DagAPI`) so they can be extended in other packages?

```js
// ipfs-core/src/components/index.js
export { DagAPI, PinAPI, BlockAPI } 
```

and 

```js
"exports": {
    "./components": {
      "types": "./src/components/index.d.ts",
      "import": "./src/components/index.js"
    }
  },
```

Then I can do

```js
// myClass.js
import { DagAPI } from 'ipfs-core/components'

class CustomDagRepo extends DagAPI { 
}
```

I've been using a deep reference for this already, so I thought I'd request this small change.

SgtPooki:
Hello @DougAnderson444,

We wanted to let you know that the issue you reported will not be resolved in js-ipfs as we are in the process of deprecating it. However, we are pleased to inform you that the issue should be resolved in our new library, **[https://github.com/ipfs/helia](https://github.com/ipfs/helia)**. We highly recommend using helia instead, as it should provide a solution to the problem you encountered.

For further assistance or if you have any questions, please refer to the helia repository at **[https://github.com/ipfs/helia](https://github.com/ipfs/helia)**. We have a [migration guide](https://github.com/ipfs/helia/wiki/Migrating-from-js-IPFS) available.

Thank you for your understanding and for considering helia as an alternative!

Best regards,
@ipfs/helia-dev

PR #4249, opened 2022-11-16
feat: re-export components

DougAnderson444:
### PR Reason:
Addresses feat request #4233.

### Background:
The IPFS Repo takes a long time to load in the browser. So I am loading just the bare essentials first so data can be loaded and saved locally, then adding the Storage and libp2p networking after the fact.

But this means I need deep access to these exports, but they aren't re-exported by the package. This simple PR addresses that.

For your consideration.


SgtPooki:
Hello @DougAnderson444,

We wanted to let you know that the issue you reported will not be resolved in js-ipfs as we are in the process of deprecating it. However, we are pleased to inform you that the issue should be resolved in our new library, **[https://github.com/ipfs/helia](https://github.com/ipfs/helia)**. We highly recommend using helia instead, as it should provide a solution to the problem you encountered.

For further assistance or if you have any questions, please refer to the helia repository at **[https://github.com/ipfs/helia](https://github.com/ipfs/helia)**. We have a [migration guide](https://github.com/ipfs/helia/wiki/Migrating-from-js-IPFS) available.

Thank you for your understanding and for considering helia as an alternative!

Best regards,
@ipfs/helia-dev

Issue #4256, opened 2022-11-22
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in node_modules/ipfs/package.json


achingbrain:
`js-ipfs@0.63` made the jump to become ESM-only.  If you are seeing the error message above, it's because you are trying to load ESM code using `require` from a CJS environment.

Instead you need to either convert your code to ESM and use `import`, or use the [dynamic `import` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) if you cannot convert to ESM.

Please see the [upgrade guide](https://github.com/ipfs/js-ipfs/blob/master/docs/upgrading/v0.62-v0.63.md#esm) for more information.

Note if you are authoring TypeScript and using `import` you also need to ensure that your `tsc` settings result in ESM and not CJS.

Please see the [TypeScript section of the upgrade guide](https://github.com/ipfs/js-ipfs/blob/master/docs/upgrading/v0.62-v0.63.md#typescript-and-esm) for more information.

SgtPooki:
js-ipfs is being deprecated in favor of [Helia](https://github.com/ipfs/helia). You can https://github.com/ipfs/js-ipfs/issues/4336 and [read the migration guide](https://github.com/ipfs/helia/wiki/%F0%9F%9A%9B-Migrating-from-js-IPFS).

Please feel to reopen with any comments by 2023-06-02. We will do a final pass on reopened issues afterward (see https://github.com/ipfs/js-ipfs/issues/4336).

PR #4265, opened 2022-12-12
fix: ipfs-http-client exports

inglkruiz:
Fix implemented for making it work when using NodeJs module.createRequire. FYI > https://nodejs.org/api/module.html#modulecreaterequirefilename

Signed-off-by: inglkruiz <12603425+inglkruiz@users.noreply.github.com>

### DIFF
--- a/packages/ipfs-http-client/package.json
+++ b/packages/ipfs-http-client/package.json
@@ -45,7 +45,8 @@
   "exports": {
     ".": {
       "types": "./dist/src/index.d.ts",
-      "import": "./src/index.js"
+      "import": "./src/index.js",
+      "require": "./src/index.js"
     }
   },
   "eslintConfig": {


inglkruiz:
Some context about this change. I'm using NodeRed for wiring a data flow and I'm using this package, recently I made a PR for fixing an issue I found after one of my flows stopped working https://github.com/node-red/node-red/pull/3984/files 

I did read about ESM and I understand the extension for `require` should be `.cjs`, although I tested this locally changing the `package.json` at `node_modules/ipfs-http-client` and the local instance stopped throwing the following error

image

Also I understand the package has been moved to be **ESM only** but many of us I believe cannot migrate right now to ESM, multiple reasons could be named, but mine is that it would take a huge effort to upgrade NodeRed to ESM, moreover, I'm not a contributor, and they have implemented a fix to overcome this > https://github.com/node-red/node-red/pull/3645

Now, I also have the same issue these packages
- https://github.com/achingbrain/it/tree/master/packages/it-all#readme
- https://github.com/achingbrain/uint8arrays#readme


Lastly, I have to admit that I'm not sure if this is the way to fix it but I can say my flows in NodeRed stopped throwing errors.

inglkruiz:
@achingbrain Hi, could you or any of the contributors review this PR? Feedback is very welcome. Also, if this does not align with the repo's vision let me know how ppl not using ESM could start using the newesgt version?

inglkruiz:
@achingbrain Hi, could you or any of the contributors review this PR? Feedback is very welcome. Also, if this does not align with the repo's vision let me know how ppl not using ESM could start using the newest version?

achingbrain:
Thanks for submitting this PR but it's not going to solve the problem - by adding `require` to the exports map and pointing it at an ESM file, it will cause use of `require` to attempt to load the ESM file.

`require` does not work with ESM files, only CJS so this will fail with `module not allowed` as you have seen in your own local testing.

If you need to import ESM from CJS you need to use the [dynamic import function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import).

See: https://github.com/ipfs/js-ipfs/issues/4256

inglkruiz:
Looking at the code in NodeRed they indeed use the dynamic import function, see > https://github.com/node-red/node-red/pull/3645/files#diff-785d7569067dc564a5a97de5578752c3ad37a42a213b55dd1956081c32294f68R147-R151 

The errors I would say is when the call:
```ts
    // To handle both CJS and ESM we need to resolve the module to the
    // specific file that is loaded when the module is required/imported
    // As this won't be on the natural module search path, we use createRequire
    // to access the module
    const modulePath = createRequire(moduleDir)
```
I debugged the code and there is where it breaks. Would you need a repo for reproducing it? ThIs package and NodeRed is very important to us (at the company where I work) since it allows to integrate different technologies using a data flow. Our users like it and it would be sad if we have to stay in one of js-ipfs old version.

achingbrain:
I don't think that code will work as intended as it's using `require.resolve` which only works with modules that have a 'main' field in their package.json and not the newer [exports](https://nodejs.org/api/packages.html#exports) field.

Does it work if you change it to just:

```js
return import(module);
```

PR #4275, opened 2022-12-28
improve dockerfiles

sevenrats:
improve dockerfiles by removing unnecessary layers and greatly reducing size. 
remove sed line that causes insecure exposure -- users must use proper configs.
adds a proper init, catatonit, selected for modern signal handling.

### DIFF
--- a/Dockerfile.latest
+++ b/Dockerfile.latest
@@ -25,4 +25,4 @@ EXPOSE 9090
 
 COPY entrypoint.sh /entrypoint.sh
 
-CMD ["catatonit", "/entrypoint.sh"]
+CMD ["catatonit", "-g", "/entrypoint.sh"]
--- a/Dockerfile.latest
+++ b/Dockerfile.latest
@@ -1,4 +1,4 @@
-FROM node:16-alpine
+FROM alpine:3.17
 
 ENV IPFS_VERSION=latest
 ENV IPFS_MONITORING=1
--- a/Dockerfile.latest
+++ b/Dockerfile.latest
@@ -5,26 +5,24 @@ ENV IPFS_MONITORING=1
 ENV IPFS_PATH=/root/.jsipfs
 ENV BUILD_DEPS='libnspr4 libnspr4-dev libnss3'
 
-RUN apk add --no-cache git python3 build-base
-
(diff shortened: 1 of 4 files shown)


Dockerfile.latest
```diff
@@ -5,26 +5,24 @@ ENV IPFS_MONITORING=1
...
-
-# Configure jsipfs
-RUN jsipfs init
-
-RUN jsipfs version
-
-# Allow connections from any host
-RUN sed -i.bak "s/127.0.0.1/0.0.0.0/g" $IPFS_PATH/config
+RUN \
+    apk add --no-cache catatonit bash procps nodejs npm \
```
achingbrain, line 9: The `node:16-alpine` container comes with node & npm so there shouldn't be a need to install more copies.

That said, this should be running on LTS node and not node 16 now.

sevenrats: It's not based on node:16-alpine anymore because then you can never clean node and npm out of the final image.  That's why the last one was almost a gigabyte.  You cannot "shrink" or "remove", in later layers, things that were added in previous layers. the only way to not have npm in the target image is to use it as a builder layer and copy out, OR do as I have done here, add and remove it in a single layer. This is the correct approach because we want to keep nodejs, otherwise we would use node:alpine as a builder and just copy out our targets.
I will check on the versions this pulls.  maybe we can rebase to alpine 17 and have the lts if this doesn't already.

sevenrats: edited my comment to reflect that we only want to scrub out npm.  obvs we need to keep nodejs in there.

sevenrats: oh lol i see what you mean now sorry.  it uses node 18.12.1 this way, which is lts.

sevenrats:
worth mentioning, probably, that this configuration removes the package manager, apk, so, it would be impractical to build containers out of this one.  I think it's appropriate to remove i if the container is not explicitly a base layer for other stuff, but it's up to you guys.

sevenrats:
contains a bug where will always init because $IPFS_PATH is not a file, change to -d or point to PATH/config

achingbrain:

Sorry we never got to this.

PR #4282, opened 2023-01-10
fix: allow reading rawLeaves in MFS

RangerMauve:
Currently, if you write to an MFS directory with `rawLeaves: true`, it'll error out when doing a read saying `Error: /example-0/example.txt was not a file`.

This accounts for that case by handling the `raw` type the same as a file.

There might be other places this should be looked at but this was an immediate blocker I was dealing with.

This might also be something we should fix in the exporter?


RangerMauve:
If someone needs a temporary workaround, either apply this patch manually, or avoid using `rawLeaves: true` with MFS.

achingbrain:
Thanks for submitting this - could you add a test please, to prevent future regressions.

RangerMauve:
@achingbrain Yeah sure, which file should I add the test to? I see that `files.read` is only used in [tests/preload.spec.js](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/test/preload.spec.js#L285)

Should I create a new file?

achingbrain:
Awesome, thanks - please add it to the interface suite: https://github.com/ipfs/js-ipfs/blob/master/packages/interface-ipfs-core/src/files/read.js

achingbrain (approved):

achingbrain (approved):

RangerMauve:
Yikes, not sure what';s up with the CI errors, but I don't think it's a result of the PR

achingbrain:
The new test is failing when run against kubo.  I've fixed it up by copying the added file to the MFS first, rather than using MFS methods to read the block using the ipfs path.

This is a bug with kubo really, it should be able to handle raw leaves.

RangerMauve:
TY for the quick turnaround time. 😁

Is this something that will be auto-released or is there somewhere I can track the release schedule?

tabcat:
@RangerMauve most of the js ipfs/libp2p repos use [release-please action](https://github.com/marketplace/actions/release-please-action) which, for each push to master, publishes an [rc](https://www.npmjs.com/package/ipfs-core/v/0.17.0-fa578bac) and preps a [release pr](https://github.com/ipfs/js-ipfs/pull/4252).

RangerMauve:
I notice some of the jobs failed, does that mean I need to wait for the next successful release for the publish? https://github.com/ipfs/js-ipfs/actions/runs/3895148505/jobs/6650213839

achingbrain:
An rc is published with every successful merge to master.

The current HEAD is passing so this is available to try out via `npm i ipfs-core@next`, or wait for the release which will go out when https://github.com/ipfs/js-ipfs/pull/4252 is merged, likely tomorrow morning.

RangerMauve:
Perfect thank you. I'll subscribe to that and update my deps accordingly. 🙇

achingbrain:
> This is a bug with kubo really, it should be able to handle raw leaves.

Actually looking back on this, I don't think it's a bug in kubo.  The test uses `/ipfs/${cid}` as the path to read, js-ipfs figures out it's an IPFS path and reads it from the blockstore.  Kubo treats it as an MFS path, which doesn't exist, hence the `file does not exist` error.

Adding the CID to the MFS with a file name and then reading that works because both js-ipfs and kubo can resolve the file name as an MFS path.

The approach Kubo takes, although less flexible, is probably easier to reason about so is probably preferable.

PR #4303, opened 2023-01-26
fix: ipfs-http-client should lowercase key type when submitting to kubo

### LINKED ISSUE #4272: Can't generate a new key using ipfs-core
- **Version**:

`json
{
  version: '0.17.0',
  commit: '',
  repo: '12',
  system: 'amd64/linux',
  golang: 'go1.19.1',
  'ipfs-http-client': '1.0.0'
}`
- **Platform**:

Linux darkside 6.0.12-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Dec 8 17:15:53 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
- **Subsystem**:

ipfs-http-client
#### Severity:

Medium
#### Description:

- I tried to generate a new key by using ipfs-http-client
- It generate a error `HTTPError: unrecognized key type: Ed25519`
- It should had generate a new key that I could use with IPNS in ed25519
- I think it happens because the keytype is Ed25519 and not ed25519 (same for RSA instead of rsa)

- Create a configuration of ipfs-http-client
- Generate a new key
```javascript
import dotenv from 'dotenv';
import { create } from 'ipfs-http-client';
dotenv.config();

const ipfs = create({
    url: process.env.IPFS_API,
})

await ipfs.key.gen("KEY");
```

RangerMauve:
Fixes #4272

Without this one cannot generate a new key in TypeScript with Kubo since the TypeScript types require uppercase key names.

Not sure if the types should be changed since I'm not sure if js-ipfs-core relies on the uppercase version.

### DIFF
--- a/packages/ipfs-http-client/src/key/gen.js
+++ b/packages/ipfs-http-client/src/key/gen.js
@@ -14,6 +14,8 @@ export const createGen = configure(api => {
   async function gen (name, options) {
     const opts = options ?? { type: 'Ed25519' }
 
+    opts.type = opts.type.toLowerCase()
+
     const res = await api.post('key/gen', {
       signal: opts.signal,
       searchParams: toUrlSearchParams({


RangerMauve:
Actually I'm not sure what to do here since I'm getting TypeScript errors now.

If I update the types to reflect what Kubo supports it'd be a major breaking change in the API. :/

Maybe we can relax the type to be `| string` to allow for arbitrary type values?

achingbrain:

Please swap out `ipfs-http-client` for `kubo-rpc-client` - it's a drop in replacement and where bug fixes & features will land in future.

PR #4341, opened 2023-04-06
fix: crash when using searchParams option

sk-gachou:
This pull attempt to fix the issue when trying to use searchParams option for http client.

According to [the doc](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-http-client#additional-options), one can provide `searchParams` (an URLSearchParams instance) to add extra query parameters. Currently, this behaviour does not work, see this gist: https://gist.github.com/sk-gachou/2dd253acb1e1663e37119120c2db4768

It crash at this line:
https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/lib/to-url-search-params.js#L42

### DIFF
--- a/packages/ipfs-http-client/src/lib/to-url-search-params.js
+++ b/packages/ipfs-http-client/src/lib/to-url-search-params.js
@@ -9,7 +9,7 @@ export function toUrlSearchParams ({ arg, searchParams, hashAlg, mtime, mode, ..
   if (searchParams) {
     options = {
       ...options,
-      ...searchParams
+      ...Object.fromEntries(searchParams)
     }
   }
 


achingbrain:
js-IPFS is in the process of being deprecated, the replacement is [Helia](https://github.com/ipfs/helia) - please see the [State of IPFS in JS](https://blog.ipfs.tech/state-of-ipfs-in-js/) post for a bit of background and the [migration guide](https://github.com/ipfs/helia/wiki/%F0%9F%9A%9B-Migrating-from-js-IPFS) for how to port your app over.

You're using `ipfs-http-client` - if you are talking to a Kubo node, please use [kubo-rpc-client](https://github.com/ipfs/js-kubo-rpc-client) instead and open any issues there.

